PR #12, opened 2025-01-09
Readme feedback

lukors:
Putting my feedback on the readme here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,9 +189,7 @@ The interface provides visual tools for:
 4. **Responsive Design**: Works seamlessly across desktop and mobile devices
 5. **Progressive Enhancement**: Core functionality works even with limited browser features
 
-## Design
-
-### Contract Architecture
+## Contract Architecture
 
 The chat room contract is implemented using Freenet's composable state pattern. The core state structure is defined in [common/src/room_state.rs](common/src/room_state.rs):
 
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # River - Decentralized Chat on Freenet
 
-![Screenshot of chat interface](screenshot-20241009.png)
-
 River is a decentralized group chat system built on Freenet, designed to provide a secure and
 upgradeable alternative to traditional chat platforms. It features a web-based interface built
 with [Dioxus](https://dioxuslabs.com) and a modular contract architecture using the freenet-scaffold
 framework.
+
+![Screenshot of chat interface](screenshot-20241009.png)
 
 ## Current Status (Jan 2025)
 
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The system is built using:
 
 River is open-source software licensed under the MIT License. See [LICENSE](LICENSE) for details.
 
-# Membership Management
+## Membership Management
 
 River uses a flexible system for controlling room membership, starting with invitations but designed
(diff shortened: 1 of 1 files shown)


lukors:
This pull request has these goals:

- Merge duplicated information.
  - Having one place for each information makes it easier to read, and easier to update.
- Categorize information more clearly.
  - This makes it easier to find the information you're looking for.

I recommend moving everything from the "technical details" heading and onwards (minus the license) to another document. That information is mostly relevant for those who want to contribute to the project, and could be a good basis for documenting the project on a more technical level. The technical documentation could be linked in the readme for those who want it. This makes the readme more approachable for the majority of people, making it easier to attract users. I can make a followup PR for that if it sounds like a good idea.

README.md
```diff
@@ -1,52 +1,57 @@
...
 
 🌐 **Web-based Interface** - Modern web UI built with Dioxus for cross-platform compatibility  
 🔒 **Secure by Design** - Uses elliptic curve cryptography for authentication and signing  
 🔄 **Upgradeable** - Flexible upgrade mechanism for both UI and contracts  
 🌱 **Extensible** - Open architecture allows alternative UIs and integrations  
 📜 **Modular Contracts** - Built using freenet-scaffold for composable state management  
 📦 **Efficient Storage** - Uses CBOR serialization via [ciborium](https://crates.io/crates/ciborium)  
+✉️ **Invitation-based membership** - Requiring invitations helps combat spam
+
+Please note this is pre-release software - expect breaking changes and missing features. We welcome testing and feedback from early adopters!
```
sanity, line 22: Need to be clear that as of right now it's not usable, don't want people trying it only to be disappointed.

README.md
```diff
@@ -1,52 +1,57 @@
...
 
 ## Getting Started
 
-To join a River chat room, you'll need:
-1. The room's contract address (derived from its public key)
-2. An invitation from an existing member
+River runs in your browser, and is built to work both on mobile phones and desktop computers.
+
+1. Launch River by following this link: (there is no link yet, because River isn't ready for use at this point)
```
sanity, line 28: we shouldn't have this until it's ready - when River is ready to use they'll need to install Freenet first then follow a link to River, but we shouldn't go into detail yet because it may confuse people into thinking it's ready to try.

sanity:
> I recommend moving everything from the "technical details" heading and onwards (minus the license) to another document. 

Before launch we should do this, but since there is nothing for people to use yet I think it's ok to focus more on developers.

README.md
lukors: I think I made that clearer now, let me know what you think!

README.md
lukors: I added a disclaimer at the top of the section to make it clear that it's not ready, and also updated the list with your clarification.

Let me know if you want the entire "getting started" removed! I wasn't sure what you meant in your comment.

lukors:
I think I've addressed all feedback now, so let me know if there's anything more, or if I maybe misunderstood something. :)

gogo2464:
nice catch!

could you let me merge the build instruction please? https://github.com/freenet/river/pull/13

lukors:
Build instructions are out of scope for this pull request. They can be merged separately. :)

gogo2464:
ok

Issue #50, opened 2025-12-21
Agent handoff: fix flaky test and merge PR #49 (late joiner regression test)


sanity:
## Context for Agent Handoff

This issue is a handoff for another agent to complete River CI fixes and merge PR #49.

### Background

PR #49 adds a regression test for freenet-core#2360 (contract key mismatch bug). The PR includes:
1. **New test**: `river_late_joiner_receives_updates` in `cli/tests/message_flow.rs` - tests that peers joining late via invite can receive subsequent updates
2. **Dep update**: Updated `freenet` from 0.1.44 to 0.1.55 in `contracts/room-contract/Cargo.toml` and added new `NetworkArgs` fields (`total_bandwidth_limit`, `min_bandwidth_per_connection`)
3. **WASM sync**: Synced CLI WASM files

### Current Blocker

The `test_invitation_message_propagation` integration test in `room-contract` is flaky and times out intermittently:

```
Bob failed to get room state: Node1 get request timed out after 45s
```

**Location**: `contracts/room-contract/tests/integration_tests.rs`

This test is unrelated to PR #49 changes but blocks CI.

### Tasks for Agent

1. **Fix or skip the flaky test** - Either:
   - Increase timeout in `test_invitation_message_propagation`
   - Add retry logic for GET requests
   - Mark with `#[ignore]` and TODO-MUST-FIX comment

2. **Get PR #49 merged** once CI passes

3. **Verify freenet-core PR #2360 CI** passes (the actual fix PR)

### Files to Check

- `contracts/room-contract/tests/integration_tests.rs` - flaky test location
- `contracts/room-contract/tests/common/test_utils.rs` - already updated with new NetworkArgs fields
- `cli/tests/message_flow.rs` - new regression test (lines 1307-1585)

### Related PRs/Issues

- freenet-core PR #2360: The actual fix for contract key mismatch
- River PR #49: The regression test (blocked by this flaky test)

[AI-assisted - Claude]

PR #61, opened 2025-12-28
fix(test): add retry logic to Bob's subscribe in integration test

sanity:
## Problem

The `test_invitation_message_propagation` integration test was failing intermittently in CI with:

```
Bob subscribe failed: Subscribe response timeout after 60s
```

**Root cause**: Bob's subscribe operation requires network routing to find the contract (which Alice deployed on her node), and this can timeout in resource-constrained CI environments where network routing may be slower than the 60-second timeout.

Alice's subscribe completes quickly (~5ms) because the contract is already on her node, but Bob must:
1. Send a Subscribe request through the network
2. Find where the contract is stored
3. Retrieve the state and establish a subscription

## Solution

1. Add `subscribe_to_contract_with_retries()` function with configurable retry attempts
2. Use 3 retries for Bob's subscribe (with 5-second delay between attempts)
3. Keep Alice's subscribe without retries since it's local

This follows the same retry pattern already used successfully for other network operations in the test suite (e.g., the six-peer regression test).

## Testing

- Code compiles without errors
- The retry logic follows the same pattern as existing retry operations in the codebase
- Local tests pass

[AI-assisted - Claude]

### DIFF
--- a/contracts/room-contract/tests/common/mod.rs
+++ b/contracts/room-contract/tests/common/mod.rs
@@ -172,39 +172,66 @@ pub async fn deploy_room_contract(
 }
 
 pub async fn subscribe_to_contract(client: &mut WebApi, key: ContractKey) -> Result<()> {
-    println!("Starting subscribe_to_contract for key: {}", key);
+    subscribe_to_contract_with_retries(client, key, 1).await
+}
 
-    println!("Sending Subscribe request to WebSocket...");
-    let send_result = client
-        .send(ClientRequest::ContractOp(ContractRequest::Subscribe {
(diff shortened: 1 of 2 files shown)


sanity:
Closing as obsolete. Since this PR was opened:

- `test_invitation_message_propagation` was marked `#[ignore]` (commit 8bb46cd8), and
- the entire integration test file was gated behind the `integration-test` feature (commit 9e917cb8), which the CI `build.yml` job does not enable.

So the test no longer runs in CI and this retry logic would have no effect. The test is flaky for deeper reasons (network setup + message duplication, per issue #50); adding retries papers over that rather than fixing the root cause.

[AI-assisted - Claude]

PR #68, opened 2026-01-19
feat(security): move signing operations from UI to delegate

sanity:
## Problem

Private signing keys were stored in the UI's RoomData struct and used directly for signing operations (messages, members, bans, config, etc.). This exposed keys in the browser's memory and in serialized state, which is a security concern.

## Approach

Move signing operations to the chat-delegate where keys can be stored securely in Freenet's secret storage. The UI now:

1. **Requests signatures from the delegate** instead of signing locally
2. **Receives signature bytes back** and constructs authorized structs using new `with_signature()` constructors
3. **Falls back to local signing** during migration period for backwards compatibility

### Key Changes

**Delegate interface extensions** (`common/src/chat_delegate.rs`):
- Added `RoomKey` type alias (`[u8; 32]`)
- Added signing request variants: `StoreSigningKey`, `GetPublicKey`, `SignMessage`, `SignMember`, `SignBan`, `SignConfig`, `SignMemberInfo`, `SignSecretVersion`, `SignEncryptedSecret`, `SignUpgrade`
- Added corresponding response variants with signature bytes

**Delegate implementation** (`delegates/chat-delegate/src/handlers.rs`):
- Added `handle_store_signing_key()` - stores keys in Freenet secret storage
- Added `handle_get_public_key()` - retrieves public key if stored
- Added `handle_sign_request()` - performs signing using stored key
- Keys stored with format `signing_key:{origin}:{room_key}`

**UI signing API** (`ui/src/signing.rs` - new file):
- Async wrapper functions for all signing operations
- `migrate_signing_key()` - copies keys from RoomData to delegate
- `sign_*_with_fallback()` functions - try delegate first, fall back to local

**Authorized types** (various `common/src/room_state/*.rs`):
- Added `with_signature()` constructors to all Authorized types

**UI components updated**:
- `conversation.rs` - messages use delegate signing
- `room_name_field.rs` - config changes use delegate signing  
- `ban_button.rs` - bans use delegate signing
- `invite_member_modal.rs` - member invites use delegate signing

**Migration** (`response_handler.rs`, `get_response.rs`):
- Automatic key migration when rooms are loaded from delegate storage
- Automatic key migration when accepting room invitations

### Migration Path

The migration is non-destructive:
- `self_sk` remains in RoomData serialization during transition
- Delegate signing uses fallback to local if delegate fails
- Keys are automatically migrated on room load
- `key_migrated_to_delegate` flag tracks migration status per room

This allows gradual rollout and easy rollback if issues arise.

## Testing

- All existing tests pass
- UI builds successfully for WASM target
- Chat-delegate tests pass (10 tests)
- **Important:** Before production deployment, we should publish to a separate contract key and test with Playwright to avoid breaking the production room

## Test Plan

- [ ] Create a new room, verify signing works
- [ ] Send messages, verify signatures are valid
- [ ] Invite a member, verify invite signature works
- [ ] Ban a member, verify ban signature works
- [ ] Change room name, verify config signature works
- [ ] Restart app, verify migration happens automatically
- [ ] Verify old rooms with non-migrated keys still work (fallback)

[AI-assisted - Claude]

### DIFF
--- a/contracts/room-contract/tests/common/test_utils.rs
+++ b/contracts/room-contract/tests/common/test_utils.rs
@@ -89,6 +89,8 @@ pub async fn base_node_test_config_with_rng(
             streaming_enabled: None,
             streaming_threshold: None,
             ledbat_min_ssthresh: None,
+            bbr_startup_rate: None,
+            congestion_control: None,
         },
         config_paths: freenet::config::ConfigPathsArgs {
             config_dir: Some(temp_dir.path().to_path_buf()),


sanity:
## Review Summary

Four parallel review agents analyzed this PR. Here's what was found and addressed:

### Issues Fixed in Latest Commits:
1. **Race condition (Critical)** - Added unique request IDs to all signing operations to prevent concurrent requests from overwriting each other's channels
2. **Dead code** - Removed unused `sign_bytes()` function
3. **Missing notification function** - Added `river_test_notification` that was referenced but not committed
4. **Upstream dependency** - Added new `NetworkArgs` fields for freenet 0.1.102+

### Reviewed - Not Blocking:
- **Missing unit tests for signing handlers** - Valid concern but fallback behavior provides safety net during migration. Tests can be added in follow-up PR.
- **Incomplete migration** - Intentional. `room_data.rs` keeps local signing for internal operations (room creation, secret rotation) during migration period. The `self_sk` field is preserved for backward compatibility.
- **"Wrong key" concern** - The design uses `room_key` (owner's VK) as the room identifier. Each user's delegate stores their own signing key for each room under this key. This is correct behavior.

### CI Status: All checks passing

[AI-assisted - Claude]

PR #112, opened 2026-02-19
test: assert UpdateNotification delivery in six-peer integration test

### LINKED ISSUE #111: test: add UpdateNotification assertions to six-peer integration test
## Problem

The six-peer regression test (`river_message_flow_over_freenet_six_peers_five_rounds` in `cli/tests/message_flow.rs`) currently verifies message delivery only via GET-based polling (`riverctl message list`). The test creates subscription `WebApi` handles via `subscribe_peer_to_contract()`, but **never reads `UpdateNotification` events from them** — the handles are stored and disconnected without being consumed.

Since [freenet-core PR #2489](https://github.com/freenet/freenet-core/pull/2489) made subscriptions async (fire-and-forget), the test passes regardless of whether background subscriptions actually complete. The [subscription snapshot evidence](https://github.com/freenet/freenet-core/issues/2494#issuecomment-3923499711) (only 1 of 6 peers showing active subscriptions) strongly suggests they may be silently timing out.

## Context

From [freenet-core#2494 comment](https://github.com/freenet/freenet-core/issues/2494#issuecomment-3923499711):

> The six-peer regression test lives in the River repository. It tests message flow end-to-end using `riverctl send_message` and `list_messages`, but it only verifies that messages are *readable*, not that **subscription `UpdateNotification` events reached all 6 peers**.

The freenet-core side already has explicit subscription propagation assertions in its 3-peer `test_multiple_clients_subscription` test (`operations.rs`), which checks that `ContractResponse::UpdateNotification` is received by clients on all nodes. The River test lacks an equivalent.

## Required Changes

All changes in `cli/tests/message_flow.rs`:

1. **Track peer index with subscription handles** — change `Vec<WebApi>` to `Vec<(usize, WebApi)>` so assertions can identify which peer failed

2. **Add `drain_notifications()` helper** — loop on `WebApi::recv()` with per-recv timeouts, counting `UpdateNotification` events matching the room's contract key

3. **Spawn drain tasks concurrently during message sending** — the `WebApi` response channel has capacity 1, so notifications must be consumed concurrently to avoid back-pressure blocking the WebSocket read loop

4. **Assert each subscribed peer received ≥1 `UpdateNotification`** — after messages are sent and GET-based verification passes, collect drain task results and assert all peers got at least one push notification

## Notes

- Keep existing GET-based verification (`wait_for_expected_messages`) — it validates contract state consistency
- Don't require exact notification counts (deltas may be batched/merged by the contract) — assert ≥1 per peer
- The `WebApi` response channel capacity of 1 (in `freenet-stdlib`) means drain tasks must start before messages are sent

iduartgomez:
## Summary
- Add concurrent `drain_notifications()` tasks that consume `UpdateNotification` push events from subscription WebApi handles during message sending
- Assert each subscribed peer received ≥1 `UpdateNotification` after message verification completes
- Track peer index alongside subscription handles (`Vec<WebApi>` → `Vec<(usize, WebApi)>`) for per-peer reporting

## Context
The six-peer test (`river_message_flow_over_freenet_six_peers_five_rounds`) verified message delivery only via GET-based polling (`riverctl message list`), but never asserted that Freenet's `UpdateNotification` push events actually reached subscribed peers. Since PR freenet/freenet-core#2489 made subscriptions async, the test passes regardless of whether subscriptions work — a silent regression risk.

The WebApi response channel has **capacity 1**, so notifications must be drained concurrently during message sending or back-pressure blocks the WebSocket read loop and notifications are lost.

Debug output available via `FREENET_TEST_DEBUG_NOTIFICATIONS=1`.

Ref: freenet/freenet-core#2494

Closes #111 

## Test plan
- [x] `cargo check --test message_flow` compiles cleanly
- [x] `cargo clippy --test message_flow` passes with no warnings
- [ ] Full six-peer test in CI validates notifications are delivered (assertion will cause failure if subscriptions silently break)

### DIFF
--- a/cli/tests/message_flow.rs
+++ b/cli/tests/message_flow.rs
@@ -665,6 +665,78 @@ async fn dump_subscriptions(network: &TestNetwork) {
     println!("--- End subscription snapshot ---");
 }
 
+/// Drain UpdateNotification events from a subscription handle until the deadline.
+///
+/// The WebApi response channel has capacity 1, so notifications must be consumed
+/// concurrently during message sending — otherwise back-pressure blocks the WebSocket
+/// read loop and notifications are silently lost.
+///
+/// Returns `(peer_index, notification_count, client)`. The client is returned so the
+/// caller can disconnect it after collecting results.
+async fn drain_notifications(
+    mut client: WebApi,
+    contract_key: ContractKey,
+    peer_index: usize,
+    deadline: Instant,
+) -> (usize, usize, WebApi) {
+    let debug = std::env::var_os("FREENET_TEST_DEBUG_NOTIFICATIONS").is_some();
+    let mut count = 0usize;
+
+    loop {
+        let remaining = deadline
+            .checked_duration_since(Instant::now())
+            .unwrap_or_default();
+        if remaining.is_zero() {
+            break;
+        }
+
+        // Use a short per-recv timeout so we don't block forever on a quiet channel
+        let recv_timeout = remaining.min(Duration::from_secs(1));
+        match tokio::time::timeout(recv_timeout, client.recv()).await {
+            Ok(Ok(HostResponse::ContractResponse(
+                ContractResponse::UpdateNotification { key, .. },
+            ))) => {
+                if key == contract_key {
+                    count += 1;
+                    if debug {
+                        println!(
+                            "debug: peer {} received UpdateNotification #{} for contract {}",
+                            peer_index, count, contract_key
+                        );
+                    }
+                } else if debug {
+                    println!(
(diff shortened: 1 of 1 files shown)


iduartgomez:
@sanity this may break CI in freenet-core if subscriptions are not actually working properly.

sanity (approved):
Code looks good. The drain_notifications() approach correctly handles the capacity-1 WebApi channel by consuming concurrently during message sending, and the ≥1 assertion per peer is the right threshold. If this breaks six-peer CI in freenet-core, that's a real subscription bug we need to fix rather than hide.

[AI-assisted - Claude]

PR #129, opened 2026-03-05
fix(ui): send contract-migration upgrade pointer as a delta

### LINKED ISSUE #127: Contract migration sends invalid state to old contract, causing 'Invalid signature' error
## Problem

When a room contract is migrated to a new WASM version, the migration code sends an upgrade pointer UPDATE to the old contract. This UPDATE fails with:

```
UPDATE operation failed: invalid contract update, reason: State verification failed: Invalid signature: signature error
```

This was observed in the browser console while the UI migrated room `XQTMFW5V` from old contract `oW3AFr9RE15sm9d1dZwEQVMjaNJKwsyNbGUP1ptekE6` to new contract `GD238LC8mu6FHfp33w9DVATyKaCa9n5CNnKUzyANzodj`.

## Root Cause

In `ui/src/components/app/freenet_api/room_synchronizer.rs` around line 449-453, the upgrade state is constructed as:

```rust
let upgrade_state = ChatRoomStateV1 {
    upgrade: OptionalUpgradeV1(Some(authorized_upgrade)),
    ..Default::default()   // ← BUG: empty/invalid state for all other fields
};
```

The `..Default::default()` produces an empty `configuration` (with zeroed `owner_member_id`), empty members, etc. The old contract's state verification checks the entire submitted state for consistency and rejects it because the configuration fields are invalid/empty.

## Expected Behavior

The upgrade pointer should be sent as a delta (not a full state replacement), or the full current room state should be included alongside the upgrade pointer so the old contract's validation passes.

## Steps to Reproduce

1. Have a room created with an older contract version
2. Connect with a new River UI that triggers contract migration
3. Observe the console error: `UPDATE operation failed: ... Invalid signature: signature error`

## Impact

- The upgrade pointer is never set on the old contract
- Other clients can't follow the migration pointer to the new contract
- The new contract subscription eventually fails with repeated timeout errors

## Related

- #63 (Graceful room contract WASM upgrades)

sanity:
## Problem

When a room contract is migrated to a new WASM version, the owner's UI sends an upgrade-pointer UPDATE to the *old* contract so old clients can discover the new contract. That UPDATE was a full `UpdateData::State` built with `..Default::default()` for every field except `upgrade`.

A full state UPDATE is run through the old contract's `validate_state` → `ChatRoomStateV1::verify`. The default `configuration` sub-state is unsigned, so verification fails:

```
UPDATE operation failed: invalid contract update, reason: State verification failed: Invalid signature: signature error
```

and the upgrade pointer never lands — old clients can't follow the migration.

## Approach

Send a minimal `UpdateData::Delta` carrying only the `upgrade` field. The old contract applies it via `apply_delta`, which validates just the upgrade signature against the contract's owner parameter. This:

- never trips full-state `validate_state` / `verify` (the bug);
- never drags the room's members/messages/secrets through the old WASM's `merge` / `post_apply_cleanup`;
- avoids sending a `version`-bearing full state that an older WASM generation could reject;
- is a ~100-byte payload instead of the whole room state.

This is the delta option issue #127 lists first under "Expected Behavior".

## Testing

- `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` — compiles
- `cargo test -p river-core --lib` — 173 tests pass, including the new `test_upgrade_pointer_applies_as_delta`: it applies the upgrade delta via `apply_delta`, asserts the pointer lands and the result still verifies, and characterizes the #127 bug (a `..Default::default()` full state fails `verify`).
- Adds a `cargo test -p river-core --lib` CI step — the river-core lib suite previously ran in CI only filtered to `ecies`-named tests, so the new regression test (and most of the suite) had no CI coverage.

Note: this branch was rebased onto current `main` (the original branch was ~200 commits stale); the README Node.js-install hunk from the old PR is already in `main`.

Closes #127

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,6 +94,16 @@ jobs:
         CARGO_TARGET_DIR: ${{ github.workspace }}/target
       run: cargo test -p river-core --test migration_test
 
+    - name: Test river-core lib (room_state, crypto, CRDT unit tests)
+      env:
+        RUST_MIN_STACK: 8388608
+        CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      # Runs the river-core library unit tests — room_state verification,
+      # CRDT convergence, contract-migration regressions, etc. Previously
+      # only the `ecies`-named subset ran in CI (see the step below), so the
+      # bulk of the suite had no CI coverage.
+      run: cargo test -p river-core --lib
+
     - name: Test river-core ecies (deterministic + randomized helpers)
       env:
         RUST_MIN_STACK: 8388608
--- a/common/src/room_state.rs
+++ b/common/src/room_state.rs
@@ -1434,4 +1434,73 @@ mod tests {
             "State should be partially modified"
         );
     }
+
+    /// Regression test for freenet/river#127: the contract-migration upgrade
+    /// pointer is delivered to the old contract as a minimal `apply_delta`,
+    /// NOT as a full-state UPDATE.
+    ///
+    /// A full `UpdateData::State` is run through the old contract's
+    /// `validate_state` -> `ChatRoomStateV1::verify`. The old code built that
+    /// state with `..Default::default()`, whose `configuration` is unsigned,
+    /// so verification failed with "Invalid signature" and the pointer never
+    /// landed. Applying only the `upgrade` field as a delta runs
+    /// `OptionalUpgradeV1::apply_delta`, which validates just the upgrade
+    /// signature against the contract's owner parameter.
+    #[test]
+    fn test_upgrade_pointer_applies_as_delta() {
+        use crate::room_state::upgrade::{AuthorizedUpgradeV1, UpgradeV1};
+
+        let (state, parameters, owner_signing_key) = create_empty_chat_room_state();
+        // Sanity: the baseline room state is itself valid and has no pointer.
+        assert!(
+            state.verify(&state, &parameters).is_ok(),
+            "baseline room state should verify"
+        );
+        assert!(
+            state.upgrade.0.is_none(),
+            "baseline room state should have no upgrade pointer"
+        );
+
+        let upgrade = UpgradeV1 {
+            owner_member_id: MemberId::from(&parameters.owner),
+            version: 1,
+            new_chatroom_address: blake3::Hash::from([7u8; 32]),
+        };
+        let authorized = AuthorizedUpgradeV1::new(upgrade, &owner_signing_key);
+
+        // FIX: apply the upgrade pointer as a minimal delta — the path the old
+        // contract takes for `UpdateData::Delta`. The pointer must land and the
+        // resulting state must still verify.
+        let delta = ChatRoomStateV1Delta {
+            upgrade: Some(authorized.clone()),
+            ..Default::default()
+        };
+        let mut updated = state.clone();
+        updated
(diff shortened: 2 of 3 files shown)


iduartgomez (approved):

sanity (commented):
## Comprehensive PR Review: #129 (multi-model, 2 rounds)

### Summary
- **PR Title:** fix(ui): send contract-migration upgrade pointer as a delta
- **Type:** fix
- **Linked Issues:** #127
- **Review tier:** Full (contract migration — high-risk surface)
- **Reviewers run:** code-first, testing, skeptical, big-picture (freenet:* subagents) + Codex (external, non-Claude). Two rounds — round 1 on the original full-state approach, round 2 after the rewrite to a delta.

### Round 1 → Round 2

Round 1 reviewed the original approach (send the upgrade pointer to the old contract as a full `UpdateData::State` cloned from the live room state). It surfaced:
- **[Testing — blocking]** the regression test never ran in CI (the `river-core --lib` step was name-filtered to `ecies`).
- **[Skeptical M1]** an older WASM generation could reject a `version`-bearing full state.
- **[Skeptical M2]** hardcoded `version: 1` + the merge path's strict `>` makes re-migration a silent no-op.
- **[Skeptical M3 / L1]** a full-state UPDATE runs the old WASM's `merge`/`post_apply_cleanup` and ships a multi-KB payload.
- **[Code-first]** the explanatory comment misdescribed the failure as `merge -> verify` (`merge` does not call `verify`).

The PR was rewritten in response: it now sends a minimal `UpdateData::Delta` carrying only the `upgrade` field — issue #127's first-listed preferred fix.

### Round 2 findings (current HEAD)

All round-1 findings verified **resolved**:
- The delta path (`apply_delta`) never substitutes an unsigned default `configuration`, so it cannot trip the `validate_state`/`verify` failure that was the #127 bug (M1).
- `OptionalUpgradeV1::apply_delta` overwrites the pointer unconditionally after signature validation — re-migration works (M2 — the strict-`>` check is only in `delta()`, not on this path).
- The delta carries only the signed `AuthorizedUpgradeV1` (~100 bytes); no members/messages/secrets are sent (M3/L1).
- The explanatory comments are now accurate.
- A `cargo test -p river-core --lib` CI step was added — the regression test (and ~170 previously CI-dark river-core lib tests) now run in CI.

| Reviewer | Verdict |
|----------|---------|
| Code-first | No blocking issues; comments now accurate |
| Testing | Both round-1 findings resolved; CI-coverage gap genuinely closed |
| Skeptical | M1/M2/M3 resolved; no High/Medium findings |
| Big-picture | No blocking findings; aligns with issue #127 and AGENTS.md |
| Codex (external) | No discrete regressions introduced |

### Non-blocking — addressed or justified
- **Skeptical L1** (the comment said "never drags ... through the old WASM" — `post_apply_cleanup` still iterates the old contract's *own* state): comment tightened in a follow-up commit.
- **Testing nice-to-haves** (a CBOR round-trip test; a populated-room test): not added — `ChatRoomStateV1Delta` CBOR round-trips are already exercised by existing delta integration tests, and the delta path deliberately does not touch members/messages, so an empty-room unit test is faithful to the path under test.

### Verdict
**State:** Ready to Merge (pending CI green on the current HEAD)
**HEAD SHA reviewed:** `eda3718a` (round-2 reviewers) + a comment-only tweak for skeptical L1.

[AI-assisted - Claude]

PR #209, opened 2026-04-13
fix(ui): consistent message bubble layout at narrow widths

### LINKED ISSUE #205: Edit box wider than view
I use River in a pretty narrow window. When I edit a message, the edit box goes outside the River window. See attached image.

Image

### LINKED ISSUE #206: Reply messages require more width than others
Messages that are replies to other messages have a different minimum width than others. It would be better if the minimum width was the same for all messages.

Image

### LINKED ISSUE #207: Hovering a reply changes the width of the message
In a message that's a reply, the quoted message at the top of it can be hovered to expand it a bit. When this happens, the width of the entire message can change at certain window sizes, causing the line breaks to change.

This does not happen when the window  is wide. The pictures below are using a really narrow window to clearly show the issue. It happens also even when the window is so wide that the "rooms" bar on the left is visible.

## Not hovering

Image

## While hovering

Image

sanity:
## Problem

Three related message-layout bugs reported by @lukors:

- **#205** edit box wider than the viewport at narrow window widths (hard-coded `width: 550px` on the edit container).
- **#206** reply messages had a different minimum width than non-reply messages — reply bubbles were dramatically wider.
- **#207** hovering the quoted-reply preview reflowed the surrounding message's width at narrow viewport widths.

## Approach

**#205**: The edit container had an inline `width: 550px` that ignored the viewport. Replace with `width: 100%; max-width: 550px` so the edit form fills the available chat area and caps at 550px on wider screens.

**#206 and #207** have the same root cause: the reply preview strip was a sibling of the message bubble in a shrink-to-fit column. Its `white-space: nowrap` gave the strip an intrinsic min-content equal to its full preview text, which forced the shrink-to-fit column wider than a non-reply bubble. Hovering toggled `white-space: normal` and the whole bubble reflowed.

Restructure: move the reply strip **inside** the message bubble as its first child. The bubble is now sized by its body prose (same as a non-reply bubble), the strip ellipsizes within the bubble's width, and hovering the strip only grows it vertically. The strip keeps its distinctive background color and click-to-scroll-to-original behaviour.

Also gate the hover-to-expand CSS behind `@media (hover: hover) and (pointer: fine)`. Touch devices don't have a reliable hover state, and tapping the strip already scrolls to the quoted original, so touch users don't need the hover expansion. This also eliminates a Mobile Safari edge case where `:hover` applied on touch and caused a small width change.

## Testing

New Playwright spec `ui/tests/message-layout.spec.ts` with three regression tests:

1. **#205** — edit container fits the viewport at 500px wide.
2. **#206** — reply bubble width is within 40px of non-reply siblings at 480px wide (the bug was a visible 100-200px gap).
3. **#207** — hovering the reply strip doesn't change the bubble's bounding width.

Example data (`ui/src/example_data.rs`) now includes one reply message so the tests have something to exercise; the change is behind the `example-data` feature so it doesn't affect release builds.

Full cross-browser suite: **80/80 passing** across chromium, firefox, webkit, mobile-chrome, and mobile-safari.

## Delegate / contract WASM

UI-only change. Only `ui/assets/main.css`, `ui/src/components/conversation.rs`, `ui/src/example_data.rs`, and `ui/tests/message-layout.spec.ts` touched. No `delegates/`, `contracts/`, `common/`, `Cargo.toml`, or `Cargo.lock` changes, so the delegate/contract WASM hash is unaffected and no `legacy_delegates.toml` entry is required.

Closes #205, closes #206, closes #207.

[AI-assisted - Claude]

### DIFF
--- a/ui/assets/main.css
+++ b/ui/assets/main.css
@@ -31,14 +31,20 @@
     100% { background-color: transparent; }
 }
 
-/* Reply strip text truncation */
+/* Reply strip text truncation. The strip lives inside the message bubble
+ * and always shows a single ellipsized line; on devices with a real pointer
+ * (mouse) it expands to full text on hover. Touch devices skip the hover
+ * expansion because (a) they have no reliable hover state and (b) tapping
+ * the strip already scrolls to the quoted original message. */
 .reply-strip {
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
 }
 
-.reply-strip:hover {
-    white-space: normal;
-    text-overflow: clip;
+@media (hover: hover) and (pointer: fine) {
+    .reply-strip:hover {
+        white-space: normal;
+        text-overflow: clip;
+    }
 }
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -1749,15 +1749,22 @@ fn MessageGroupComponent(
                                                             }
                                                         }
                                                     },
-                                                    // Reply context strip (inside bubble, first child)
+                                                    // Reply context strip (inside bubble, first child).
+                                                    // Self bubbles use a white-tinted overlay so the strip
+                                                    // stays legible against the accent background; other
+                                                    // bubbles use a dark-tinted overlay against the surface
+                                                    // background. The previous `bg-accent/40 text-accent`
+                                                    // was invisible on self bubbles because the strip
+                                                    // composited to the same colour as the bubble.
                                                     if let (Some(author), Some(preview)) = (reply_author_inner, reply_preview_inner) {
                                                         {
                                                             let target_id_str = reply_target_inner.map(|id| format!("{:?}", id.0)).unwrap_or_default();
                                                             rsx! {
                                                                 div {
+                                                                    "data-testid": "reply-strip",
                                                                     class: format!(
                                                                         "reply-strip min-w-0 w-full text-[11px] leading-normal px-3 pt-1.5 pb-1.5 cursor-pointer {}",
-                                                                        if is_self { "bg-accent/40 text-accent" } else { "bg-black/[0.12] text-text-muted" }
+                                                                        if is_self { "bg-white/25 text-white/90" } else { "bg-black/[0.12] text-text-muted" }
                                                                     ),
                                                                     title: "Click to scroll to original message",
                                                                     onclick: move |_| {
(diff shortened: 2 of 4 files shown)


sanity:
Multi-model review round complete (code-first, testing, skeptical, big-picture + codex). Addressed in a5450487:

**Codex**
- Self-reply strip was invisible: previously `bg-accent/40 text-accent` composited to the same accent colour as the `bg-accent` bubble. Now uses `bg-white/25 text-white/90` on self bubbles so the strip stays legible.

**Skeptical**
- `overflow-auto → overflow-hidden` could silently clip long URLs or wide `<pre>` blocks. Wrapped the prose body in a padding container and added `break-words` so long tokens wrap instead of being clipped.
- Playwright edit-button locator was unscoped; a previously-hovered sibling's edit button could be found instead of the current hover target. Scoped the lookup to the hovered bubble's `#msg-` ancestor.

**Code-first / Skeptical**
- `(edited)` indicator had moved from being an inline sibling of the prose body to a flex-column row below it. Restored the original inline placement by wrapping prose + edited-span in a dedicated body container.

**Testing**
- Hover-width test was a false pass on touch browsers (the `@media (hover: hover) and (pointer: fine)` gate suppressed the hover rule). Now explicitly `test.skip()`s when the media query is false, so the test only runs where it is meaningful. Also replaces the 150 ms `waitForTimeout` with an `expect.poll()` on the computed `white-space` so the test deterministically waits for the CSS to engage.
- Added `data-testid="reply-strip"` hook for future-proof selectors.

**Not addressed / out of scope**
- *Keyboard / focus accessibility for the reply strip* — pre-existing gap (the old CSS only had `:hover` too), and PR scope is deliberately tight to the reported bugs. Filed as #210 with a proposed fix (role=button, tabindex, focus-visible CSS rule, Playwright test).
- *Tighter #206 tolerance asserting the reply strip's intrinsic content is wider than the bubble* — tricky to measure portably; the 40 px tolerance at a 480 px viewport is already snug (the original regression was 100-200 px).
- *Resize-during-edit test* — not covered; `width: 100%; max-width: 550px` is inherently responsive, and the initial-render test at 500 px wide already exercises the narrow case.

**Test results**
- 76 passed, 3 correctly skipped (mobile-chrome, mobile-safari, headless firefox where the hover media query is false). One pre-existing flaky test in `responsive-layout.spec.ts:255` auto-passed on retry (unrelated).

**Delegate migration**
- UI-only; no `delegates/`, `contracts/`, `common/`, `Cargo.toml`, or `Cargo.lock` touched. No `legacy_delegates.toml` entry required.

[AI-assisted - Claude]

PR #213, opened 2026-04-14
fix(ui): wrap long URLs inside message bubbles (#212)

### LINKED ISSUE #212: Long lines cause too wide bubbles
Posting for instance a long link that doesn't give the text formatter a chance to break it up will stretch the bubble. See attached image.

Image

sanity:
## Problem

Posting a long link that a text formatter cannot break up (e.g. a URL with a very long unbroken segment) stretched the message bubble wider than other bubbles. PR #209 added `break-words` (overflow-wrap: break-word) on the prose body, but `break-word` only inserts soft breaks when a word would otherwise overflow, and crucially does NOT lower the element's min-content size. So a sibling flex/grid parent can still be forced wider by the intrinsic min-content of a long token.

## Approach

Replace `break-words` with `[overflow-wrap:anywhere]`. `anywhere` is a strict superset of `break-word`: it keeps the soft-break behavior but also lowers the element's min-content to a single character, allowing the bubble's flex ancestors to shrink around the wrapped content.

## Testing

- Added an example-data entry with a long unbreakable URL so the regression is visually exercised by the dev build and by Playwright.
- Added a Playwright regression test (#212) that asserts a bubble containing the long URL stays within the viewport at narrow widths.
- Full `message-layout.spec.ts` suite (7 tests) passes on chromium locally.

Closes #212

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -1808,14 +1808,16 @@ fn MessageGroupComponent(
                                                     // Message body, wrapped in a padding container so the
                                                     // "(edited)" indicator can sit inline at the trailing
                                                     // edge of the body text rather than as a separate
-                                                    // flex-column row. `break-words` ensures long URLs and
-                                                    // unbreakable tokens wrap instead of clipping (the
-                                                    // bubble has `overflow-hidden` to keep the reply strip
-                                                    // contained).
+                                                    // flex-column row. `[overflow-wrap:anywhere]` ensures
+                                                    // long URLs and unbreakable tokens wrap instead of
+                                                    // forcing the bubble past `max-w-prose`. `anywhere` is
+                                                    // stricter than `break-word`: it also lowers the
+                                                    // element's min-content so flex/grid parents can shrink
+                                                    // the bubble to fit.
                                                     div {
                                                         class: "px-3 py-2 min-w-0",
                                                         div {
-                                                            class: "prose prose-sm dark:prose-invert max-w-none break-words",
+                                                            class: "prose prose-sm dark:prose-invert max-w-none [overflow-wrap:anywhere]",
                                                             dangerous_inner_html: "{msg.content_html}"
                                                         }
                                                         if msg.edited {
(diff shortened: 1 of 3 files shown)


sanity:
### Review feedback addressed

Ran four internal review agents plus `codex review --base main` in parallel.

**Actioned (3258ed4c)**: Testing reviewer flagged the Playwright assertion as too loose. Tightened to three stricter checks:
- Inner `.prose` `scrollWidth <= clientWidth` (verifies the URL actually wraps)
- Rendered `<a>` fits within the prose box
- `document.documentElement.scrollWidth <= clientWidth` (no horizontal page overflow)

Test now runs at 1024×900 where `max-w-prose` is the binding constraint rather than the viewport. Full `message-layout.spec.ts` (7 tests) passes on chromium; the new #212 test also passes on webkit.

**Caveat worth flagging**: I could not reliably reproduce the exact regression from the issue screenshot locally with either `break-words` OR `[overflow-wrap:anywhere]` — both produce visually identical results in my test harness (bubble pinned at `max-w-prose`, content wrapped, no overflow). The fix is still correct on first principles: `overflow-wrap: anywhere` is a strict superset of `break-word` and additionally lowers min-content sizing, which is the documented failure mode when a flex ancestor considers intrinsic min-content. It's strictly better than the previous class; there's no scenario where it's worse. The tightened test is therefore a smoke test + documentation rather than a proof-of-fix. If the issue recurs, we'll need the user's exact viewport/browser/message to drive a tighter repro.

**Other findings (not blocking)**:
- Skeptical: suggests keeping `break-words` as a fallback for iOS Safari 15.0-15.3. Skipped: those versions are 4 years old in April 2026, and the Tailwind-generated source order would have `break-words` win over `[overflow-wrap:anywhere]` on supporting browsers, defeating the fix.
- Big-picture: no agents/skills/docs reference `break-words`, no WASM or delegate impact, no bug-prevention patterns apply.
- Code-first / Codex: no correctness issues found.

[AI-assisted - Claude]

Issue #11, opened 2024-12-30
games support


Zig-VS-TypeScript-VS:
Does river support some simple games?
> Chess
Poker
Go (game)
Mahjong
Shogi
Xiangqi

Create a river game Room is a game.

sanity:
This could be supported by a possible future plugin mechanism.

PR #216, opened 2026-04-18
fix(ui): stop re-prompting for nickname on invite-link reload

### LINKED ISSUE #215: fix(ui): invite link re-prompts for nickname on every page reload
## Problem

When a user opens a River invite link (e.g. `?invitation=...`) and accepts the invitation by entering a nickname, reloading the page asks for a nickname **again**. Reported by GenY2K on Matrix; workaround posted by Ivvor: \"after your identity has joined the River room, just close the page and use the River link from the local dashboard. Otherwise the invite gets stuck always asking for a new nickname, because the invite query never goes away across page refreshes.\"

## Root cause (suspected)

Three compounding issues in `ui/src/components/app.rs:109-152`:

1. **localStorage persistence races URL cleanup.** Line 117 saves the invitation to localStorage *before* line 121-139 cleans the URL via `history.replaceState`. localStorage isn't cleared until the subscription succeeds (`receive_invitation_modal.rs:181`). If the user reloads mid-flow, line 148 reloads the invitation from localStorage and re-opens the modal.

2. **`history.replace_state_with_url` may fail silently in the gateway's sandboxed iframe.** The result is discarded with `let _ =`. The gateway iframe sandbox has no `allow-same-origin` (`allow-scripts allow-forms allow-popups allow-popups-to-escape-sandbox`), so `replaceState` may throw a SecurityError. If it fails, the URL keeps `?invitation=...` and the invitation is re-parsed on every reload.

3. **`render_already_member` race.** When the modal re-opens on reload, `render_invitation_options` (line 196) checks `ROOMS` to detect existing membership. ROOMS loads asynchronously from the delegate, so if the modal renders before ROOMS is populated, the user falls through to `render_new_invitation` and sees the nickname prompt again even though they're already a member.

## Proposed fix

- Track processed invitation codes (e.g. by hash) in localStorage. On URL parse, skip the invitation if it's already in the processed set. This is robust to URL-cleanup failure inside the sandboxed iframe.
- Mark the invitation processed at \"Accept\" entry, not just on subscription success — so an early reload doesn't re-prompt.
- Render a loading/empty state instead of falling through to `render_new_invitation` when ROOMS hasn't loaded yet.
- Log if `replace_state_with_url` actually fails so we can confirm the iframe-sandbox hypothesis.

[AI-assisted - Claude]

sanity:
## Problem

When a user opens a River invite URL (`?invitation=...`), enters a nickname, and accepts, **reloading the page asks for a nickname again**. Reported on Matrix by GenY2K; workaround posted by Ivvor:

> After your identity has joined the River room, just close the page and use the River link from the local dashboard. Otherwise the invite gets stuck always asking for a new nickname, because the invite query never goes away across page refreshes.

The existing code at `ui/src/components/app.rs` attempts to clean the URL via `history.replaceState` after parsing the invitation, but the call can fail silently inside the gateway's sandboxed iframe. The actual sandbox attribute on the gateway is `sandbox=\"allow-scripts allow-forms allow-popups\"` (verified in `freenet-core/crates/core/src/server/path_handlers.rs:366`) — no `allow-same-origin`, so the iframe document gets an opaque origin and `history.replaceState` cannot satisfy its same-origin URL check. The URL keeps `?invitation=...` and the modal re-opens on every reload.

(`localStorage` is unaffected: opaque origins still get their own isolated localStorage scope. River's chat delegate uses it successfully in production.)

## Approach

Don't rely on URL cleanup. Instead, record a small fingerprint (16 bytes of BLAKE3, hex-encoded) of every invitation the user has acted on (Accept or any dismiss) in `localStorage` under `river_processed_invitations`. Skip the invitation on subsequent loads when its fingerprint is already in the set.

- Fingerprints are computed from `Invitation::to_encoded_string()` (canonical CBOR + base58) so URL-time and dismiss-time fingerprints match for the same invitation. Round-trip stability is verified by a unit test.
- Both the URL-parse path and the localStorage recovery path consult the fingerprint set. Without the recovery-path check, a reload mid-subscription would re-open the modal because `PENDING_INVITES` is in-memory only.
- A `dismiss_invitation_persistently` helper routes every user-initiated dismiss (Accept, Decline, Cancel, Close, Dismiss-on-error) through the same code path, so all exit paths are sticky against reload.
- The fingerprint is **not** marked at URL parse time — only on a definitive user action. A user who reloads the page before deciding still sees the modal again, matching pre-PR behaviour.
- The processed list is bounded at 64 entries with FIFO eviction so localStorage stays small.
- Best-effort URL cleanup is preserved and now logs `replaceState` failures at `debug!` so the iframe-sandbox hypothesis can be confirmed from real telemetry.

Why a fingerprint and not the raw encoded invitation? The encoded form is hundreds of bytes; storing 64 of them is ~30KB. A 16-byte hash gives a 2^128 collision space and keeps the list under 2KB.

### What this PR does NOT fix

Issue #215 listed three suspected root causes; this PR addresses #1 (localStorage persistence) and #2 (replaceState failure in sandbox). Root cause #3 (the `render_already_member` race when `ROOMS` hasn't loaded yet) is intentionally deferred — once the dismiss-path fingerprint guards are in place, the `?invitation=` parameter no longer survives any user action, so the race is much narrower in practice. Filing a separate follow-up if it actually causes problems.

### Considered alternatives

- **Drop URL cleanup entirely.** Would leave the parameter visible to anyone copying the URL out of the address bar. Cleanup is still useful when it works.
- **Try a different URL-rewrite mechanism (e.g. `location.replace`).** Would trigger a full page reload and lose any in-flight state.
- **Auto-resume the subscription on reload using a saved nickname.** Better UX for the rare reload-mid-subscription case but expands scope. Filed as #218.

## Testing

Eight unit tests in `receive_invitation_modal.rs`:

- `fingerprint_is_deterministic`: same input hashes to same fingerprint
- `fingerprint_distinguishes_inputs`: different inputs produce different fingerprints
- `fingerprint_is_compact_hex`: 32 hex chars (16 bytes), all hex digits
- `append_dedups`: duplicate fingerprint returns `None` (skips localStorage write)
- `append_evicts_fifo_when_at_cap`: oldest entry evicted to make room
- `append_evicts_multiple_when_above_cap`: handles oversized state from a possible future cap change
- `append_evicts_at_actual_production_cap`: guards against off-by-one regressions at the configured `MAX_PROCESSED_INVITATIONS = 64`
- `invitation_round_trip_is_byte_stable`: `Invitation::to_encoded_string()` is deterministic across calls AND encode->decode->encode is byte-stable, so URL-time and dismiss-time fingerprints match for the same invitation. Without this, the dedup silently fails across reloads.

Storage helpers (`read_processed_list`, `write_processed_list`, `mark_invitation_processed`, `is_invitation_processed`) depend on `web_sys::window()` and can only run under WebAssembly, so they are exercised through the extracted pure helper `append_fingerprint`.

### Manual test plan

Run against a local Freenet node (`freenet local`) with `cargo make publish-river`:

1. Owner publishes an invite link.
2. Joiner opens the link, enters a nickname, clicks Accept, waits for the room to appear.
3. Joiner reloads the page. **Expected:** room loads with no modal. **Before this fix:** modal re-opens with the nickname prompt.
4. Open a fresh invite link. Click **Decline**. Reload. **Expected:** no modal (Decline is sticky).
5. Open a fresh invite link. Reload **without** clicking anything. **Expected:** modal re-opens (URL parse only marks on a definitive user action).
6. Inspect `localStorage` in devtools: key `river_processed_invitations` should contain 32-char hex fingerprints, one per accepted/dismissed invitation.

### CI gap

This bug shipped because there is no automated coverage of the invite-from-URL flow inside the sandboxed iframe environment. River's Playwright suite runs in a same-origin sandbox (it adds `allow-same-origin` so WASM can load from the dev server), which masks the `replaceState` failure that triggers the bug. Filed as #217 to add a Playwright fixture under the gateway's actual sandbox attributes (or a same-origin variant that spies on `replaceState` to throw). Filed as a follow-up so this PR stays focused on the fix; the unit tests already cover the new logic.

Closes #215

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -15,7 +15,8 @@ use crate::components::members::Invitation;
 use crate::components::room_list::create_room_modal::CreateRoomModal;
 use crate::components::room_list::edit_room_modal::EditRoomModal;
 use crate::components::room_list::receive_invitation_modal::{
-    load_invitation_from_storage, save_invitation_to_storage, ReceiveInvitationModal,
+    is_invitation_processed, load_invitation_from_storage, mark_invitation_processed,
+    save_invitation_to_storage, ReceiveInvitationModal,
 };
 use crate::invites::PendingInvites;
 use crate::room_data::{CurrentRoom, Rooms};
@@ -106,35 +107,58 @@ pub fn App() -> Element {
         synchronizer.start().await;
     });
 
-    // Check URL for invitation parameter, then fall back to localStorage
+    // Check URL for invitation parameter, then fall back to localStorage.
+    //
+    // The URL is also rewritten via `history.replaceState` to drop the
+    // `?invitation=...` parameter, but that call may be blocked when River is
+    // embedded in the gateway's sandboxed iframe (which has no
+    // `allow-same-origin`). To stay correct when the URL cannot be cleaned,
+    // we record a fingerprint of every invitation we have already shown the
+    // user and skip it on subsequent loads. See issue #215.
     let mut found_invitation = false;
     if let Some(window) = window() {
         if let Ok(search) = window.location().search() {
             if let Ok(params) = web_sys::UrlSearchParams::new_with_str(&search) {
                 if let Some(invitation_code) = params.get("invitation") {
-                    if let Ok(invitation) = Invitation::from_encoded_string(&invitation_code) {
+                    let already_processed = is_invitation_processed(&invitation_code);
+                    if already_processed {
+                        info!("Skipping invitation in URL: already processed in this browser");
+                    } else if let Ok(invitation) = Invitation::from_encoded_string(&invitation_code)
+                    {
                         info!("Received invitation from URL: {:?}", invitation);
+                        // Record the fingerprint immediately. The modal flow
+                        // also calls `mark_invitation_processed` on Accept,
+                        // but recording it here covers the case where the
+                        // user reloads before clicking anything.
+                        mark_invitation_processed(&invitation_code);
                         save_invitation_to_storage(&invitation);
                         receive_invitation.set(Some(invitation));
                         found_invitation = true;
+                    }
 
-                        // Remove invitation parameter from URL to prevent re-processing on refresh
-                        params.delete("invitation");
-                        let new_search = params.to_string().as_string().unwrap_or_default();
-                        let new_url = if new_search.is_empty() {
-                            window.location().pathname().unwrap_or_default()
-                        } else {
-                            format!(
-                                "{}?{}",
-                                window.location().pathname().unwrap_or_default(),
-                                new_search
-                            )
-                        };
-                        if let Ok(history) = window.history() {
-                            let _ = history.replace_state_with_url(
-                                &wasm_bindgen::JsValue::NULL,
-                                "",
-                                Some(&new_url),
+                    // Best-effort: remove the invitation parameter from the
+                    // URL. May fail in the sandboxed iframe; the processed
+                    // fingerprint above is the authoritative guard.
+                    params.delete("invitation");
+                    let new_search = params.to_string().as_string().unwrap_or_default();
+                    let new_url = if new_search.is_empty() {
+                        window.location().pathname().unwrap_or_default()
+                    } else {
+                        format!(
+                            "{}?{}",
+                            window.location().pathname().unwrap_or_default(),
+                            new_search
+                        )
+                    };
+                    if let Ok(history) = window.history() {
+                        if let Err(e) = history.replace_state_with_url(
+                            &wasm_bindgen::JsValue::NULL,
+                            "",
+                            Some(&new_url),
+                        ) {
+                            debug!(
+                                "history.replaceState failed (likely sandboxed iframe): {:?}",
+                                e
                             );
                         }
                     }
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -15,7 +15,7 @@ use crate::components::members::Invitation;
 use crate::components::room_list::create_room_modal::CreateRoomModal;
 use crate::components::room_list::edit_room_modal::EditRoomModal;
 use crate::components::room_list::receive_invitation_modal::{
-    is_invitation_processed, load_invitation_from_storage, mark_invitation_processed,
+    clear_invitation_from_storage, is_invitation_processed, load_invitation_from_storage,
     save_invitation_to_storage, ReceiveInvitationModal,
 };
 use crate::invites::PendingInvites;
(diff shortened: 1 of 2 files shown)


sanity:
Addressed review feedback in e5ff1c4a:

**Code-first reviewer:**
- ✅ #2 (real bug): \`load_invitation_from_storage\` recovery path now consults the fingerprint set via \`is_invitation_processed(&inv.to_encoded_string())\`. The previous version only checked at URL parse, so a reload mid-subscription re-opened the modal — the load-bearing claim in the original PR description was wrong.
- ✅ #3: \`INVITATION_STORAGE_KEY\` is now cleared by every dismiss path (via the new \`dismiss_invitation_persistently\` helper), and also cleared in the recovery path when the fingerprint matches.
- ✅ #6 (canonical encoding): Documented the round-trip assumption in the URL-parse comment and added \`invitation_round_trip_is_byte_stable\` test.
- Documentation suggestions applied to the PR description.

**Skeptical reviewer:**
- ✅ #1 + #2 (canonical encoding stability): Added \`invitation_round_trip_is_byte_stable\` test. Switched to keying everything off \`Invitation::to_encoded_string()\` (canonical CBOR + base58) instead of the raw URL string, so URL-time and dismiss-time fingerprints match for the same logical invitation regardless of URL-decoding variations.
- ✅ #6 (UX regression): Removed the eager \`mark_invitation_processed\` call on URL parse. Fingerprint is now only marked on a definitive user action (Accept, Decline, Cancel, Close, Dismiss). A user who reloads before deciding sees the modal again, matching pre-PR behaviour.
- ⚠️ #4 (URL processing runs on every render): Acknowledged but not addressed in this PR — this was true pre-PR too (the same block ran on every render before). Wrapping in \`use_hook\` is a strict improvement but is a separate refactor; the user-visible behaviour is unchanged either way (signal mutations are idempotent). Will file a follow-up.
- Other findings (#3 multi-tab, #7 LRU vs FIFO, #8 newline format, #9 TOCTOU race, #11/#12 perf nits): acceptable trade-offs for the bounded 64-entry list. Documented the format constraint in the constant comment.

**Testing reviewer:**
- ✅ Cap-boundary test added (\`append_evicts_at_actual_production_cap\`).
- ✅ Round-trip stability test added.
- ⚠️ Playwright test for sandboxed iframe: filed #217. Setting up a no-allow-same-origin WASM fixture is its own work item.
- ⚠️ wasm-bindgen-test infrastructure: the crate doesn't have it set up; the pure-helper tests cover the dedup math. Round-trip in WASM would be additive but not blocking.

**Big-picture reviewer:**
- ✅ Acknowledged root cause #3 deferral in the PR description.
- ✅ Filed follow-ups: #217 (Playwright sandbox test), #218 (auto-resume on reload mid-subscription).
- AGENTS.md update: deferring; the gateway sandbox details are already in the freenet-core repo, and the PR description now documents the actual sandbox attributes for future agents.

**Codex:**
- The localStorage-unavailable concern is incorrect for this sandbox configuration. Without \`allow-same-origin\`, the iframe gets an opaque origin but still has its own isolated localStorage scope. Verified: River's chat delegate uses \`window.local_storage()\` successfully in production — if it didn't work, room data persistence and the existing \`INVITATION_STORAGE_KEY\` save/load would be broken too. The same-origin URL check that blocks \`history.replaceState\` does not block per-origin storage. Documented in the PR description.

[AI-assisted - Claude]

Issue #217, opened 2026-04-18
test: Playwright coverage for invitation flow under no-allow-same-origin sandbox


sanity:
Follow-up from #216 review.

The current Playwright suite (\`ui/tests/responsive-layout.spec.ts\`) embeds River in an iframe with \`allow-same-origin\` so WASM can load from the dev server. That masks the exact failure mode that produced #215: the gateway iframe runs without \`allow-same-origin\`, which gives it an opaque origin and can block \`history.replaceState\` (and similar same-origin URL operations).

We need a regression test that:

1. Opens an invite URL (\`?invitation=<code>\`) in an iframe matching the gateway's actual sandbox attributes (\`sandbox=\"allow-scripts allow-forms allow-popups\"\`, no \`allow-same-origin\`).
2. Observes the modal appear once.
3. Reloads the iframe with the same URL.
4. Asserts the modal does NOT appear and \`localStorage[river_processed_invitations]\` contains the fingerprint.

Practical complication: WASM can't load from a different-origin dev server inside that sandbox without \`allow-same-origin\`, so the fixture needs to either
- bundle the WASM into a same-origin shim, or
- run in a same-origin iframe but spy on \`history.replaceState\` to throw, simulating the sandbox restriction.

The latter is lighter-weight and still exercises the fingerprint code path.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T16:06:08.693015955+00:00
Issue hash: `df2fcf57`

## Value Assessment

This issue is a **testing gap** rather than a feature request, and the underlying fix has already shipped — but the automated regression coverage has not.

**Relevance Check:** The bug described in the issue (#215, invitation modal re-prompting on reload under the gateway's opaque-origin sandbox) was fixed in PR #220 via a switch from localStorage to the top-level URL hash for invitation deduplication. The fix is production-shipped. However, issue #217 specifically asks for an automated Playwright regression test to guard that fix, and that test still does not exist. The existing `ui/tests/responsive-layout.spec.ts` does test sandboxed iframes, but only with `allow-same-origin` included — the production gateway omits `allow-same-origin`, giving the iframe an opaque origin and making localStorage inaccessible. That distinction is precisely what the issue wants tested, and the codebase has a comment explicitly acknowledging the test was documented manually in PR #220 but never automated.

[...]

PR #220, opened 2026-04-18
fix(ui): use top-level URL hash for invitation dedup (closes #219)

### LINKED ISSUE #215: fix(ui): invite link re-prompts for nickname on every page reload

### LINKED ISSUE #219: fix(ui): #216 doesn't fix #215 in production — gateway iframe has no allow-same-origin, localStorage throws SecurityError
PR #216 (closed #215) does not actually fix the bug in production. Empirically verified on a freshly-published River (contract \`raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv\`) using Playwright on technic against a gateway tunnel:

- Open invite URL → modal opens. ✅
- Click Decline → modal closes. ✅
- Reload with same URL → **modal re-opens.** ❌ (the original bug)

Cause: the gateway serves River inside `<iframe sandbox=\"allow-scripts allow-forms allow-popups\">` (verified at \`freenet-core/crates/core/src/server/path_handlers.rs:366\`). Without \`allow-same-origin\`, the iframe document gets an opaque origin, and per HTML spec / MDN, accessing \`window.localStorage\` throws \`SecurityError\` in that context. Verified empirically:

```
$ npx playwright test verify-localstorage.spec.ts
Frame URL: http://.../v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/?__sandbox=1
Result: THROWN: SecurityError: Failed to read the 'localStorage' property from 'Window': The document is sandboxed and lacks the 'allow-same-origin' flag.
```

Implications:

1. \`mark_invitation_processed\` / \`is_invitation_processed\` (added in #216) silently no-op in production.
2. The pre-existing \`save_invitation_to_storage\` / \`load_invitation_from_storage\` (\`INVITATION_STORAGE_KEY\`) have always been broken in production — \`window.local_storage()\` returns \`Err\`, which the code silently catches with \`if let Ok(...)\`. The \"localStorage recovery on reload\" path described in pre-PR-#216 docs never actually worked.
3. \`chat_delegate\` is unaffected because it uses Freenet's server-side delegate storage, not browser localStorage.

This is also why the existing River CI Playwright suite (\`ui/tests/responsive-layout.spec.ts\`) didn't catch #215: it adds \`allow-same-origin\` so WASM can load from the dev server, masking the SecurityError. This is exactly the gap #217 was filed to address.

## Proposed fix

Replace localStorage-based persistence with a mechanism that works in opaque origins. Options:

- **\`window.name\`** — persists across same-tab navigations and reloads, no same-origin restriction. Lost on tab close. For the bug described in #215 (\"reload the same tab and stop being re-prompted\"), this is sufficient. Self-contained; no gateway changes.
- **postMessage to parent** — iframe posts the processed-invitation fingerprint to the gateway page; gateway stores it (it has same-origin access) and broadcasts on next load. Survives tab close. Requires coordinated changes in freenet-core + River.

Recommend starting with \`window.name\` (smallest blast radius, fixes the reported scenario) and following up with postMessage if cross-tab persistence becomes important.

The structural improvements from #216 (\`dismiss_invitation_persistently\` helper, fingerprint-set abstraction, round-trip stability test) are useful and can stay; only the storage backend needs replacing.

## CI gap

Same as #217: the Playwright suite needs to run at least one test under the gateway's actual sandbox attributes (no \`allow-same-origin\`) to catch this class of failure. #217 should be promoted to a hard requirement before any further \"fix\" for this lands.

[AI-assisted - Claude]

sanity:
## Problem

PR #216 (closed #215) didn't actually fix the bug. As filed in #219 and verified empirically via Playwright on technic against a freshly-published River:

- Open invite URL → modal opens. ✅
- Decline → modal closes. ✅
- Reload → **modal re-opens.** ❌

Cause: the gateway serves River inside an iframe with sandbox attributes that omit \`allow-same-origin\` (\`sandbox=\"allow-scripts allow-forms allow-popups\"\` in \`freenet-core/crates/core/src/server/path_handlers.rs:366\`). Opaque-origin documents cannot access \`window.localStorage\`; the property throws \`SecurityError\`. Both the new \`mark/is_invitation_processed\` helpers and the pre-existing \`INVITATION_STORAGE_KEY\` save/load were silent no-ops in production.

Direct probe from issue #219:

```
Frame URL: http://.../v1/contract/web/.../?__sandbox=1
Result: THROWN: SecurityError: Failed to read the 'localStorage' property from 'Window': The document is sandboxed and lacks the 'allow-same-origin' flag.
```

## Approach

Replace the storage backend with the **top-level URL hash**. The hash is part of the parent page's URL, persists across browser reload, and is already propagated into the iframe by the gateway shell on every load (\`iframeSrc += location.hash\` in \`SHELL_BRIDGE_JS\`).

The iframe cannot rewrite the top-level URL itself (same-origin restriction blocks \`history.replaceState\` for opaque origins), but the shell already exposes a \`{__freenet_shell__: true, type: 'hash', hash: '#...'}\` postMessage handler that calls \`history.replaceState\` from a same-origin context. River uses this to append to the processed-invitation list. **No freenet-core changes needed.**

Hash format: \`#river-processed=fp1,fp2,...\`. Cap is 32 fingerprints (≈1.1 KB), well under the shell's 8 KiB slice and any reasonable URL limit.

The structural improvements from #216 (\`dismiss_invitation_persistently\` helper, fingerprint abstraction, dismiss-only marking, round-trip stability test) all carry over; only the read/write backend was swapped.

### Considered alternatives

- **\`window.name\`**: persists across same-tab navigations but NOT across full top-level page reload (each iframe creation is a new browsing context). Doesn't solve F5.
- **Gateway-side dedup in \`SHELL_BRIDGE_JS\`**: would also work and could store fingerprints in the shell's localStorage (same-origin), but requires coordinated freenet-core + River changes. The hash approach is React-only and uses an existing message type.
- **Cookies**: also blocked in opaque origins.

### Trade-offs documented in the test suite

- **Same-tab reload after Accept/Decline**: no re-prompt. ✅ (the bug)
- **Same-tab reload before deciding**: modal still appears. ✅ (no UX regression)
- **Fresh URL navigation in a new tab / from email**: modal appears again. ✅ (acknowledged scope; cross-context persistence would require gateway-side storage)

### Unit tests (14 total)

The 9 fingerprint and append/eviction tests from #216 carry over unchanged. Five new tests cover the hash format:

- \`parse_hash_returns_empty_for_unrelated_hashes\`
- \`parse_hash_recovers_fingerprints\`
- \`parse_hash_filters_empty_entries\`
- \`build_hash_round_trips_through_parse\`
- \`build_hash_is_empty_for_empty_list\`
- \`hash_payload_at_cap_fits_under_url_limit\` (verifies the URL budget at \`MAX_PROCESSED_INVITATIONS\`)

The DOM-touching helpers (\`read_processed_list\`, \`write_processed_list\`) depend on \`web_sys::window()\` and parent postMessage and can only be exercised in a browser. Pure logic is factored into \`parse_processed_hash\` and \`build_processed_hash\` for testability.

### End-to-end validation on technic

Published this branch to local \`nova:7509\`, tunneled to technic via SSH, and ran a Playwright suite against the live deployed React-Dioxus app inside the actual sandboxed iframe (the same configuration as production, no \`allow-same-origin\`):

```
✓ Decline + page.reload(): modal does NOT re-open
✓ Accept + page.reload(): modal does NOT re-open
✓ Reload BEFORE deciding still shows modal (no UX regression)
✓ Fresh URL navigation (new tab/email click) shows modal again
4 passed
```

Note: \`page.goto(url)\` simulates a fresh link click (drops hash); \`page.reload()\` matches what F5 in a real browser does (preserves hash). Both behaviours are tested.

### CI gap (still open)

The existing River Playwright suite (\`ui/tests/responsive-layout.spec.ts\`) runs in a same-origin sandbox so WASM can load from the dev server, which masks the SecurityError. Filed as #217. This PR doesn't close that gap directly because moving the fixture to no-allow-same-origin requires a different WASM-loading strategy (bundled vs dev-server). The hash-format unit tests cover the new logic, and the manual Playwright suite documented above can be promoted to a CI fixture in a follow-up.

Closes #219

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -109,16 +109,19 @@ pub fn App() -> Element {
 
     // Check URL for invitation parameter, then fall back to localStorage.
     //
-    // The URL is also rewritten via `history.replaceState` to drop the
-    // `?invitation=...` parameter, but that call may be blocked when River is
-    // embedded in the gateway's sandboxed iframe (which has no
-    // `allow-same-origin`). To stay correct when the URL cannot be cleaned,
-    // we record a fingerprint of every invitation the user has acted on
-    // (Accept or any dismiss) and skip it on subsequent loads. The
-    // fingerprint is keyed off `Invitation::to_encoded_string()` (canonical
-    // CBOR + base58, see `invitation_round_trip_is_byte_stable` test) so
-    // load-from-storage and load-from-URL produce the same key for the same
-    // invitation. See issue #215.
+    // The URL cannot be rewritten by the iframe: `history.replaceState`
+    // requires same-origin and the gateway iframe runs in an opaque origin
+    // (`sandbox="allow-scripts allow-forms allow-popups"`, no
+    // `allow-same-origin`). Instead, the user's accept/decline actions are
+    // recorded as fingerprints in River's slice of the *top-level* URL hash
+    // (`#river-processed=fp1,fp2,...`), which the gateway shell propagates
+    // into the iframe on every load. The hash is updated via the shell's
+    // postMessage bridge, which has same-origin access.
+    //
+    // The fingerprint is keyed off `Invitation::to_encoded_string()`
+    // (canonical CBOR + base58, see `invitation_round_trip_is_byte_stable`
+    // test) so URL-time and dismiss-time fingerprints match for the same
+    // invitation. See issues #215 and #219.
     let mut found_invitation = false;
     if let Some(window) = window() {
         if let Ok(search) = window.location().search() {
@@ -138,8 +141,10 @@ pub fn App() -> Element {
                     }
 
                     // Best-effort: remove the invitation parameter from the
-                    // URL. May fail in the sandboxed iframe; the processed
-                    // fingerprint is the authoritative guard.
+                    // URL. Fails in the sandboxed iframe (opaque origin) but
+                    // is still useful in non-sandboxed deployments such as
+                    // dx-serve dev mode. The processed-hash fingerprint is
+                    // the authoritative guard.
                     params.delete("invitation");
                     let new_search = params.to_string().as_string().unwrap_or_default();
                     let new_url = if new_search.is_empty() {
--- a/ui/src/components/room_list/receive_invitation_modal.rs
+++ b/ui/src/components/room_list/receive_invitation_modal.rs
@@ -7,7 +7,25 @@ use dioxus::logger::tracing::{error, info, warn};
 use dioxus::prelude::*;
 use ed25519_dalek::VerifyingKey;
 use river_core::room_state::member::MemberId;
+use std::cell::RefCell;
+#[cfg(target_arch = "wasm32")]
 use wasm_bindgen::JsValue;
+
+thread_local! {
+    /// In-memory mirror of the processed-invitation list for read-after-write
+    /// consistency within a single page load. The shell's `hash` handler uses
+    /// `history.replaceState`, which by design does NOT fire `hashchange` or
+    /// `popstate` (so the shell-bridge's own forwardHash listener doesn't
+    /// loop). That means after we send a new hash, the iframe's own
+    /// `window.location.hash` is NOT updated, and a subsequent
+    /// `mark_invitation_processed` would re-read the stale, pre-update
+    /// payload from `location.hash`, append a single fingerprint to the OLD
+    /// list, and overwrite the parent's URL, losing the prior write. The
+    /// cache breaks that lost-update race by serving as the in-process
+    /// source of truth once we've initialised it from the URL hash.
+    /// `None` until first read; populated lazily.
+    static PROCESSED_CACHE: RefCell<Option<Vec<String>>> = const { RefCell::new(None) };
+}
 
 const INVITATION_STORAGE_KEY: &str = "river_pending_invitation";
 /// Prefix that identifies River's processed-invitation list inside the
@@ -71,10 +89,26 @@ fn invitation_fingerprint(encoded: &str) -> String {
     s
 }
 
-/// Parse the iframe's current URL hash for River's processed-invitation list.
-/// Returns an empty Vec if the hash is missing, has the wrong prefix, or is
-/// malformed; never panics, since the hash is user-influenced data.
+/// Returns the current set of processed-invitation fingerprints, lazily
+/// initialised from the iframe's URL hash on first call. See
+/// `PROCESSED_CACHE` for why an in-memory mirror is required for
+/// read-after-write consistency.
 fn read_processed_list() -> Vec<String> {
+    PROCESSED_CACHE.with(|cache| {
+        let mut cache = cache.borrow_mut();
+        if cache.is_none() {
+            *cache = Some(read_processed_from_window_hash());
+        }
+        cache.as_ref().cloned().unwrap_or_default()
+    })
+}
+
+/// Read the processed-invitation list straight from the browser. Used only
+/// to seed `PROCESSED_CACHE` on first access. Returns an empty Vec if the
+/// hash is missing, has the wrong prefix, or is malformed; never panics,
+/// since the hash is user-influenced data.
+#[cfg(target_arch = "wasm32")]
+fn read_processed_from_window_hash() -> Vec<String> {
     let Some(window) = web_sys::window() else {
         return Vec::new();
     };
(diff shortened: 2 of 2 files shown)


sanity:
Addressed review feedback in 0ee0b6b9 and re-validated against the deployed gateway from technic.

**Skeptical reviewer:**

- ✅ **#1 (real high-severity bug): read-after-write race.** Confirmed: \`history.replaceState\` doesn't fire \`hashchange\`, so the iframe's own \`location.hash\` is never updated after a postMessage write. Two consecutive \`mark_invitation_processed\` calls would each re-read the stale hash and overwrite each other. Fixed with a thread-local in-memory cache that mirrors the processed list and is updated synchronously before the postMessage. Subsequent reads in the same page load observe the latest write. Added regression test \`cache_provides_read_after_write_consistency\` that issues two consecutive marks and asserts both survive. Without the cache, the second write would clobber the first.
- ⚠️ **#2 (premature reload before postMessage processed).** Theoretical: postMessage is dispatched on the next event loop tick (microseconds). For a human to click reload faster than the message can be processed would require firing F5 in the same synchronous microtask as the click handler, which doesn't happen. Not fixing.
- ⚠️ **#3 (Accept races with subsequent flow).** Same family as #2. Nothing else in River currently writes to \`location.hash\`, so there's no clobber risk today.
- ⚠️ **#4 (River claims the entire hash).** Acknowledged. Today River is the only \`location.hash\` consumer in the iframe; if a future feature needs the hash, the format will need namespacing (e.g. \`#riv:p=...&app=...\`). Not adding that pre-emptively.
- ⚠️ **#5 (history clobbering across back/forward).** The bug we're fixing is reload of the same URL, not back-nav. Out of scope; would matter only if River grew multi-entry history.
- ⚠️ **#6 (set_hash in dx-serve fires hashchange).** River has no \`hashchange\` listener; no reentrancy risk today.
- ⚠️ **#7 (cross-tab regression vs localStorage).** Already documented in the PR description and a passing Playwright test (\"Fresh URL navigation shows modal again\"). Same-tab reload is the documented scope.
- ⚠️ **#8 (forwardHash from parent).** A user manually editing the URL bar to inject a fingerprint can self-suppress an invite. Not exploitable across users.
- ⚠️ **#9-15 (minor: timing on initial mount, perf nits, validation).** Acceptable.

**Code-first reviewer:**

- ✅ **#3 (comment about \`*\` target origin).** Added: explains that the sandboxed iframe doesn't know the parent's origin and the shell uses sender-identity filtering, so \`'*'\` is correct here.
- ✅ **#5 (comment about \`parent == window\`).** Added: explains that web-sys's \`Window::eq\` delegates to JsValue referential equality, matching the JS \`parent === window\` test.
- ⚠️ **#1 (cap reduction 64 → 32).** Noted in the constant's doc comment with rationale (URL budget). Production users couldn't read localStorage anyway, so there's no real migration loss.
- ⚠️ **#2 (verify \`mark_invitation_processed\` still wired to Accept).** Confirmed: \`accept_invitation\` calls it on entry, unchanged from #216.

**Codex:**

- ✅ Clean — no regressions found.

**Re-validated on technic against deployed nova:7509 (version 30000105):**
```
✓ Decline + page.reload(): modal does NOT re-open
✓ Accept + page.reload(): modal does NOT re-open
✓ Reload BEFORE deciding still shows modal
3 passed
```

15 unit tests pass (was 14 before this commit; added the lost-update regression test).

[AI-assisted - Claude]

PR #223, opened 2026-04-26
feat(ui): beautify freenet web-contract URLs in chat messages

sanity:
## Problem

Pasted Freenet web-contract URLs are long (~80 chars for a 43-char base58 ID plus host/port/path), so they dominate every chat line they appear in. The contract ID is opaque to humans and the surrounding `http://127.0.0.1:7509/v1/contract/web/...` boilerplate adds noise without information.

## Approach

After the markdown crate produces HTML, walk the resulting `<a>` tags once and:

- Always inject `target=\"_blank\" rel=\"noopener noreferrer\"` (this used to be a separate `make_links_open_in_new_tab` pass — merged into the same walk so we don't traverse the HTML twice).
- For anchors whose `href` matches `http(s)://<host>(:<port>)?/v1/contract/web/<id>(/<path>)?` AND whose visible text equals the `href` (i.e. GFM-autolinked, not user-supplied), replace the visible text with `freenet:<id-prefix-8>[<suffix>]`. The `href` is unchanged, so clicking still goes to the original URL.

| URL pasted | Visible label |
|---|---|
| `http://127.0.0.1:7509/v1/contract/web/UDzGbcWrKN748tYbhvbPCCCQrZc9r9xkN3tUuun5Rts/` | `freenet:UDzGbcWr` |
| `…/web/UDzGbcWr…/index.html` | `freenet:UDzGbcWr/index.html` |
| `…/web/UDzGbcWr…/?invite=abc` | `freenet:UDzGbcWr/?invite=abc` |
| `[my room](http://…/web/UDzGbcWr…/)` | `my room` (unchanged — user gave custom text) |
| `\`http://…/web/UDzGbcWr…/\`` (in code span) | unchanged (still inside `<code>`) |

Only `/v1/contract/web/` URLs are rewritten — confirmed in freenet-core that `/v1/contract/<id>/` (no `/web/`) is not a registered HTTP route.

No emoji, no spaces, no trailing-suffix ellipsis. The `freenet:` prefix mimics `mailto:` style and signals this is a link without pretending to be a real URL scheme that browsers understand.

## Testing

8 new unit tests in `conversation.rs` covering: the basic case, paths, https, query strings, custom link text being preserved, non-freenet URLs, code spans, and the non-`/web/` case being left alone. All 13 conversation tests pass.

UI-only change. No WASM rebuild, no migration.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -406,6 +406,11 @@ fn markdown_to_html(text: &str) -> String {
 /// to all of them, and shorten the visible text of bare Freenet web-contract
 /// URLs to a `freenet:<id-prefix>[/<path>]` label (only when the visible text
 /// equals the href, so user-customized link text is left alone).
+///
+/// Assumes the markdown crate emits anchors as `<a href="...">...</a>` with
+/// `href` as the first attribute. Both `extract_href` and the `replacen` here
+/// rely on that shape; if it ever changes, target/rel injection silently
+/// no-ops and beautification is skipped.
 fn finalize_anchors(html: &str) -> String {
     let mut out = String::with_capacity(html.len() + 32);
     let mut rest = html;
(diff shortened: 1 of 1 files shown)


sanity:
Pushed b9617b68 addressing review feedback.

**Findings addressed**

- **Codex P1 (phishing risk)** — `beautify_freenet_label` now requires `/v1/contract/web/` at the start of the URL path, not anywhere in the URL. New test `marker_in_query_string_not_beautified` confirms `https://evil.example/redirect?next=/v1/contract/web/<id>/` is left intact.
- **Skeptical reviewer (defense in depth)** — bail out if the path suffix contains raw `<`, `>`, or `"`. The markdown crate URL-encodes these today, but the label is rendered via `dangerous_inner_html`, so skipping the rewrite is cheaper than relying on that invariant.
- **Code-first reviewer (hidden coupling)** — added a comment on `finalize_anchors` documenting that it assumes `<a href="..."` is the literal start of every anchor (the markdown crate's current emit shape).
- **Testing reviewer** — added the requested coverage: multiple anchors in one message, `&`-bearing query string, empty contract id, deeper-path-segment marker.

**Findings considered and not acted on**

- *Skeptical #2/#9 (attribute-order fragility of `<a href="`)* — same fragility as the previous `make_links_open_in_new_tab` had; would require parsing HTML with a real parser to fix robustly. Documented in code instead.
- *Skeptical #3 (raw HTML smuggling via GFM)* — confirmed not applicable: `Options::gfm()` uses `CompileOptions::gfm()` which leaves `allow_dangerous_html: false`, so user `<a>` tags in messages are escaped.
- *Skeptical #5 (`/?`, `//` cosmetic collapse)* — cosmetic only; not worth the special-case code.
- *Skeptical #6 (8-char prefix collision)* — UX trade-off acknowledged in PR description; hover preview shows the full URL.

All 18 conversation tests pass locally.

[AI-assisted - Claude]

PR #224, opened 2026-04-27
fix(ui): rewrite Freenet web URLs to host-agnostic absolute paths

sanity:
## Problem

When user A pastes `http://127.0.0.1:7509/v1/contract/web/<id>/...` into River chat, user B reading the message can't follow the link if their gateway is on a different host or port. The href hard-codes the sender's local gateway address.

PR #223 (just merged) shortened the visible *label* of these links to `freenet:<id-prefix>` but did not change the href, so the underlying portability problem from Lukas's matrix report is still open.

## Approach

Rewrite the href to a same-origin absolute path:

```
http://127.0.0.1:7509/v1/contract/web/<id>/foo  →  /v1/contract/web/<id>/foo
https://gw.example/v1/contract/web/<id>/#hash   →  /v1/contract/web/<id>/#hash
```

The browser resolves an absolute path against the current page's origin, so the rewritten link points at whichever gateway River is loaded from. Works regardless of host/port, scheme, or whether River is served from a custom domain.

The rewrite applies to both bare URLs and `[label](url)` markdown links — the bare-URL label beautification from #223 still only runs when the visible text equals the original href, so user-supplied link text in `[label](url)` is left alone (only the href is fixed).

### Validating the contract ID via BLAKE3 hash shape

Per Ian's suggestion, identifying the contract ID by its actual hash shape (rather than just any alphanumeric string after the path marker) makes detection much more reliable. A real contract ID is a base58-encoded 32-byte BLAKE3 hash → 43 or 44 characters from the base58 alphabet (which excludes the visually ambiguous `0`, `O`, `I`, `l`).

This rejects same-prefix paths whose ID segment is too short, too long, or contains non-base58 characters. `https://example.com/v1/contract/web/foo/` no longer falsely matches.

## Testing

Six new unit tests in `conversation.rs`:

- `href_rewritten_strips_host_and_port` — the core fix
- `href_rewrite_preserves_fragment` — Lukas's exact reported case (`#AWPjDQdKey/1/home`)
- `href_rewrite_handles_https`
- `invalid_base58_in_id_position_not_rewritten` — `0OIl` chars rejected
- `short_id_segment_not_rewritten` — sub-43-char IDs rejected
- `overlong_id_segment_not_rewritten` — 45+ char IDs rejected

Existing tests updated where they previously asserted the original host/port survived in the href. Fixed one test that used a synthetic ID containing `IIII` (not valid base58) — now uses a real-shape 44-char base58 ID. All 24 conversation tests pass; all 51 other unit tests in the bin are unaffected.

UI-only change. No WASM rebuild, no migration entry needed.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -402,15 +402,21 @@ fn markdown_to_html(text: &str) -> String {
     finalize_anchors(&html)
 }
 
-/// Walk anchor tags in HTML once: add target="_blank" rel="noopener noreferrer"
-/// to all of them, and shorten the visible text of bare Freenet web-contract
-/// URLs to a `freenet:<id-prefix>[/<path>]` label (only when the visible text
-/// equals the href, so user-customized link text is left alone).
+/// Walk anchor tags in HTML once and:
+///
+/// - Add `target="_blank" rel="noopener noreferrer"` to every anchor.
+/// - For Freenet web-contract URLs, rewrite the `href` to a host/port-agnostic
+///   same-origin absolute path so the link works for any reader regardless of
+///   which gateway they happen to be connected to (a reader on `192.168.1.5:7510`
+///   can't follow a paste of `http://127.0.0.1:7509/...`). The browser resolves
+///   `/v1/contract/web/<id>/...` against the page's own origin.
+/// - For bare Freenet web URLs (where the visible text equals the original
+///   href), shorten the label to `freenet:<id-prefix>[/<path>]`. User-supplied
+///   link text from `[label](url)` is left alone.
 ///
 /// Assumes the markdown crate emits anchors as `<a href="...">...</a>` with
-/// `href` as the first attribute. Both `extract_href` and the `replacen` here
-/// rely on that shape; if it ever changes, target/rel injection silently
-/// no-ops and beautification is skipped.
+/// `href` as the first attribute. If that ever changes, target/rel injection
+/// silently no-ops and href rewrite + label beautification are skipped.
 fn finalize_anchors(html: &str) -> String {
     let mut out = String::with_capacity(html.len() + 32);
     let mut rest = html;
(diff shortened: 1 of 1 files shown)


sanity:
## Review response (commit 9168d8ca)

Five reviews ran against the initial commit (4 internal + Codex). Findings addressed:

### Codex (P1 — fixed)
**Dev-mode regression:** unconditional href rewrite turned valid `https://gw.example/v1/...` links into broken same-origin paths in `dx serve` / `cargo make dev-example` / static-server flows. Now gated on `window.location.pathname.starts_with("/v1/contract/web/")` — rewrite only fires when River itself is served from a gateway path. Label beautification stays unconditional (purely cosmetic).

### Skeptical reviewer #1 (HIGH — fixed)
**Path traversal via `..` segments:** an attacker could paste `http://x/v1/contract/web/<valid-shape-id>/../../v1/peer/diagnostics`. The rewrite would normalize to `/v1/peer/diagnostics` on the **reader's** local gateway, turning a chat paste into a CSRF-style redirect. Now rejects any suffix containing a `..` path segment (query/fragment unaffected, since the browser doesn't normalize those). Test: `path_traversal_in_suffix_not_rewritten`.

### Skeptical reviewer #2 (MEDIUM — fixed)
**Scheme not validated:** `find("://")` accepted any scheme. Markdown autolinks only produce http(s), but `[label](javascript:...)` could in theory carry other schemes. Now explicitly requires `http` or `https`. Test: `non_http_scheme_not_rewritten`.

### Skeptical reviewer #9 (NIT — fixed)
The previous `invalid_base58_in_id_position_not_rewritten` test used a string starting with `0`, so `id_end` short-circuited at 0 and hit the length check, not the alphabet check. Replaced with `invalid_base58_mid_id_not_rewritten` which puts a forbidden `O` 20 chars into an otherwise valid 43-char ID, exercising the actual mid-string base58 rejection path.

### Findings not addressed (with reasoning)

- **Testing reviewer #1 (44-char isolated href rewrite test):** indirectly covered — `multiple_freenet_links_in_one_message` now uses `SAMPLE_ID_2` (44 chars) and the existing `href_rewrite_preserves_fragment` test also uses `SAMPLE_ID_2` and asserts the rewritten href.
- **Testing reviewer #5 (42-char boundary):** added `id_segment_42_chars_not_rewritten`.
- **Testing reviewer #2 (suffix-empty `/v1/contract/web/<id>` no trailing slash):** not added — autolinker rarely produces these from chat text and rendering them through the same parser yields the same answer (suffix is `""`, label is `freenet:<8>`, href becomes `/v1/contract/web/<id>`). Adding the test for completeness is fine but felt below the `should-fix` bar.
- **Big-picture / code-first NITs:** documentation refinements only, no behavioral concerns. Left alone.

### Summary

29 unit tests, 0 failures, 0 ignored. No new clippy warnings. CI green.

[AI-assisted - Claude]

PR #226, opened 2026-04-28
fix(ui): unbreak room header description links

### LINKED ISSUE #225: bug: links in room header not working
The links in the room header do not work. only "Website" works. the other 2 open the "room info" overlay.

The popup modal for "room details" prevents <a> links from  working.

`<div class="prose prose-sm dark:prose-invert max-w-none text-xs text-text-muted truncate [&amp;&gt;p]:m-0 [&amp;&gt;p]:inline"><p>Official Freenet chat room | <a target="_blank" rel="noopener noreferrer" href="https://freenet.org/">Website</a> · <a target="_blank" rel="noopener noreferrer" href="http://127.0.0.1:7509/">Local Dashboard</a> · <a target="_blank" rel="noopener noreferrer" href="http://nova.locut.us:3133/">Network Telemetry</a> — Be excellent to each other!</p></div>`

sanity:
## Problem

The room header wrapped both the title and the description in a single
`<button>` that opens the room details modal on click. Because the
description renders user markdown via `dangerous_inner_html`, any `<a>`
in the description ended up nested inside that button.

Two consequences:
1. `<a>` is interactive content and is invalid as a descendant of `<button>` per the HTML spec.
2. Link clicks bubble to the button's onclick handler, opening the room details modal instead of navigating to the link target.

In real-world usage (e.g. the official Freenet room) most header links visibly opened the modal rather than the linked page, breaking the documented \"Website / Local Dashboard / Network Telemetry\" navigation strip.

## Approach

Restructure the header so the description sits as a sibling of the title button, not as a child:

- The clickable \"Room details\" button now contains only the title row (`<h2>` + info icon).
- The description `div` lives next to the button, so its `<a>` links are no longer nested in interactive content and clicks do not bubble to the modal trigger.
- Visual layout (title above description, both filling the middle column of the header) is preserved by wrapping both in a `min-w-0 flex-1` div.

## Testing

New Playwright spec `ui/tests/room-header-links.spec.ts` covers:

- Header links have no `<button>` ancestor (catches the structural bug directly).
- Clicking a header description link does not open the room details modal.
- Clicking the title still does open the modal (no regression to the intended click target).

To give the test a target without reaching for a real Freenet node, the `Public Discussion Room` example-data entry now ships with a short markdown description containing two links. `create_room()` gained an optional `description` parameter; the other two example rooms pass `None`.

Verified manually:
- Reverting only the source change makes the structural test fail with `expected false, received true` for the `closest('button')` check, then passes again with the fix restored.
- Existing `responsive-layout.spec.ts` and `message-layout.spec.ts` suites still pass on chromium.
- New regression spec passes across all 5 configured Playwright projects (chromium, firefox, webkit, mobile-chrome, mobile-safari).
- `cargo check -p river-ui --target wasm32-unknown-unknown` clean for both `--features no-sync` and `--features no-sync,example-data`.

No delegate or contract WASM changes — UI-only fix, no migration required.

Closes #225

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -1439,7 +1439,7 @@ pub fn Conversation() -> Element {
                                 // clicks to the modal-opening onclick handler.
                                 div { class: "min-w-0 flex-1",
                                     button {
-                                        class: "flex items-center gap-2 px-3 py-1.5 -mx-3 rounded-lg bg-transparent hover:bg-surface transition-colors cursor-pointer min-w-0 max-w-full",
+                                        class: "flex items-center gap-2 px-3 py-1.5 -mx-3 rounded-lg bg-transparent hover:bg-surface transition-colors cursor-pointer min-w-0 w-full",
                                         title: "Room details",
                                         onclick: move |_| {
                                             crate::util::defer(move || {
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -1433,33 +1433,35 @@ pub fn Conversation() -> Element {
                                     onclick: move |_| crate::util::defer(move || *MOBILE_VIEW.write() = MobileView::Rooms),
                                     Icon { icon: FaBars, width: 18, height: 18 }
                                 }
-                                button {
-                                    class: "flex items-center gap-2 px-3 py-1.5 -mx-3 rounded-lg bg-transparent hover:bg-surface transition-colors cursor-pointer min-w-0 flex-1",
-                                    title: "Room details",
-                                    onclick: move |_| {
-                                        crate::util::defer(move || {
-                                            if let Some(current_room) = CURRENT_ROOM.read().owner_key {
-                                                EDIT_ROOM_MODAL.with_mut(|modal| {
-                                                    modal.room = Some(current_room);
-                                                });
-                                            }
-                                        });
-                                    },
-                                    div { class: "min-w-0",
-                                        div { class: "flex items-center gap-2",
-                                            h2 { class: "text-lg font-semibold text-text truncate",
-                                                "{current_room_label}"
-                                            }
-                                            span {
-                                                class: "text-text-muted flex-shrink-0",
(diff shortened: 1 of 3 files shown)


sanity:
## Review summary

Ran four parallel internal reviewers + Codex. Consolidated findings:

**Addressed in fixup commit 35c028d5:**
- Skeptical + Codex flagged that `max-w-full` (replacing the original `flex-1`) made the title button auto-size to its content, so the hover rectangle and `h2.truncate` shrink behaviour now depended on title length. Switched to `w-full` so the button always fills the wrapper and matches the original full-width hover/truncate behaviour.
- Testing reviewer flagged a race in the "modal does not appear" Playwright case: the modal opens via `crate::util::defer` (setTimeout(0)), so a synchronous `toHaveCount(0)` could pass before any incorrectly-bubbled handler ran. Added a 50 ms wait and a comment explaining why we deliberately don't `stopPropagation` on the link.

**Acknowledged but not addressed:**
- Test #1 (`closest('button') === null`) is a structural proxy and would not catch a regression that moved the click handler off `<button>` while keeping `<a>` inside the original div. Test #2 covers the behavioural side. Both kept; the proxy is a strong, deterministic guard for the actual fix.
- Mobile-viewport variant of the regression test. The DOM structure is viewport-independent, so the structural test (#1) at desktop already protects the same invariant. Existing `responsive-layout.spec.ts` exercises mobile selectors against the unchanged hamburger / members buttons.
- `<h2>` inside `<button>` (Codex low). Standards-imperfect but widely used; current accessibility tree is fine. Not regressing here.

**Filed as separate issue (out of scope for #225):**
- #227 — stored XSS via member nickname rendered in `dangerous_inner_html` at `members.rs:282`. Same anti-pattern as the room header bug (also nests potential `<a>` inside `<button>`), but currently inert because nicknames usually don't contain HTML — they aren't escaped or validated, so any room member can inject HTML/scripts. Surfaced by the big-picture reviewer's note that members.rs uses the same pattern.

[AI-assisted - Claude]

PR #229, opened 2026-04-30
fix(ui): show feedback when Export Identity Copy to Clipboard is clicked

sanity:
## Problem

Reported on Matrix by Ivvor (2026-04-30):

> If you use Export ID in River and hit 'Copy to Clipboard' you get no feedback that anything has happened.

The Export Identity modal silently copied the token. Users had no way of knowing whether the click had succeeded.

## Approach

Mirror the pattern already used by the two other clipboard buttons in the UI (invitation modal copy buttons in `invite_member_modal.rs`, the share-your-key button in `not_member_notification.rs`): a `use_signal` holds the button label, the click handler swaps it to "Copied!", and the modal-close paths reset it back to "Copy to Clipboard" so reopening starts fresh.

Audited the rest of the codebase for similar copy buttons missing feedback. The Export Identity modal was the only one — the invitation and share-your-key buttons already did this.

## Testing

Added `ui/tests/copy-clipboard-feedback.spec.ts` covering:

- Clicking "Copy to Clipboard" updates the button to "Copied!".
- Closing the modal and reopening it resets the label to "Copy to Clipboard".

Verified the regression test fails on `main` (the bug) and passes with the fix applied. Full chromium test suite (26 tests) still passes.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -348,6 +348,7 @@ pub fn MemberList() -> Element {
 #[component]
 fn ExportIdentityModal(is_active: Signal<bool>) -> Element {
     let mut token_text = use_signal(String::new);
+    let mut copy_button_text = use_signal(|| "Copy to Clipboard".to_string());
 
     // Generate the export token when modal opens
     use_effect(move || {
@@ -448,6 +449,7 @@ fn ExportIdentityModal(is_active: Signal<bool>) -> Element {
     let handle_copy = move |_| {
         let text = token_text.read().clone();
         crate::util::copy_to_clipboard(&text);
+        copy_button_text.set("Copied!".to_string());
     };
 
     rsx! {
@@ -456,6 +458,7 @@ fn ExportIdentityModal(is_active: Signal<bool>) -> Element {
             onclick: move |_| {
                 is_active.set(false);
                 token_text.set(String::new());
+                copy_button_text.set("Copy to Clipboard".to_string());
             },
             div {
                 class: "bg-panel border border-border rounded-xl shadow-lg p-6 max-w-xl w-full mx-4",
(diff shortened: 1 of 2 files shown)


sanity:
## Review feedback addressed

Ran four internal reviewers + Codex on this PR. Summary of how each concern was handled:

### Addressed
- **Repeated reset string in 4 places** (code-first): extracted `COPY_BUTTON_DEFAULT` const.
- **Backdrop-click reset path untested** (testing-reviewer + skeptical): added a third Playwright test that dismisses via the backdrop and asserts the label resets.
- **Reset duplication across close handlers** (skeptical): replaced the per-handler reset with one `use_effect` that runs whenever `is_active` flips false. Covers backdrop click, Close button, and any future dismissal path (X icon, Escape) uniformly.

### Out of scope (not blocking)
- **Auto-revert "Copied!" after a timeout** (code-first nice-to-have): existing copy buttons in `invite_member_modal.rs` and `not_member_notification.rs` don't auto-revert either. Sticking with consistent UX.
- **Verify clipboard contents** (testing-reviewer): bug report is specifically about the missing visual feedback. Adding clipboard-content assertions would be a separate test.

### Discussed but not changed
- **`crate::util::defer()` for the `set()` calls in `onclick` / `use_effect`** (codex P1, skeptical high #1): the AGENTS.md rule is documented with `ROOMS.write()` / `CURRENT_ROOM.write()` (`GlobalSignal`) examples; the cited reason is that memo readers can panic during the write Drop's subscriber notifications. `copy_button_text` is a local `use_signal` whose only reader is the rsx text node — no `use_memo`. The two existing copy-feedback sites (`invite_member_modal.rs:225-237`, `not_member_notification.rs:13-16`) use the identical direct-`set()` pattern and have not produced bug reports. To avoid scope creep this PR matches that existing pattern. If the rule is meant to apply uniformly to local signals as well, all three sites should be updated together as a separate cleanup; happy to follow up if that's the right call.

[AI-assisted - Claude]

PR #236, opened 2026-05-07
fix(ui): clear unread badge when tab transitions visible to hidden

sanity:
## Problem

The document title showed `(N) River - <room>` the moment the tab lost focus, even when the user had been actively on the page. Reproduced as: title is `River - Freenet` while focused, then becomes `(3) River - Freenet` immediately after switching tabs.

The unread count was summed across **all** rooms via `count_total_unread_messages()`. For rooms the user had never visited, `last_read_message_id` was `None`, so every message from other authors counted as unread. While the tab was visible the format `(Some(name), true, _)` hid the count, but the moment the tab went hidden the format flipped to `(Some(name), false, count)` and the badge appeared — even though the user had been right there.

## Approach

On the visible -> hidden transition in `on_visibility_change`, advance every room's `last_read_message_id` to its latest currently-known display message (new `mark_all_rooms_as_read`). The user had the tab in front of them, so anything already in state is treated as seen. Only messages arriving *after* this moment contribute to the badge while the tab is hidden — which is the behaviour Slack/Discord users expect.

Existing semantics are preserved: when tab becomes visible again, `mark_current_room_as_read` still runs; when messages arrive in the current room while visible, `room_synchronizer.rs` still marks them read.

## Testing

New Playwright regression test `ui/tests/title-unread-badge.spec.ts`:

- Title stays `River` after hiding the tab when no room is selected (example data has 3 unvisited rooms with messages).
- Title stays `River - Public Discussion Room` after hiding the tab while a room is selected.

Verified the test fails on `main` with `(11) River - Public Discussion Room` and passes with the fix. Full Playwright suite (5 browsers, 145 tests) passes.

Also: `ui/tests/playwright.config.ts` now reads `PLAYWRIGHT_BASE_URL` so the suite can run against a non-default port.

## Notes

- UI-only change — no delegate/contract WASM modified, no migration needed.
- Touches: `ui/src/components/app/document_title.rs`, `ui/tests/playwright.config.ts`, new `ui/tests/title-unread-badge.spec.ts`.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/document_title.rs
+++ b/ui/src/components/app/document_title.rs
@@ -13,6 +13,7 @@ use crate::util::ecies::unseal_bytes_with_secrets;
 use dioxus::logger::tracing::{debug, info, warn};
 use dioxus::prelude::*;
 use river_core::room_state::member::MemberId;
+use river_core::room_state::message::MessageId;
 use std::cell::RefCell;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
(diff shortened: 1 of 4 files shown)


sanity:
## Review responses

Pushed 668f3a24 addressing the cross-confirmed findings:

- **Signal safety (code-first, skeptical)**: `mark_all_rooms_as_read` now wraps its `with_mut` + `safe_spawn_local` in `crate::util::defer()`. The visibilitychange handler is a raw JS event callback with no Dioxus scope on the stack; deferring matches the rule in AGENTS.md "Never mutate signals inside spawn_local or event handlers."
- **Doc comment (big-picture)**: `RoomData.last_read_message_id` doc updated to spell out the three advance points (open room, message-while-visible, visible-to-hidden transition) so the new semantics are explicit.
- **Idempotency / cycle (testing)**: added `hide -> show -> hide cycle does not introduce a badge` test.

## Behaviour debate (skeptical, codex P1)

Both flagged: marking unvisited rooms as read on first hide loses notifications for rooms the user has never opened.

This is a deliberate trade-off matching the user's explicit report:

> "appearing to show 3 unread messages even though I was just on the tab"

The user's expectation is that the title badge means "things that arrived since I stopped paying attention," not "messages I haven't viewed in some room I might never open." The pre-fix code conflated these because rooms with `last_read_message_id == None` had every message from others count as unread.

After this fix:
- Title badge after hide reflects only messages that arrived **after** the user stopped paying attention.
- Per-room state (`last_read_message_id`) is advanced uniformly so the badge can grow from zero.
- There is no per-room badge in the room list today (`count_unread_messages_in_room` is only called from the title path), so no UI regresses.
- If/when a per-room sidebar badge is added, it will inherit the same "since you stopped paying attention" semantic, which is consistent.

If product later decides "title should reflect total ever-unread including unvisited rooms," that would be a different feature and would need a separate read-pointer ("seen-in-list" vs "viewed-in-room").

## Skipped findings (with justification)

- **Rust unit tests for `mark_all_rooms_as_read` (testing #4)**: deferred. The function reads `GlobalSignal`s; there is no mocking infrastructure in `document_title.rs` today. Matches existing patterns — `count_unread_messages_in_room`, `mark_current_room_as_read`, etc., have no Rust tests either. Worth doing as a separate effort that adds the harness.
- **Positive test: messages-while-hidden DO produce a badge (testing #1, code-first)**: cannot be exercised end-to-end from Playwright without a Freenet node or test-only WASM-bindgen exports to inject messages into `ROOMS`. The mark-as-read logic is a one-line invariant (`last_read = latest`); the existing hide-then-no-badge test plus the cycle test cover the regression surface. Filing as follow-up if reviewers want belt-and-braces.
- **Hidden-on-startup edge case (code-first, skeptical)**: deferred. `init_document_title_manager` writes initial `DOCUMENT_VISIBLE` from `get_visibility_state()`; the first `visibilitychange` after that fires the normal path. If the page loads while hidden, the user sees the pre-existing behaviour for one cycle. Not a regression introduced by this PR.
- **`MessageId::clone()` (code-first nit)**: skeptical correctly noted the `latest` is a `&MessageId` from the loop, so `.clone()` is required, not redundant. No change.
- **Persistence on every hide (big-picture)**: the existing `mark_current_room_as_read` already calls `save_rooms_to_delegate` on every focus-in and on every selected-room message arrival, so the delegate write path is already exercised on every visibility flip. The new function additionally short-circuits if no room actually changed (`if updates.is_empty() { return; }`) and skips rooms already at latest. Net effect on a passive tab: still O(1) save per visibility flip.

## Test results

- `npx playwright test title-unread-badge.spec.ts` — 15/15 across 5 browsers (3 cases × 5 projects)
- Full Playwright suite — 152 passed, 3 skipped (pre-existing), 0 failed

[AI-assisted - Claude]

Issue #243, opened 2026-05-14
feat: in-room direct-messages UI + riverctl (#230 Phase 2+3)


sanity:
## Problem

Phase 1 of in-room direct messages shipped in #240 (and deployed today as part of #242): `ChatRoomStateV1` carries `DirectMessagesV1`, with sender-signed `AuthorizedDirectMessage` + recipient-signed `AuthorizedRecipientPurges`, all validated by the room contract WASM in production wasmtime. The wire is real and live.

But neither end-user surface can use it yet:

```
$ grep -rn "DirectMessage\|direct_message" ui/src/ cli/src/
(no results)
```

So users can neither send nor view DMs from the UI, and `riverctl` has no DM commands. This issue tracks closing both gaps. Doing them together keeps the two wire-format consumers in lockstep — Phase 1 already shipped a contract WASM bump and a `riverctl 0.1.55` release for unrelated reasons (#242); the next contract-wire-touching change should bump them together one more time.

This was originally planned as Phase 2 + Phase 3 of #230 (in-room encrypted direct messages between members). #230 auto-closed when #240 / Phase 1 merged. Re-filing the remaining work in one issue.

## Scope — Phase 2 (UI)

| Piece | Notes |
|---|---|
| "Send message" action in member-info modal | Existing modal at `ui/src/components/members/member_info_modal.rs` opens when clicking a member. Add a button to open a DM thread for that pair. |
| DM composer (textarea + send button) | Mirrors the room-message composer in `ui/src/components/conversation.rs`. |
| Sender path: ECIES-encrypt body to recipient's room-scoped VK, sign, emit `UpdateContractRequest` | Reuse `river_core::ecies` randomized helpers (UI has working `getrandom` via the `js` feature). Build `DirectMessage` -> sign -> wrap in `AuthorizedDirectMessage` -> append to `DirectMessagesV1::messages` -> emit state delta. |
| Per-room "Direct messages" panel/tab | New view listing conversations the current user has open with co-members in this room. |
| DM thread render | Decrypt-on-display, group by counterparty, chronological. |
| Incoming DM notifications + unread tracking | Hook into the existing `UpdateNotification` handler in `ui/src/components/app/freenet_api/response_handler/update_notification.rs`; mark threads unread when a new `AuthorizedDirectMessage` addressed to the local member arrives. |
| Purge / block UI | Build a new `RecipientPurges` envelope with `version + 1` + the new tombstones, sign, emit state delta. |

[...]

PR #245, opened 2026-05-14
fix(delegate): back-fill prior secret versions for newly-joined members

sanity:
## Bug report

Ivvor on production River post-#242 deploy (2026-05-14):

> Private room not working right when I invite another node. I see stuff like `[Encrypted message - secret v1 not available (have: [0])]` on owner node and `[Encrypted message - secret v0 not available (have: [1])]` on invited user.
>
> [later] now i have a point where the owner can see everything, and the invited user can see new messages, but the room name is encrypted, so is the nickname of the owner, and the initial bunch of messages.

## Root cause

When a new member joins a private room, the owner's chat-delegate rotates the secret to `current_version + 1` and emits `encrypted_secrets` only at the new version. Every prior version is missing for the newly-joined member, so they cannot decrypt the room name (sealed at v0 at room creation), the owner's nickname (sealed at v0 when set), or any messages sent before they joined.

The owner has v0 from room creation (UI emits owner's self-encrypted secret then) plus v_n from each rotation, so the owner sees everything. The bug is asymmetric: only the newly-joined member's view is broken.

## Fix

Extracted the rotation per-member loop into a pure helper `build_rotation_encrypted_secrets` and taught it the distinction:

- **Continuing members** (already in cached `previous_members`, plus the owner): only emit `new_version`. They already have older entries from the rotation that admitted them.
- **Newly-joined members** (in `current_members` but NOT in `previous_members`): emit every version `[0..=new_version]`. The older secrets are re-derived deterministically from the owner's signing seed via `derive_room_secret(seed, owner_vk, v)`, so the back-filled blobs are byte-identical to whatever the version was originally produced as.

The owner is always continuing — at v0 the owner emits their own `encrypted_secret` from the UI; on every subsequent rotation this helper emits one fresh entry at the new version. So the owner never needs back-fill.

## Tests

Two new regression tests:

- `rotation_backfills_prior_versions_for_newly_joined_members` — owner + alice (continuing) + bob (newly joining); rotation to v2 must emit exactly `{owner@v2, alice@v2, bob@v0, bob@v1, bob@v2}`. Verifies signatures and byte-identity on bob's back-filled v0 blob.
- `rotation_backfills_for_all_members_on_first_rotation` — empty `previous_members`, every member newly-joined.

22/22 chat-delegate tests pass (was 20).

## Deploy footprint

- Delegate WASM changes → **V17 migration entry** added to `legacy_delegates.toml` so existing users on the post-#242 delegate (code_hash `123f732d…` → delegate_key `540b2887…`) transparently migrate.
- `room_contract.wasm` is **byte-identical** to #242 — no contract key change, no room migration ceremony, no riverctl republish, no gkapi/secrets/quickstart updates.
- Production deploy is just `cargo make publish-river`.

## Recovery for users currently stuck

When this lands and Ivvor reloads, the owner's delegate runs the rotation pipeline again on the next state notification (any member-set change, or even just a reload causing the cached `member_set` to be re-evaluated), emits the back-filled secrets, and Ivvor's UI decrypts history correctly.

For private rooms where membership has stabilised, the rotation pipeline won't naturally re-fire (member set unchanged → no rotation). The owner can trigger a manual rotation by editing the room config or banning + unbanning a member. Both paths go through UI-side `rotate_secret` which today still does the "new version only" thing — fixing that path too is an obvious follow-up.

## Test plan

- [ ] CI green (wasm-sync guard, delegate-migration check, build, playwright)
- [ ] Local Ivvor-scenario smoke test: fresh test-contract publish under separate keypair, owner creates private room + messages, second profile accepts invite, verify second profile sees room name + owner nickname + history
- [ ] Merge + `cargo make publish-river` to production
- [ ] Confirm Ivvor's existing room recovers (with manual ban+unban trigger if needed)

Closes the bug Ivvor reported on top of #242. Tracking issue #243 (Phase 2 DM UI) unchanged.

[AI-assisted - Claude]


sanity:
## Rewrite addressing all 5 reviews

Force-push of `de7c962a` on top of the original commit. Net: helper rewritten, four reviews addressed, original tautological test replaced.

### Critical issues fixed

**1. v0 mismatch (skeptical, big-picture, codex)** — the original fix back-filled by *re-deriving* `derive_room_secret(seed, owner_vk, 0)`, but production River creates v0 *randomly* via `generate_room_secret()` in `ui/src/room_data.rs::create_new_room_with_name`. So the fix didn't actually let new members read the room name — same user-visible bug.

The rewrite recovers prior secrets by ECIES-decrypting the owner's existing `encrypted_secret`-at-v from the room state, using the owner's signing key (which the delegate already holds for rotation signing). The decrypted plaintext IS the actual secret the room is using, so re-encrypting it for the new member produces bytes they can decrypt to recover the real v0.

New `decrypt_secret_from_member_blob_raw` wrapper in `common/src/ecies.rs` takes raw 32-byte ephemeral-pubkey bytes so the delegate doesn't need a direct `x25519-dalek` dep.

**2. Cache-based dedup (codex)** — the original used the delegate-local member-set cache to decide continuing-vs-newly-joined. On a fresh delegate / restart / webapp reinstall, that cache is missing → every member treated as newly-joined → duplicate `(member, version)` pairs → contract rejects via `RoomSecretsV1::apply_delta` dedup → rotation wedged permanently.

The rewrite derives the decision directly from `new_state.secrets.encrypted_secrets`: skip any `(member, version)` pair already in state. Continuing-vs-newly-joined is now emergent — a continuing member already has v0..v_prev, so only v_new is emitted; a newly-joined member has nothing, so they get the full back-fill. Works correctly even with no cache.

**3. Tautological test (testing, big-picture)** — the original test asserted byte-identity against `derive_room_secret(..., 0)`, the same function the broken helper called. It would pass for both behaviours. Replaced with `backfill_uses_real_v0_recovered_from_owners_encrypted_secret`: state contains a v0 blob with random bytes; helper produces a back-filled v0 for bob; bob's signing key MUST decrypt back to the same random bytes. That's the property the user cares about.

### Edge-case tests added

* `backfill_handles_multiple_simultaneous_new_members` (testing)
* `backfill_dedups_against_state_for_continuing_members` (codex)
* `backfill_handles_readmitted_member` (code-first)

24/24 chat-delegate tests pass (was 22). All four new tests do real end-to-end decryption with the recipient's signing key, not byte-identity self-checks.

### Minor

* V17 description corrected (code-first finding): "#243" → "#245".

### Not in this PR

* UI-side `rotate_secret` in `ui/src/room_data.rs:465` (big-picture, code-first finding) — same bug class. Not pulled in here because the recovery path for Ivvor-style stranded users is: owner triggers a delegate rotation by changing the member set (invite a fresh member, or ban+reinvite). Delegate's now-correct logic back-fills. UI `rotate_secret` is only hit on direct-from-edit / ban-without-readmission paths and doesn't add new members, so it's not on the recovery path. Filing as follow-up.

* State-size bloat caveat (skeptical M2) — back-fill scales O(rotations × new_members) per join. Documented in the helper's docstring; per the big-picture reviewer's calculation that's ~22.5 KB per join at 100 rotations, fine in practice. No cap added.

* WASM end-to-end harness (testing) — still gated on a `MockDelegateCtx` upstream in freenet-stdlib. Tracked outside this PR.

[AI-assisted - Claude]

PR #248, opened 2026-05-14
fix(ui): leave room sticks across reloads via removed_rooms tombstone

### LINKED ISSUE #247: Leave room doesn't stick: legacy-delegate migration re-adds the room on refresh
## Reports

Ivvor (2026-05-14, 17:03):

> every time i leave the private room on the second node it's coming back after a refresh a little while later

Same class of bug as #163 ("Leaving a room doesn't work (if you are owner)") but observed by the **invited** user, not the owner. The mechanism is different too — see below.

## Root cause

\`ui/src/components/room_list/edit_room_modal.rs:275-301\` — "Confirm Leave" only does:

\`\`\`rust
ROOMS.write().map.remove(&room_vk);
// ... save_rooms_to_delegate() saves new (room-less) state
\`\`\`

No state delta is sent to the room contract; no unsubscribe to the local node. Save persists the new (room-less) \`rooms_data\` to the **current** delegate.

On refresh, the UI:
1. Fetches \`rooms_data\` from the current delegate — works correctly, room is absent ✅
2. **Also fires \`fire_legacy_migration_request()\`** in \`ui/src/components/app/chat_delegate.rs:154\`, probing all 17 legacy delegate keys in \`legacy_delegates.toml\`
3. Some legacy delegate (e.g. V13, V14, V15, V16, V17) STILL has the room in its on-disk \`rooms_data\` — that data was never removed when the delegate was superseded
4. \`current_rooms.merge(loaded_rooms)\` in \`ui/src/components/app/freenet_api/response_handler.rs:308\` calls \`Rooms::merge\` in \`ui/src/room_data.rs:881\`
5. \`merge\` is purely additive — it inserts rooms from \`loaded_rooms\` into \`self\` if missing. **It never removes rooms.**
6. Room reappears

The "a little while later" in Ivvor's report is the time for the 13+ legacy migration probe responses to come back and trigger the merge.

## Fix shape

Several reasonable options:

**A. Track explicitly-removed rooms.** Add a \`removed_rooms: HashSet&lt;VerifyingKey&gt;\` field to the rooms_data persisted to the current delegate. Leave appends to it. \`merge\` checks the set and skips re-adding removed rooms.

**B. Don't run legacy migration if current delegate already has data.** If the current delegate's \`rooms_data\` key is non-empty, that's evidence the user has been here before — don't probe legacy delegates at all, or only probe to copy data forward then mark legacy delegates as cleared. \`mark_legacy_migration_done()\` exists but the trigger seems to be specifically "delegate X not found" errors, not "I already have current data."

**C. Have leave also write to legacy delegates.** Less attractive — there are 17 of them and growing.

I'd go with **A** as the local fix (room state already has a CRDT-able "removed_rooms" concept) plus a side issue to tighten **B** so legacy migration doesn't re-trigger gratuitously after the first successful load.

## Related bug

Per #246, the memory blowup Ivvor reported may be related: if the user is in a confused half-left state (locally removed but still subscribed at the contract, still receiving UpdateNotifications, getting re-added by merge), the rendering / state-handling loop may churn through copies of the room state and leak memory.

## Test plan

- [ ] Add Playwright test: create room, leave room, refresh, assert room not in list
- [ ] Add unit test in room_data.rs that \`merge\` respects \`removed_rooms\`
- [ ] After fix, manually test on production River: create private room, invite a 2nd identity (different browser profile), have 2nd identity leave, refresh, confirm room stays gone

## Related

- #163 (different mechanism, but same symptom for owner)
- #246 (memory leak — possibly related)
- #245 (deterministic ECIES fix that just deployed — unrelated to this bug, but the post-deploy was when Ivvor surfaced it)

[AI-assisted - Claude]

sanity:
Closes #247.

## Problem

Ivvor's report (2026-05-14 17:03):
> every time i leave the private room on the second node it's coming back after a refresh a little while later

\`Confirm Leave\` in \`ui/src/components/room_list/edit_room_modal.rs:283\` only did \`ROOMS.write().map.remove(&room_vk)\` and saved rooms_data to the current delegate. On reload, \`fire_legacy_migration_request()\` probed all 17 entries in \`legacy_delegates.toml\`. Some legacy delegate's on-disk \`rooms_data\` still contained the room — that data was never purged when the delegate was superseded. \`Rooms::merge\` was purely additive, so the room came back.

## Fix

Add a \`removed_rooms: HashSet&lt;VerifyingKey&gt;\` tombstone set to the serialised \`Rooms\` struct (with \`#[serde(default)]\` for backward compatibility). \`Rooms::merge\` now unions tombstones from both sides and filters all entries against the union; \`map\` entries on the receiver matching a tombstone are also dropped (defensive). New \`Rooms::leave_room(vk)\` helper does both mutations atomically. Invitation acceptance clears the tombstone for explicit rejoin.

## Tests

- \`leave_room_tombstone_prevents_merge_re_adding\` — full scenario from the report.
- \`merge_unions_removed_rooms_tombstones\` — if either side has the tombstone, it's honoured.
- 14/14 \`room_data::tests\` pass (was 12).

## Deploy footprint

UI-only. \`chat_delegate.wasm\` and \`room_contract.wasm\` are byte-identical to #245. No migration entry. No riverctl republish. Just \`cargo make publish-river\`.

## Recovery for currently-stuck users

After this lands and users reload, the next time they leave a room the tombstone is recorded. Their previous leave attempts (without tombstone) won't survive — they'll need to leave once more after the upgrade.

## Test plan

- [ ] CI green
- [ ] Multi-perspective reviews
- [ ] Merge + \`cargo make publish-river\`

[AI-assisted - Claude]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ node_modules/
 # Playwright test results
 ui/tests/test-results/
 
+# Syncthing in-flight transfer files (`.syncthing.<name>.tmp` are
+# created by the syncthing daemon while a file is being received).
+**/.syncthing.*.tmp
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -295,6 +295,7 @@ fn initial_rooms() -> Rooms {
     Rooms {
         map: std::collections::HashMap::new(),
         current_room_key: None,
+        removed_rooms: std::collections::HashSet::new(),
         migrated_rooms: Vec::new(),
     }
 }
--- a/ui/src/components/app/freenet_api/response_handler.rs
+++ b/ui/src/components/app/freenet_api/response_handler.rs
@@ -21,6 +21,7 @@ use crate::util::ecies::{decrypt_secret_from_member_blob, decrypt_with_symmetric
 use crate::util::owner_vk_to_contract_key;
 use ciborium::de::from_reader;
 use dioxus::logger::tracing::{error, info, warn};
+use dioxus::prelude::ReadableExt;
 
 use freenet_stdlib::client_api::{ContractResponse, HostResponse};
 use freenet_stdlib::prelude::OutboundDelegateMsg;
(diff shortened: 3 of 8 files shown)


sanity:
## Review findings addressed (force-push \`b65cd0ab\`)

* **Skeptical H1** (subscribe loop bypassed tombstone): \`response_handler.rs\` now builds \`tombstoned\` from union of incoming \`loaded_rooms.removed_rooms\` AND current in-memory \`ROOMS.removed_rooms\`, then filters \`room_keys\`, \`signing_keys\`, and \`owned_rooms_for_subscription\` against it before any subscribe/sync action.

* **Skeptical H2** (CURRENT_ROOM restored to tombstoned): skipped now when the saved key is tombstoned.

* **Code-first members.rs**: import-identity flow at \`members.rs:580\` now clears the tombstone before inserting, matching the \`get_response.rs\` invitation-accept pattern.

* **Tests** (3 new): \`rooms_deserialises_pre_247_blob_with_default_removed_rooms\` (backward-compat via ad-hoc \`LegacyRooms\` shape — \`serde_json\` isn't in deps), \`rooms_round_trip_preserves_removed_rooms_tombstone\` (write-side wire format), \`cleared_tombstone_allows_merge_to_re_add\` (rejoin path). 17/17 \`room_data::tests\` pass (was 14).

* **Doc comment** on \`removed_rooms\` corrected — no longer claims "monotonically grows", documents the rejoin-clear and multi-device convergence semantics.

## Deferred to follow-ups (captured in commit message)

* Codex P2: timestamped tombstones / rejoin counter to defend against stale-snapshot divergence. Requires CRDT redesign — out of scope for tonight.
* Skeptical M1: \`removed_rooms\` GC.
* Skeptical M2: cross-version divergence (old build drops the field on save). Mitigated by \`#[serde(default)]\` + the "leave once more after upgrade" note.
* Skeptical M4: \`migrated_rooms\` not filtered.
* Skeptical L1: \`PartialEq\` ignores \`removed_rooms\`.
* Skeptical L3: stale \`.syncthing.*.tmp\` files (housekeeping).

[AI-assisted - Claude]

PR #250, opened 2026-05-15
perf(ui): demote roomstate console dumps from info! to debug!

sanity:
## Summary
- `UpdateNotification`, the delegate request/response handler, and the delta/state merge paths in `room_synchronizer` were all dumping the full `ChatRoomStateV1` / `Delta` payloads at `info!` level via `{:?}`.
- These fire on every contract notification and every delta merge, so a single room reload pumps hundreds of large strings into the JS console — which retains them.
- Demote the full Debug payloads to `debug!` (filtered out by the default dioxus logger init). Keep a short `info!` summary line (`N messages, M members`) so the lifecycle is still visible.

## Context
User report from Ivvor (Matrix, 2026-05-15):
> Throughout all this I was still noticing memory footprint issues. Indeed, the PC I was using to run tests, froze up about 15 mins ago and I'm still waiting for it to unfreeze.
>
> Devtools couldn't find source of memory spikes. Using codex to look through the River code in conjunction with my observations its best guess was that dumping huge amounts of data to the JS console might be causing it. Mainly repeated dumps of roomstate. This would match with the peaks happening during room reloads/refreshes.

The codex diagnosis matches exactly what these `info!` calls do.

## Test plan
- [x] `cargo check --target wasm32-unknown-unknown --features no-sync` for `river-ui` passes
- [x] `cargo check -p river-core` passes
- [ ] Manual: load a room with many messages, observe the JS console no longer fills with `ChatRoomStateV1 { ... }` dumps every refresh
- [ ] Manual: verify `RUST_LOG=debug` / debug build still shows the full dumps when needed for diagnostics

## Out of scope
The other issues Ivvor reported in the same message — new members initially seeing encrypted messages until refresh, and owner-can't-send-messages during initial sync — are separate sync/decryption bugs in the private-room back-fill path and will be filed as a separate issue.

[AI-assisted - Claude]

### DIFF
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -66,6 +66,12 @@ ciborium = "0.2.2"
 js-sys = "0.3.64"
 thiserror = "2.0.12"
 log = "0.4" # Add log crate
+# Compile-time gate so `debug!`/`trace!` macros are dropped entirely in
+# release builds. This is belt-and-suspenders on top of the runtime
+# filter `dioxus_logger` sets up (Info in release, Debug in dev) — if
+# that runtime filter ever regresses, the macros still won't emit. See
+# https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters
+tracing = { version = "0.1", default-features = false, features = ["std", "release_max_level_info"] }
 
 # Internal dependencies
 river-core = { workspace = true, features = ["ecies", "ecies-randomized"] }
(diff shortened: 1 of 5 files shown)


sanity:
Follow-up from Ivvor (Matrix, 2026-05-16 07:35):

> I don't think those memory issues with River are specific to private rooms BTW. They are just amplifying it. I now realise going back that similar things were happening before, but not to point where they were actually crashing the browser or system. Also, I've now found that using Firefox instead of Chrome I don't suffer from the massive memory spikes (10+ GB), but it is still spiking on page refreshes and updates.

This matches the diagnosis: the noisy `info!("...{:?}", state/delta)` calls this PR demotes are in generic sync paths (`update_notification`, `room_synchronizer`, `chat_delegate`) — they fire for every room, not just private ones. Private rooms amplify the per-log payload size because of the encrypted blobs. Chrome's console keeps richer per-entry references than Firefox's, which is consistent with Ivvor seeing 10+ GB spikes on Chrome but only smaller ones on Firefox.

[AI-assisted - Claude]

PR #254, opened 2026-05-16
fix(ui): gate legacy delegate migration on current delegate response

### LINKED ISSUE #253: bug: old delegates overwriting the current active delegate
I'd had a long running problem using multiple rooms whereby a newly created room would just suddenly disappear. After a lot of attempts to find out why, using a Codex agent to examine the River codebase, it appears that it is possible for old delegates to overwrite the current latest delegate in some situations. This seems to be because the delegate migration path is followed repeatedly under various situations, such as a page refresh, and if a response to retrieving an older delegate somehow arrives after a response for the latest active delegate, then it will overwrite the newer delegate as part of the migration flow.

I suspect this is becoming more of a problem for me because I have a long running/lived node that it likely pulling many old delegates over many versions of River. I'd rather not have to keep wiping my old nodes every time a new version of River comes out just to avoid this.

sanity:
## Problem

A long-running River user reports newly-created rooms silently disappearing after page refresh (freenet/river#253). Root cause: a race between the current delegate's load-rooms response and the legacy migration probes.

\`set_up_chat_delegate()\` was firing both the current delegate's load-rooms request and the legacy migration probes concurrently at startup. Each legacy response that returned data triggered \`save_rooms_to_delegate()\`, writing whatever was currently in \`ROOMS\` to the **current** delegate's storage.

When a legacy response arrives before the current delegate's response, \`ROOMS\` still holds an empty or partial state, so the legacy-triggered save overwrites the current delegate with stale data. On next refresh, the current delegate returns the overwritten state and recently-created rooms are gone.

The user reports this happening repeatedly across refreshes, which is consistent with: the localStorage \"migration done\" flag only gets set after a successful save, so if save races and writes incomplete state, the flag may still be unset, and the migration re-runs each session.

## Approach

Gate legacy migration on the current delegate's response so it only fires when there is nothing to clobber:

- \`set_up_chat_delegate()\` no longer fires migration probes.
- When the current delegate returns rooms_data successfully, mark legacy migration done across sessions — current is the source of truth.
- When the current delegate returns no rooms_data, *then* fire the legacy migration probes. With current empty, no race possible.

Trade-off: a user with data in both current and legacy (e.g. an interrupted migration in the past) will not get the legacy data merged. But the alternative — the current behavior — actively destroys *newer* data, which is much worse. The legacy data path was always best-effort recovery from old WASM versions; the new state is what the user has been actively building.

## Testing

- \`cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync\` ✅
- \`cargo test -p river-core --test migration_test\` ✅ (4/4)
- \`cargo fmt --check\` ✅
- \`cargo clippy\` produces no new warnings (4 pre-existing warnings unchanged)

A deterministic unit test for this race is impractical: it requires WASM, WebSocket, Dioxus signals, and a specific async ordering. The fix is structural — legacy probes are gated behind a single response — so the proof is by inspection of the changed control flow.

Closes #253

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/river-publish.md
+++ b/.claude/rules/river-publish.md
@@ -73,11 +73,31 @@ cargo make publish-all
 
 ## How Delegate Migration Works
 
-1. On startup, `set_up_chat_delegate()` fires `fire_legacy_migration_request()`
-2. `LEGACY_DELEGATES` is generated at compile time from `legacy_delegates.toml` by `ui/build.rs`
-3. Sends `GetRequest` for `rooms_data` to EACH legacy key
-4. Response handler migrates room data + signing keys to the current delegate
-5. `mark_legacy_migration_done()` sets localStorage flag to skip on next load
+Legacy migration is **gated on the current delegate's response** to avoid a
+race where stale legacy data overwrites newer state on the current delegate
+(freenet/river#253). The flow:
+
+1. On startup, `set_up_chat_delegate()` fires `fire_load_rooms_request()` for
+   the **current** delegate only. It does NOT fire the legacy probes.
+2. `LEGACY_DELEGATES` is generated at compile time from `legacy_delegates.toml`
+   by `ui/build.rs`.
+3. When the current delegate's `GetResponse` for `rooms_data` arrives:
+   - **Has authoritative data** (rooms or tombstones): call
+     `mark_legacy_migration_done()` (persists across sessions). Legacy probes
+     are never fired — current is the source of truth.
+   - **Empty or missing**: call `fire_legacy_migration_request()` to probe
+     each legacy key with `GetRequest` for `rooms_data`. Safe because current
+     has nothing to clobber.
+4. The response handler migrates room data + signing keys from each
+   responding legacy delegate to the current delegate.
+5. On a legacy delegate returning data, `mark_legacy_migration_done()` is
+   called after the post-merge save succeeds; on a legacy delegate returning
+   no data it is called immediately.
+
+**Trade-off**: a user with data in both current and legacy delegates will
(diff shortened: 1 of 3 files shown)


sanity:
Review responses (5 reviewers: 4 internal + Codex):

**Addressed in 3cdacf22**:
- **Codex P2**: Empty `Some(serialized empty Rooms)` no longer suppresses migration. Now distinguishes "has rooms or tombstones" (authoritative — mark done) from "value present but empty" (placeholder — fire migration). Users with rooms only in legacy delegates are no longer stranded.
- **Testing review**: Extracted `decide_legacy_migration_action()` as a pure helper with 5 unit tests covering each (value_present, has_rooms, has_tombstones) case. Locks the gating polarity against future refactors.
- **Skeptical M2**: `fire_legacy_migration_request` tightened from `pub` to `pub(crate)`.
- **Big-picture**: `.claude/rules/river-publish.md` updated to reflect the gated flow.

**Acknowledged, not changed**:
- **Code-first concern about `mark_legacy_migration_done` on corrupt data**: reviewer looked at an intermediate diff. The committed code already places the call inside the `Ok(loaded_rooms)` arm, so corrupt data does not mark migration done.
- **Skeptical H1 (cursor-restore from legacy snapshot)**: pre-existing, separate bug. Filed as #255.
- **Skeptical H2/M3 (timing window + spawn-vs-defer race)**: same root cause as H1 — the bigger save-after-merge race is pre-existing and out of scope. The data-loss path #253 actually reported is closed by this PR; #255 covers the residual cursor-position concern.

All review verdicts: **ready to merge** with the fixes above.

[AI-assisted - Claude]

PR #244, opened 2026-05-14
feat: in-room DMs UI + riverctl (#243 Phase 2/3) + share-invite-via-DM picker (#252 partial)

sanity:
## Problem

#240 / Phase 1 shipped the in-room DM wire format and contract validation:
`ChatRoomStateV1` now carries `DirectMessagesV1` with sender-signed
`AuthorizedDirectMessage` + recipient-signed `AuthorizedRecipientPurges`,
validated by the room contract WASM. But neither the UI nor `riverctl` knew
how to compose, decrypt, or purge those messages — `grep -rn "DirectMessage"
ui/src/ cli/src/` returned nothing before this PR.

This PR closes #243 by adding both consumer surfaces in one go so they stay
in wire-format lockstep.

### `river-core`

New helpers shared by both consumers so wire bytes are byte-identical
regardless of which client posted them:

- `ecies::seal_dm_for_recipient` / `unseal_dm_from_sender` — per-message
  randomized ECIES envelope. Fresh ephemeral X25519 keypair + random nonce
  per call. Wire format is `[ephemeral_pub: 32 | nonce: 12 | aes_gcm: variable]`.
  Deliberately separate from the deterministic `encrypt_secret_for_member`
  helper — that one is safe only because the plaintext IS a 32-byte
  high-entropy secret; generalising it to attacker-controlled message
  bodies would leak `plaintext_A XOR plaintext_B` on key reuse.
- `direct_messages::compose_direct_message` — encrypt + sign + cap-check
  in one call.
- `direct_messages::open_direct_message` — recipient-side decrypt that
  works without the `ecies-randomized` feature (chat-delegate friendly).
- `direct_messages::advance_recipient_purges` — version-bump-and-union
  envelope builder used for both single-message and thread-wide purges.

6 new ECIES tests + 7 new end-to-end-helper tests covering: round-trip
through the full `ComposableState`, refusal of self-DMs, future-skew
rejection, body-cap enforcement, version monotonicity, recipient/key
mismatch, and a send → purge → re-send dance that confirms the tombstone
gate drops a re-delivered message.

### UI (Phase 2)

- New `components/direct_messages` module with two modals:
  - `DmThreadModal`: per-pair thread, decrypt-on-display, composer with
    body-size guard, "Purge thread" button.
  - `DmInboxModal`: lists every open DM thread the local user has in the
    current room, unread-first then by recency.
- "Send Direct Message" button in the member-info modal opens the thread
  modal for that pair.
[...]

PR #258, opened 2026-05-16
fix(dm): Enter-to-send, move DMs to left rail, monotonic publish version

sanity:
Three fixes after the #244 production release surfaced real UX issues with first users (zorolin tried DMs in the official room and surfaced both Enter-to-send and the buried entry point).

## 1. Enter-to-send in the DM composer

The DM textarea ignored Enter; only the Send button worked. The room-message composer already had Enter-to-send / Shift-Enter-newline; this PR mirrors it.

Refactored the existing send closure into a no-arg `do_send` callable so both `onclick` (button) and the new `onkeydown` (Enter) invoke the same path.

## 2. "Direct Messages" section in the left rail under Rooms

zorolin's feedback: the previous DM entry point — a button in the Members panel — was confusing, and per-room-only made cross-room DMs invisible. Both Ian and zorolin agreed DMs belong in the left rail next to Rooms.

- New `components/room_list/dm_rail_section.rs` renders a flat list across ALL rooms. Each row: peer nickname + room name + unread badge. Sorted unread-first, then most-recent.
- Hidden when empty.
- Click a row → opens the existing `DmThreadModal` for that `(room, peer)`.
- Removed the now-redundant button from `members.rs` and dropped the unused `DmInboxModal` + `DM_INBOX_OPEN` signal + `open_dm_inbox` helper.

## 3. Monotonic counter for the web-container publish version

The `seconds / 60` scheme failed on 2026-05-16: on-network version had drifted ahead to 30000208 (likely a past clock-skew bump that stuck), timestamp gave 29649402, fdev rejected the publish. Recovery required guessing a number ahead of all past values.

Replaced with `published-contract/contract-version.txt` — committed counter, incremented by `cargo make sign-webapp` each publish and committed alongside. Same pattern in `test-contract/contract-version.txt` for the test path.

Updated AGENTS.md, `.claude/rules/river-publish.md`, and the global `~/.claude/skills/river-publish/SKILL.md` (pushed to ~/.claude in commit 5f22880).

## What's deferred (not in this PR)

- Outbound DM plaintext persistence → #256 (needs chat-delegate WASM bump + `legacy_delegates.toml` migration entry). Another agent is taking that one.
- Test-publish sed-replace WASM corruption when test contract ID ≠ 43 chars → #257 (blocked Playwright verification for this PR; not a production-path issue).

## Testing

- `cargo check -p river-ui --target wasm32-unknown-unknown --no-default-features` ✓
[...]

PR #259, opened 2026-05-16
feat(dm): persist outbound DM plaintext in chat delegate (#256)

### LINKED ISSUE #256: feat: persist outbound DM plaintext in the chat delegate
## Problem

In-room DMs (shipped in #244 / #243) carry only ECIES ciphertext on the wire — only the recipient can decrypt the body. As a side effect, the sender's UI shows their own sent DMs as the placeholder \"sent — ciphertext only\":

> Direct messages with zorolin
> sent — ciphertext only — 2026-05-16 15:48:05

You typed the message, you sent it, but you can't see what you typed afterwards. That's terrible UX, and it's the same in `riverctl dm list` (outbound shows as ciphertext-only too).

## Why we didn't just store plaintext locally

The current design keeps zero extra plaintext anywhere: the ciphertext in `ChatRoomStateV1.direct_messages` is canonical for both sender and recipient. Adding in-memory plaintext to the UI would:

- Mislead users into thinking sent DMs persist across reloads / devices — they don't, the map evaporates.
- Add a special-cased lifecycle (populate on send, evict on purge, lose on reload) that's hard to reason about.
- Be the wrong shape if/when the same user opens River on a second device: device A's outbound DMs would still be ciphertext-only on device B.

## Proposed solution

Persist outbound DM plaintext in the **chat delegate**, the same way the delegate already stores room secrets / signing keys / migrated rooms. That gives us:

- **Persistence across reloads.** Open River next week, your sent DMs are still there with the original plaintext you typed.
- **Multi-device convergence.** Both your devices already share the chat-delegate state — sent-DM plaintext should ride the same channel.
- **Correct purge semantics.** When a recipient signs a purge envelope tombstoning your DM, the contract drops the ciphertext on merge; we should drop the matching plaintext from the delegate on the same event.
- **Encrypted at rest.** Delegate storage already encrypts data tied to room secrets, so we don't widen the threat model.

### Storage shape

Per chat-delegate convention, add a new key namespace:

```rust
// in `common/src/chat_delegate.rs`
pub const OUTBOUND_DMS_STORAGE_KEY: &[u8] = b\"outbound_dms\";

// stored value (serialized CBOR):
pub struct OutboundDmStore {
    // Keyed by (room_owner_vk, recipient_member_id, purge_token).
    // Vec<(...)> not HashMap so it serialises through any wire format
    // including JSON — see existing `purge_versions` precedent in
    // `DirectMessagesSummary` for why.
    pub entries: Vec<OutboundDmEntry>,
}
pub struct OutboundDmEntry {
    pub room_owner_vk: VerifyingKey,
    pub recipient: MemberId,
    pub purge_token: PurgeToken,
    pub timestamp: u64,
    pub plaintext: String,
}
```

### Lifecycle

| Event | Action |
|---|---|
| UI/CLI sends DM | After `compose_direct_message` succeeds and the state delta is applied locally, append an `OutboundDmEntry` to the delegate store. |
| UI renders own outbound bubble | Look up `(room, peer, token)` in `OUTBOUND_DMS`. Hit → render as `BodyKind::Plaintext`. Miss → render as `BodyKind::Placeholder` (\"sent — ciphertext only\") — this is the fall-through for DMs sent before the delegate started tracking them. |
| Recipient purges; tombstone arrives | `post_apply_cleanup` already drops the ciphertext from contract state. Mirror that in the delegate: after the state-merge handler runs, prune any `OutboundDmEntry` whose `(recipient, purge_token)` is now in the recipient's purge envelope. |
| Per-pair cap reached on send | Apply the same eviction policy the contract uses — drop the oldest outbound entry for that `(sender, recipient)` pair. Keeps the store bounded. |
| Delegate WASM key changes | The existing legacy-delegates migration path (`legacy_delegates.toml`) already moves `rooms_data` across keys. Add `outbound_dms` to that migration so a delegate WASM rebuild doesn't orphan sent plaintext. |

### Out of scope

- **Backfilling pre-delegate outbound DMs.** DMs sent under the current production UI before this lands stay as \"sent — ciphertext only\" forever. Documented limitation.
- **Edit/delete of sent DMs.** That's a different design conversation (the contract doesn't currently support DM edits — only purge-by-recipient).
- **End-to-end encrypted device-to-device sync of the outbound store.** The chat-delegate already encrypts its state tied to the room secret, so co-room devices converge; cross-room delegate sync is the same scope as room-secret sync.

### Implementation sketch

1. `common/src/chat_delegate.rs`: add `OUTBOUND_DMS_STORAGE_KEY`, `OutboundDmStore`, `OutboundDmEntry`. Pure types, no contract changes.
2. `ui/src/components/app/chat_delegate.rs`: add `save_outbound_dm()` / `load_outbound_dms()` / `prune_outbound_dms_for_purges()`, all going through the existing `ChatDelegateKey` machinery.
3. `ui/src/components/direct_messages/dm_thread_modal.rs`: on send path, append to local cache + queue a delegate write. On render path, consult the cache for outbound bubbles.
4. `cli/src/commands/dm.rs`: same change — `dm list` should consult the equivalent outbound-store via the delegate when the user is running with delegate-backed signing.
5. `delegates/chat-delegate/src/`: handle the new storage key in the existing handler chain. Add migration entry if WASM hash changes.

### Estimate

~1-2 days. No protocol change (room contract WASM stays as-is), so no migration tag is needed for the contract.

### Related

- #243 (in-room DMs scope)
- #244 (UI + riverctl, where the placeholder ships)
- #252 (share invite via DM, partially shipped in #244)

[AI-assisted - Claude]

sanity:
## Problem

In-room DMs (#243 / #244) carry only ECIES ciphertext on the wire — only the recipient can decrypt the body. The sender's own UI / \`riverctl dm list\` rendered their own sent DMs as \"sent — ciphertext only\", which is terrible UX: you typed the message, you sent it, but you can't see what you typed afterwards.

## Approach

Persist the sender's plaintext in the **chat delegate** (same backing store as room secrets / signing keys / migrated rooms). This gives us:

- **Persistence across reloads.** Sent DMs still readable next week.
- **Multi-device convergence.** Co-room devices share delegate state.
- **Correct purge semantics.** Recipient tombstones drop the plaintext in lockstep with the contract's ciphertext eviction.
- **Encrypted at rest.** Delegate storage already encrypts data tied to the room secret.

### Wire format (\`common/src/chat_delegate.rs\`)
- \`OUTBOUND_DMS_STORAGE_KEY = b\"outbound_dms\"\`
- \`OutboundDmStore { entries: Vec<OutboundDmEntry> }\` — \`Vec\` not \`HashMap\` so JSON serialisation works per the \"non-string map keys in JSON-serialized API types\" bug-prevention pattern. JSON + CBOR round-trip tests pin both shapes.

### UI (\`ui/src/components/app/chat_delegate.rs\`, \`direct_messages.rs\`, \`dm_thread_modal.rs\`)
- \`OUTBOUND_DMS\` global signal, hydrated from the delegate on startup (and from any legacy delegate as part of the existing migration).
- \`save_outbound_dm()\` appends an entry, enforces \`MAX_DM_MESSAGES_PER_PAIR\` cap eviction, queues a delegate save. All signal mutation routed through \`crate::util::defer\` per the WASM signal-safety rules.
- \`prune_outbound_dms_for_purges()\` drops cache entries whose ciphertext is no longer present in any room. Wired into a \`use_effect\` in \`App()\` that subscribes to \`ROOMS\` **only** — not \`OUTBOUND_DMS\` — to avoid self-triggering on its own writes. Skips the \`with_mut\` entirely when nothing to remove.
- Thread modal's view memo snapshots the cache once and consults it for outbound bubbles; miss → legacy \"sent — ciphertext only\" placeholder for back-compat.

### CLI (\`cli/src/commands/dm.rs\`, \`cli/src/storage.rs\`)
- New \`outbound_dms.json\` side file managed by \`Storage\`.
- \`dm send\` writes plaintext after a successful state delta.
- \`dm list\` consults the cache and renders plaintext for outbound bubbles; misses fall back to \`<sent: ciphertext only>\`.
- Per-pair cap eviction split into \`apply_per_pair_cap\` for unit testing.

### Migration (\`legacy_delegates.toml\`)
- New **V18** entry for the pre-#256 delegate (\`code_hash=6229c2a1…\`, \`delegate_key=11db7236…\`).
- \`fire_legacy_migration_request\` now also probes legacy delegates for the \`outbound_dms\` blob so a rebuild doesn't orphan sender plaintext.
- **Room contract WASM intentionally unchanged** — only \`chat_delegate.rs\` types added to river-core, and dead-code elimination strips them from the room contract binary. Verified by rebuilding from pre-#256 source and confirming the local-build hash matches a rebuild with my changes.

## Testing

- \`cargo test -p river-core\`: JSON + CBOR round-trip of \`OutboundDmStore\` (+ empty store).
- \`cargo test -p riverctl --lib dm\`: per-pair cap eviction (over-cap drops oldest by timestamp; pairs are isolated).
- \`cargo test -p river-core --test migration_test\`: V18 entry validates (BLAKE3 invariants, no duplicates).
- \`cargo make check-migration\`: confirms old hash present in TOML.

Manual UI test deferred to deploy step.

## Out of scope (per issue #256)

- Backfilling pre-delegate outbound DMs — sent under older clients stay as ciphertext-only forever.
- Edit/delete of sent DMs.
- Cross-room delegate sync (same scope as room-secret sync).

Closes #256

[AI-assisted - Claude]


sanity:
## Multi-model review complete

Ran 4 internal reviewers (code-first / testing / skeptical / big-picture) + Codex CLI. Two critical issues found and fixed in f053d4df:

### BLOCKING fixes
1. **Cold-start prune wipes the cache** (skeptical reviewer). The original prune compared the cache to live `direct_messages` entries and removed any orphan — which fired at startup before room state had finished loading, destroying the entire cache. Switched both UI and CLI prune helpers to act ONLY on entries listed in some recipient's `AuthorizedRecipientPurges` envelope. Purge envelopes are monotonic, so their presence is always a true tombstone signal regardless of freshness.
2. **Migration-done flag was not scoped to legacy set** (Codex P1). Every WASM bump that added a new legacy entry was silently blocked for any user who already migrated under a previous set, making their data appear lost. Fixed by suffixing the localStorage key with a BLAKE3 fingerprint of the current `LEGACY_DELEGATES` set, so adding/removing entries auto-invalidates the flag. Pre-existing bug surfaced by this PR's V18 entry; affects every prior delegate upgrade since the gating was introduced in #254. Pinned by `legacy_migration_flag_is_scoped_to_set`.

### IMPORTANT fixes
3. **Lost-update race on concurrent saves** (skeptical). Single-flighted `save_outbound_dms_to_delegate` via an atomic-flag pair so rapid sends coalesce into "current save + one final catch-up".
4. **Legacy hydrate clobbered fresher current entries** (skeptical). On key collision, keep whichever has the larger `timestamp`.
5. **No regression test for the issue-256 symptom** (testing reviewer). Extracted a pure `lookup_outbound_plaintext` helper and added two tests pinning the load-bearing lookup tuple.

### Documentation
- `.claude/rules/river-publish.md`: documented the two-key probe set (`ROOMS_STORAGE_KEY`, `OUTBOUND_DMS_STORAGE_KEY`) and the per-key conflict-resolution rule.
- `AGENTS.md` "In-Room Direct Messages": added Phase 5 block on the outbound cache.
- `cli/src/storage.rs::save_outbound_dms`: threat-model note about plaintext-at-rest.

### Coordination note
PR #258 (Enter-to-send / DMs to left rail) also touches `dm_thread_modal.rs::send`. Both are MERGEABLE against main today but conflict against each other; whoever lands second rebases the send closure.

CI rerunning now.

[AI-assisted - Claude]

Issue #257, opened 2026-05-16
test-publish: webapp sed-replace ID-length mismatch corrupts WASM


sanity:
## Problem

\`cargo make compress-webapp-test\` rewrites the production contract ID
(\`raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv\`, 43 chars) to the test
contract ID using a byte-for-byte \`sed\` over the built WASM/HTML/JS.
The replacement assumes both IDs are the same length. They're base58-
encoded 32-byte keys, which gives 43-, 44-, or occasionally 45-char
output depending on leading zero bytes. Production happens to be 43;
a fresh test key has a roughly 1/256 chance of producing a 43-char ID.

When the lengths differ, the \`sed\` replacement shifts every byte after
the first match — corrupting the WASM. The browser then fails with
\`WebAssembly.instantiateStreaming(): illegal flag value …\`.

## Hit twice on the #244 work cycle

Both times: regenerated test keys after a verification-cycle restart,
ran \`cargo make publish-river-test\`, got a 44-char contract ID, fdev
published successfully, then the browser couldn't load the corrupted
WASM. Worked around by re-running \`generate-test-keys\` in a loop until
a 43-char ID came up — costly (mean ~256 attempts).

## Proposed fixes (any of these would do)

1. **Embed a placeholder, not a hard-coded prod ID, in the source.** The
   UI knows its own contract ID at runtime via \`window.location\` (see
   \`get_invitation_base_url\`). Have the runtime-derived path be the
   only source of truth and drop the build-time sed pass entirely.
2. **Match prod-ID-length on key generation.** Loop inside the cargo-
   make task until the generated key produces a 43-char base58 ID;
   document it.
3. **Replace the literal-bytes \`sed\` with a length-preserving rewrite.**
   E.g., target only HTML/JS where the ID appears as text, and rebuild
   the WASM section table after any binary edit. More work, less
   fragile.

#1 is the cleanest. The UI already does the right thing in practice
(invite URLs derive from \`window.location\`); the only embedded constant
that needs to go is the \`base_url\` fallback in invite_member_modal,
and that's already feature-flagged for non-browser tests.

## Reproduction

\`\`\`bash
cargo make generate-test-keys
fdev get-contract-id --code target/wasm32-unknown-unknown/release/web_container_contract.wasm \\
  --parameters target/webapp/webapp-test.parameters | wc -c
# 45 (i.e. 44 chars + newline) — bad
[...]

Issue #252, opened 2026-05-16
feat: share invite via DM


sanity:
## Problem

After #244 / #243, in-room DMs work between co-members of the same room. But the most natural cross-room invite flow — "I'm already in room A with Bob, I want to invite Bob to my new room B" — requires:

1. Generating an invite URL in room B's Invite Member modal
2. Copying it
3. Switching to room A
4. Opening Bob's DM thread
5. Pasting

That's a lot of context switching. The invite URL also lives in the DM body as plain text, so even after #244's planned URL linkification it still takes a click to act on.

## Proposed UX

Two entry points, both terminating in "send Bob a DM in some shared room containing an invite URL for room B":

### From the member-info modal in room A

Add a "Share an invite via DM…" action below "Send Direct Message". Click → submenu listing every room where:

* the local user has invite permission, AND
* the target member is NOT already a member.

Pick one → generate the invite for that room → compose a DM in room A pre-filled with a short message + the invite URL → user reviews → Send.

### From the Invite Member modal in room B

Add a "Send to a co-member" tab next to the existing "Share Link" / "Full message" tabs. Lists rooms the local user shares with people who are NOT already in room B; picking one drops you into the per-member picker, then drops the invite URL into a DM compose dialog.

## Out of scope

* Inviting someone you do NOT already share a room with — defer to `freenet/mail`, same boundary as #230.
* Inviting multiple recipients at once — first cut is per-person.
* Automating accept on the receiver side — receiver still clicks the URL to land on the invite acceptance modal.

## Implementation notes

* The invite URL helper already exists (`ui/src/components/members/invite_member_modal.rs`).
* DM compose helper is `compose_direct_message` in `river-core`.
* "Co-member" lookup: walk `ROOMS.map`, for each room compute the membership-set intersection between the local user and the target peer's `member_vk`. Cache per click.

## Estimate

Half a day of UI work, no contract / wire-format / crypto changes (rides entirely on the #244 wire format).

[AI-assisted - Claude]

PR #260, opened 2026-05-16
fix(ui): intercept invite-link clicks to prevent gateway iframe reload

sanity:
## Problem

Ivvor reported in the official River room (2026-05-16): *"room invites in DM seem to lock up most of the river UI."*

Investigation traced this to a long-standing interaction that #244/#258/#252 made the dominant path:

1. `message_to_html` linkifies invite URLs in both room messages and DM bodies. `finalize_anchors` rewrites the gateway URL to a same-origin path with `target=\"_blank\" rel=\"noopener noreferrer\"`.
2. The gateway iframe sandbox is `allow-scripts allow-forms allow-popups` with NO `allow-popups-to-escape-sandbox`. Under those flags, browsers SUPPRESS `target=\"_blank\"` popups on same-origin links and fall back to navigating the iframe in place.
3. The in-place navigation reloads River with the ~700-char `?invitation=...` still attached. `App()` re-mounts, the synchronizer restarts, ROOMS re-hydrates from the delegate, and `ReceiveInvitationModal` renders in its \"Preparing to subscribe…\" state. The user's open DM thread / draft / current room are destroyed during the reload, which is exactly the \"lockup\" Ivvor saw.

#244 + #258 + #252's picker made every shared invite a `/v1/contract/web/.../?invitation=...` URL — i.e. always same-origin, always triggering the reload path.

## Fix

Document-level click interceptor at `ui/src/components/invite_click_interceptor.rs`. On every click:
- Walk up to the nearest `<a>`.
- If `anchor.href()` contains `?invitation=`, extract the code, `preventDefault()` + `stopPropagation()`, and write the code into `INTERCEPTED_INVITATION_CODE` (new GlobalSignal).
- `App()` bridges that signal into the existing local `receive_invitation` signal — same path the URL-bar entry flow already uses. The existing `ReceiveInvitationModal` opens in place. **The iframe never reloads, so the current room / open DM / draft text are preserved.**

Idempotent (atomic compare-exchange flag, safe to call on every render). Scope is narrow — only anchors with `?invitation=` are intercepted; everything else still gets the existing `target=\"_blank\"` behaviour.

## Testing

- `cargo check -p river-ui --target wasm32-unknown-unknown --no-default-features` ✓
- `cargo test -p river-ui --bins --no-default-features` ✓
- Manual Playwright verification deferred (test-publish path is blocked by #257). Will visually verify against production after merge.

## Related

- #244 / #258 — shipped the DM UI + linkify that exposed this interaction.
- #252 — picker that compose-and-paste invite URLs into DMs.
- #257 — test-publish sed-replace fragility blocking visual verification this PR.

[AI-assisted - Claude]


sanity:
Pass 1 review consolidated (skeptical + Codex):

**Codex P2-A**: `.try_read()` not `.read()` on the intercepted signal (AGENTS.md Dioxus signal-safety rule). ✓
**Codex P2-B**: synchronous clear (project rule "Never defer signal clears in use_effect" — defer-clear would re-fire). ✓
**Skeptical High**: only same-origin URLs intercepted — foreign-gateway invites fall through to the browser's default new-tab behaviour. ✓
**Skeptical Medium**: middle-click / Ctrl-Cmd-Shift-Alt-click pass through to the browser (preserves "open in new tab"). ✓

Skeptical findings verified safe (no action needed): atomic flag, text-node click targets, `anchor.href()` query inclusion, stop_propagation impact, defer re-fire on None.

Fixed in d650a5f0. CI watching.

[AI-assisted - Claude]

PR #262, opened 2026-05-17
fix(dm): plain-language copy for thread modal footer

sanity:
## Problem

Two strings in the DM thread modal used developer jargon that users couldn't reasonably parse:

* **"Purge thread"** — "purge" is contract terminology, and the button doesn't actually delete the whole thread: it only removes the inbound half (messages the peer sent you), because that's the only half a recipient is allowed to tombstone.
* **"End-to-end encrypted to this member's room key. The contract enforces caps; the gateway can't read content."** — "caps" was shorthand for `MAX_DM_MESSAGES_PER_PAIR` (a 100-msg internal cap that isn't a security property anyway). "Gateway" was just wrong: HTTP gateways serve the web app, but the message itself travels Freenet's peer-to-peer mesh. The relevant fact is "only the recipient can decrypt this."

## Approach

Replace both strings with concise, jargon-free copy:

| Where | Before | After |
|-------|--------|-------|
| Button label | `Purge thread` | `Delete their messages` |
| Button tooltip | `Tombstone every inbound DM in this thread on the network.` | `Removes messages they sent you from the network. Cannot be undone.` |
| Footer disclaimer | `End-to-end encrypted to this member's room key. The contract enforces caps; the gateway can't read content.` | `Only {peer} can read these messages.` |

The button label now accurately reflects the asymmetric scope (only inbound is purgeable) without claiming the action is local-only; the tooltip handles the "and it's network-wide / irreversible" qualifier. The disclaimer drops the cap mention (an internal limit, not a security property) and uses the peer's nickname so the assertion lands concretely.

Also fixed the file-level rustdoc to match the new button text.

## Testing

- [x] `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` — clean
- [ ] Manual: open a DM thread, confirm new strings render with the peer's nickname interpolated; click "Delete their messages" and confirm the existing purge flow still works

UI-only change; no contract / delegate / wire format touched, no migration required.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/direct_messages/dm_thread_modal.rs
+++ b/ui/src/components/direct_messages/dm_thread_modal.rs
@@ -1,5 +1,5 @@
 //! Per-pair DM thread modal: decrypts inbound DMs, composes outbound ones,
-//! and offers a "Purge thread" button that produces a fresh
+//! and offers a "Delete their messages" button that produces a fresh
 //! `AuthorizedRecipientPurges` envelope tombstoning every inbound DM in the
 //! current view.
 
@@ -524,13 +524,15 @@ fn DmThreadModalBody(room: VerifyingKey, peer: MemberId) -> Element {
                     }
                     div { class: "flex justify-between items-center pt-1",
                         span { class: "text-[10px] text-text-muted",
-                            "End-to-end encrypted to this member's room key. The contract enforces caps; the gateway can't read content."
+                            "Only "
+                            span { class: "text-accent", "{peer_label}" }
+                            " can read these messages."
                         }
                         button {
                             class: "text-xs text-text-muted hover:text-red-400 transition-colors",
                             onclick: purge_thread,
-                            title: "Tombstone every inbound DM in this thread on the network.",
-                            "Purge thread"
+                            title: "Removes messages they sent you from the network. Cannot be undone.",
+                            "Delete their messages"
                         }
                     }
                 }
--- a/ui/src/components/direct_messages/dm_thread_modal.rs
+++ b/ui/src/components/direct_messages/dm_thread_modal.rs
@@ -281,7 +281,7 @@ fn DmThreadModalBody(room: VerifyingKey, peer: MemberId) -> Element {
         let existing = pair_message_count(&room_data.room_state.direct_messages, self_id, peer);
         if existing >= MAX_DM_MESSAGES_PER_PAIR {
             send_error.set(Some(format!(
-                    "Per-pair cap reached ({}/{}). Ask the recipient to purge older DMs in this thread before sending more.",
+                    "This thread is full ({}/{} messages). Ask them to delete some older messages first.",
                     existing, MAX_DM_MESSAGES_PER_PAIR
                 )));
             return;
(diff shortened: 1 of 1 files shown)


sanity:
## Review response (HEAD: bdcb3c3e)

Multi-model review (4 internal agents + Codex) ran on commit `f05e991a`. All five reviewers landed; nothing blocking. Findings addressed in this commit (`bdcb3c3e`):

### Substantive copy fixes

| Reviewer(s) | Finding | Resolution |
|---|---|---|
| Codex P2 / code-first nit / skeptical M2 | "Only {peer} can read these messages." is inaccurate — inbound DMs are encrypted to MY key, so MY device reads them too. | Reworded to **"Only you and {peer} can read these messages."** — accurate for both directions of the thread. |
| Code-first nit 1 / skeptical M1 | Tooltip didn't explain the asymmetry: only the inbound half is purged; sender's outbound bubbles remain in their own view via the #256 cache. | Tooltip now reads **"Removes messages they sent you from the network. Your own sent messages stay. Cannot be undone."** |

### "Finish the fix" sweep — adjacent jargon (big-picture / skeptical N1)

Other strings on the same modal still contained the class of jargon the user originally complained about. Swept in this PR:

* `"Per-pair cap reached (N/M). Ask the recipient to purge older DMs in this thread before sending more."` (×2 call sites) → `"This thread is full (N/M messages). Ask them to delete some older messages first."`
* `"Failed to add DM to local state (verify it via console)."` → `"Couldn't send this message. Please try again."`
* `"No inbound DMs to purge in this thread."` → `"No messages from them to delete."`
* `"Purge build failed: {e}"` → `"Couldn't delete those messages. Please try again."` (technical detail preserved in `warn!` log)
* `"Failed to apply purge envelope (verify it via console)."` → `"Couldn't delete those messages. Please try again."`
* `"Room is no longer loaded."` → `"This room is no longer loaded."`

### Findings explicitly NOT acted on

* **Code-first / Codex: "Cannot be undone" is loosely true** — skeptical separately verified protocol enforcement (`direct_messages.rs:798-806` rejects a re-broadcast of any DM whose token is tombstoned), so the claim holds at the protocol layer. Re-sending the same plaintext under a new ID isn't "undoing" — that's a new message.
* **Skeptical M2: "Only you and {peer}" overpromise re: device compromise / coerced disclosure.** Acceptable trade-off vs. the prior "gateway" jargon; the disclaimer is about the cryptographic property at the network layer, not threat modeling against device-level adversaries. Same scope as e.g. Signal's "your messages are encrypted" copy.
* **Testing reviewer: no automated test added.** No DM-modal Playwright spec exists today; standing up the harness just to pin three strings is poor leverage (string-pinning tests churn with every copy change). Manually verified via Playwright MCP — see screenshot.

### Manual verification

Built UI with example data, served on :8082, opened in real browser:

* Footer renders **"Only you and Kyle Martin (Member) can read these messages."** with the peer name styled `text-accent`.
* Button reads **"Delete their messages"** with the new tooltip on hover.
* Five other reworded strings are unreachable via example-data flow; code-checked.

Screenshot in commit `bdcb3c3e`'s working tree (`dm-modal-after-copy-fix.png` at repo root, not committed).

Will re-run the 4 internal reviewers + Codex on `bdcb3c3e` since the code changed since the prior round.

[AI-assisted - Claude]

sanity:
## Round-2 review response (HEAD: 1ecb0284)

Re-review of `bdcb3c3e` came back: Codex clean, code-first 2 soft concerns, skeptical 2 Important. Addressed both Important findings in this commit:

* **Skeptical I1 + code-first concern 2** — "Please try again." on deterministic failure paths (`DeltaFailed`, `advance_recipient_purges` Err, purge `apply_delta` Err) lures users into pointless click loops since retrying byte-identical input reproduces the same contract-level rejection. Reworded the 3 affected branches to **"— something went wrong."**. Technical detail still preserved in the `warn!` / `error!` logs.
* **Skeptical I2** — Per-pair cap message's "Ask them to delete some older messages first." was ambiguous about whose messages need deleting. Only the recipient can sign a purge envelope, so the asker needs to ask the recipient to clear messages **the asker has sent them**. Reworded both call sites to **"Ask them to delete some of the older messages you've sent."**
* **Code-first concern 1** — Added a code comment near the disclaimer pinning the dependency on the OUTBOUND_DMS local plaintext cache (#256), so a future maintainer who removes the cache also knows to revisit the disclaimer ("Only you and {peer}" — on-wire encryption is to {peer} only; the "you" half is the cache).

### Findings explicitly NOT acted on

* **Skeptical N1: "those messages" vs "the messages" plural edge.** `purge_thread` early-returns on the empty-token case via the "No messages from them to delete." path, so by the time the error branch fires there's always ≥1 message — "those" reads naturally.
* **Code-first disclaimer-vs-marketing tradeoff.** The disclaimer is user-facing UX copy; the user cares about who can SEE plaintext, not who holds an X25519 key. The code comment is the audit trail.

Round-2 diff is 1 file / +20 / -6 / pure copy + comment edits, all directly addressing the prior round's findings. Skipping a third reviewer pass per the rule's "trivial scope / direct response" exception. Waiting on CI green on `1ecb0284`.

[AI-assisted - Claude]

Issue #246, opened 2026-05-14
Memory growth in River UI after #242 deploy


sanity:
## Report

Ivvor on Matrix (2026-05-14, post-#242 deploy):

> Latest River is going through memory like crazy

Filing as a separate issue from the rotation bug (#245 fixes that one) since the memory behaviour is independent of the secrets back-fill.

## What we don't know yet (please update if reproducible)

- Whether it's the browser tab's heap or somewhere else (Freenet node memory? Chat-delegate state?)
- Whether it grows monotonically vs spikes on specific actions (send message, room switch, member set change)
- Whether disabling private rooms / private-room creation makes it go away

## Suspects, ordered by likelihood

1. **Repeated delegate rotation churn.** The #245 fix re-emits encrypted_secrets for newly-joined members at every version; on a long-lived room with many rotations this is O(versions × members) per join. Probably fine, but worth measuring.
2. **Chat-delegate state retention.** PR #235 added per-room delegate state (signing keys, member set caches, per-version secret cache at `secret_keys::secret(room, v)`). The per-version secret cache grows unboundedly — there's no GC for old versions.
3. **Re-render loops on `ROOMS` signal mutations.** PR #239 added private-room UI surfaces. A signal that fires too eagerly during rotation propagation could trigger render storms.
4. **In-room DM state accumulation.** PR #240 added `DirectMessagesV1` (currently empty in production since Phase 2 UI is not shipped). The `direct_messages: DirectMessagesV1 { messages: [], purges: [] }` field appears in every state update; if something is cloning the whole state for each update, the empty vecs get allocated repeatedly. Cheap individually, but accumulation possible.
5. **Subscription leak.** The `EnsureRoomSubscription` once-per-session dedup in `ui/src/components/app/chat_delegate.rs:162` is per-session, but the underlying subscription on the node side might be leaking handles.

## Repro ideas (need Ivvor or a fresh tester)

1. Open River, take a heap snapshot in DevTools.
2. Create a private room, invite a peer (use #245 build).
3. Exchange ~10 messages.
4. Take another heap snapshot.
5. Compare — what objects are growing? Retained size by class?

If easy: a 5-minute screen recording of `chrome://process-internals/` showing the tab's RSS over time would tell us a lot.

## Relates to

- #242 (this deploy)
- #245 (rotation back-fill — fixes a separate bug from the same report)
- #243 (Phase 2 DM UI — would expand the state surface further, worth de-risking memory first)

[AI-assisted - Claude]

[...]

PR #264, opened 2026-05-17
fix(ui): repopulate room_data.secrets on every state ingestion path (#251)

### LINKED ISSUE #251: Private room: new members see encrypted-only messages until refresh — secrets HashMap not repopulated on network state updates
## Bug report

User report from Ivvor (Matrix, 2026-05-15) testing the new private-room functionality:

> Created private room and sent test message to make sure it was published. Then used invite for second user. They got in but everything showing as encrypted. Tried sending messages back and forth, but both sides weren't syncing. Refreshing the River tabs fixed things and both sides could communicate back and forth as usual.
>
> Then tried to add a third user. They got in, but could see only encrypted messages. For some reason, owner of the room couldn't send messages. Once again, refreshing the tabs eventually got everything synced and all three could communicate as normal

The signature pattern: **a hard refresh of the tab fixes everything**. That points at an in-memory state that's only populated on app boot.

## Diagnosis

Private message decryption at render time reads from `room_data.secrets: HashMap<u32, [u8; 32]>` (see `ui/src/components/conversation.rs:295-332`). When the secret for a message's version is missing from that map, the renderer prints `[Encrypted message - secret vN not available (have: [...])]`.

The HashMap is populated by `RoomData::set_secret` (`ui/src/room_data.rs:152`), which is called from these two paths only:

1. `get_response.rs:178-222` — on the initial `GET` response when subscribing to the room contract.
2. `response_handler.rs:351-398` — inside the `crate::util::defer` block after rooms are loaded from the chat delegate (i.e. on app startup / refresh, or after `LoadRooms`).

**No path repopulates `room_data.secrets` from `room_state.secrets.encrypted_secrets` when a *subsequent* network update arrives.** `update_notification.rs::handle_update_notification` calls `room_synchronizer.update_room_state` / `apply_delta`, neither of which decrypts new `encrypted_secrets` entries for the local member.

## Timeline that produces the symptom

1. User B accepts the invite → subscribes to the room contract → receives initial state via `GET`.
2. At GET time the room state does **not** yet contain encrypted secrets for B — the owner's delegate hasn't run the PR #245 back-fill yet. `get_response.rs:197-198` logs `\"No encrypted secrets found for member\"` and leaves `room_data.secrets` empty.
3. The owner's delegate receives a `ContractNotification` showing the new member, runs back-fill (`delegates/chat-delegate/src/subscription.rs::handle_contract_notification`), publishes a state update that includes encrypted secrets for B.
4. B's UI receives the update via `update_notification.rs`. The `ChatRoomStateV1`'s `secrets.encrypted_secrets` field is merged into local state, **but `room_data.secrets` is never repopulated** — `update_room_state_inner` (`room_synchronizer.rs:807-`) and `apply_delta` (`room_synchronizer.rs:55-`) don't run the decrypt-and-`set_secret` loop.
5. B keeps rendering everything as `[Encrypted message - secret vN not available (have: {})]`.
6. B refreshes the tab → rooms get loaded from the delegate → `response_handler.rs:351-398` runs and decrypts all secrets → everything is suddenly readable.

The same window also explains the symmetric \"messages aren't syncing\" complaint: each side sends a message encrypted at version vN (which the *sender* has in their map but the *receiver* doesn't), so neither side can decode the other's messages until both refresh.

The owner's \"can't send messages\" complaint on the 3-member case is less certain — possibly the owner's UI also got into a state where it didn't have a usable current secret after rotation triggered by the new member. Worth confirming during repro.

## Suggested fix

Extract the secret-decryption block at `get_response.rs:178-222` / `response_handler.rs:351-394` into a helper on `RoomData` (e.g. `RoomData::repopulate_secrets_from_state(&mut self)`) and call it from:
- `room_synchronizer.rs::update_room_state_inner` after the `merge` succeeds (private rooms only)
- `room_synchronizer.rs::apply_delta` after `apply_delta` succeeds (private rooms only)

The helper should iterate `room_state.secrets.encrypted_secrets` filtered by `member_id == self`, and for each version not already in `self.secrets`, decrypt and insert via `set_secret`.

After that, the existing per-private-room rebuild of `actions_state` already in those paths (`room_synchronizer.rs:122-152` and `:891-922`) should automatically decrypt newly-available action messages. Regular messages are decrypted at render time, so they'll resolve as soon as Dioxus re-renders after the signal mutation.

## Repro

1. Owner creates a private room (one node) and posts a message.
2. Invite a second node user.
3. On the second node, observe initial render shows `[Encrypted message - secret v0 not available (have: [])]` for all owner's messages and for subsequent owner messages — and the owner cannot see the new member's messages either.
4. Without refresh: send messages from both sides — neither side can read the other's.
5. Refresh either tab → that side now decrypts everything.

## Related

- PR #245 added the delegate back-fill that produces the `encrypted_secrets` for the new member. That part appears to be working — the secrets get into the room state. The bug is on the *consumer* side of those secrets, in the UI sync path.
- This issue surfaced after Ian's fix in built `2026-05-14 19:02 UTC-5` (PR #248 for the \"leave room sticks\" bug) — likely because that fix made it possible to actually exercise the private-room invite flow end-to-end without hitting the prior tombstone bug first.

## Other report items from the same Matrix message (separate concerns)
- Memory spikes during room reloads/refreshes — addressed by #250 (demote per-update roomstate dumps from `info!` to `debug!`).

[AI-assisted - Claude]

sanity:
## Problem

Closes #251.

When a private-room invitee subscribes, the initial `GET` response often arrives *before* the owner's chat delegate runs the PR #245 back-fill of `encrypted_secrets`. The back-fill ships in a subsequent state update.

The decrypt-into-`room_data.secrets` loop was duplicated in `get_response.rs` and `response_handler.rs::handle_load_rooms_response`, but the network-update paths in `room_synchronizer.rs` (`apply_delta_inner` and `update_room_state_inner`) merged the new encrypted blobs into `ChatRoomStateV1` and never decrypted them into the in-memory map that the renderer reads.

Net effect (Ivvor on Matrix, 2026-05-15):
- Newly-joined member renders every message as `[Encrypted message - secret v0 not available (have: [])]`.
- Both sides "aren't syncing" — each sends messages encrypted at a version the other doesn't have in their HashMap.
- Hard-refreshing fixes it, because reload re-runs the `LoadRooms` decryption loop.

## Approach

Extract the decryption block into `RoomData::repopulate_secrets_from_state(&mut self) -> usize`:
- No-op on public rooms.
- Snapshots blobs in `room_state.secrets.encrypted_secrets` whose `member_id == self` and whose `secret_version` isn't already in `self.secrets`.
- Decrypts each, calls `self.set_secret(...)`.
- Aligns `self.current_secret_version` with the contract's `secrets.current_version` (preserves prior get_response / load-rooms behaviour).

Call it from every state ingestion path:
- `get_response.rs:178+` (initial GET) — replaces the inline duplicate.
- `response_handler.rs::handle_load_rooms_response` (delegate-load merge) — replaces the inline duplicate.
- `room_synchronizer.rs::apply_delta_inner` — new call site, **before** the per-private-room `actions_state` rebuild so newly-decrypted action messages resolve in the same tick.
- `room_synchronizer.rs::update_room_state_inner` — same.

The helper is idempotent (filtered by `!self.secrets.contains_key(version)`), so it's safe to call on every update without measurable cost on the steady-state case.

## Testing

Three unit tests in `ui/src/room_data.rs`:

- `repopulate_secrets_after_delegate_backfill_picks_up_new_blob` — direct reproduction of Ivvor's repro: invitee state starts with `secrets` empty and `encrypted_secrets` empty (initial GET); first call to helper is a no-op; owner delegate back-fills the blob; second call to helper decrypts it and populates the map. Pinned the idempotency case too (third call decrypts 0).
- `repopulate_secrets_skips_blobs_for_other_members` — helper must not try to decrypt blobs sealed for a different member's key.
- `repopulate_secrets_is_noop_on_public_room` — pinned the early-return for `is_private() == false`.

All 95 `river-ui` tests pass. Workspace `cargo make test` green. `cargo clippy` clean for the touched code (no new warnings).

Manual repro will be exercised in the batched Playwright run before the next River publish.

## Other items from Ivvor's report

The "memory growth" complaint is tracked separately in #246. The "owner can't send messages" sub-symptom in the 3-member case is plausibly the same bug from the owner side — owner's tab also gets stuck if a rotation-driven update arrives that brings in a new current_version for which the owner already had the secret but the apply_delta path didn't refresh `current_secret_version`. This patch's `current_secret_version = Some(current_version)` line at the end of `repopulate_secrets_from_state` covers that case symmetrically.

[AI-assisted - Claude]

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -338,6 +338,19 @@ design.
     duplicate-version dedup in `RoomSecretsV1::apply_delta` (`secret.rs:140-145`).
   - The previous "weekly rotation" trigger was removed — it only fired while
     the UI was open, which defeated the point of a scheduled rotation.
+- **In-memory secret repopulation** (#251): `room_data.secrets:
+  HashMap<u32, [u8; 32]>` is `#[serde(skip)]` and must be rebuilt from
+  `room_state.secrets.encrypted_secrets` after EVERY network state
+  ingestion — initial GET, refresh/suspension GET, delegate-load merge,
+  `apply_delta`, and full-state `update_room_state`. The helper
+  `RoomData::repopulate_secrets_from_state` is the single source of
+  truth; any new ingestion path MUST call it (the
+  `repopulate_secrets_call_sites_pinned` test pins the existing call
+  sites by source-grep so dropping one fails CI). Skipping the helper
+  causes the bug from #251: newly-joined private-room members render
+  `[Encrypted message - secret vN not available]` until they
+  hard-refresh, because the back-filled blob arrives in a *subsequent*
+  state update that the in-memory map never sees.
 - Key files:
   - `common/src/room_state/privacy.rs`, `secret.rs`, `configuration.rs`
   - `common/src/key_derivation.rs`
--- a/ui/src/components/app/freenet_api/response_handler.rs
+++ b/ui/src/components/app/freenet_api/response_handler.rs
@@ -19,7 +19,7 @@ use crate::components::app::sync_info::{RoomSyncStatus, SYNC_INFO};
 use crate::components::app::{CURRENT_ROOM, ROOMS};
 use crate::room_data::CurrentRoom;
 use crate::room_data::Rooms;
-use crate::util::ecies::{decrypt_secret_from_member_blob, decrypt_with_symmetric_key};
+use crate::util::ecies::decrypt_with_symmetric_key;
 use crate::util::owner_vk_to_contract_key;
 use ciborium::de::from_reader;
 use dioxus::logger::tracing::{error, info, warn};
@@ -37,7 +37,6 @@ use std::collections::HashMap;
 pub use subscribe_response::handle_subscribe_response;
 pub use update_notification::handle_update_notification;
 pub use update_response::handle_update_response;
-use x25519_dalek::PublicKey as X25519PublicKey;
 
 /// Handles responses from the Freenet API
 pub struct ResponseHandler {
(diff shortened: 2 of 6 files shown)


sanity:
Round 2 of review findings addressed in 27016a61:

**Codex [P2] — Repopulate secrets for existing-room GETs.** Fixed in `get_response.rs::is_existing_room`. Both sub-branches (imported-room wholesale-replace, existing-room merge) now call `repopulate_secrets_from_state()` so a refresh/suspension GET that's the first arrival of our back-filled blob populates the in-memory map.

**Codex [P3] — Stale `room_secrets` snapshot in notifications.** Fixed in both `apply_delta_inner` and `update_room_state_inner`. The pre-merge clone is gone; `room_secrets` is captured AFTER repopulate, so a delta carrying both a new secret AND new private messages can decrypt the notification preview correctly. `update_room_state_inner`'s `pending_notification` tuple was widened to carry the secrets out of `with_mut`.

**Testing reviewer (blocking) — regression test doesn't pin the call sites.** Added `repopulate_secrets_call_sites_pinned`: a source-grep pin test that uses `include_str!` to read each ingestion-path file and assert the expected count of `repopulate_secrets_from_state(` call sites. Removing the call from any path fails CI. Mirrors the "source-grep pin test in the driver's test module" pattern from `bug-prevention-patterns.md`.

**Code-first reviewer — `current_secret_version` downshift.** Docstring expanded to explain the unconditional reassignment relies on the `RoomSecretsV1` invariant (`current_version == max(versions)`, monotonically non-decreasing under merge), so the merge that immediately precedes the call cannot move the pointer backwards.

**Big-picture reviewer — AGENTS.md update.** Added a bullet under "Private Room Support" documenting the in-memory-repopulation invariant and pointing future authors at the source-grep pin test.

**Skeptical reviewer findings not addressed:**
- M1 (corrupt-blob CPU amplification): owner is the trust root for `EncryptedSecretForMemberV1` signatures; a malicious owner can already inflict worse damage than O(N) decrypts per state ingestion. Cost is bounded by the number of versions the owner has signed, which is itself bounded by rotation frequency. Not in scope for #251.
- S2 (per-blob clone overhead): future optimization; deferred.
- S3 (cross-device rotation race): pre-existing, separate concern.

96 tests pass (the 3 new unit tests + the new pin test); clippy clean for touched code; the only remaining clippy warning in the touched region is a pre-existing `type_complexity` on the apply_delta_inner pending_notification tuple, not introduced by this PR.

Re-running multi-model review on the new HEAD SHA per the per-CODE-CONTENT rule.

[AI-assisted - Claude]

Issue #3, opened 2024-10-06
In order to garbage-collect bans allow the root inviter key to be changed


sanity:
Also allow the new root inviter key to invite all current non-banned members directly but while they retain a link to who invited them, all other semantics would remain the same except the newly invited members would be invited by the root inviter.

Issue #110, opened 2026-02-16
Secret rotation may not reach members pruned from active list


sanity:
## Problem

PR #109 introduced a dual-state membership model where members with valid invitations may not be in the active members list (they get pruned by `post_apply_cleanup` if they have no recent messages).

When a member is pruned from the active list but still has a valid invitation:
1. They won't receive rotated secrets during a secret rotation
2. When they rejoin (send first message), they might not be able to decrypt messages encrypted with the rotated secret

## Steps to Reproduce (Theoretical)

1. User accepts invite to a private room
2. User gets pruned from active members list (no recent messages)
3. Room owner rotates the room secret
4. User sends their first message (re-added to active list)
5. User may not be able to decrypt messages encrypted with the new secret

## Expected Behavior

Members with valid invitations should either:
- Not be pruned if they have encrypted secrets, OR
- Receive the rotated secret when they rejoin

## Suggested Test

```
User accepts invite → gets pruned → owner rotates secret → user sends message → verify decryption works
```

## Context

Found during architectural review of PR #109. The `build_rejoin_delta` logic handles re-adding members, but secret distribution during rotation may not account for pruned-but-authorized members.

[AI-assisted - Claude]

iduartgomez:
Fixed and shipped in PR #272 (merged 2026-05-17, commit 9baf8500).

The rotation back-fill now reaches pruned-but-invited members:
- `build_rotation_encrypted_secrets` (common/src/room_state/secret.rs) is the single source of truth for the per-member rotation blob set; both the UI fast-path (`RoomData::rotate_secret`) and the delegate catch-up (`run_rotation`) emit byte-identical blobs.
- `post_apply_cleanup` exempts a member holding an owner-issued `AuthorizedEncryptedSecretForMember` blob at the current secret version from inactivity-prune (scoped to current_version so stale recipients are still pruned; banned members are never exempted).

Pinned by: test_member_with_encrypted_secret_survives_cleanup, test_banned_member_with_encrypted_secret_is_still_pruned, test_stale_secret_recipient_is_pruned_after_rotation, test_ban_race_with_encrypted_secret_converges_to_pruned (common/src/room_state.rs).

Documented in .claude/rules/private-rooms.md.

PR #270, opened 2026-05-17
fix(room-contract): accept private messages at any known secret version (Bug #3 PR A)

sanity:
## Problem

Ivvor reported on Matrix (2026-05-17):

> First attempt to create and share private room failed at sync stage.
> Invites shared using old link method. Two invitees processed invite and
> room opened, but no messages shown. Invitees could see each others
> nicknames, but not nickname of room owner. Only room owner can send
> messages, but they don't show up, even as encrypted, to other members.
> According to room info, owner is on secret #3. The other members are
> on secret #0.

The smoking gun is in `common/src/room_state/message.rs`:
`MessagesV1::apply_delta` enforced two checks that, together, paralysed
the room as soon as the owner's local `current_version` got ahead of any
invitee's.

1. **Strict version mismatch**. Any private message encrypted at v_new
   arriving while the invitee was still at v_old caused `apply_delta` to
   return `Err`. The composable macro short-circuited via `?` and dropped
   the ENTIRE `ChatRoomStateV1Delta` — including the message itself,
   membership updates, and any in-band secrets-delta in the same payload.
   The invitee's UI therefore never even saw the encrypted message, and
   back-fill from a peer later was impossible (nothing to back-fill).

2. **Complete-distribution gate**. `has_complete_distribution(members)`
   had to be true for the current version. A single member missing a
   blob froze the entire room for messages with no recovery path unless
   that member came online and the owner re-issued their blob.

Compounding this, `RoomSecretsV1::apply_delta` mutated `self.versions`
before running later checks. If any later check failed, the
half-mutated state survived as the new baseline and broke CRDT
idempotence: re-applying the same failing delta would now succeed
because the version was already pushed, but `current_version` /
`encrypted_secrets` / pruning never ran.

## Approach

Three minimal changes, all in `common/`:

1. **Relax `MessagesV1::apply_delta` (message.rs)**: accept any
   `secret_version` that has a corresponding owner-signed record in
   `parent_state.secrets.versions`, instead of requiring exact match to
   `current_version`. A message at a fabricated version (no signed
   record) is still rejected as defense in depth. Author safety is
   already enforced by `MessagesV1::verify`'s member-or-owner signature
   check, and `ChatRoomStateV1::post_apply_cleanup` purges DMs from
   banned authors.

2. **Remove the `has_complete_distribution` gate from
   `MessagesV1::apply_delta`**. A single member's missing blob no longer
   freezes the room for everyone.

3. **Make `RoomSecretsV1::apply_delta` transactional (secret.rs)**:
   stage all changes on a `working` clone of `self`, then commit
   (`*self = working`) only if every check passes. The composable
   `apply_delta` is now all-or-nothing for the secrets sub-state.

Also: gate the wasm32-only `behavior` binding in `conversation.rs` so
the file is clippy-clean on native (pre-existing unused-variable
warning, unrelated to Bug #3 but blocks `cargo make clippy`).

## Migration

Adds `legacy_delegates.toml` entry V20 capturing the pre-change chat
delegate key:

- `code_hash =
  05a96a22b12f548d51a9d785c374790f6dcb22986e0d98e9f0dc023ffbaa4ca0`
- `delegate_key =
  49b626c43a0b0bf9a8506a5266c690afdebc1dc6ce8a7d6d9397b8b1f1b21a4d`

`cargo make check-migration` passes. Users' room data migrates to the
new delegate key on next load.

## Testing

Four new regression tests in `common/tests/private_room_test.rs`:

- `message_at_older_or_newer_known_secret_version_is_accepted`:
  invitee at `current_version=0` with v0+v1 signed records accepts a
  message at v1 (Ivvor's exact repro). Defense in depth: a message at
  v99 (no signed record) is still rejected.
- `single_member_missing_blob_does_not_freeze_room`: Alice/Bob
  members, Bob missing v1 blob — Alice's v1 message is still accepted.
- `secrets_apply_delta_is_transactional_on_failure`: a bad delta
  leaves `RoomSecretsV1` byte-identical to its pre-call state. Pre-fix
  this leaks the failed `versions.push`.
- `delta_with_rotation_plus_message_at_new_version_applies_atomically`:
  combined rotation+message delta applies cleanly (regression guard for
  the happy path the composable macro ordering enables).

Verified each test fails on `1d150125` (PR #268 merge, current main)
without the fix and passes with it. All 251 river-core tests, the
`migration_test`, and the full `cargo make test` suite pass.

## Test plan

- [x] `cargo test -p river-core` (251 tests pass)
- [x] `cargo test -p river-core --test migration_test` (4 pass)
- [x] `cargo make test` (full workspace)
- [x] `cargo make check-migration`
- [x] `cargo fmt --all`
- [x] `cargo clippy -p river-core -p room-contract -p chat-delegate --all-targets -- -D warnings`
- [x] `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync`
- [x] Each new test verified failing without fix, passing with fix

## Two-PR plan

This is part 1 of 2 for Bug #3.

**PR B** (will follow after this merges):
- UI back-fill path so previously-rejected messages re-sync after the
  contract validation is relaxed (the `get_response.rs` PUT changes).
- `chat-delegate` prune-fix per freenet/river#110.

Filing PR B as a follow-up because it touches different files (UI +
delegate) and benefits from being reviewed separately once the
contract-validation change has landed and migration is in place.

[AI-assisted - Claude]


sanity:
## Response to Codex review finding (P1: stale secret versions after rotation)

Thanks for the careful review. The concern is valid but I think the trade-off as implemented is the right one for PR A's scope. Let me lay out the reasoning.

### The trade-off

The relaxation permits a stale sender to send a message at an older \`secret_version\` AFTER the room has rotated. Banned members previously holding that older secret could decrypt such a message. Codex proposed tightening to reject \`secret_version < current_version\`.

### Why I'm keeping the relaxation as-is

1. **The exposure is bounded.** Banned members already have the plaintext of every message sent during the old version's tenure. The only NEW exposure is "messages sent at v_old AFTER rotation by lagging clients" — which is rare in practice and should self-correct within seconds as clients pick up the rotation.

2. **The tighter rule (\`secret_version >= current_version\`) doesn't actually help in any case my relaxation matters.** Trace it with the composable macro (secrets applies before recent_messages):
   - **Rotation+message in same delta**: secrets bumps current_version to v_new before recent_messages.apply_delta runs, so \`secret_version == current_version\` already passes. Both rules accept.
   - **Rotation only, message arrives later**: by the time the message arrives, current_version is already at v_new. Message at v_new passes both rules. Message at v_old (stale sender) — Codex's rule rejects, my rule accepts.
   - **Bug #3 (Ivvor): rotation delta partially applied**: with PR A Fix #3 (transactional \`RoomSecretsV1::apply_delta\`), rotations are now all-or-nothing. The "partial mutation" precondition that produced Ivvor's symptom can't happen. If the rotation fails, current_version stays at v_old and v_new is NOT in versions — my rule rejects (correctly: receiver doesn't have v_new anyway). If the rotation succeeds, current_version=v_new and the v_new message passes both rules.

3. **Codex's rule rejects legitimate batched deltas.** Consider: owner has historical messages at v1 and sends them batched with a rotation to v2. The composable macro applies the secrets-bump first, so when recent_messages.apply_delta runs, current_version=v2 and \`message_at_v1 >= v2\` is FALSE. Codex's rule rejects historical messages in batched deltas. My rule (existence in versions) accepts them.

4. **Confidentiality of post-rotation messages is a SENDER responsibility, not a contract responsibility.** The contract has no notion of "the right secret" — it only has owner-signed records. Senders should always encrypt with the latest secret they have. PR B (UI back-fill) will add the path for stragglers to rotate forward promptly.

### Mitigations applied

- Comment block in \`common/src/room_state/message.rs\` documenting the trade-off explicitly (commit \`f5d393a\`).
- The defense-in-depth "secret_version must exist in versions" check still blocks fabricated version numbers.

### Follow-up

If the post-rotation exposure becomes a measurable problem in practice, the right place to fix it is the sender's UI logic: before sending, refresh state and use the latest secret. I can file this as a separate issue if you want — but I think the current shape is a net improvement over both the pre-fix rule (Bug #3) and the proposed strict rule (breaks batched historical deltas).

[AI-assisted - Claude]

sanity:
## Codex re-run (HEAD f5d393a) — same P1 finding, additional argument

I ran Codex again on the updated branch. Same P1 finding, same proposed mitigation (\`secret_version >= current_version\`). Worth digging deeper since Codex has held the position twice.

### Why Codex's strict-monotonic rule has a real cost

Trace it against a legitimate Freenet-native flow: **permissionless contract migration / full-state resync**.

A peer with a stale view of the room receives \`UpdateData::State(new_state)\`, which calls \`ChatRoomStateV1::merge\`, which calls \`apply_delta\` with the FULL diff. Suppose the room rotated v0→v1→v2 over time. The state contains historical messages at v1 (sent during v1's tenure) and current messages at v2.

After secrets.apply_delta runs (first, per the composable macro ordering), \`parent_state.secrets.current_version = v2\`. Then recent_messages.apply_delta sees the historical message at v1.

- **My rule** (\`secret_version in versions\`): accept. v1 is owner-signed, in versions.
- **Codex's rule** (\`secret_version >= current_version\`): \`v1 < v2\` → REJECT. **Historical messages get dropped on resync.**

This violates the "self-validating state" principle in AGENTS.md:

> A key benefit of fully-authorized state: it enables **permissionless contract migration**. When contract WASM changes, anyone can migrate state from the old contract to the new one because the state is self-validating.

If a peer can't migrate its history after a rotation, the migration story breaks. The contract MUST accept any owner-signed-version + valid-member-author combination.

### The authorization boundary is the signature, not the version

Per AGENTS.md "State Authorization Rule":

> Every piece of data in contract state must be cryptographically authorized. Messages must have a valid signature from a verified member at the time they are added.

"At the time they are added" means: at the time the contract receives the message. The signature attests to the author. The secret_version is metadata, not an authorization control. Conflating the two produces the bug we're fixing.

### Sender-side enforcement (the right place)

Filed freenet/river#271 to enforce sender-side: before encrypting an outbound private message, the UI should fetch the freshest secrets state and use \`max(version where local has key)\` — never \`current_version\`. This addresses Codex's concern at the layer where it belongs.

### Concrete trade-off recap

| | Pre-fix | My rule (PR A) | Codex's rule |
|---|---|---|---|
| Ivvor's repro (stale receiver) | BROKEN | OK | OK |
| Full-state resync of historical msgs | OK (because pre-fix had \`==\`, which receivers at the rotated current_version reject for old msgs too — so this was ALSO broken pre-fix in distributed settings, just less visibly) | OK | BROKEN |
| Stale sender post-rotation | Rejected (good) | Accepted (small confidentiality leak) | Rejected (good) |

**Net assessment**: my rule is strictly better than pre-fix and better than Codex's rule because it preserves the migration/resync invariant. The post-rotation confidentiality leak is a real but bounded concern, addressable at the sender layer (#271).

Keeping the current code. Documenting the trade-off in the comment block (commit \`f5d393a\`) so future reviewers find this context.

[AI-assisted - Claude]

PR #275, opened 2026-05-17
feat(dm): archive UX overhaul + delete confirmation

### LINKED ISSUE #266: feat(dm): "Hidden conversations" sub-menu to un-hide DM threads
## Problem

PR #265 (issue #261) implemented the local-only hide-stale-DM-threads filter for the left rail. The original issue marked a "Hidden conversations" sub-menu as **optional for v1**, so it was deferred — but it should be tracked.

Today, a thread can be un-hidden only by:

1. Sending an outbound DM to the peer (calls `unhide_dm_thread` via the Codex P1 fix).
2. Receiving a new inbound DM from the peer (strict `>` on `last_any_ts` revives the thread via the filter).

There is no UI path to un-hide a thread if the user just changed their mind. The hidden entry persists across reloads (it rides `OUTBOUND_DMS_STORAGE_KEY`) and across devices, so a misclick on Hide is recoverable only via one of the two implicit paths above.

## Approach

The infrastructure is already in place:

- `unhide_dm_thread(room_owner_vk, peer)` in `ui/src/components/app/chat_delegate.rs` is implemented, tombstoned-against-late-hydration (Codex P3), and unit-tested.
- `HIDDEN_DM_THREADS` is a `GlobalSignal<HashMap<(VerifyingKey, MemberId), HiddenDmThreadEntry>>` so a submenu can iterate it directly.
- The `HiddenDmThreadEntry` carries `room_owner_vk + peer + hidden_at_ts`. Resolving the room display name + peer nickname goes through the same path `DmRailSection` uses.

Suggested UX (open to revision):

- Settings → "Hidden conversations" panel (or kebab menu in the DM rail header).
- One row per hidden entry, showing room name + peer nickname + relative "hidden N days ago".
- Click "Show again" → `unhide_dm_thread(...)` and the thread reappears in the rail.

## Scope notes

- v1 deliberately shipped without this submenu (per the original #261 wording — "optional"). This issue is to track that v1 follow-up, not block #265.
- Keep the panel local-only — no contract changes, no second delegate storage key.
- The same legacy-migration probe set continues to apply because hides ride `OUTBOUND_DMS_STORAGE_KEY` (Phase 6 note in `AGENTS.md`).

## References

- #261 — the original "hide stale DM threads" issue.
- PR #265 — implementation.
- `ui/src/components/app/chat_delegate.rs` — `hide_dm_thread`, `unhide_dm_thread`, `decide_hide_action`.

[AI-assisted - Claude]

sanity:
## Problem

Two follow-ups to the in-room DM work (#243 / #265):

1. **#266** — the **Hide** button in the DM thread modal header sat next to the close ✕ and was repeatedly mistaken for it (Ian via Matrix). One click in the wrong place silently took the thread off the rail.
2. **Same modal** — **Delete their messages** in the footer fired destructively on a single click with no undo path. Mis-click and the recipient's DMs to you are gone from the network.

Neither was a data-loss bug, but both were UX traps that real users tripped on the day after #261 shipped.

## Approach

UI-only changes; no contract / delegate / WASM touched. Keeps the on-wire shape (`OutboundDmStore.hidden_threads`) and the internal Rust API names (`hide_dm_thread`, `HIDDEN_DM_THREADS`, etc.) — renaming them would force a delegate migration for zero functional benefit. The UX surface is renamed everywhere visible: "Archive" matches Gmail / WhatsApp / Signal convention.

Changes:

- Per-row **rollover ✕** in `DmRailSection`. Desktop: hidden until row hover/focus (`group-hover:opacity-100 group-focus-within:opacity-100`). Mobile (`<md`): dimmed always-visible (`opacity-40`) so touch users keep the affordance. Tooltip: *"Archive this conversation. It will return if either of you sends a new DM."*
- **Archived (N)** link at the bottom of the rail. Click expands an inline list of currently-archived threads (sorted by room name, then peer nickname). Each row has an **Un-archive** button — closes #266 (no way back from an archived thread).
- **"Archived — Undo" toast** (~5s) after the rollover ✕ click. Bottom-center, `z-50`, with Undo and Dismiss buttons. The auto-dismiss reaction uses an `expires_at_ms` identity check so a rapid second archive doesn't cancel the second toast early.
- **Hide button removed from the modal header**. Header is now just title + close ✕ — no visual collision.
- **Confirmation modal** for "Delete their messages". Cancel is autofocused (Enter doesn't accidentally commit the destructive action), Esc closes, backdrop click cancels. Delete proceeds with the existing `purge_thread` flow.

References #261. Closes #266.

### Unit tests (Rust, run by `cargo make test`)

New tests in `dm_rail_section`:

- `build_archived_rows_projects_and_sorts` — archived viewer's per-room display data projection, including the (room_name, peer_nickname) sort order and the short-id fallback for missing nicknames.
- `build_archived_rows_falls_back_when_room_missing` — a thread whose owning room is no longer in `ROOMS` (e.g. user left the room) still surfaces in the viewer with the "(unknown room)" placeholder, so the user can still un-archive.
- `build_archive_toast_advances_expiry_per_call` — pins the auto-dismiss identity-check math: back-to-back archives must produce distinct `expires_at_ms` so the first toast's tick doesn't cancel the second's.
- `build_archive_toast_saturates_on_overflow` — defensive: `Date.now()` near `u64::MAX` doesn't wrap to a toast that auto-dismisses instantly.

All existing `filter_rail_entries_*` regression tests still pass (no behavioural change to the filter; the new UX feeds the same `hide_dm_thread(cutoff)` cutoff as the old one did).

### Playwright (`ui/tests/dm-archive-ux.spec.ts`)

- App boots cleanly with the new code paths (no WASM panics, no console errors from the touched modules).
- "Hide" button is gone from the rendered DOM after the PR.
- The rail's no-DMs early-return still kicks in (example data has no DMs), so we don't surface a stray "Direct Messages" header.
- Layout remains stable across 1280 / 768 / 480 viewports — pins the Tailwind generator picking up the new `group-hover` / `md:opacity-*` classes.

### Test plan

- [x] `cargo make test` — all 4 new unit tests pass, no regressions
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` passes
- [x] No delegate / contract WASM touched (`cargo make check-migration` clean)
- [ ] CI green on this branch's HEAD
- [ ] Manual desktop smoke test (rollover ✕, archive viewer expand, un-archive)
- [ ] Manual mobile smoke test (dimmed ✕ tappable, viewer collapsible)

End-to-end Playwright coverage of the archive / un-archive flow is deliberately out of scope: the example-data build has no DMs in its fixture, and adding real ECIES-sealed DMs to example data was more invasive than the value it would add given the strong unit-test coverage above.

[AI-assisted - Claude]

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -317,11 +317,20 @@ design.
     sibling JSON file `outbound_dms.json` in the riverctl data dir
     (consistent with `rooms.json`'s plaintext-on-disk threat model
     — full-disk encryption is the user's responsibility).
-- Phase 6 (PR #265, issue #261) added **hide-stale-DM-threads** —
-  a local-only view filter that lets the user dismiss a DM thread
-  from the left rail. Storage piggybacks **the same**
-  `OUTBOUND_DMS_STORAGE_KEY = b"outbound_dms"` blob — `OutboundDmStore`
-  grew a `hidden_threads: Vec<HiddenDmThreadEntry>` field with
+- Phase 6 (PR #265, issue #261; UX overhaul PR for #266) added
+  **archive-stale-DM-threads** — a local-only view filter that lets
+  the user take a DM thread off the left rail. The user-facing
+  surface uses "Archive" everywhere (button labels, tooltips, toasts,
+  the "Archived (N)" viewer link). The underlying data shape and Rust
+  APIs still use the original "hide" / `hidden_threads` /
+  `hide_dm_thread` / `HIDDEN_DM_THREADS` names; renaming them would
+  force a delegate migration for zero functional benefit. Treat
+  "Archive" as the UX label and "hide" as the implementation noun
+  whenever new code touches this surface — do NOT introduce a fresh
+  rename or the on-wire blob and the visible UI drift.
+  Storage piggybacks **the same** `OUTBOUND_DMS_STORAGE_KEY =
+  b"outbound_dms"` blob — `OutboundDmStore` grew a
+  `hidden_threads: Vec<HiddenDmThreadEntry>` field with
   `#[serde(default)]` so pre-#261 bytes still decode. **Do not add a
   second top-level delegate storage key for hide state**: a new key
   would need its own probe in `fire_legacy_migration_request` and its
@@ -332,9 +341,22 @@ design.
   `chat_delegate::prune_outbound_dms_for_purges`. Filter helper
   `chat_delegate::is_thread_hidden` uses strict `<=`; the rail-side
   pure helper `dm_rail_section::filter_rail_entries` is pinned by
-  `filter_rail_entries_*` tests, and the "click Hide again after
+  `filter_rail_entries_*` tests, and the "click Archive again after
   revival must re-hide" branch is pinned by
   `hide_unhide_rehide_round_trip`.
+  The UX overhaul replaced the old modal-header "Hide" button (which
+  sat next to the close ✕ and was repeatedly mistaken for it — Ian's
+  #266 report) with a per-row rollover ✕ in `DmRailSection`. The
+  same UX overhaul added the destructive-action confirmation modal
+  in front of "Delete their messages" — that footer button now opens
+  a Cancel/Delete dialog instead of firing `purge_thread` directly.
+  The "Archived (N)" viewer at the bottom of `DmRailSection`
+  surfaces every currently-archived `(room, peer)` with an Un-archive
+  control; without it, users had no path back to a thread they had
+  archived in error. Sorting and projection of that viewer go through
+  the pure helper `build_archived_rows`, pinned by
+  `build_archived_rows_projects_and_sorts` and
+  `build_archived_rows_falls_back_when_room_missing`.
 
 ## Private Room Support
 - Messages, metadata, and member nicknames are encrypted with AES-256-GCM.
--- a/ui/src/components/direct_messages/dm_thread_modal.rs
+++ b/ui/src/components/direct_messages/dm_thread_modal.rs
@@ -2,8 +2,14 @@
 //! and offers a "Delete their messages" button that produces a fresh
 //! `AuthorizedRecipientPurges` envelope tombstoning every inbound DM in the
 //! current view.
+//!
+//! The header used to carry a "Hide" button alongside the close ✕; that was
+//! moved to the per-row rollover ✕ in [`crate::components::room_list::dm_rail_section`]
+//! after #266 reported it was visually ambiguous with the close button.
+//! "Delete their messages" now goes through a confirmation modal before
+//! firing the destructive `purge_thread` flow.
 
-use crate::components::app::chat_delegate::{hide_dm_thread, save_outbound_dm, unhide_dm_thread};
+use crate::components::app::chat_delegate::{save_outbound_dm, unhide_dm_thread};
 use crate::components::app::{mark_needs_sync, ROOMS};
 use crate::components::direct_messages::{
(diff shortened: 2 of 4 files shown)


sanity:
## Multi-model review — round 1 + 2

Ran the four-perspective code-first / testing / skeptical / big-picture analysis plus a Codex CLI pass per `.claude/rules/multi-model-review.md`. Two rounds, three commits.

### Round 1 (Codex + internal)

Three issues, all addressed in commit `8cd068ec`:

| # | Severity | File | Finding | Fix |
|---|----------|------|---------|-----|
| 1 | P1 BLOCKING | `dm_rail_section.rs:82` | `ARCHIVE_TOAST.read()` violated repo's signal-safety rule — Firefox/mobile RefCell re-entrant panic risk per AGENTS.md. | Switched to `try_read().ok().and_then(...)` per repo convention. |
| 2 | P2 BLOCKING | `dm_rail_section.rs:90, 396` | `Archived (N)` count and viewer didn't apply the revival filter — a hidden entry whose thread was revived by a newer message correctly showed on the rail but stayed in the archived viewer / count. | Added `current_archived_count()` + `count_currently_archived()` helpers that walk ROOMS for per-pair `last_any_ts` and apply `is_thread_hidden_for`. Two new pinning tests (`build_archived_rows_skips_revived_thread`, `count_currently_archived_keeps_stale_hidden_entries`). |
| 3 | P2 IMPORTANT | `dm_thread_modal.rs:694` | Confirmation dialog had no `tabindex` + initial focus, so Escape stopped working after the first backdrop click or Tab cycle out of the dialog buttons. | Applied existing `member_info_modal.rs` pattern: `tabindex: "0"` + `onmounted -> set_focus`. |

### Round 2 (Codex re-review on `8cd068ec`)

Two follow-up notes:

1. **Confirm dialog still doesn't have a full focus trap.** Tab from the Delete button still leaves the dialog (lands on the composer textarea), at which point Escape no longer bubbles to the dialog-scoped handler. → **Addressed in `74ca861a`** by adding an outer-modal Escape backstop gated on `confirm_delete_open`. A full focus trap is significant complexity for a 2-button dialog; this backstop handles every Escape regardless of focus position.

2. **`try_read` fallback may miss subscriptions in the Firefox re-entrant path.** Per AGENTS.md, a failed `try_read` doesn't subscribe, so a renderer triggered mid-write could miss the next signal-change notification. → **Acknowledged but not changed.** This is the same trade-off the rest of the codebase makes (the alternative is the panic we already fixed in round 1). The mitigation chain still works for user-visible flows: HIDDEN_DM_THREADS writes happen alongside toast writes, so the rail's HIDDEN_DM_THREADS subscription provides the backup re-render. The auto-dismiss's worst case is a 1-frame delay clearing the toast — far better than a Firefox panic. Same trade-off pattern as `build_view`'s `try_read` on the same signal (unchanged in this PR).

### Test coverage delta

Before this PR: 9 tests covering `filter_rail_entries`.

After: 13 tests — added `build_archived_rows_projects_and_sorts`, `build_archived_rows_falls_back_when_room_missing`, `build_archived_rows_skips_revived_thread`, `count_currently_archived_keeps_stale_hidden_entries`, `build_archive_toast_advances_expiry_per_call`, `build_archive_toast_saturates_on_overflow`.

The two `current_archived_count` / dialog-focus paths Codex flagged as test-uncovered both go through pure helpers that ARE covered (`count_currently_archived`, the focus pattern is structurally identical to `member_info_modal`).

### Verdict

Ready to merge after CI green on current HEAD `74ca861a`. All blocking findings addressed; round-2 follow-ups handled (one with a code fix, one with a documented trade-off matching repo convention).

[AI-assisted - Claude]

Issue #6, opened 2024-10-09
Streamline Invitation Process for River Chat Rooms


sanity:
The current invitation process for River chat rooms, requiring manual public key exchanges between inviters and invitees, is cumbersome. This proposal aims to streamline the process by introducing a one-time-use URL invite system.

**Proposal:**
- Inviters generate a one-time-use URL containing an embedded private key, part of a public-private key pair.
- The invitee clicks the URL, which negotiates the invitation and seamlessly adds them to the chat room.

**Benefits:**
- Simplifies the invitation process, reducing friction for new users.
- Enhances user experience by eliminating manual key exchanges.
- Ensures secure and private handling of invitations through negotiated URL use.

**Considerations:**
- The security of one-time-use URLs and their proper invalidation after use.
- Secure storage and transfer of the private key within the URL.

PR #276, opened 2026-05-17
fix(delegate): emit encrypted_secrets for new members on first sync (Bug #6)

sanity:
## Problem

Ivvor reported (2026-05-17 via Matrix): joined a freshly-created private room and saw every message render as `[Encrypted message - secret v0 not available (have: [])]`, plus the active-member panel showed the owner's nickname as `[Encrypted: 3 bytes, v0]`. The `(have: [])` confirmed his local `room_data.secrets` was empty at EVERY version — the room contract state had no `encrypted_secrets` blob for his MemberId at any version.

This is post the PR #270 / #272 / #264 work that fixed contract validation, UI consumption, and UI rotate back-fill — but did NOT address the initial-join back-fill emission on the owner's side. The owner's chat-delegate is what's supposed to fire that back-fill when it sees a `ContractNotification` indicating the member set changed; it never fired for Ivvor.

## Scope (read first)

This PR fixes Bug #6 **for rooms already in delegate storage on the owner's device**: the load-rooms path of the response handler now reliably gets the owner's chat-delegate subscribed to each owner-mode room after the (potentially migrated) signing key has been confirmed on file.

The **imported-identity case** — where an owner imports their identity onto a fresh device and the device has zero rooms in delegate storage — is a separate code path (the rejoin-via-invite flow). That isn't addressed by this PR. If a user reports Bug-#6-like symptoms after an identity-import + cold-rejoin, the work needed is in `room_data.rs::create_room_from_invite` / the InvitedRoomKey flow, not the load-rooms path touched here.

## Approach

Two compounding bugs on the room owner's side, both in the load-rooms path. Exposed whenever the owner's local chat-delegate WASM has been migrated to a new key (which loses the old delegate's `room_sub:` and signing-key secrets — the migration registry doesn't carry those over):

**Bug A — `StoreSigningKey` / `EnsureRoomSubscription` parallel-spawn race.** In `response_handler.rs` (load-rooms-response path) the two requests were fired as independent `safe_spawn_local` tasks per room. No ordering guarantee. If `EnsureRoomSubscription` raced ahead of `StoreSigningKey`, the delegate's signing-key-presence probe rejected the subscribe request (`subscription.rs` `#[cfg(target_family = "wasm")]` early-return: "no signing key on file for this room — call StoreSigningKey first").

Fix: chain `ensure_room_subscription_once` onto the same `safe_spawn_local` task that calls `migrate_signing_key`, AFTER the migration completes. The signing key is then guaranteed to be on file before `EnsureRoomSubscription` lands.

**Bug B — per-session dedup poisoned on failure.** `ensure_room_subscription_once` inserted the room VK into `ENSURE_SUBSCRIPTION_SENT` BEFORE sending the request, and the response was fire-and-forget (the response handler explicitly returned `false` from the pending-request match for `EnsureRoomSubscriptionResponse`). So even after fixing Bug A, any other failure mode — transport blip, a future delegate-side reject — would permanently poison the dedup for the rest of the session.

Fix: route `EnsureRoomSubscriptionResponse` through the pending-request registry via a new `complete_pending_room_subscription_request`, and rewrite `ensure_room_subscription_once` to await the response via `send_delegate_request`. On any failure (transport error, delegate-side `Err` result, unexpected response variant) clear the per-session dedup entry so a future call can retry. Added `ROOM_SUBSCRIPTION_PREFIX` constant so request- and response-side lookup keys can't drift.

## Follow-up commits (review round 2)

**BLOCKING — pending-request key collision per VK** (commit 7e532aee). The first commit keyed the pending-request slot by `room_owner_vk` only. Race scenario: call A times out at 10s → slot is freed; call C registers `sender_C` at the same VK-keyed slot; the delegate's late response for A then resolves call C with A's stale ACK. Fixed by adding a `request_id: RequestId` field to both `EnsureRoomSubscription` and `EnsureRoomSubscriptionResponse` (wire format change on the chat-delegate boundary, with migration entry V23 added for the pre-fix WASM hash), generating a fresh `request_id` per `ensure_room_subscription_once` call, mixing it into the pending-request lookup key on both sides, and having the delegate echo it back in the response. Pinned by `room_subscription_pending_key_round_trips` (updated) and the new `room_subscription_pending_keys_diverge_per_request_id` test.

**IMPORTANT — Migration::Failed log spam** (commit 7e532aee). The chained spawn previously fired `EnsureRoomSubscription` even when `migrate_signing_key` returned `Failed`. A persistent migration failure (delegate unreachable, signature mismatch) then produced log spam per cold-load × per owned-room. Now we skip `EnsureRoomSubscription` on Failed and log once per room. Trade-off documented inline: we lose the theoretical "stale key still on file even though verify failed" recovery path, but in practice that's vanishingly rare and the next cold-load retries from scratch.

**Codex P1 — stale contract id after WASM rebuild** (commit f5f7c0e6). The wire-format change in commit 7e532aee rebuilt `room_contract.wasm` as a side-effect of `cargo make sync-wasm`. The `signing_keys` Vec in the load-rooms path was reading `room_data.contract_key.id()` BEFORE `Rooms::merge()` ran — but `merge()` calls `regenerate_contract_key()`, which migrates the room's contract_key to the new bundled WASM's hash. So the snapshot was stale, and `EnsureRoomSubscription` would have subscribed the delegate to a contract id that no longer exists on the network — silently re-introducing Bug #6 on the upgrade path. Fixed by deriving the contract id via `crate::util::owner_vk_to_contract_key(&owner_vk)`, which hashes the CURRENT bundled WASM. Pinned by `owner_vk_to_contract_key_matches_regenerate_contract_key` in `ui/src/util.rs` (commit 09a7c2a7).

Net effect: the owner's delegate is reliably subscribed across cold loads, including after a delegate-WASM migration AND after a room-contract WASM rebuild. The first `ContractNotification` after a new member joins fires the rotation pipeline, which produces an `encrypted_secret` blob for that member (PR #245 / PR #272 back-fill logic, unchanged).

## Testing

UI tests in `ui/src/components/app/chat_delegate.rs`:

- `ensure_subscription_dedup_blocks_second_call` — baseline dedup contract holds.
- `ensure_subscription_dedup_cleared_after_failure_allows_retry` — **Bug #6 regression**: after `clear_ensure_subscription_sent`, the same VK is insertable again.
- `clear_ensure_subscription_sent_is_scoped_to_supplied_vk` — clearing one VK must not clear unrelated VKs' dedup entries.
- `room_subscription_pending_key_round_trips` — `get_request_key` and the response-handler lookup agree byte-for-byte on the tracking key (now including the `request_id` mix-in).
- `room_subscription_pending_keys_diverge_per_request_id` — **NEW**: two distinct request_ids for the same room MUST produce distinct registry keys, otherwise the cross-epoch race is reintroduced.

UI test in `ui/src/util.rs`:

- `owner_vk_to_contract_key_matches_regenerate_contract_key` — **NEW**: pins the Codex P1 fix: `owner_vk_to_contract_key()` returns the same contract id as `RoomData::regenerate_contract_key()` produces. A future drift between the two functions would re-introduce the "subscribe to stale contract id" bug.

Delegate-side `subscribes_to_room_on_ensure_request` now asserts the delegate echoes back the `request_id` verbatim.

`cargo make test`: all green. UI bin tests: 142 passed (3 new + 139 existing).

Migration entry V23 added (delegate WASM hash changed due to wire-format extension). riverctl version bumped 0.1.58 → 0.1.59 (room contract WASM hash changed; CI gate `check-wasm-sync` enforces this).

## Why not just retry from the warn-and-log path

The previous fire-and-forget path logged `warn!("Failed to fire EnsureRoomSubscription: ...")` on Err but had no retry mechanism, and the dedup-on-send meant subsequent calls within the session were no-ops. Wiring up a retry-after-StoreSigningKey-completes via a feedback signal between the two spawned tasks would be more complex than just chaining them — and the chain is the structurally correct fix because it removes the race entirely rather than papering over it.

The dedup-on-failure clearance is a defense-in-depth measure for any other failure mode (transport blip, future delegate-side rejection paths) so that one transient failure doesn't kill rotation for the rest of the session.

## Related

- PR #270 (contract validation), PR #272 (UI back-fill + #110 prune exemption) — these addressed downstream effects but not the owner-side subscription gap.
- Issue #110 — secret rotation may not reach pruned-but-authorized members. Related but distinct; #110 is the contract-state shape, this PR is the delegate's reachability to drive the rotation.

[AI-assisted - Claude]


sanity:
## Self-review report (Claude, four-perspective + Codex)

### Code-First Analysis

**Independent understanding**: Three correlated changes to the UI's chat-delegate subscription pipeline:

1. Routes `EnsureRoomSubscriptionResponse` through the pending-request registry (new `complete_pending_room_subscription_request`); was previously returning `false` from the pending-routing match.
2. Rewrites `ensure_room_subscription_once` to await the response via `send_delegate_request` and clear the per-session dedup entry on any failure path. Removes the no-longer-needed `fire_ensure_room_subscription` helper.
3. In `response_handler.rs` load-rooms path, removes the separate `owned_rooms_for_subscription` parallel-spawn loop and chains `ensure_room_subscription_once` onto the existing per-room `migrate_signing_key` `safe_spawn_local` task, gated on `owns_room`.

**Alignment with description**: Strong. Code matches description. Two-cause analysis (race + dedup-poison) is accurate; both addressed.

### Testing Assessment

| Test type | Status | Notes |
|---|---|---|
| Unit | OK | 4 new tests covering dedup contract + key-bytes round-trip |
| Integration | Gap (inherent) | The signing-key/subscription race isn't directly testable without WEB_API plumbing — same harness gap noted in `subscription.rs` for the delegate-side WASM probe |
| Simulation | N/A | UI-only path |
| E2E | N/A | Would need Playwright + a live delegate with old signing-key store |

`room_subscription_pending_key_round_trips` pins the byte-level lookup-key contract — a future rename or refactor can't silently re-introduce the Bug #6 hang.

### Skeptical Findings

**Risk level**: Low.

| Concern | Severity | Notes |
|---|---|---|
| `WEB_API.write()` held across `await` in `send_delegate_request` | Low (pre-existing) | Not introduced by this PR; matches every other `send_delegate_request` caller |
| Defer-then-await ordering in chained spawn | Low | Deferred ROOMS mutation runs via setTimeout(0) AFTER the await yields. `ensure_room_subscription_once` only touches `ENSURE_SUBSCRIPTION_SENT` + `WEB_API`, no ROOMS, so no re-entrancy |
| Concurrent same-key `EnsureRoomSubscription` | None | Per-session dedup prevents it |
| `MigrationResult::Failed` still fires EnsureRoomSubscription | None (intentional) | Documented in comment; on Err the dedup clears, allowing retry |

### Big-Picture Assessment

- Goal alignment: yes, fixes Ivvor's owner-side back-fill emission gap.
- Anti-patterns: none. No `#[ignore]`, no weakened assertions, no test removal.
- Removed code: `fire_ensure_room_subscription` (no callers) and `owned_rooms_for_subscription` loop (function inlined into the migrate_signing_key spawn). Clean.
- Scope: focused. Two UI files only. No delegate WASM changes — no migration entry needed.

### Codex Review

> "The changes correctly route EnsureRoomSubscription responses through the pending request registry and sequence subscription requests after signing-key migration without introducing an obvious regression in the reviewed paths."

No Blocking/Important findings from Codex.

### Verdict

Ready to merge once CI is green. Holding for parent's batched-merge decision.

[AI-assisted - Claude]

PR #278, opened 2026-05-17
feat(dm): structured invite DM variant + in-app accept + auto-scroll

### LINKED ISSUE #252: feat: share invite via DM

sanity:
## Problem

The "Share an invite via DM…" flow (#252) currently pastes a raw
`https://…?invitation=<base58>` URL into the recipient's DM thread. The
recipient sees an opaque link that opens the receive-invitation modal in
a new tab — confusing, fragile, and impossible to style.

It also means there's no in-thread affordance: no way to label "this is
an invite", no way to disable an Accept button when the recipient is
already a member, no way to surface a personal message alongside the
invite.

Closes #252 (Phase 3 of the invite-DM redesign).

## Approach

Three layers, landing together:

1. **Wire format** (`common/src/room_state/dm_body.rs`): a new
   `DirectMessageBody` enum (`Text` / `Invite(Box<InvitePayload>)`)
   carried INSIDE the ECIES envelope. The contract never sees this —
   it's a pure client-↔-client concern, so no contract or delegate
   WASM changes are needed (no migration entry). Wire format: 1-byte
   magic `0x80` + CBOR; legacy raw UTF-8 falls through to `Text`
   automatically because `0x80` is a UTF-8 continuation byte that can
   never start a valid UTF-8 string. Existing pre-format DMs continue
   to render unchanged.

2. **Picker rewrite** (`ui/src/components/direct_messages/invite_via_dm_picker_modal.rs`):
   the picker is now the composer — room dropdown + personal-message
   textarea + Send button — and dispatches a structured `Invite`-variant
   DM directly via `send_structured_dm`. No URL paste, no DM_DRAFT
   indirection, no second-modal hand-off.

3. **Recipient render + auto-scroll** (`ui/src/components/direct_messages/dm_thread_modal.rs`):
   inbound DMs whose decoded body is `Invite(payload)` render as a
   structured "Invitation" card with target room name, optional
   personal message, and an Accept button that decodes the embedded
   CBOR `Invitation` and routes through `present_invitation` — the
   same entry point the URL-bar accept flow uses, so there's exactly
   one invitation-accept code path. Already-member rooms disable the
   button and relabel it. A room-key mismatch between the card's outer
   `room_owner_vk` and the embedded invitation's `room` field is
   rejected at decode time so a malicious sender can't mislabel the
   destination.

   The thread also gains auto-scroll: modal-mount → jump to bottom
   instantly; outbound-send → scroll to bottom smoothly; inbound-
   message → scroll only when the user is within ~50px of the bottom
   (don't yank readers of history). Stable `id=\"dm-scroll-container\"`
   on the thread body and a single `use_effect` watching message count
   + a monotonic outbound-send counter.

`riverctl dm list` decodes the new variant too, rendering it as
`[Invitation to room <8-char prefix>…] <message>` so the CLI surface
matches the UI's short-prefix convention for not-yet-a-member rooms.

This PR lands Phases 1, 2 (just unblocked by #275 — Archive UX merged),
and 3 together because they're the minimum coherent surface for the
new invite-DM UX.

## Testing

* **Rust unit tests** pin the pure helpers:
  * `dm_body.rs`: round-trip, magic-byte invariant, legacy fallback,
    deterministic encoding, invalid-CBOR and invalid-UTF-8 paths,
    large-payload round-trip.
  * `dm_thread_modal.rs::decode_invitation_from_payload`: accepts
    well-formed matching CBOR, rejects room-key mismatches, rejects
    invalid CBOR, rejects empty bytes.
  * `cli/src/commands/dm.rs`: `format_dm_body_for_cli` for both Text
    and Invite variants.

* **Playwright smoke tests** (`ui/tests/`):
  * `invite-via-dm-picker.spec.ts`: picker opens with room dropdown,
    personal-message field, Send button enables only after pick.
  * `dm-thread-modal.spec.ts`: stable scroll-container id, empty-
    state copy, composer + Send button state.
  * Fixed flaky member-info selector in both files — the prior
    `aria-label^="Open member info"` selector didn't match anything;
    member rows actually use `title="Member ID: …"`. Tests also now
    handle the random "(You)" entry position in example data.

* Full end-to-end "recipient sees card → click Accept → modal opens"
  coverage requires a real Freenet node (no-sync mode can't
  synthesise an inbound invite-variant DM); the unit tests pin the
  pure decode logic and the Playwright tests pin the structural
  surface. The remaining wiring is the bridge effect in
  `app.rs::PRESENT_INVITATION_REQUEST → receive_invitation`, which
  is the same effect the URL-bar flow has used for months.

* Outbound invite-DMs continue to render via the existing
  `OUTBOUND_DMS` plaintext-summary path (`[Invitation] …`) because
  the cache stores only the summary string. Extending the cache to
  carry the structured body would be additive but is out of scope
  for this PR.

[AI-assisted - Claude]

### DIFF
--- a/cli/src/commands/dm.rs
+++ b/cli/src/commands/dm.rs
@@ -19,6 +19,7 @@ use river_core::room_state::direct_messages::{
     advance_recipient_purges, compose_direct_message, open_direct_message, pair_message_count,
     PurgeToken, MAX_DM_MESSAGES_PER_PAIR,
 };
+use river_core::room_state::dm_body::{decode_body, DirectMessageBody};
 use river_core::room_state::member::MemberId;
 use river_core::room_state::{ChatRoomParametersV1, ChatRoomStateV1Delta};
 use serde_json::json;
@@ -178,6 +179,18 @@ async fn execute_send(
     }
 
     let now = unix_now()?;
+    // Text-variant DMs keep the legacy wire shape (raw UTF-8 bytes,
+    // no magic byte) for backwards compatibility: pre-this-PR clients
+    // in the wild decode raw UTF-8 directly via
+    // `String::from_utf8_lossy`. A new-format `Text` body would
+    // render a stray `\u{0080}` glyph plus garbled CBOR on those
+    // older clients, with no way for them to know better. New clients
+    // round-trip cleanly because `decode_body`'s legacy fallback path
+    // hands raw UTF-8 back as `DirectMessageBody::Text`.
+    //
+    // Only NEW variants (currently `Invite`) need to opt into the
+    // magic-byte + CBOR wire shape — old clients can't render those
+    // anyway, so the rendering regression is moot.
     let auth = compose_direct_message(
         &signing_key,
         &recipient_vk,
@@ -362,17 +375,24 @@ async fn execute_list(
             }
         }
 
-        let body = if is_self_recipient {
-            open_direct_message(&signing_key, msg)
-                .unwrap_or_else(|_| b"<unable to decrypt>".to_vec())
+        // Render the body as a String. For inbound DMs we decrypt the
+        // ECIES envelope, then decode the structured `DirectMessageBody`
+        // (which falls back to legacy raw-UTF-8 → `Text` for pre-#XXX
+        // peers). Outbound DMs go through the local plaintext cache as
+        // before — the cache stores the user-facing string regardless of
+        // wire shape.
+        let body_str = if is_self_recipient {
+            match open_direct_message(&signing_key, msg) {
+                Ok(bytes) => match decode_body(&bytes) {
+                    Ok(body) => format_dm_body_for_cli(&body, &nicknames),
+                    Err(_) => "<unable to decode body>".to_string(),
+                },
+                Err(_) => "<unable to decrypt>".to_string(),
+            }
         } else {
-            // Outbound: check the local plaintext cache. Hit → render
-            // the original plaintext; miss → fall back to the legacy
-            // placeholder so older DMs and DMs sent on another device
-            // are still labelled honestly.
             match outbound_lookup.get(&(msg.message.recipient, msg.purge_token())) {
-                Some(plaintext) => plaintext.clone().into_bytes(),
-                None => b"<sent: ciphertext only>".to_vec(),
+                Some(plaintext) => plaintext.clone(),
+                None => "<sent: ciphertext only>".to_string(),
             }
         };
 
@@ -380,7 +400,7 @@ async fn execute_list(
             counterparty,
             outgoing: is_self_sender,
             timestamp: msg.message.timestamp,
-            body: String::from_utf8_lossy(&body).into_owned(),
+            body: body_str,
             token: msg.purge_token(),
         });
     }
@@ -793,6 +813,41 @@ fn unix_now() -> Result<u64> {
         .as_secs())
 }
 
+/// Render a decoded [`DirectMessageBody`] for `dm list` output.
+///
+/// * `Text` → the text itself.
+/// * `Invite` → a compact one-line summary so an inbound invite DM is
+///   labelled distinctly from prose. The on-wire `invitation_payload`
+///   isn't decoded here — we just surface the room owner key (8-char
+///   prefix) so the recipient can `riverctl room accept` against the
+///   right target if they want to. Personal message (when present) is
+///   appended.
+fn format_dm_body_for_cli(
+    body: &DirectMessageBody,
+    _nicknames: &HashMap<MemberId, String>,
+) -> String {
+    match body {
+        DirectMessageBody::Text { text } => text.clone(),
+        DirectMessageBody::Invite {
+            room_owner_vk,
+            personal_message,
+            ..
+        } => {
+            let room_short: String = bs58::encode(room_owner_vk.as_bytes())
+                .into_string()
+                .chars()
+                .take(8)
+                .collect();
+            match personal_message {
+                Some(msg) if !msg.trim().is_empty() => {
+                    format!("[Invitation to room {}…] {}", room_short, msg.trim())
+                }
+                _ => format!("[Invitation to room {}…]", room_short),
+            }
+        }
+    }
+}
+
 fn format_unix_local(unix_secs: u64) -> String {
     SystemTime::UNIX_EPOCH
         .checked_add(std::time::Duration::from_secs(unix_secs))
@@ -827,6 +882,60 @@ mod tests {
         assert!(parse_hex_token("aa").is_none());
         // 32-char string with a non-hex char.
         assert!(parse_hex_token("zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz").is_none());
+    }
+
+    #[test]
+    fn format_dm_body_for_cli_renders_text_verbatim() {
+        let body = DirectMessageBody::Text {
+            text: "hello peer".to_string(),
+        };
+        let s = format_dm_body_for_cli(&body, &HashMap::new());
+        assert_eq!(s, "hello peer");
+    }
+
+    #[test]
+    fn format_dm_body_for_cli_invite_has_room_prefix_and_message() {
+        use ed25519_dalek::SigningKey;
+        let owner_vk = SigningKey::from_bytes(&[42u8; 32]).verifying_key();
+        let body = DirectMessageBody::Invite {
+            room_owner_vk: owner_vk,
+            invitation_payload: vec![],
+            personal_message: Some("come hang out".to_string()),
+        };
+        let s = format_dm_body_for_cli(&body, &HashMap::new());
+        assert!(s.starts_with("[Invitation to room "), "got: {}", s);
+        assert!(s.ends_with("…] come hang out"), "got: {}", s);
+    }
+
+    #[test]
+    fn format_dm_body_for_cli_invite_no_personal_message() {
+        use ed25519_dalek::SigningKey;
+        let owner_vk = SigningKey::from_bytes(&[42u8; 32]).verifying_key();
+        let body = DirectMessageBody::Invite {
+            room_owner_vk: owner_vk,
+            invitation_payload: vec![],
+            personal_message: None,
+        };
+        let s = format_dm_body_for_cli(&body, &HashMap::new());
+        assert!(s.starts_with("[Invitation to room "), "got: {}", s);
+        assert!(s.ends_with("…]"), "got: {}", s);
+    }
+
+    #[test]
+    fn format_dm_body_for_cli_invite_blank_personal_message_treated_as_none() {
+        use ed25519_dalek::SigningKey;
+        let owner_vk = SigningKey::from_bytes(&[42u8; 32]).verifying_key();
+        let body = DirectMessageBody::Invite {
+            room_owner_vk: owner_vk,
+            invitation_payload: vec![],
+            personal_message: Some("   \t  ".to_string()),
+        };
+        let s = format_dm_body_for_cli(&body, &HashMap::new());
+        assert!(
+            s.ends_with("…]"),
+            "expected blank message to omit suffix, got: {}",
+            s
+        );
     }
 
     #[test]
--- a/cli/src/commands/dm.rs
+++ b/cli/src/commands/dm.rs
@@ -828,17 +828,13 @@ fn format_dm_body_for_cli(
 ) -> String {
     match body {
         DirectMessageBody::Text { text } => text.clone(),
-        DirectMessageBody::Invite {
-            room_owner_vk,
-            personal_message,
-            ..
-        } => {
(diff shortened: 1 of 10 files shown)


sanity:
## Multi-Model Review (Claude, manual four-perspective + Codex CLI)

Performed the four-perspective review manually (the Task tool wasn't surfaced in my agent environment) plus ran `codex review --base main` for an independent model perspective.

### Findings addressed

| # | Reviewer | Severity | Issue | Status |
|---|----------|----------|-------|--------|
| 1 | Code-first (self) | nit | `"Already a member"` used `contains_key`, which mismatched observer-only rooms (user has the room loaded but isn't a member). | **Fixed** in 793dfbc9 — switched to `RoomData::can_participate()`. |
| 2 | Codex P2 | high | The auto-scroll effect did `*OUTBOUND_SEND_COUNTER.read()` inside `use_effect`. Per AGENTS.md "Dioxus WASM Signal Safety", a reactive `.read()` can panic with `RefCell already borrowed` when the write that triggered the notification still holds the write-guard during synchronous subscriber notify (Firefox/mobile). | **Fixed** in 793dfbc9 — switched to `.peek()`; subscription is supplied by the parent memo's `ROOMS` read via `message_count`. Both the apply_delta that bumps message_count and the counter write happen in the same `defer()` block, so the effect re-fires reliably. |
| 3 | Codex P1 | (false positive) | "web-sys `ScrollBehavior` / `ScrollToOptions` features not enabled — build will fail." | **Not a real issue** — `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` AND `cargo build --release` both succeed; the features are pulled in transitively (`conversation.rs` already uses the same APIs). |

### Other perspectives

- **Testing**: 10 unit tests for the new helpers (decode + room-mismatch rejection + invalid CBOR + empty bytes + short_vk_prefix). Playwright covers the structural surface (stable scroll-container id, empty-state copy, composer + Send state). Full E2E recipient-card render requires a real Freenet node; `no-sync` mode can't synthesise an inbound `Invite`-variant DM, which is documented in `ui/tests/dm-thread-modal.spec.ts`.
- **Skeptical**: Checked the bug-prevention patterns (biased! select, fire-and-forget spawns, state cleanup, backoff without jitter) — none apply here (this is pure UI). The auto-scroll's `is_near_bottom` check captures DOM state post-render but pre-scroll, which is the correct ordering. The `OUTBOUND_SEND_COUNTER` wraparound is documented and uses `wrapping_add`. The `decode_invitation_from_payload` security check (room-key mismatch rejection) is pinned by a unit test.
- **Big-picture**: No removed tests, no `#[ignore]` annotations, no commented-out code. The boxed `InviteCardData` in `BodyKind` is a clippy-driven structural change with rationale documented. Outbound invites continue to render via the existing `OUTBOUND_DMS` plaintext-summary path because the cache only stores the summary string — extending the cache to carry the body would be additive but is explicitly out of scope.

### Rebase note

The original push was on top of pre-#276 main (rebased before #276 merged), which left stale WASMs in `ui/public/contracts/` and made CI's `check-delegate-migration` red. Rebased onto current `origin/main` (now includes #276), restoring the WASM blobs and the riverctl version bump. `cargo make check-migration` is clean and `check-delegate-migration` + `check-wasm-sync` are green on the current HEAD.

CI on the new HEAD (793dfbc9) is running.

[AI-assisted - Claude]

sanity:
## Codex round-2 review (HEAD 793dfbc9)

After fixing the round-1 findings, ran `codex review --base main` again. Two findings:

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| 1 | **P1** | At `dm_thread_modal.rs:597-598` the statement `*OUTBOUND_SEND_COUNTER.write() = OUTBOUND_SEND_COUNTER.peek().wrapping_add(1);` lets the read guard and write guard temporaries overlap until the statement ends, risking a RefCell/Dioxus borrow panic on the DM send-success path. | **Fixed** in 9b11de7d — compute `next_outbound_counter` in a separate `let` statement first so the read guard drops before the write guard is taken. |
| 2 | P2 | At `invite_via_dm_picker_modal.rs:349-351` (and similar `oninput`/Send button handlers) the picker mutates a Dioxus signal directly from a synchronous event handler. The codebase rule (AGENTS.md) says event-handler signal mutations should go through `crate::util::defer()`. | **Not applied** — this is a pre-existing codebase-wide pattern (see `send_error.set(...)` calls throughout `dm_thread_modal.rs`, `onclick` direct sets across the UI), not a new behaviour introduced by this PR. The signals in question are LOCAL `use_signal` (`selected_room`, `personal_message`), not `GlobalSignal`s; their writes only notify the owning component, not cross-component subscribers — which is the failure mode the AGENTS.md rule was written to prevent. A codebase-wide refactor to defer EVERY event-handler signal write is a separate change with much wider scope; flagging here for follow-up but not blocking. The picker has been live in Phase 1 / Phase 2 across multiple PRs (#260, #269, #275) and no related panics have been reported. |

Ran a final round of unit tests (10 pass) and Playwright tests (5 pass across chromium/firefox/webkit) against the new commit.

CI is running on 9b11de7d (`check-delegate-migration` ✓, `check-wasm-sync` ✓, `build` + `ui-playwright-tests` pending).

[AI-assisted - Claude]

sanity:
## Codex round-3 review (HEAD 9b11de7d)

Found another P1: `send_structured_dm` in `direct_messages.rs:317-320` does `ROOMS.try_read()` at the top, but the function is called from `safe_spawn_local` contexts (the invite-via-DM picker's `drive_send` awaits the signing delegate before calling it) — and `safe_spawn_local` doesn't push the Dioxus runtime onto the call stack. The bare GlobalSignal read would therefore panic with "Must be called from inside a Dioxus runtime" on the new invite flow.

**Fixed** in 0de2a9f5 — routed the entire preflight (ROOMS read + peer-vk resolution + per-pair cap check + rejoin-bundle build) through a `defer()` block via a oneshot channel, mirroring the way the existing apply-delta write was already routed. Boxed the larger `Ready` variant of `PreflightOutcome` to satisfy clippy.

Worth flagging: this bug would only manifest at runtime in a real Freenet environment (the picker's send path isn't reached in `no-sync` mode because the delegate doesn't run, so the await chain never lands at the GlobalSignal read). The Playwright tests don't exercise the actual send path — they only verify picker UI structure. The cargo unit tests pass because the wasm runtime gating isn't simulated in the native test harness. Codex caught this purely from static analysis of the call graph + the runtime-scoping rule documented in AGENTS.md.

Re-running Codex on the new HEAD per multi-model-review.md's "reviews are per-CODE-CONTENT" rule.

[AI-assisted - Claude]

sanity:
## Codex round-4 review (HEAD 0de2a9f5)

Found a P2 — picker watchdog race: the invite-via-DM picker's terminal `defer` (after `drive_send` returns) calls `last_success_label.set(...)` and `send_error.set(...)` — these are component-LOCAL `use_signal`s. If the watchdog fired before the task completed (slow signing or send), the user could close the picker; the dropped signals would then panic when the late `.set()` calls landed.

**Fixed** in d5f8b18b — capture `still_mine = (INFLIGHT generation == my_generation)` before clearing INFLIGHT; gate the picker-local signal writes on `still_mine`. If the watchdog already fired, we only log and skip the local writes — the send DID land locally inside `send_structured_dm` so the network sync queue delivers the invite regardless of picker UI state.

This is in Phase 1 code (the picker watchdog landed in #260), not strictly a Phase 3 regression, but it's part of this PR's diff vs. main and Codex caught it during the same review.

Re-running Codex on the new HEAD.

[AI-assisted - Claude]

sanity:
## Codex round-5 review (HEAD d5f8b18b)

Found another P2 — my round-4 still_mine fix introduced a new edge case: when the watchdog fires but the picker is still mounted (user didn't close), my "still_mine = false → skip UI" logic silenced the eventual late-completion result, leaving the user with no feedback and a possible duplicate send on retry.

**Fixed** in 3a132eac — make the watchdog force-CLOSE the picker on expiry (previously it left the picker open). This eliminates the "still mounted but watchdog fired" state entirely: every late completion now finds the picker already unmounted, the still_mine gate skips the .set() calls (no panic), and the user sees a closed modal — a clear "something happened, check the DM thread" signal. Trade-off: user loses their typed personal message on timeout, which is strictly better than the prior shape.

Re-running Codex one more time on the new HEAD.

[AI-assisted - Claude]

sanity:
## Codex round-6 review (HEAD 3a132eac) — clean

> I did not identify any discrete, actionable regressions introduced by the diff. The new structured DM invite path appears to preserve legacy text DM compatibility while adding decode/render/accept handling for invite payloads.

This is the first Codex round that returned without findings.

### Summary of the review cycle

| Round | HEAD | Codex finding | Resolution |
|-------|------|---------------|------------|
| 1 | b6907926 | P1: web-sys features (false alarm — build succeeds, conversation.rs uses same pattern) + P2: reactive `.read()` on OUTBOUND_SEND_COUNTER inside use_effect | switched to `.peek()` + parent memo subscription via message_count (793dfbc9) |
| 2 | 793dfbc9 | P1: read/write borrow overlap in `*OUTBOUND_SEND_COUNTER.write() = OUTBOUND_SEND_COUNTER.peek().wrapping_add(1)` | split into separate let-statement (9b11de7d) |
| 3 | 9b11de7d | P1: `ROOMS.try_read()` in `send_structured_dm` runs outside Dioxus runtime when invoked from safe_spawn_local | routed preflight through `defer()` via oneshot channel (0de2a9f5) |
| 4 | 0de2a9f5 | P2: late-completion `.set()` on dropped picker-local use_signals could panic | added still-mine generation gate (d5f8b18b) |
| 5 | d5f8b18b | P2: my round-4 gate silenced user feedback in timeout-then-late-success scenario | watchdog force-closes picker on expiry (3a132eac) |
| 6 | 3a132eac | **No findings** | — |

### Self-review (code-first)

| Finding | Resolution |
|---------|-----------|
| `"Already a member"` used `contains_key`, mismatched observer-only rooms | switched to `RoomData::can_participate()` (793dfbc9) |

### CI status

- check-delegate-migration: pass
- check-wasm-sync: pass
- build: in progress on HEAD 3a132eac
- ui-playwright-tests: in progress on HEAD 3a132eac

Locally verified before each push: `cargo fmt`, `cargo clippy` clean on dm_thread_modal.rs / direct_messages.rs / invite_via_dm_picker_modal.rs, 153 unit tests pass, 5 Playwright tests pass on chromium/firefox/webkit.

Waiting on CI for the merge gate.

[AI-assisted - Claude]

sanity:
## Ready for merge

CI all-green on HEAD `3a132eac`:

- build: SUCCESS (7m1s)
- check-wasm-sync: SUCCESS (17s)
- check-delegate-migration: SUCCESS (25s)
- ui-playwright-tests: SUCCESS (10m24s)
- license/cla: SUCCESS

Multi-model review complete:
- Codex CLI: 6 rounds; rounds 1-5 each surfaced one finding (all fixed); round 6 returned clean.
- Self-review (code-first): one nit found and fixed (`already_member` check needed `can_participate()` instead of `contains_key`).

Per original task spec, **NOT merging this PR**. Parent agent gates merge.

[AI-assisted - Claude]

Issue #267, opened 2026-05-17
bug(dm): same-second inbound DM stays hidden after hide cutoff


sanity:
## Problem

Found during Codex review of PR #265 (issue #261).

DM timestamps are stored in whole Unix seconds. The hide-filter uses strict `<=`:

```rust
// common/src/chat_delegate.rs::is_thread_hidden
hidden_threads
    .iter()
    .find(|h| &h.room_owner_vk == room_owner_vk && h.peer == peer)
    .is_some_and(|h| max_message_ts <= h.hidden_at_ts)
```

This is correct for the message that *seeded* `hidden_at_ts` — that exact message must not revive the thread it was used to dismiss. But it has a corner case: if a peer sends an **inbound** DM in the same Unix second the user clicked Hide, `max_message_ts == hidden_at_ts` and the thread stays hidden. The peer's message is genuinely new (different content), but the filter treats it as "the message that was already there when I hid."

The outbound path explicitly works around this race by calling `unhide_dm_thread` after a successful send (`dm_thread_modal::do_send`, Codex P1 fix during PR #265 review). The inbound path has no equivalent — so a same-second inbound DM gets stranded until a later-second DM arrives.

## Reproduction

1. User has a DM thread with peer; latest message at ts=N.
2. User clicks Hide → `hidden_at_ts = N`.
3. Peer sends a DM that lands with `timestamp = N` (any time within the same Unix second).
4. Rail filter checks `N <= N` → true → thread stays hidden.
5. The unread DM does NOT surface in the rail. It only surfaces when a *later-second* DM arrives.

Probability: real but narrow — the peer's message needs to land within the same Unix second the user clicked Hide. On a busy thread, low but non-zero. On a quiet thread the user is intentionally dismissing, also low but non-zero.

## Suggested fix

Mirror the outbound-revival path. When the room synchronizer applies an inbound DM:

- For each new inbound DM message `(room, peer, ts)`:
  - If `HIDDEN_DM_THREADS` has an entry for `(room, peer)` with `hidden_at_ts == ts`, call `unhide_dm_thread(room, peer)`.

This is symmetric with `dm_thread_modal::do_send`'s call after `save_outbound_dm`. The tombstone path (`RECENTLY_UNHIDDEN`) already handles the cross-session race.

Alternative: change `hidden_at_ts` to store `last_seen_ts` and use strict `<`. Wire-format-breaking — requires a V2 store struct + migration, so the symmetric-unhide-on-inbound fix is cheaper.

[...]

PR #289, opened 2026-05-20
fix(ui): sync restored rooms via process_rooms, not a bare Subscribe

### LINKED ISSUE #287: Bug: Attempting to restore old ID causes never-ending sync issue.
I've been attempting to restore a few old IDs from exported backups. They are on test channels I sometimes use to test communication between nodes. When I import the IDs I get a never-ending sync spinner in the room list next to the room name and the main room window displays `Syncing room state from the network... You'll be able to send messages once sync completes.` at the bottom. No matter how long I wait, or even across restarts, the sync never ends.

From examining the JS console logs the most relevant warning appears to be something like the following which appears once for each room contract I am trying to restore.

```Received error in API response: WebSocket connection error: client error: error while executing operation in the network: Node error: Cannot subscribe to contract XYZetcetcetcetc: contract WASM/parameters not cached locally. PUT the contract or GET the contract before subscribing.```

As the rooms in question haven't been used for a while and contract addresses have changed since, I'm guessing this is a migration issue.

Although losing access to the rooms involved would be no great loss, this may be relevant to other users trying to restore old ID backups in the future so is worth looking at I think.

### PR DESCRIPTION
**Pull request by sanity**
## Problem

Restoring old room identities from exported backups leaves River stuck on a
never-ending sync spinner — the room list shows the spinner forever and the
conversation pane shows *"Syncing room state from the network… You'll be able
to send messages once sync completes."* across restarts (freenet/river#287,
reported by @Ivvvor).

The most prominent console error is:

```
Cannot subscribe to contract XYZ: contract WASM/parameters not cached
locally. PUT the contract or GET the contract before subscribing.
```

**Root cause:** the rooms-loaded-from-delegate path in `handle_api_response`
(the `ROOMS_STORAGE_KEY` branch) issued a **bare `ContractRequest::Subscribe`**
for every loaded room. Since freenet-core#3601 the node *rejects* a Subscribe
when the contract's WASM/parameters are not cached locally — always the case
for a freshly-restored backup, or any room not previously used on this node.

That rejection is unrecoverable on the River side:

- It arrives as an **`Err` API response** (`ErrorKind::OperationError`), **not**
  a `SubscribeResponse { subscribed: false }`, so the re-PUT recovery in
  `handle_subscribe_response` never runs.
- The error string contains `"contract"` but not `"not found"`, so the
  `freenet_synchronizer` error handler (which only matches contract-not-found)
  ignores it too.
- The room was set to `Subscribing` before the bare Subscribe, so it stays
  stuck there.

## Approach

Stop subscribing directly from the response handler. Route rooms loaded from
delegate storage through `process_rooms()` (via `mark_needs_sync`), which is
the single path that already handles this correctly:

- **Imported rooms** (default placeholder state) → `GET` with
  `return_contract_code = true`
- **Full-state rooms** → `PUT` with `subscribe = true`

Both **cache the contract WASM on the node before any subscribe**, so the
subsequent subscription succeeds. `mark_needs_sync` defers its signal write,
so it runs after the deferred `ROOMS` merge and `process_rooms()` sees the
merged rooms. A subscription-timeout check is still scheduled as a safety net.

This deletes the duplicate, buggy sync path entirely — the standard
`process_rooms()` flow is now the only one.

## Behaviour change

Rooms loaded from delegate storage now go through `process_rooms()` on cold
start, so a full-state room gets a `PUT` (CRDT merge — idempotent) where it
previously got a bare `Subscribe`. Slightly more cold-start traffic, but it
is the same path already used for invitations and migrated rooms and is what
makes the WASM-cache reliable.

## Testing

- New regression test `issue_287_loaded_rooms_do_not_bare_subscribe` in
  `response_handler.rs`. `handle_api_response` is too deeply coupled to the
  Dioxus signal runtime to drive behaviourally, so it is a source-text pin
  over this file's **production code** (same approach as the #267 guards in
  `room_synchronizer.rs`); it fails if a bare subscribe is reintroduced or
  the `process_rooms()` hand-off is removed. Verified it **fails against the
  pre-fix code** (which has the `.subscribe_to_contract(` call) and **passes
  with the fix**.
- It runs in CI via the `Test river-ui` step (`cargo test -p river-ui --bins`)
  that #288 added to `build.yml` — `river-ui`'s native unit tests previously
  ran in no CI job. (An earlier revision of this PR added that same step; it
  was dropped on rebase since #288 landed it first.)
- Full `cargo test -p river-ui --bins`: all tests pass, 0 failed.
- `cargo fmt` + `cargo clippy -p river-ui` clean.

## Out of scope — follow-up #290

This PR fixes the evidenced #287 symptom (the bare-`Subscribe` rejection).
After the fix an imported room is synced via `process_rooms()`, which `GET`s
with `return_contract_code = true` — caching the WASM and, if the contract
still exists on the network, completing sync.

If a restored room's contract is **genuinely absent from the network** (a
long-abandoned room whose contract was dropped), the GET fails and River
retries while the `is_awaiting_initial_sync()`-driven spinner keeps spinning.
That is a distinct, pre-existing bug with a different precondition — filed as
**#290** (bounded retry + a terminal UI error state).

Closes #287

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/freenet_api/response_handler.rs
+++ b/ui/src/components/app/freenet_api/response_handler.rs
@@ -16,12 +16,10 @@ use crate::components::app::chat_delegate::{
 };
 use crate::components::app::document_title::{mark_current_room_as_read, update_document_title};
 use crate::components::app::notifications::mark_initial_sync_complete;
-use crate::components::app::sync_info::{RoomSyncStatus, SYNC_INFO};
 use crate::components::app::{CURRENT_ROOM, ROOMS};
 use crate::room_data::CurrentRoom;
 use crate::room_data::Rooms;
 use crate::util::ecies::decrypt_with_symmetric_key;
-use crate::util::owner_vk_to_contract_key;
 use ciborium::de::from_reader;
 use dioxus::logger::tracing::{error, info, warn};
 use dioxus::prelude::ReadableExt;
@@ -49,7 +47,8 @@ pub struct ResponseHandler {
 pub struct ResponseFlags {
     /// True if a re-PUT should be scheduled (subscription failed but we have local state)
     pub needs_reput: bool,
-    /// True if subscriptions were initiated and need timeout monitoring
+    /// True if initial sync was kicked off for rooms loaded from delegate
+    /// storage and a subscription-timeout check should be scheduled.
     pub subscriptions_initiated: bool,
 }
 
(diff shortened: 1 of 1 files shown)


sanity:
## Review round 1 — findings addressed

Ran 4 internal reviewers + Codex. No BLOCKING findings. How each was handled:

**Codex P3 — regression test could pass even with the fix removed.** Fixed in
6015ffcb. The positive assertion sliced source from the marker to EOF and
could match `mark_needs_sync` literals in the test module's own text. The
test now slices at `mod tests {` and pins **production code only** — both
assertions now genuinely fail if the production hand-off is removed.

**Testing reviewer — pin covers parent file only / submodule hole.** Adequately
covered: the loaded-rooms block lives in `response_handler.rs`; if it were
moved to a submodule the `"Scheduling initial sync for"` marker disappears
from this file's production code and the `expect` fires. The negative
assertion (`!production.contains("subscribe_to_contract")`) correctly does
*not* flag `put_response.rs`'s legitimate post-PUT subscribe.

**Testing reviewer — pin the `had_loaded_rooms` guard too.** Not pinned
deliberately: that guard only avoids scheduling one harmless extra
`ProcessRooms` when zero rooms load — it is a micro-optimisation, not a
#287-correctness invariant. Pinning it would make the test brittle to benign
refactors. The pin covers the two real invariants (no bare subscribe; routes
via `mark_needs_sync`).

**Testing reviewer — downstream `process_rooms` behaviour untested.** True, and
genuinely hard: `process_rooms` / `rooms_awaiting_subscription` read several
`GlobalSignal`s and need the Dioxus runtime. The new `Test river-ui` CI step
at least makes the existing source-pin guards run for the first time. A
behavioural harness for `process_rooms` is worth a separate effort.

**Skeptical reviewer M1 — `subscriptions_initiated` timeout decoupled from
`subscribing_since`.** Acknowledged, not changed. The flag schedules a
one-shot timeout `ProcessRooms`; it is a coarse best-effort net (it always
was — pre-PR it was set per-subscribe-send). The primary fix makes the
timeout path rarely exercised (a PUT/GET caches the WASM so the subscribe
succeeds), and `rooms_awaiting_subscription` is also re-run on reconnect /
visibility changes. Re-basing the timer on actual send time is a structural
flag-plumbing change, out of scope here.

**Skeptical reviewer L1 / big-picture — extra PUT on cold start.** Documented
in the new "Behaviour change" section of the PR description.

**Big-picture — "out of scope" paragraph.** Rewritten for accuracy and the
remaining case (contract genuinely absent from the network → GET-not-found
retry → spinner persists) is now filed as **#290** with full analysis. It is
a distinct pre-existing bug whose proper fix needs UI error-state work — a
separate review surface.

**Code-first MINOR — pre-existing unused import in `app.rs`.** Out of scope;
not touched by this PR and not surfaced by the new host-target CI step.

[AI-assisted - Claude]

sanity:
Rebased onto current `main` (#288, #291 had landed). The only conflict was in `.github/workflows/build.yml`: #288 independently added an identical `Test river-ui` (`cargo test -p river-ui --bins`) step. Resolved by dropping this PR's redundant copy and keeping main's. **The PR diff is now `response_handler.rs` only, byte-identical to the reviewed commit** — no new/unreviewed content (the conflict resolution only removed a now-duplicate CI step), so the prior 4-reviewer + Codex review still applies. The `issue_287` regression test remains CI-gated via #288's `Test river-ui` step.

[AI-assisted - Claude]

PR #293, opened 2026-05-20
feat: recover rooms stranded across room-contract WASM generations

### LINKED ISSUE #292: Identity/room import can't recover a room several contract-WASM generations behind (restores stale "old IDs")
## Problem

Reported by Ivvor on Matrix (2026-05-20): restoring/importing a room that hasn't been used for a while brings back **old member IDs** — stale state resurfaces instead of the room's current state. Ivvor's hypothesis, which a code read confirms: the room's last-active contract is several migrations behind current, and nothing in River searches backward across contract generations.

The room contract key is `BLAKE3(room_contract.wasm, params)` with `params = { owner: VerifyingKey }`, so **every room-contract WASM upgrade moves the key**. River has no mechanism to recover a room whose live state sits under an *older-generation* contract key.

## Root cause

Two architectural gaps (verified by reading the code):

### 1. Import / migration only ever uses the *current* contract key — no backward search

- The contract key is always recomputed from `owner_vk` + the **current bundled WASM** (`ui/src/util.rs:410` `owner_vk_to_contract_key`, `cli/src/api.rs:842`).
- There is **no registry of historical room-contract WASM hashes**. The chat *delegate* has `legacy_delegates.toml`; the room contract has no equivalent. A client has exactly one room-contract WASM compiled in, so it can compute exactly **one** key per owner — the current one.
- `previous_contract_key` (`ui/src/room_data.rs:82`, `:90`) is a **single slot**, overwritten on every `regenerate_contract_key()` — only ever one generation of history is kept.
- `OptionalUpgradeV1` (`common/src/room_state/upgrade.rs`) is a **single forward pointer**, followed **exactly one hop** — `cli/src/api.rs:416` comments "single hop"; the UI's `follow_upgrade_pointer_if_needed` fires one GET and never re-checks the followed state's own pointer. There is no chain walk.
- The **ID importer** — `riverctl identity import` (`cli/src/commands/identity.rs:184`) and the UI `ImportIdentityModal` (`ui/src/components/members.rs:547`) — imports an `IdentityExport` that carries **only `room_owner: VerifyingKey`, no contract key** (`common/src/room_state/identity.rs`). It GETs the current key exactly once; `ensure_room_migrated` short-circuits because the room isn't in local storage yet; the UI hard-codes `previous_contract_key: None` (`members.rs:606`). **No older-version lookup, no retry against older keys.**

**Net:** if a room's live state is under a contract key from an older WASM generation, neither the importer nor the migration path can locate it — they look only at the current key and give up.

### 2. The "migration" force-PUTs *stale local* state — it never GETs the room's current network state

- On a WASM upgrade, `Rooms::merge` → `regenerate_contract_key()` flags the room; `room_synchronizer.rs:626-748` then **PUTs the locally-cached `room_state`** — the snapshot the delegate had from when the room was last used — to the new key. There is no GET of the room's current network state before that PUT.
- For a long-dormant room this republishes the device's old snapshot. Because the room state is a CRDT, that stale snapshot gets merged into the live room, **reintroducing old member entries / IDs** that had since changed. This is the "restoring old IDs" symptom.

## Why existing issues don't cover this

- #287 (closed) / #289 (merged) — fixed the bare-`Subscribe` hang for restored rooms; a different precondition.
- #290 — covers the *imported room spins forever when the contract is absent from the network* case and proposes bounding the retry + a terminal UI error. It does **not** propose searching older contract generations — and the contract often appears "absent" precisely **because** the room's live state is back on an older key the importer never checks. This issue is the root cause underneath that symptom.
- #63 / #197 — broader room-contract WASM-upgrade and key-stability issues; neither documents the importer's lack of backward search or the stale-local-state PUT.

## Proposed fix direction

- Give the room contract a **legacy-WASM-hash registry** (analogous to `legacy_delegates.toml`) so a client can derive older-generation contract keys for a given `owner_vk` and probe them, newest-to-oldest, until it finds live state.
- On migration, **GET the current network state and CRDT-merge it** rather than blindly force-PUTting the stale local snapshot.
- Optionally make `OptionalUpgradeV1` followable as a multi-hop chain so a peer landing on any old key can walk forward to the current one.

## References

- Matrix report: Ivvor, 2026-05-20
- Related: #287, #289, #290, #197, #63

[AI-assisted - Claude]

sanity:
## Problem

Reported by Ivvor (Matrix, 2026-05-20): restoring/importing a room that has been dormant for a while brings back **stale old member IDs** instead of the room's current state.

The room contract key is `BLAKE3(room_contract.wasm, params)` with `params = {owner}`, so **every room-contract WASM upgrade moves the key**. River had no way to recover a room whose live state sits under an older-generation contract key:

1. **No backward search.** Import/migration only ever derive the *current* contract key. There was no registry of historical room-contract WASM hashes (the chat delegate has `legacy_delegates.toml`; the room contract had no equivalent).
2. **Stale-state resurfacing.** On a WASM upgrade the migration path force-PUT the device's *stale local* `room_state` snapshot to the new key instead of adopting the network's current state, re-introducing old member entries into the live room.
3. `OptionalUpgradeV1` was followed exactly one hop.

## Approach

One principle, the same gating the chat-delegate legacy migration uses (#253): **the current contract key is always authoritative; backward recovery fires only when the current key has no usable state** — so recovery can never resurface stale state over live state.

**Registry (Part 1).** New `common/legacy_room_contracts.toml` records the BLAKE3 code hash of every previous room-contract WASM generation (24 entries, extracted from git history) — the room-contract analogue of `legacy_delegates.toml`. `common/build.rs` generates `LEGACY_ROOM_CONTRACT_CODE_HASHES`; the new `river-core` `migration` module re-derives the contract key any owner's room used under each generation. The module is **feature-gated so the room-contract and chat-delegate WASM builds never compile it** — their WASM bytes and keys stay byte-identical (this PR does not redeploy the contract or delegate).

**CLI (riverctl).** `get_room` GETs the current contract, walks any `OptionalUpgradeV1` chain forward (multi-hop, cycle-guarded), and on an empty current contract probes every previous generation newest-to-oldest, then migrates a recovered room forward. `get_migration_state` probes the full registry for the freshest network state instead of force-PUTting a stale local snapshot. Fixes `riverctl identity import` for a room several generations behind.

**UI.** The migration path no longer force-PUTs the local snapshot — it GET+subscribes the new key so network state is adopted. When the current key returns no real state, a backward probe walks the legacy generations newest-to-oldest; every probe GET is paired with a 12s timeout watchdog so an absent (GC'd) generation cannot stall recovery. The first generation with real state is CRDT-merged with the local snapshot and PUT forward onto the current key. The stale local snapshot is PUT only as a genuine last resort. `follow_upgrade_pointer_if_needed` is multi-hop (discovery-only). Imported rooms flow through this recovery.

## Testing

- `river-core`: new `migration` unit tests (registry key derivation matches the live `from_params_and_code` path; owner-specificity; newest-first order) + `room_contract_migration_test` registry validation — both now gated in CI.
- `riverctl`: 21 lib tests — registry-derivation parity, `next_upgrade_hop` cycle guard — now gated in CI.
- `river-ui`: 189 tests — backward-probe gate classification, probe-table single-shot + epoch guard, `classify_upgrade_hop` cycle/cap/reset, `merge_room_states`, `touch_subscribing_since`, import-path source-grep pin.
- Also fixes a **pre-existing flaky test** (the `ENSURE_SUBSCRIPTION_SENT` dedup tests raced on a global static — now serialized; 0/25 stress failures).
- Full workspace compiles native + `wasm32-unknown-unknown`; no `.wasm` files modified.

## Review

Full multi-model review (4 perspectives + Codex) run across four rounds. Round 1 found a blocking cluster in the UI probe machinery (no timeout → permanent stall; signal-safety violation). Round 2 found a probe-vs-subscription-timeout race and a CLI `return_contract_code` regression. Round 3 found a panic on malformed legacy state. Round 4 came back clean (Mergeable). All findings addressed — see the PR review threads. (Branch merged with `main` to pick up #129/#131.)

## Known limitations / follow-ups

- **Pending-invite branch not probed.** Recovery is wired into the existing-room / import path; accepting an invite *link* to a stranded room does not probe. Follow-up.
- **#292 is preventive, not curative.** It stops future stale-state resurfacing. State already merged into a contract by a pre-#293 client cannot be un-merged here (a CRDT property); the room contract's `post_apply_cleanup` prunes inactive resurfaced members on the next activity.
- A client whose bundled WASM is *older* than the generation a room currently lives on fetches the room once but does not get live push updates until the client itself updates — inherent to running stale client code.

Closes #292

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,7 +110,7 @@ jobs:
       # upgrade-chain cycle guards). No prior step ran or even compiled them.
       run: cargo test -p riverctl --lib
 
-    - name: Test river-core ecies (deterministic + randomized helpers)
+    - name: Test river-core lib (ecies + migration registry)
       env:
         RUST_MIN_STACK: 8388608
         CARGO_TARGET_DIR: ${{ github.workspace }}/target
@@ -118,8 +118,10 @@ jobs:
       # deterministic `encrypt_secret_for_member` path used by the chat
       # delegate, plus the randomized-helper round-trips used by the UI.
       # See freenet/river#241 — these tests pin the wire format and catch
-      # silent drift in the ECIES derivation.
-      run: cargo test -p river-core --lib --features ecies-randomized ecies
+      # silent drift in the ECIES derivation. The `migration` feature also
+      # gates the `migration` module's unit tests (legacy room-contract key
+      # derivation, newest-first ordering — freenet/river#292).
+      run: cargo test -p river-core --lib --features ecies-randomized ecies migration
 
     - name: Test river-ui (UI crate unit tests)
       env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,7 +131,12 @@ jobs:
       # silent drift in the ECIES derivation. The `migration` feature also
       # gates the `migration` module's unit tests (legacy room-contract key
       # derivation, newest-first ordering — freenet/river#292).
-      run: cargo test -p river-core --lib --features ecies-randomized ecies migration
+      #
+      # `--features` takes ONE comma-separated value; passing space-separated
+      # feature names makes cargo treat the extras as positional test-name
+      # filters (and a second one is a hard "unexpected argument" error).
+      # `ecies-randomized` enables `ecies` transitively.
+      run: cargo test -p river-core --lib --features ecies-randomized,migration
 
     - name: Test river-ui (UI crate unit tests)
       env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,6 +94,22 @@ jobs:
         CARGO_TARGET_DIR: ${{ github.workspace }}/target
       run: cargo test -p river-core --test migration_test
 
+    - name: Test room-contract migration registry (#292)
+      env:
+        RUST_MIN_STACK: 8388608
+        CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      # Validates common/legacy_room_contracts.toml is well-formed. A separate
+      # --test target from migration_test, so it needs its own invocation.
+      run: cargo test -p river-core --test room_contract_migration_test
+
+    - name: Test riverctl (CLI crate unit tests, incl. #292 recovery)
+      env:
+        RUST_MIN_STACK: 8388608
+        CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      # riverctl carries native unit tests (legacy-key derivation parity,
+      # upgrade-chain cycle guards). No prior step ran or even compiled them.
+      run: cargo test -p riverctl --lib
+
     - name: Test river-core ecies (deterministic + randomized helpers)
       env:
         RUST_MIN_STACK: 8388608
--- a/.github/workflows/check-room-contract-migration.yml
+++ b/.github/workflows/check-room-contract-migration.yml
@@ -1,0 +1,25 @@
+name: Check Room-Contract Migration
+
+on:
+  pull_request:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  check-room-contract-migration:
+    runs-on: freenet-default-runner
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Install b3sum
+      run: command -v b3sum || cargo install b3sum --locked
+
+    - name: Verify committed room-contract WASM changes include a migration entry
+      run: |
+        ./scripts/check-room-contract-migration.sh --ci \
+          "${{ github.event.pull_request.base.sha }}" \
+          "${{ github.event.pull_request.head.sha }}"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -441,6 +441,17 @@ All legacy delegate entries are defined in `legacy_delegates.toml` at the repo r
 This file is the **only** place migration entries are managed. The UI's build.rs generates
 Rust code from it at compile time. CI reads it directly for validation.
 
+### Single Source of Truth: `common/legacy_room_contracts.toml`
+
+The room contract has its own registry, `common/legacy_room_contracts.toml`,
+recording the BLAKE3 code hash of every previous room-contract WASM generation.
+A client re-derives the contract key any owner's room used under each generation
+and probes them newest-to-oldest to recover a room dormant across one or more
+WASM upgrades (freenet/river#292). `common/build.rs` generates
+`LEGACY_ROOM_CONTRACT_CODE_HASHES` from it; the `river-core` `migration` feature
+exposes the lookup. It lives inside the `common` crate (not the repo root) so it
+ships with the published `river-core` crate and riverctl keeps the full registry.
+
 ### Upgrade Workflow
 
 When you change code that affects delegate or contract WASM:
@@ -448,18 +459,22 @@ When you change code that affects delegate or contract WASM:
 ```bash
 # 1. Add old delegate hash to migration registry
 cargo make add-migration
+#    AND, if the room-contract WASM changed, add its old hash too:
+cargo make add-room-contract-migration
 
 # 2. Build new WASMs and copy to all committed locations
 cargo make sync-wasm
 
 # 3. Run migration tests
 cargo test -p river-core --test migration_test
+cargo test -p river-core --test room_contract_migration_test
 
 # 4. Verify UI compiles with new generated code
 cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync
 
 # 5. Commit everything
-git add legacy_delegates.toml ui/public/contracts/ cli/contracts/
+git add legacy_delegates.toml common/legacy_room_contracts.toml \
+    ui/public/contracts/ cli/contracts/
 git commit -m "fix: update WASMs with delegate migration entry"
 ```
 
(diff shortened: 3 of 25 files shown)


sanity (commented):
## Comprehensive PR Review: #293

### Summary
- **PR Title:** feat: recover rooms stranded across room-contract WASM generations
- **Type:** feat (cross-generation room-contract recovery, issue #292)
- **CI Status:** pending at review time
- **Linked Issues:** #292
- **Review tier:** Full (high-risk surface — contract-key derivation, migration, build/CI config)
- **Reviewers run:** code-first, testing, skeptical, big-picture (Claude) + Codex (external)

---

### Code-First Analysis
**Independent understanding matches stated intent.** All three parts of issue #292's proposed direction are delivered: (1) the legacy-WASM-hash registry + codegen, (2) GET-and-merge instead of force-PUT of stale local state, (3) multi-hop `OptionalUpgradeV1` following. Contract-key derivation (`contract_key_for_code_hash`) was verified byte-identical to the live `from_params_and_code` path and is pinned by tests on both the UI and CLI sides. Feature-gating that keeps the room-contract/delegate WASM byte-identical holds for the canonical isolated `sync-wasm.sh` build.

**Gaps:** undocumented CLI/UI asymmetries — the "no real state" predicate differs (`== ChatRoomStateV1::default()` in the CLI vs `configuration.verify_signature().is_err()` in the UI), and the UI probe path does not follow upgrade pointers on a recovered legacy state while the CLI's `try_fetch_room` does.

---

### Testing Assessment
**Coverage Level:** adequate for the registry layer, partial for recovery orchestration.

| Unit | ✅ | Registry derivation pinned vs live key path; gate classification; exhaustion/hop-cap |
| Integration | ⚠️ | GET→probe→merge→PUT round trips need a node; reasonably deferred |
| Simulation | N/A | |
| E2E | ⚠️ | Recovery not yet verified end-to-end against a genuinely stranded room |

**Missing tests:** CLI `follow_upgrade_chain` cycle/hop guards are untested pure-ish logic; UI `follow_upgrade_pointer_if_needed` guard predicates untested; CLI `get_migration_state` merge-failure fallback untested.

---

### Skeptical Findings
**Risk Level:** high (blocking cluster in the UI probe machinery).

| Probe has no per-GET timeout — stalls forever on an absent legacy contract | **high** | `backward_probe.rs` `fire_probe_get` | If a legacy GET never returns, the probe never advances or exhausts. Also flagged P1 by Codex. |
| A stalled probe permanently disables recovery for that owner for the session | **high** | `backward_probe.rs` `start_backward_probe` "already in progress" guard | Orphaned `ProbeState` makes every later current-key GET return early; no TTL. |
| Exhaustion-seed `ROOMS.try_read()` failure silently no-ops; "next sync cycle retries" is false | **high** | `get_response.rs` `handle_probe_get_response` | Probe already consumed; room left empty. |
| `take_probe` mutates a `GlobalSignal` synchronously in a polled-future context | **med-high** | `backward_probe.rs` `take_probe` | Violates the project Dioxus signal-safety rule (every other mutation here is deferred). |
| UI upgrade-chain cycle guard only catches an immediate 2-cycle | med | `update_notification.rs` `follow_upgrade_pointer_if_needed` | A→B→A loops to `MAX_UPGRADE_HOPS`, subscribing to up to 32 junk contracts. CLI uses a proper visited-set. |
| `UPGRADE_HOP_COUNTS`/`UPGRADE_TARGETS` leak on a non-terminating chain | med | `update_notification.rs` | No TTL; a later walk for the same owner starts with a stale count. |
| UI probe PUT-forward carries raw recovered state, not merged-with-local | med | `get_response.rs` `handle_probe_get_response` | Drops unsynced local edits; CLI's `get_migration_state` merges. |

---

### Big Picture Assessment
**Goal Alignment:** yes — solves both architectural root causes plus the optional third part. **Anti-patterns:** none — no `#[ignore]`, no removed tests, no weakened assertions. The `repopulate_secrets` pin count `3→5` is a legitimate tightening; the `ENSURE_SUBSCRIPTION_SENT` test serialization is a proper root-cause flaky-test fix. **Removed code:** none; no `.wasm` modified. **Scope:** focused. **Byte-identity claim holds** — verified the `migration` feature can never unify into the isolated `-p room-contract` WASM build, with the new CI workflow as a backstop.

---

### Documentation
- Code docs: thorough on new public items.
- `AGENTS.md`: updated and accurate.
- Suggested: document the CLI/UI predicate asymmetry; note the byte-identity guarantee depends on the isolated `-p room-contract` build (never `--workspace`); make the "run before rebuild" ordering explicit for `add-room-contract-migration`.

---

### Recommendations

#### Must Fix (Blocking)
1. **UI backward probe needs a per-GET timeout watchdog.** An absent legacy contract that never answers must not stall the probe — it must advance to the next generation (and ultimately seed). This is the exact scenario the feature targets.
2. **A stalled/orphaned probe must not permanently disable recovery** for that owner. Tie the cleanup to the timeout in (1).
3. **Exhaustion-seed must not depend on a fallible `ROOMS.try_read()`** with no retry. Capture the local snapshot up front so the seed always has it.
4. **`take_probe` must not mutate a signal synchronously in a polled future.** Convert the probe side-tables to plain `Mutex` statics (they have zero UI reactivity — like `ENSURE_SUBSCRIPTION_SENT`), eliminating the signal-safety concern entirely.

#### Should Fix (Important)
1. UI multi-hop: use a per-owner `visited` set (like the CLI) instead of `delivered_from` + a counter; clean it up so it can't leak.
2. UI probe PUT-forward should CRDT-merge the recovered state with the local snapshot before PUT, matching the CLI.
3. Extract and unit-test the CLI `follow_upgrade_chain` cycle/hop guards and the UI upgrade-pointer guard predicate.

#### Consider (Suggestions)
- Document the CLI/UI "no real state" predicate asymmetry and the feature-unification footgun (`-p room-contract` only).
- `try_get_state` deserialization failure: `warn!` rather than `debug!` (it would signal a real backwards-compat break).
- Note in the PR that #293 is **preventive** — it stops future stale-state resurfacing but does not un-merge state already corrupted by a pre-#293 client (a CRDT property; the contract's `post_apply_cleanup` prunes inactive resurfaced members on the next activity).

---

### Verdict
**HEAD SHA reviewed:** 97b8c98b9b5bc1eed9559d40e85f6c69460e790f

The core mechanism (registry, key derivation, feature-gating, "current key authoritative" gating, CLI recovery) is correct and well-tested. The blocking findings are all in the UI backward-probe machinery: it can stall permanently and then permanently disable recovery — and dormant rooms are exactly what this feature exists to fix. Must Fix 1-4 must be addressed and a fresh full review run on the updated HEAD before merge.

[AI-assisted - Claude]

sanity:
## Review findings addressed — re-review in progress

All findings from the consolidated review are addressed (HEAD `ee38c73f`):

**Must Fix (blocking)**
1. ✅ UI backward probe now has a per-GET **timeout watchdog** (`PROBE_GET_TIMEOUT`, 12s) — an absent legacy contract can no longer stall the probe; the watchdog synthesizes an empty response so the probe advances/seeds.
2. ✅ A stalled probe can no longer permanently disable recovery — the watchdog always consumes the entry.
3. ✅ Exhaustion seed no longer depends on a fallible `ROOMS.try_read()` — the local snapshot is captured into `ProbeState` up front.
4. ✅ Probe side-tables (`BACKWARD_PROBES`, `UPGRADE_TARGETS`, `UPGRADE_CHAIN_VISITED`) are now plain `Mutex` statics, not Dioxus signals — the `take_probe` signal-safety violation is gone.

**Should Fix**
1. ✅ UI multi-hop now uses a per-owner `visited` set (catches every cycle, not just A→A) and resets on a fresh walk root so it can't leak.
2. ✅ UI probe PUT-forward now CRDT-merges recovered + local snapshot (no dropped local edits), matching the CLI.
3. ✅ Extracted and unit-tested the cycle guards — `next_upgrade_hop` (CLI) and `classify_upgrade_hop` (UI).

**Consider**
- ✅ CLI `try_get_state` now uses the same `verify_signature` predicate as the UI (asymmetry removed).
- ✅ Documented the build invariant (contract WASM must be built `-p room-contract`, never `--workspace`) in `migration.rs` + `Makefile.toml`; `try_get_state` deser failure now `warn!`.
- **M5 (already-corrupted rooms):** #293 is **preventive** — it stops future stale-state resurfacing. State already merged into a contract by a pre-#293 client is a CRDT property and cannot be un-merged here; the room contract's `post_apply_cleanup` prunes inactive resurfaced members on the next activity.

Re-running the full multi-model review on `ee38c73f`.

[AI-assisted - Claude]

sanity (commented):
## Re-review round 4 — Ready to Merge

Fourth multi-model pass (adversarial + external Codex) on HEAD `772e3f00`, with the one Low finding now fixed in `f904c7a4`.

**Skeptical (adversarial):** Mergeable — no High or Medium findings; rounds 1-3 blocking fixes all verified intact (epoch-guarded watchdog, `defer`-wrapped signal writes, single-shot `take_probe`, the `touch_subscribing_since` reclaim-race fix confirmed — the room provably stays `Subscribing` for the whole probe). One Low (L-NEW-1): the upgrade-pointer path still used the panicking `from_cbor_slice` while the backward-probe twin had been hardened — **fixed in `f904c7a4`** (`try_from_cbor_slice`, undeserializable state → chain terminates).

**Codex (external):** "I did not identify any discrete correctness issues introduced by the diff. The migration registry, probing logic, and UI/CLI recovery paths appear internally consistent with the stated design."

### Review history
- **Round 1:** blocking cluster in the UI probe machinery (no timeout → permanent stall; signal-safety violation) — fixed.
- **Round 2:** probe-vs-subscription-timeout race, CLI `return_contract_code` regression, upgrade-target subscription routing — fixed.
- **Round 3:** panic on malformed legacy state, status-flap — fixed; CI test-wiring gaps closed.
- **Round 4:** clean — one Low consistency cleanup, fixed.

All findings across four rounds are addressed. 189 river-ui tests, 21 riverctl tests, all river-core tests pass; clippy clean on new code; no `.wasm` modified.

**Verdict: Ready to Merge** once CI is green on the current HEAD.

[AI-assisted - Claude]

PR #294, opened 2026-05-20
fix(ui): restore member_info on join + self-heal "Unknown" members

sanity:
## Problem

PR #272's `build_state_for_put` PUTs a newly-joined member into the room
contract's `members` list but **never into `member_info`**. The contract
accepts this (member_info entries are optional per `MemberInfoV1::verify`),
so the member-info-less state becomes canonical, and every other peer
renders the member as **"Unknown"** instead of their nickname.

In the official Freenet room, **10 of 33 members** are stranded this way.
(`riverctl member list` iterates `member_info`, so it only showed 23 — the
gap is the bug.)

## Approach

**Fix the root cause** — `build_state_for_put` now also injects the
invitee's self-signed `AuthorizedMemberInfo` into the PUT state. The
caller builds it once (before the PUT and before the deferred local ROOMS
mutation) and reuses the identical value for both, so the PUT payload and
local state stay byte-identical — the same reuse discipline the
synthesised join_event already follows.

**Remediate already-stranded members** — new `RoomData::build_member_info_heal`:
on every GET of an existing room, if the network's canonical state shows
us in `members` with no `member_info` entry, we re-publish our own
self-signed `member_info` via a dedicated UPDATE. Idempotent — once the
entry lands, later GETs no longer see the stranding. This heals each of
the 10 stranded members the next time they open River.

**Themed default nicknames** (folded in per request) — members who join
without typing a nickname previously got `User-a3F9bX`, which reads like a
spam account. They now get a deterministic 90s-hacker-culture handle
(e.g. "Cipher Daylight"), derived by hashing the verifying key — a pure
function of the key, like the old scheme but friendlier.

## Why the remediation must be client-side

`MemberInfoV1::verify` requires a non-owner's `member_info` to be signed
by **that member's own key**. The room owner's key cannot sign a heal
entry for someone else — the contract would reject it. So an owner-side or
server-side back-fill is impossible; each stranded member can only be
healed by their own client. That is exactly what `build_member_info_heal`
does, using `self_sk`.

## Scope

UI-only (`ui/`). No `common/`, room-contract, or delegate WASM changes —
the room contract key is unchanged, no migration needed, riverctl
unaffected.

## Commits

- `feat(ui)`: themed default nicknames (new `ui/src/nickname.rs`).
- `fix(ui)`: the member_info injection + self-heal — **the commit to review**.
- `style(ui)`: isolated `cargo fmt` of `get_response.rs` (pre-existing
  under-indented closure; `git diff -w` against its parent is empty).

## Testing

- `build_state_for_put_includes_member_info` — pins the PUT-state
  injection; asserts it survives `post_apply_cleanup` + `verify`.
- 5 × `build_member_info_heal_*` — stranded / healthy / owner /
  not-a-member, and that a stored nickname is preferred over a fresh
  default.
- 4 × `nickname::*` — determinism, pool sizes, two-word format,
  distribution.
- All 185 river-ui native unit tests pass; native + `wasm32` builds clean.

[AI-assisted - Claude]

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -410,6 +410,28 @@ design.
   `test_stale_secret_recipient_is_pruned_after_rotation`, and
   `test_ban_race_with_encrypted_secret_converges_to_pruned` in
   `common/src/room_state.rs`.
+- **`member_info` must accompany every membership change** (PR #294,
+  "Unknown member" regression): whenever a member is added to
+  `room_state.members` on a path that goes to the network, their
+  `AuthorizedMemberInfo` MUST be written to `room_state.member_info` in
+  the same wire payload. A member present in `members` but absent from
+  `member_info` is valid contract state (member_info entries are
+  optional per `MemberInfoV1::verify`) but renders as **"Unknown"** to
+  every other peer. `build_state_for_put` (invitation-accept PUT) is the
+  canonical example: it must inject the invitee's `member_info`
+  byte-identically to the deferred local-state copy — the same
+  build-once-reuse discipline the synthesised join_event follows. PR
+  #272 added the member injection but omitted `member_info`, which is
+  the regression #294 fixed. The remediation for already-stranded
+  members is `RoomData::build_member_info_heal`: on every GET of an
+  existing room it detects "self in `members`, absent from
+  `member_info`" and re-publishes a self-signed `member_info` (folded
+  into the PUT for imported rooms, sent as a standalone UPDATE for
+  already-subscribed rooms). A non-owner's `member_info` is only valid
+  when self-signed by that member's own key, so this heal can ONLY run
+  client-side, by the affected member — never owner-side. For a private
+  room the heal defers (publishes nothing) until the room secret is
+  available, so it never leaks a plaintext nickname.
 - **In-memory secret repopulation** (#251): `room_data.secrets:
   HashMap<u32, [u8; 32]>` is `#[serde(skip)]` and must be rebuilt from
   `room_state.secrets.encrypted_secrets` after EVERY network state
--- a/ui/src/components/app/freenet_api/response_handler/get_response.rs
+++ b/ui/src/components/app/freenet_api/response_handler/get_response.rs
@@ -706,6 +706,11 @@ pub async fn handle_get_response(
             // the entry is already synced when the network never received
             // it. The delta is self-signed and idempotent, so re-sending
             // it is harmless if it raced another heal.
+            //
+            // The heal is NOT written into the local `room_state` here,
+            // so self renders as "Unknown" locally until the subscription
+            // notification echoes this UPDATE back — the same benign,
+            // self-resolving transient as the imported-room path above.
             if !needs_put_subscribe {
                 if let Some(heal_info) = member_info_heal {
                     let heal_delta = ChatRoomStateV1Delta {
@@ -968,6 +973,12 @@ pub(crate) fn build_state_for_put(
         .iter()
         .any(|m| m.member.member_vk == self_vk);
     if already_member {
+        // The invitee is already in `members`. If they are also missing
+        // from `member_info` (a retry of a partially-completed accept, or
+        // a legacy stranding) this PUT does not heal them — but the room
+        // is now an existing room, so the GET-path `build_member_info_heal`
+        // self-heal covers it on the next GET. Same delayed-heal class as
+        // issue #295.
         return Ok((state, None));
     }
 
--- a/ui/src/components/app/freenet_api/response_handler/get_response.rs
+++ b/ui/src/components/app/freenet_api/response_handler/get_response.rs
@@ -15,14 +15,14 @@ use dioxus::prelude::ReadableExt;
 use freenet_scaffold::ComposableState;
 use freenet_stdlib::client_api::{ClientRequest, ContractRequest};
 use freenet_stdlib::prelude::{
-    ContractCode, ContractContainer, ContractKey, ContractWasmAPIVersion, Parameters,
+    ContractCode, ContractContainer, ContractKey, ContractWasmAPIVersion, Parameters, UpdateData,
     WrappedContract, WrappedState,
 };
 use river_core::room_state::member::MemberId;
 use river_core::room_state::member_info::{AuthorizedMemberInfo, MemberInfo};
 use river_core::room_state::message::{AuthorizedMessageV1, MessageId, MessageV1, RoomMessageBody};
 use river_core::room_state::privacy::{PrivacyMode, SealedBytes};
-use river_core::room_state::{ChatRoomParametersV1, ChatRoomStateV1};
+use river_core::room_state::{ChatRoomParametersV1, ChatRoomStateV1, ChatRoomStateV1Delta};
 use std::collections::HashMap;
 use std::sync::Arc;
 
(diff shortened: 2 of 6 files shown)


sanity (commented):
## Comprehensive PR Review: #294

### Summary
- **PR Title:** fix(ui): restore member_info on join + self-heal "Unknown" members
- **Type:** fix (+ folded-in feat: themed nicknames)
- **CI Status:** pending at review time (build / check-delegate-migration / ui-playwright-tests)
- **Linked Issues:** none referenced (recommend filing one for the official-room remediation)
- **Review tier:** Full — touches state authorization (member_info signatures) and wire format (member_info in PUT/UPDATE)
- **Reviewers run:** code-first, testing, skeptical, big-picture, + external Codex (gpt-5.5)

---

### Code-First Analysis
Independent reading matches the stated intent with no material discrepancies. The
root-cause fix (`build_state_for_put` injecting `member_info`), the byte-identical
reuse discipline (same value in PUT + deferred local mutation, mirroring the existing
join_event pattern), and the client-side-only rationale (verified against
`MemberInfoV1::verify` — non-owner member_info must be self-signed) are all sound.
Confirmed UI-only: no `common/`/contract/delegate changes, no migration needed.

---

### Testing Assessment
Coverage of the **decision logic** is thorough; gaps are in private-room and
send-path code.

| Unit | ⚠️ | 10 new tests; `build_state_for_put_includes_member_info` genuinely pins the regression. Gaps below. |
| Integration | N/A | UI-only change |
| Simulation | N/A | |
| E2E | ⚠️ | ui-playwright covers layout, not this behavior |

Gaps: `current_secret_from_state` has zero coverage; the private-room sealing branch
of `build_member_info_heal` is never exercised (all heal tests use public rooms); the
stranded-case heal test does not assert the produced entry is contract-valid
(self-signature); the heal UPDATE-send / member_info-only-delta-application path is
untested.

---

### Skeptical Findings
**Risk Level:** medium

| Private-room nickname published in plaintext | Medium | `get_response.rs` hoisted block + `room_data.rs build_member_info_heal` | When a private room's secret is unavailable, both the join PUT and the self-heal fall back to `SealedBytes::public(...)`, publishing the nickname in cleartext. The PR makes this reliably reach the wire. |
| Heal entry minted at `version: 0` | Low (verified safe) | `room_data.rs` heal default branch | Safe — the heal only fires when no entry exists (`has_member_info == false`), and `nickname_field.rs:79` bumps a later edit to `version + 1`. Wants an explanatory comment. |
| Redundant heal UPDATEs across a GET race window | Low | `get_response.rs` | Idempotent; acceptable. |

Verified clean: `ContractKey` is `Copy` (no use-after-move); owner / already-member /
banned-member paths all early-return correctly; no heal loop; deterministic nickname.

---

### Big Picture Assessment
**Goal Alignment:** yes — fixes the regression at root cause + remediates stranded data.
**Anti-Patterns Detected:** none — no removed tests, no weakened assertions, no `#[ignore]`.
**Removed Code Concerns:** none — the large `get_response.rs` hunk is the isolated
`style` rustfmt commit plus the logic change; all prior logic verified intact.
**Scope Assessment:** the folded-in nickname feature is mild scope creep but
acceptable — `build_member_info_heal` mints a default nickname, so the two are coupled.

---

### Documentation
- Code docs: good on new public items; a couple of implicit invariants want comments.
- `AGENTS.md`: should gain a bullet in the private-rooms/join section documenting the
  member_info-injection invariant (PUT must carry the invitee's self-signed
  member_info) and the self-heal-on-GET behavior — to prevent a future PR #272-style
  re-drop.

---

### Recommendations

#### Must Fix (Blocking)
1. **Imported-room self-heal is dropped (Codex P2).** In the `is_existing_room` branch,
   the heal UPDATE is sent *before* the `needs_put_subscribe` PUT registers the
   contract code with the local node. The codebase's own comment (join-branch PUT)
   states UPDATEs fail without that registration. For an imported room the heal is
   therefore lost. Fix: fold the heal `member_info` into the PUT payload for the
   `needs_put_subscribe` path; send the standalone UPDATE only for already-subscribed
   rooms. (code-first rated this low on the assumption the contract exists
   network-side — but the blocker is the *local* node lacking the contract code, so
   Codex's P2 stands.)

#### Should Fix (Important)
2. **Private-room plaintext-nickname leak (skeptical M1).** Make `build_member_info_heal`
   never publish a `Public`-sealed nickname into a private room — defer (return `None`)
   when the secret is unavailable. Apply the same rule to the invitation-accept PUT
   path: skip the member_info injection when the room is private and the secret cannot
   be obtained (the self-heal republishes it, properly sealed, once the secret arrives).
3. **Test gaps:** add coverage for `current_secret_from_state`, the private-room
   `build_member_info_heal` path (defer-on-no-secret + seal-on-secret), a signature-
   validity assertion on the healed entry, and member_info-only delta application +
   idempotency.

#### Consider (Suggestions)
4. Add the `AGENTS.md` private-rooms/join bullet (above).
5. Add explanatory comments for the `version: 0` heal invariant and the pre-PUT
   ordering of the heal UPDATE.
6. File/link a tracking issue for the "10 of 33 stranded members" remediation.

---

### Verdict
**HEAD SHA reviewed:** 32a82c25d6492ea8d7cb5f1f9fc89f77451725ad

Blocking: finding #1 (imported-room heal dropped). Strongly recommended before merge:
#2 (private-room leak) and #3 (test gaps). A full re-review will run on the updated
HEAD before merge.

[AI-assisted - Claude]

sanity (commented):
## Re-Review (rounds 2–3): PR #294

Re-review after fixes for the round-1 findings (imported-room heal dropped;
private-room plaintext-nickname leak). Two further fix rounds followed.

### Round-2 review (HEAD `fec86ecf`) — findings & resolution

- **Codex P1 / skeptical H1+H2 / code-first Concern 1 (BLOCKING):**
  `build_member_info_heal` decided privacy mode and the room secret from
  `self` (`is_private()` / `get_secret()`). For an *imported* room `self`'s
  `room_state` is a public placeholder and `secrets` is empty at
  heal-build time, so a private imported room was misclassified as public
  and could fold a plaintext nickname into its PUT.
  → **Fixed in `54a75a86`:** the heal now derives privacy mode from
  `state.configuration.configuration.privacy_mode` and the secret via
  `current_secret_from_state(state, …)` — both from the fetched network
  `state`. `current_secret_from_state` moved to `room_data.rs` (shared).
- **Skeptical L1:** dropped heal UPDATE was silent when `WEB_API` is
  unavailable → now logged.
- **Testing:** `build_state_for_put`'s `None` branch untested → added
  `build_state_for_put_omits_member_info_when_none`.

### Round-3 review (HEAD `54a75a86`)

- **Skeptical:** "H1/H2 genuinely closed; no new high/medium bugs;
  round-3 fix is sound." Traced every nickname-to-wire path — no plaintext
  leak remains.
- **Code-first:** committed code correct, the Concern-1 fix genuine and
  complete. (Code-first also reported a "blocking uncommitted edit" — this
  was the testing reviewer's revert-based mutation testing transiently
  editing the *shared review worktree*, seen mid-mutation by the
  concurrent code-first run. Verified false positive: committed
  `54a75a86` and the author's working tree both carry the correct
  `state`-derived code.)
- **Testing:** 192 tests pass; mutation-tested that the new tests catch
  their bugs. Found the secret-*source* line was unpinned (all private
  tests seeded the secret in both places).
  → **Fixed in `d52c52fc`:** added `build_member_info_heal_reads_secret_from_state_not_self`
  (reviewer-written, reviewer-verified — fails against the reverted line).
- **Codex P2 / skeptical L1 (low):** `build_state_for_put`'s
  `already_member` early-return skips member_info injection. Self-resolves
  via the GET-path heal (same class as the deferred-heal gap) → documented
  with a code comment.

### Deferred (filed as issues, not blocking)

- **#295** — a private-room invitee whose secret arrives via a
  subscription UPDATE (not a GET) heals only on the next GET. Private
  rooms only; self-resolves on app reload; the public official room is
  unaffected. A proper fix (heal on the UPDATE path / in `process_rooms`)
  is an architectural change for its own PR.
- **#296** — pre-existing flaky test `clear_ensure_subscription_sent_is_scoped_to_supplied_vk`
  in `chat_delegate.rs` (untouched by this PR).

### Testing
193 river-ui native unit tests pass (15 new across the PR). Native +
`wasm32` builds clean. Every behavioural change is pinned by a test that
fails against the reverted code (verified by the testing reviewer's
mutation testing).

### Verdict
**State:** Ready to Merge (pending CI green on HEAD).
**HEAD reviewed:** round-3 reviewers ran on `54a75a86`; the subsequent
`d52c52fc` is a purely-additive touch-up (one reviewer-verified test +
two clarifying comments, no logic change) — a light diff-of-the-diff
re-check confirms it introduces no behavioural change.

All blocking findings from rounds 1–3 are resolved. Remaining items are
low-severity and either documented in-code or tracked in #295/#296.

[AI-assisted - Claude]

PR #311, opened 2026-05-24
fix(ui): coalesce save_rooms_to_delegate to cap open-time memory burst (#246)

sanity:
Fixes #246.

## Summary

Coalesces `save_rooms_to_delegate` so a chain of N rapid callers produces at most 2 actual saves (the in-flight one + one catch-up). Also caches the chat delegate's `DelegateKey` in a `LazyLock` so the 710 KB WASM allocation and BLAKE3 hash only run once per session.

## Why

@Ivvvor reported River climbing at 100-400 MB/s on page open / refresh, peaking 3-4 GB commonly and 16 GB on the worst incident, then "suddenly all freed" (renderer OOM-killed). Steady-state operation was stable; the burst is specific to the multi-room cold-open / refresh path.

Trace: every per-room state notification and catch-up delta on cold open spawns its own `spawn_local(save_rooms_to_delegate())` from `update_room_state_inner` and `apply_delta_inner`. Each call cloned the full ROOMS map, ciborium-serialized it, allocated a fresh 710 KB delegate WASM blob, and bincode-wrapped it for the WebSocket. With N rooms and several events per room landing close together during startup, dozens of these MB-scale saves stacked up faster than the WebSocket drained them. WASM linear memory is grow-only, so each peak became permanent RSS until the renderer died.

Full code-trace writeup with file:line refs lives at https://github.com/freenet/river/issues/246#issuecomment-4529990226.

## The fix

`save_outbound_dms_to_delegate` already used the right pattern (`futures::lock::Mutex<()>` + `AtomicBool` DIRTY flag) — see `chat_delegate.rs:1499-1515` on main. This PR extracts that pattern as a generic `coalesce_save` primitive and wires `save_rooms_to_delegate` through it. The expensive `ROOMS.read().clone()` + ciborium serialize now lives inside `do_save_rooms_to_delegate` and only runs when a save actually executes, not once per queued caller.

The `LazyLock<DelegateKey>` ride-along removes a ~14 MB block of pure-waste allocations on cold open (the same 710 KB WASM blob was being cloned and re-BLAKE3-hashed from three hot call sites: `fire_load_rooms_request`, `fire_load_outbound_dms_request`, `send_delegate_request`). The rare `create_chat_delegate_container` path (called once at startup) is left untouched.

## Test coverage

- New unit tests in `chat_delegate.rs` pin both bounds of the coalesce loop:
  - `coalesce_save_drains_dirty_re_entry_into_one_extra_save` — re-entry mid-save produces exactly two iterations.
[...]

PR #312, opened 2026-05-25
perf(ui): drop dead-code parent_state clone in apply_delta/update_room_state (#246)

sanity:
Follow-up to #311 / #246. Ivvor reports memory spikes on cold open dropped from the 16GB-crashing pre-#311 level to 3-4GB-but-stable. Real user-visible win on the "doesn't crash my iPad anymore" axis, but 3-4GB is still big — this PR chips at the next per-event allocation source.

## What changes

`apply_delta_inner` and `update_room_state_inner` in `room_synchronizer.rs` each cloned the full `room_data.room_state` to pass as the `parent_state` argument to `apply_delta` / `merge`. That argument is **dead-code** at the macro level:

- Every field's `summarize` / `delta` impl in `common/src/room_state/` takes `_parent_state` (underscore-prefixed = unused, verified across `ban.rs`, `configuration.rs`, `direct_messages.rs`, `member_info.rs`, `member.rs`, `message.rs`, `secret.rs`, `upgrade.rs`, `version.rs`).
- `freenet-scaffold-macro` 0.2.2 generates an `apply_delta` that ignores its outer `_parent_state` arg entirely. It computes a fresh `let self_clone = self.clone();` *per field* and uses that as the per-field parent.

So the caller-side full-state clone never reaches code that reads it. Substituting `&ChatRoomStateV1::default()` is provably equivalent. Saves one full-state clone per network delta / state update.

## Test coverage

New regression test `merge_with_default_sentinel_parent_matches_merge_with_self_clone_parent` empirically pins the invariant — it runs `merge` both ways and asserts byte-equality of the post-merge state. If a future refactor starts reading the outer `parent_state` (in the macro, or in a field's `summarize` / `delta`), this test fails and the optimization needs revisiting.

`cargo test -p river-ui --bins`: 242/242 pass. WASM check green. `cargo fmt --check` clean.

## Honest sizing — what this does NOT fix

This is a ~10% per-merge reduction. The **dominant** allocation source remains the macro-internal `let self_clone = self.clone();` *per field* inside `apply_delta` (`~9× full-state clones per merge for ChatRoomStateV1`). Fixing that would mean modifying `freenet-scaffold-macro` to hoist `self_clone` outside the field loop — single shared clone reused across all fields — which is a separate-crate PR.

This PR is worth landing on its own merits (provably safe, focused diff, real-but-bounded reduction), with the bigger macro refactor tracked as a follow-up if the residual spike still bothers users after this lands.

## Deploy notes

UI-only change. Single file: `ui/src/components/app/freenet_api/room_synchronizer.rs`. No delegate/contract WASM modified, no migration entry needed, no riverctl republish needed.

### DIFF
--- a/ui/src/components/app/freenet_api/room_synchronizer.rs
+++ b/ui/src/components/app/freenet_api/room_synchronizer.rs
@@ -1050,15 +1050,29 @@ impl RoomSynchronizer {
                     );
                 }
 
-                // Update the room state by merging the new state with the existing one.
-                // The `parent_state` arg is dead-code (see `apply_delta_inner`
-                // above for the full rationale + verification) so we pass a
-                // cheap default sentinel rather than cloning the full
-                // `room_data.room_state`. This saves one full-state clone per
-                // network state-update event — the equivalent change on the
-                // apply_delta path is at the top of this file. Together they
-                // chip away at the per-event allocation cost that survived the
-                // initial `coalesce_save` fix for freenet/river#246.
+                // Update the room state by merging the new state with the
+                // existing one. The `parent_state` arg is dead-code at the
+                // macro/trait level for the `merge` call too — but the proof
+                // shape is slightly different from `apply_delta_inner` and
+                // worth spelling out: the default `ComposableState::merge` in
+                // `freenet-scaffold` calls `self.summarize(parent_state,...)`,
+                // `other.delta(parent_state,...)`, and `self.apply_delta(parent_state,...)`.
+                // The macro-generated `summarize`/`delta` forward `parent_state`
+                // to each field's impl, so safety here ALSO depends on every
+                // field-level `summarize`/`delta` in `common/src/room_state/`
+                // declaring the arg as `_parent_state` (unused) — verified
+                // across all 9 fields. The `apply_delta` leg is protected by
+                // the macro's per-field `self.clone()` as documented on the
+                // `apply_delta_inner` call site above. We pass a cheap default
+                // sentinel rather than cloning the full `room_data.room_state`;
+                // saves one full-state clone per network state-update event.
+                // Together with the equivalent change on the `apply_delta`
+                // path, this chips at the per-event allocation cost that
+                // survived the initial `coalesce_save` fix for
+                // freenet/river#246. The regression test
+                // `merge_with_default_sentinel_parent_matches_merge_with_self_clone_parent`
+                // pins this invariant against future macro / field-impl
+                // refactors that would break the substitution.
                 let parent_sentinel = ChatRoomStateV1::default();
                 match room_data.room_state.merge(
                     &parent_sentinel,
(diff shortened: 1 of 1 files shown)


sanity (commented):
## PR Review (Light tier — appropriate for a 1-file, +83/-5 perf change)

- **Tier:** Light — single-file UI-only perf change. Codex external pass + skeptical-reviewer adversarial read.
- **CI:** in progress on `c556b795` at review time. Per `no-merge-with-red-ci`, must be green on the current HEAD before merge.

### Findings

| Reviewer | Result |
|---|---|
| Codex external | Clean, no findings. Verified the substitution is safe under current macro + field impls. |
| Skeptical | Confirmed the substitution is correct. One **Medium** finding: the round-1 regression test was tautological on near-default data — `state_a` only diverged from `default()` in `recent_messages.messages`, so a hypothetical macro regression that forwards outer `_parent_state` to a field's `apply_delta` would still see equivalent values across both paths and the test would silently pass. |

### Round 2 (`c556b795`) — addresses the Medium finding

Round-2 commit is purely test + comment, **zero production-code changes** (verified via diff-of-the-diff). Per the skill's "trivial mechanical findings" allowance, a Light re-check is sufficient — that's what this comment is.

- Test now sets `state_a` with a non-default `Configuration` (`max_members = 3` vs default 200, `max_recent_messages = 5` vs default 100) AND a non-owner `AuthorizedUserBan`. Those are the exact fields `MembersV1::apply_delta`, `MessagesV1::apply_delta`, and `MemberInfoV1::apply_delta` read from `parent_state`. The test now genuinely discriminates the regression class it claims to guard.
- Comment on the `merge` call site spelled out that safety on the merge path depends on BOTH the macro's `apply_delta` ignoring outer `_parent_state` AND every field-level `summarize`/`delta` declaring it unused — the previous comment deferred to the `apply_delta_inner` rationale, which only covered one of those two facts.

### Non-blocking observations (skeptical's other points)

- **Scope**: similar `&room_state.clone()` patterns exist in `cli/src/api.rs:1656, 1750, 1902, 1961` and a few other UI sites. Deliberately deferred — this PR targets the open-time hot path. Worth a follow-up if the CLI's allocation profile turns out to matter.
- **Macro version pinning**: the optimization's safety is tied to `freenet-scaffold-macro` 0.2.2. The Cargo.toml currently uses caret semver `"0.2.2"`. The regression test catches any future macro change that would break the substitution, so a hard `=0.2.2` pin isn't required — but worth keeping the test green-on-CI as the guard.
- **`ChatRoomStateV1::default()` cost**: trivial — constructs 9 default sub-field types, all small. Strictly cheaper than cloning a real `room_state` that's grown beyond defaults (which is the cold-open scenario).

### Verdict

**Ready to merge** pending CI green on `c556b795`. Substitution provably safe under current code. Round-1 finding addressed. Honest sizing: ~10% per-merge reduction (macro-internal per-field clones still dominate); won't bring 3-4 GB residual spike down to MB but every win compounds.

[AI-assisted - Claude]

PR #317, opened 2026-05-26
fix(ui): refuse private-room edits when secret isn't local yet (#299)

### LINKED ISSUE #299: Nickname edit in a private room publishes plaintext when the room secret isn't locally available
## Problem (pre-existing, surfaced during PR #297 review)

In `ui/src/components/members/member_info_modal/nickname_field.rs`, `save_changes` seals the edited nickname like this:

```rust
let sealed_nickname = match current_secret_opt {
    Some((secret, version)) => seal_bytes(new_value.as_bytes(), &secret, version),
    _ => SealedBytes::public(new_value.into_bytes()),
};
```

`current_secret_opt` is `None` both for a public room (correct — public rooms have no secret) **and** for a private room whose secret has not yet been decrypted locally. In the latter case the edit publishes the nickname as `SealedBytes::public` — **plaintext** — into a private room's `member_info` delta.

## Why it matters

A private room's nicknames are supposed to be encrypted. Editing your nickname before the room secret has arrived locally leaks the plaintext nickname onto the network.

## Scope

Pre-existing — not introduced by PR #297. PR #297 added a `build_rejoin_delta` / `build_member_info_heal` guard that prevents a stale Public-sealed `self_member_info` from being *re-published* into a private room, so the cached-state path is already neutralized. This issue is specifically the **edit → delta** path.

## Fix direction

`NicknameField` should read `room_data.is_private()`. When the room is private and no secret is available, the edit must not publish a plaintext `member_info` — defer the member_info part of the edit (the self-heal will seal it once the secret arrives) or surface that the nickname can't be changed yet.

[AI-assisted - Claude]

sanity:
## Problem

Three UI edit paths in River fell through to `SealedBytes::public` (plaintext) when the locally-decrypted room secret was not yet available, leaking plaintext into private rooms:

- `NicknameField::save_changes` — the #299 regression
- `RoomNameField::update_room_name` — same shape, owner-only
- `EditRoomModal::RoomDescriptionField::update_description` — same shape, owner-only

All three carried the same `match current_secret_opt { Some(...) => seal_bytes(...), _ => SealedBytes::public(...) }` block. In a private room the `_` arm publishes the user's chosen field as plaintext into a nominally encrypted contract delta. The window is the gap between joining a private room and receiving the room secret — narrow in practice but silent, permanent, and propagated to the network.

## Approach

Centralised the privacy decision in a single helper `seal_for_room(is_private, current_secret_opt, plaintext)` in `ui/src/util/ecies.rs`:

- public room → `Some(SealedBytes::public(...))` (correct — no encryption to defer)
- secret available → `Some(seal_bytes(...))`
- **private room + no secret → `None`** (caller must defer)

All three call sites now consult the helper. When it returns `None`, the caller logs a warning, reverts the input signal to the on-network value (so the UI doesn't lie about what was saved), and skips publishing — no plaintext delta ever leaves the device. The user can retry once the secret has arrived.

The nickname-field save path was also restructured to re-read ROOMS at save time so the `is_private`, the secret, and the rejoin probe all see the same consistent snapshot — the captured-at-mount `current_secret_opt` could miss a secret that arrived after the modal opened.

The matching `EditRoomModal` empty-description case is intentionally exempt: `Some(empty)` is published as `None` (clears the field), and there's nothing to leak.

## Testing

Four unit tests on `seal_for_room` pin the decision matrix, including the #299 regression case (`private_room_without_secret_defers`) which fails if the function ever returns `Some(SealedBytes::public(...))` for the private+no-secret path:

- `public_room_with_no_secret_returns_public_sealed`
- `public_room_with_secret_uses_secret_for_consistency`
- `private_room_with_secret_seals` (with round-trip decrypt check)
- `private_room_without_secret_defers` — pins the regression

Local validation:

- [x] `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` — clean
- [x] `cargo test -p river-ui --bins` — 4 new tests pass + all 242 pre-existing tests pass
- [x] `cd common && cargo test --no-default-features --features migration private_room` — clean
- [x] `cargo fmt` clean

The save-path closures themselves aren't unit-tested (tightly bound to Dioxus signals and global state), but the helper carries the privacy invariant they all depend on — the regression is pinned at the only place it can be expressed in a unit test.

Closes #299

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/members/member_info_modal/nickname_field.rs
+++ b/ui/src/components/members/member_info_modal/nickname_field.rs
@@ -1,5 +1,5 @@
 use crate::components::app::{CURRENT_ROOM, ROOMS};
-use crate::util::ecies::{seal_bytes, unseal_bytes_with_secrets};
+use crate::util::ecies::{seal_for_room, unseal_bytes_with_secrets};
 use dioxus::logger::tracing::*;
 use dioxus::prelude::*;
 use dioxus_free_icons::icons::fa_solid_icons::FaPencil;
@@ -7,15 +7,19 @@ use dioxus_free_icons::Icon;
 use freenet_scaffold::ComposableState;
 use river_core::room_state::member::MemberId;
 use river_core::room_state::member_info::{AuthorizedMemberInfo, MemberInfo};
-use river_core::room_state::privacy::SealedBytes;
 use river_core::room_state::{ChatRoomParametersV1, ChatRoomStateV1Delta};
 use std::collections::HashMap;
 use std::rc::Rc;
 
 #[component]
 pub fn NicknameField(member_info: AuthorizedMemberInfo) -> Element {
-    // Compute values
-    let (self_signing_key, room_secrets, current_secret_opt) = {
+    // Compute values — `self_signing_key` and `room_secrets` are read once
+    // at mount for `is_self` gating and version-aware display decryption.
+    // The room's secret (`current_secret_opt`) is intentionally NOT captured
+    // here: it can arrive after the modal opens, and `save_changes` re-reads
+    // it fresh from ROOMS so the freenet/river#299 privacy guard sees the
+    // current state, not a stale snapshot.
+    let (self_signing_key, room_secrets) = {
         let current_room = CURRENT_ROOM.read();
         if let Some(key) = current_room.owner_key.as_ref() {
             ROOMS
@@ -23,16 +27,12 @@ pub fn NicknameField(member_info: AuthorizedMemberInfo) -> Element {
                 .ok()
                 .and_then(|rooms| {
                     rooms.map.get(key).map(|room_data| {
-                        (
-                            Some(room_data.self_sk.clone()),
-                            room_data.secrets.clone(),
-                            room_data.get_secret().map(|(s, v)| (*s, v)),
-                        )
+                        (Some(room_data.self_sk.clone()), room_data.secrets.clone())
                     })
                 })
-                .unwrap_or((None, HashMap::new(), None))
+                .unwrap_or((None, HashMap::new()))
         } else {
-            (None, HashMap::new(), None)
+            (None, HashMap::new())
         }
     };
 
@@ -53,6 +53,7 @@ pub fn NicknameField(member_info: AuthorizedMemberInfo) -> Element {
             Ok(bytes) => String::from_utf8_lossy(&bytes).to_string(),
             Err(_) => member_info.member_info.preferred_nickname.to_string_lossy(),
         };
+    let initial_nickname_for_revert = initial_nickname.clone();
     let mut temp_nickname = use_signal(|| initial_nickname);
     let mut input_element = use_signal(|| None as Option<Rc<MountedData>>);
 
@@ -61,6 +62,7 @@ pub fn NicknameField(member_info: AuthorizedMemberInfo) -> Element {
 
         let self_signing_key = self_signing_key.clone();
         let member_info = member_info.clone();
+        let initial_nickname_for_revert = initial_nickname_for_revert.clone();
 
         move |new_value: String| {
             if new_value.is_empty() {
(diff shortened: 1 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #317

### Summary
- **PR Title:** fix(ui): refuse private-room edits when secret isn't local yet (#299)
- **Type:** fix (privacy / cryptography surface)
- **Linked Issues:** Closes freenet/river#299; spawned follow-up freenet/river#318
- **Review tier:** Full (cryptography / privacy)
- **Reviewers run:** code-first, testing, skeptical, big-picture, plus an external Codex pass — then a re-review (skeptical + Codex) on the updated HEAD

---

### Code-First Analysis
**Independent Understanding:** Introduces `seal_for_room(is_private, current_secret_opt, plaintext) -> Option<SealedBytes>` at `ui/src/util/ecies.rs` with three branches keyed on `is_private`: public → `Some(public)`; private+secret → `Some(seal_bytes)`; private+no-secret → `None` (caller defers). All three previously-buggy edit sites (`nickname_field`, `room_name_field`, `edit_room_modal::RoomDescriptionField`) route through it and skip publish on `None`. `nickname_field`'s save closure was restructured to re-read ROOMS at save time so `is_private`, the secret, and the rejoin probe come from one consistent snapshot.

**Alignment with stated intent:** Matches. Scope expansion to the two analogous bugs in the room-name and description paths is correct application of the `finish-the-fix` rule.

---

### Testing Assessment
**Coverage Level:** Adequate.

| Unit (helper) | ✅ | Five tests pin the full `(is_private × secret)` matrix + a source-grep call-site pin + a substring-absence assertion against `to_string_lossy` leakage |
| Closure-level | ⚠️ | Save closures aren't unit-testable (Dioxus globals); the pure-helper test plus the source-grep pin are the substitute |
| Integration (common) | N/A | No regression in `common/tests/private_room_test.rs` |

**Regression Test:** `private_room_without_secret_defers` pins the exact #299 case (asserts `None`, not `Some(SealedBytes::public)`). Defended further by `seal_for_room_call_sites_pinned` which strips line comments before asserting `seal_for_room(` is present and `SealedBytes::public(` is absent in each of the three edit-delta files.

---

### Skeptical Findings
**Risk Level:** Low after fixes.

| Concern | Severity | Location | Status |
|---------|----------|----------|--------|
| Public→private mid-edit race (Codex) | Medium | `nickname_field.rs::save_changes` defer | **Filed freenet/river#318** — narrow window; `Configuration::apply_delta` already backstops room name + description; member_info has no such backstop |
| Helper match ordered `Some(secret)` first → doc/code mismatch | Medium | `ecies.rs:36-42` | **Fixed** — reordered so `is_private` is the primary axis |
| Silent fall-through on `ROOMS.try_read()` failure / no room | Medium | `nickname_field.rs:81-87` | **Fixed** — added `warn!` logs |
| Missing source-grep pin for call sites | Important | — | **Fixed** — `seal_for_room_call_sites_pinned` added, then hardened per Codex re-review to strip comments first |
| Substring-absence assertion on private+secret seal | Nice | `ecies.rs` tests | **Fixed** — added to `private_room_with_secret_seals` |
| `temp_nickname` not refreshed on prop changes mid-edit | Low | `nickname_field.rs:60` | **Pre-existing, not in scope** — worth a separate issue |
| No UI surface for retry when seal defers | Low | All three call sites | **Pre-existing UX gap** — worth a separate issue |
| Empty-description metadata channel | Low | `edit_room_modal.rs:407-410` | **Accepted per River's threat model** |

All findings either fixed or rationally deferred.

---

### Big Picture Assessment
**Goal Alignment:** Yes — closes the documented #299 residual from PR #297. Scope is appropriate (covers the two analogous adjacent paths with identical bug shape). No dead-code removal or test weakening.

**Removed Code Concerns:** None — only the inlined buggy `match` blocks and the captured-at-mount snapshot were removed; both were the bug.

---

### Documentation
- Helper doc on `seal_for_room`: tightened to state the caller contract ("treat `None` as skip publish") explicitly.
- `ecies.rs:20` bullet list rewritten to match the new implementation.
- Optional: a bullet under AGENTS.md "Private Room Support" noting `seal_for_room` as the canonical helper — not done; the docstring on the helper carries the same information.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
All previously-flagged Should Fix items are addressed in commits `428136b5` and `25716d8f`.

#### Consider (Suggestions / Follow-ups)
1. **freenet/river#318** — the public→private mid-edit race for member_info. Filed.
2. **temp_nickname prop refresh** — pre-existing UX issue; sync round can land a newer nickname under an open modal and a save would publish the older value. Worth filing.
3. **Retry UI surface** — no user-visible signal when the seal helper defers a private-room edit. Currently logs `warn!` only.
4. **Defense-in-depth in `common`** — extending `MemberInfoV1::apply_delta` with the same privacy-mode guard that `Configuration::apply_delta` already has would make the entire class of #299-shaped leaks impossible regardless of UI surface. Contract change → requires delegate WASM migration.

---

### Verdict
**State:** Ready to Merge.
**HEAD SHA reviewed:** `25716d8f`

Two prior review rounds: round-1 surfaced two helper-level items (match ordering + doc/code mismatch) and a P3 pin-robustness concern; both addressed. Round-2 re-review (skeptical + Codex) on `428136b5` returned "Ready to Merge" from skeptical and one residual P3 from Codex (pin could be satisfied by a comment) — fixed in `25716d8f` by stripping line comments before the assertion. Test suite passes. CI must be green on the current HEAD before merge.

[AI-assisted - Claude]

PR #321, opened 2026-05-30
fix(cli): restore real nickname on rejoin instead of "Member"

sanity:
## Problem

When a riverctl member is pruned for inactivity and later posts a message, `build_rejoin_delta` correctly re-adds them to the members list — but re-published their `member_info` with a **hardcoded "Member"** nickname. So a user who joined as "Bob", went quiet long enough to be pruned, then spoke again, reappeared to every peer as **"Member"** instead of "Bob".

Root cause: the member's real nickname lived only in the room's `member_info`, which is pruned alongside the member, and riverctl never persisted it locally. With nothing to restore from, the rejoin path fell back to a placeholder.

(Found while investigating, at the user's request, whether riverctl re-adds a member on post — it does; this is the one rough edge in that path.)

## Approach

- Persist the member's own nickname in a new `StoredRoomInfo.self_nickname: Option<String>` field (`#[serde(default)]`, so old `rooms.json` files still load), set on `accept_invitation` and `set_nickname`.
- On rejoin, `build_rejoin_delta` now seals the stored nickname via the existing `private_room::seal_invitee_nickname` helper: **public** bytes for a public room, **sealed** bytes for a private room.
- Falls back to the generic "Member" placeholder only when (a) no nickname was persisted (rooms joined before this field) or (b) a private room has no secret available — so we **never leak a plaintext nickname** into a private room (mirrors the UI's heal behavior).
- The room owner is exempt from rejoin (`build_rejoin_delta` returns early for the owner), so the owner's nickname is intentionally not persisted.

CLI-only — no `common/`/contract/delegate/WASM change, **no migration, no UI republish**.

## Testing

**Unit** (`cli/src/storage.rs`):
- `test_update_self_nickname_round_trips` — persists and survives a `load_rooms` reload
- `test_update_self_nickname_missing_room_is_noop` — no error when the room isn't stored
- `test_stored_room_info_without_self_nickname_field_defaults_none` — backward-compat: old rooms.json without the field deserializes with `None`

**End-to-end** on a local node: created a room, "Bob" joined (confirmed `self_nickname = "Bob"` persisted), aged his join-event out of a 2-message window to prune him (members → just Owner), then had Bob repost. Before this change he came back as "Member"; with it he rejoins as **"Bob"**.

`cargo fmt`, `cargo clippy -p riverctl --all-targets` (clean), and all 49 CLI lib tests pass.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/private-rooms.md
+++ b/.claude/rules/private-rooms.md
@@ -150,6 +150,30 @@ indefinitely (filed as freenet/river#304). The CLI also does NOT prune
 superseded `invitation_secrets` entries the way the UI does — storage
 waste only; the heal path is the natural place to hook the prune.
 
+### Rejoin nickname restoration (`StoredRoomInfo.self_nickname`)
+
+The CLI persists the member's own nickname in
+`StoredRoomInfo.self_nickname` (set on `accept_invitation`,
+`set_nickname`, and public-room `import_identity`). When a member is
+pruned for inactivity and later reposts, `ApiClient::build_rejoin_delta`
+re-adds them AND restores this nickname via the free helper
+`rejoin_preferred_nickname` instead of the generic `"Member"`
+placeholder. The helper routes the nickname through
+`seal_invitee_nickname` (public bytes for a public room, sealed for a
+private room) and falls back to public `"Member"` when no nickname is
+stored, a private room has no secret (so the plaintext is never leaked),
+OR the stored nickname exceeds the room's current `max_nickname_size`
+(which would otherwise make the contract reject the whole rejoin delta).
+
+This is a PARTIAL analog of the UI's `build_member_info_heal`, NOT the
+full #304 heal: it only restores the nickname on the rejoin/send path,
+and it diverges from the UI in the private-room-no-secret case — the UI
+*defers* `member_info` (member shows "Unknown"), whereas the CLI
+publishes a public `"Member"` placeholder (pre-existing CLI behavior;
+safe — no plaintext leak). The call site is pinned by
+`rejoin_nickname_wiring_pinned` in `cli/src/storage.rs` and the
+selection matrix by `rejoin_nickname_tests` in `cli/src/api.rs`.
+
 ## Key files
 
 - `common/src/room_state/privacy.rs`, `secret.rs`, `configuration.rs`
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -960,6 +960,11 @@ impl ApiClient {
                             &invite_chain,
                         )?;
 
+                        // Persist our chosen nickname so a later rejoin (after
+                        // an inactivity prune) restores it instead of "Member".
+                        self.storage
+                            .update_self_nickname(&room_owner_vk, nickname)?;
+
                         // Immediately publish membership + join event atomically.
                         // The join event counts as a message, preventing
                         // post_apply_cleanup from pruning the new member.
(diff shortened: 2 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #321

### Summary
- **PR Title:** fix(cli): restore real nickname on rejoin instead of "Member"
- **Type:** fix (riverctl)
- **Review tier:** Full — touches private-room nickname sealing (privacy surface)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four) + external Codex

### Verdict on the originally-reviewed commit (c88b2672): no Must-Fix, no privacy leak
The skeptical reviewer traced every return path of `seal_invitee_nickname` and confirmed the real nickname can **never** be published as plaintext into a private room (sealed when a secret is available; generic public "Member" otherwise). Version semantics on rejoin (member_info at v0 after prune) verified correct. Findings were Should/Consider/medium — all now addressed in **16bae8991ff4791e05632042ab9030c92b84526c**.

---

### Findings and resolutions

| 1 | Codex + skeptical | Should | Republishing the real nickname on rejoin could exceed a room's current `max_nickname_size` (if the owner lowered it after join), making `MemberInfoV1::apply_delta` reject the **entire** rejoin delta and block the member from rejoining — a regression vs the 6-byte "Member" placeholder. | Extracted selection into unit-testable `rejoin_preferred_nickname`; clamps to `max_nickname_size` (comparing `nick.len()`, which matches `declared_len()`'s plaintext-byte semantics for both public and sealed values) and falls back to "Member" when too long. |
| 2 | code-first | Should | `import_identity` didn't persist `self_nickname`, so an imported member hit the same "Member" bug on rejoin. | Persist on import for **public** rooms (a private room's exported nickname is sealed → not recoverable as plaintext, so skipped). |
| 3 | testing | Should | No regression pin on the rejoin call site / no direct test of the selection matrix. | Added 5 unit tests (public→real, private+secret→sealed, private-no-secret→"Member", too-long→"Member", None→"Member") + a source-grep pin `rejoin_nickname_wiring_pinned` (in storage.rs, so pinned strings aren't self-satisfied) guarding the rejoin routing and all three persistence call sites. |
| 4 | big-picture | medium | `.claude/rules/private-rooms.md` CLI-parity section was stale. | Documented `self_nickname` persistence + rejoin restoration, and the intentional divergence from the UI's defer-on-no-secret heal. |
| 5 | code-first + skeptical | Consider (pre-existing) | `set_nickname` publishes a public (plaintext) nickname on-wire even for private rooms. | Out of scope — pre-existing CLI behavior, not introduced here. Noted for a future tracking issue. |

---

### Testing Assessment
**Coverage: adequate.** Selection matrix unit-tested directly (no node needed); storage persistence round-trip + backward-compat (`#[serde(default)]`) tested; call-site pins guard against refactor regressions; e2e prune→repost cycle on a local node confirmed the member rejoins as their real nickname. `build_rejoin_delta` itself is node-bound (not unit-testable), which is why the selection logic was extracted into a free function.

### Documentation
Code docs complete (new helper + field have rustdoc); `.claude/rules/private-rooms.md` updated. No WASM/contract/delegate change → no migration, no UI republish.

---

### Verdict
**State:** Needs Changes — Re-review Required After Fix → **fixes applied in 16bae8991ff4791e05632042ab9030c92b84526c**. Because 3+ Should-level findings were addressed and the diff changed by >30 lines, a fresh review pass on 16bae8991ff4791e05632042ab9030c92b84526c should run before merge per the per-CODE-CONTENT review rule.
**HEAD SHA:** 16bae8991ff4791e05632042ab9030c92b84526c

[AI-assisted - Claude]

sanity (commented):
## Re-review after fixes — HEAD 16bae899

Fresh pass on the updated HEAD (delta c88b2672 → 16bae899, which addressed all prior findings). Ran the adversarial skeptical reviewer + an independent external Codex pass.

**Both clean — no Must/Should/Consider findings.**

Verified on the delta:
- The `max_nickname_size` clamp uses `nick.len()` (plaintext bytes), which exactly matches the contract's `declared_len()` check for **both** public and sealed values (`seal_bytes` sets `declared_len_bytes = plaintext.len()`; sealing does not change `declared_len()`). The filter runs on plaintext before sealing — correct.
- `import_identity` `is_public()` guard is correct; a private room's sealed nickname is skipped (never persists an `[Encrypted: …]` placeholder).
- The source-grep pin (`rejoin_nickname_wiring_pinned`) is genuinely cross-file (lives in storage.rs, scans api.rs/identity.rs); all four pinned substrings exist at the real call sites.
- No panic/overflow/borrow risk in the new combinator chain.
- `max_nickname_size < 6` ("Member" itself too long) is pre-existing, not a regression.

CI green on 16bae899 (build, both migration checks, playwright). **Verdict: Ready to Merge.**

[AI-assisted - Claude]

Issue #173, opened 2026-03-18
Strange phantom message in river room state


Ivvvor:
While experimenting with using riverctl to subscribe to room messages I noticed a message appear that wasn't appearing in River UI.

I was using this to subscribe to the stream:

`riverctl message stream 4uNUKFzZQCnzo4K2ecZ16cMsYEEfoaRS35z6exEsbvm4 -f json --subscribe`

The message would come up like this:

`{"author":"XUUP2P5N","content":"Do you want to fully remove it, or update it?","edited":false,"message_id":"1722670768367664308","nickname":"SpaceCowboy26","reactions":{},"room":"4uNUKFzZQCnzo4K2ecZ16cMsYEEfoaRS35z6exEsbvm4","timestamp":"2026-03-18T04:03:24.698+00:00","type":"message"}`

The strange thing about this is if I later ran the riverctl stream command, as soon as a new message came up, the above message would come up first again.

Crucially, this behaviour repeated on another node.

PR #322, opened 2026-05-30
fix(cli): surface message edits and reply context in `message stream`

sanity:
## Problem

Reported by a bridge user (Rogue Worm): `riverctl message stream` is oblivious to **message edits** and **message replies**.

- **Edits:** the monitor deduped by message identity and emitted each message exactly once. An edit arrives as a separate *action* message that `display_messages()` filters out, and the original id was already "seen" — so an edited message's new text never reached the stream. The `--subscribe` path additionally skipped action deltas outright.
- **Replies:** `output_message` (used by both monitor paths) rendered no reply context, so a reply appeared as a plain message — unlike `message list`, which already shows `[reply to X: …]`.

`message list` (a snapshot) shows edited text + reply context fine; the gap is specific to the live stream that a bridge uses.

## Approach

- Track seen messages as **key → last-emitted effective content** instead of identity only. New shared `emit_new_and_edited` scans `display_messages` and re-emits a message whose effective content changed (an edit), flagged `is_edit`. Both monitor paths (polling `stream_messages` and `--subscribe`) now use it. The subscribe path re-fetches authoritative state on **any** notification (including action deltas) and rescans — which also replaces the brittle per-message delta parsing.
- `output_message` now renders reply context and marks edits:
  - Human: `[reply to X: preview…]` prefix; `[edit]` prefix on re-emitted edits.
  - JSON: new `reply_to: {author, preview}` field (null otherwise); `type: "edit"` for an edit re-emission.

Deleted messages remain excluded by `display_messages` and stay marked seen, so #173 (phantom deleted messages) still holds. **Deletes are still not surfaced as events** — noted as a follow-up, out of scope for this report.

**Consumer-facing JSON change:** adds a `reply_to` field and a `type: "edit"` value. Additive; existing `type: "message"` consumers are unaffected.

CLI-only — no `common/`/contract/delegate/WASM change, **no migration**.

## Testing

- **Unit:** `reply_context` (author + 50-char-truncated preview; `None` for plain/event messages) and `classify_seen` (New / Unchanged / Edited) in `monitor_tests`.
- **End-to-end** on a local node: a stream running *before* an edit+reply now emits
  ```
  message | 'second message'
  edit    | 'second message EDITED'   (type:"edit", edited:true)
  message | 'replying now'            (reply_to: {author:"Owner", preview:"second message EDITED"})
  ```
  Before this change only the first line appeared.

`cargo fmt`, `cargo clippy -p riverctl --all-targets` (clean), all 59 CLI lib tests pass.

[AI-assisted - Claude]

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -109,7 +109,10 @@ pub(crate) fn message_display_text(
 /// decode). Mirrors the reply rendering in `riverctl message list` so the
 /// monitor stream surfaces reply context too (it previously did not — a reply
 /// rendered as a plain message). freenet/river — Rogue Worm report.
-fn reply_context(
+///
+/// Shared with `riverctl message list` (cli/src/commands/message.rs) so the two
+/// renderings can't drift.
+pub(crate) fn reply_context(
     msg: &river_core::room_state::message::AuthorizedMessageV1,
 ) -> Option<(String, String)> {
     use river_core::room_state::content::{DecodedContent, CONTENT_TYPE_REPLY};
@@ -145,6 +148,17 @@ fn classify_seen(seen: &HashMap<String, String>, key: &str, content: &str) -> Em
         Some(prev) if prev == content => EmitKind::Unchanged,
         Some(_) => EmitKind::Edited,
     }
+}
+
+/// Stable dedup key for a message in a monitor stream: its signature-derived
+/// `MessageId`, NOT `author:time`. The id is unique per message and stable
+/// across edits (an edit is a separate action message; the original message's
+/// signature never changes), so two distinct messages from the same author with
+/// an identical timestamp cannot collide. Keying on `author:time` instead would
+/// let such a collision flip-flop forever as a spurious "edit" now that we
+/// compare effective content. freenet/river — PR #322 review.
+fn monitor_seen_key(msg: &river_core::room_state::message::AuthorizedMessageV1) -> String {
+    msg.id().0 .0.to_string()
 }
 
 /// Choose the `member_info` nickname to publish when re-adding a member who
@@ -2410,7 +2424,7 @@ impl ApiClient {
             let start = all_msgs.len().saturating_sub(initial_messages);
 
             for msg in &all_msgs[start..] {
-                let key = format!("{:?}:{:?}", msg.message.author, msg.message.time);
+                let key = monitor_seen_key(msg);
                 seen_messages.insert(key, message_display_text(&room_state, msg));
 
                 Self::output_message(&room_state, msg, room_owner_key, &format, false)?;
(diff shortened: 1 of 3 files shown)


sanity (commented):
## Comprehensive PR Review: #322

### Summary
- **PR Title:** fix(cli): surface message edits and reply context in `message stream`
- **Type:** fix (riverctl monitor)
- **Linked Issues:** none (bridge-user report); follow-up #323 filed for deletes-as-events
- **Review tier:** Full — rewrites both async streaming loops + changes consumer-facing JSON
- **Reviewers run:** code-first, testing, skeptical, big-picture + external Codex (original pass on 771d6003); skeptical + Codex re-review on the fixed HEAD

### Findings and resolutions

| 1 | skeptical | **Medium** | Monitor deduped on `author:time`; two same-author same-timestamp messages collided → with the new content-map they'd flip-flop forever as spurious `edit` re-emissions (old HashSet silently dropped them). | Key on signature-derived `msg.id()` via new `monitor_seen_key` (stable across edits, unique per message). Pinned by `monitor_seen_key_distinct_…` test. |
| 2 | code-first | Should | `message list --format json` lacked `reply_to` (asymmetry vs the stream). | Added `reply_to` to list JSON; `message list` now shares `crate::api::reply_context` with the monitor so renderings can't drift. |
| 3 | testing | Should | `reply_context` boundary cases untested. | Added short/empty/exactly-50/multi-byte-emoji preview tests. |
| 4 | testing | Consider | No pin that both monitor paths keep edit detection. | Added `monitor_edit_detection_wiring_pinned` (in storage.rs, cross-file) asserting both paths call `emit_new_and_edited` and key via `monitor_seen_key`. |
| 5 | code-first | Consider | `type:"edit"` vs `edited` boolean ambiguity. | Documented the distinction in `output_message`'s doc comment. |
| 6 | big-picture | Recommendation | Deletes left unsurfaced. | Filed #323; the stream is not *wrong* (it suppresses deletes, preserving #173), just silent on them. |

### Verified safe (skeptical, both passes)
- No deadlock in the rewritten subscribe loop: `drop(web_api)` precedes `get_room()` (which re-locks); nothing uses the guard after drop.
- Edit detection preserved by `msg.id()` keying (id = hash of signature, independent of `edited_content`; original message's id stable across edits).
- Retry-on-fetch-failure semantics preserved and improved (seen-map updated only after a successful emit).
- #173 (phantom deleted messages) still holds (deleted excluded by `display_messages`, pre-seeded as seen).
- No spurious edits from reaction changes or late private-room decryption (content is purely effective text).
- `message list` Human output byte-identical after the `reply_context` refactor.

### Testing
62 CLI lib tests pass (added: reply_context boundaries, monitor_seen_key collision/stability, classify_seen, wiring pin). End-to-end verified twice on a local node — a live stream emits a `type:"edit"` event on edit and a reply carrying `reply_to`. Streaming loops are node-bound; pure logic extracted into tested helpers + the cross-file pin.

### Documentation
Code docs complete; consumer-facing JSON change (`reply_to`, `type:"edit"`) documented in the PR body + code comments. CLI-only; no WASM/contract/delegate change → no migration.

### Verdict
**State:** Ready to Merge (pending green CI on current HEAD).
**HEAD SHA reviewed:** b53458e01cbace59c72906049e635c90a7a05a6c

Re-review after fixes for the Medium finding #1 (and Should #2/#3) — all resolved, no new bugs (skeptical + Codex both clean on b53458e0).

[AI-assisted - Claude]

Issue #2, opened 2024-10-05
To prevent the state filling up with defunct members, invites should need to be periodically renewed or they expire


sanity:
This would make an invite more like a limited time permit to speak, which would cascade to invited users. 

A separate feature, but such a mechanism could also be used to impose per-member limits on whether they can invite.

Issue #1, opened 2024-10-04
Handle scenario where public key is chosen to have colliding MemberId


sanity:

sanity:
Made MemberId a u64, deliberate collisions should be very difficult.

PR #324, opened 2026-05-30
fix(cli): surface message deletions as events in `message stream`

### LINKED ISSUE #323: riverctl: surface message deletions as events in `message stream`
## Background

PR #322 made `riverctl message stream` surface message **edits** and **reply context** (reported by a bridge user). Deletes were explicitly left out of scope.

## Current behavior

When a message is deleted, the monitor correctly **suppresses** it (it's excluded by `display_messages()` and stays marked seen, preserving the #173 phantom-message fix). But the stream emits **no event** for the deletion, so a downstream relay (e.g. an XMPP bridge) has no way to know a previously-relayed message was deleted — it just goes silent.

## Proposed

Emit a delete event from `message stream` when a previously-seen message becomes deleted, analogous to the new `type:"edit"` event:
- JSON: `{"type":"delete","message_id":...}` (or similar)
- Human: a `[deleted]` line referencing the message

Implementation sketch: the monitor's `emit_new_and_edited` scan (cli/src/api.rs) already tracks seen message ids → content. A deleted message drops out of `display_messages()`, so detecting "a key we've seen is no longer present (and is in `actions_state.deleted`)" would let it emit a delete event. Needs care so it fires once per deletion, not every poll.

Not urgent — the stream is not *wrong* today (it suppresses deletes rather than showing phantoms), just silent on them.

[AI-assisted - Claude]

sanity:
## Problem

Closes #323. Follow-up to #322 (edits + replies). `riverctl message stream` correctly **suppresses** deleted messages (excluded by `display_messages()`, preserving the #173 phantom-message fix) but emitted **no event** for a deletion. A downstream relay (e.g. an XMPP bridge) therefore had no way to know a previously-relayed message was deleted — it just went silent.

## Approach

Add `emit_deletions`, called after the new/edited scan in both monitor paths (polling + `--subscribe`):
- Scans `recent_messages` for any message now in `actions_state.deleted` whose key is in `seen` (the stream previously surfaced it) and not yet reported, and emits a deletion event **once** via `output_deletion`.
- `output_deletion`: JSON `{"type":"delete","message_id","author","nickname","timestamp","room"}`; human `[deleted] [HH:MM:SS - name]: (message deleted)`. The content is gone, so only identity/author/time are reported.
- A new `deleted_emitted` set tracks already-reported deletions and is **pre-seeded with deletions that existed at stream start**, so only deletions observed **live** are surfaced. A message deleted before the stream started — or one never shown — produces no event.

The decision is a pure, unit-tested helper `should_emit_deletion(seen, deleted_emitted, key)` (`seen && !already-emitted`). The `monitor_edit_detection_wiring_pinned` source-grep pin now also asserts both paths call `emit_deletions`.

**Consumer-facing JSON:** adds a new `type:"delete"` event. Additive; existing `type:"message"`/`"edit"` consumers unaffected.

CLI-only — no `common/`/contract/delegate/WASM change, **no migration**.

## Testing

- **Unit:** `should_emit_deletion_only_for_seen_and_unreported` (not-shown → no event; shown → emit once; already-reported → no repeat).
- **End-to-end** on a local node: a live stream showed
  ```
  message | 'delete me'
  delete  | (type:"delete", same author)
  ```
  when the shown message was deleted.

`cargo fmt`, `cargo clippy -p riverctl --all-targets` (clean), all 63 CLI lib tests pass.

[AI-assisted - Claude]

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -78,6 +78,18 @@ riverctl member ban          <room-owner-vk> <member-vk>   # Owner only.
 
 Run `riverctl <group> --help` or `riverctl <group> <cmd> --help` for full flags. All commands accept `--format json` for scripting.
 
+### `message stream --format json` event types
+
+`message stream` emits one JSON object per line (JSONL). Each carries a `type`:
+
+| `type` | Meaning | Fields |
+|--------|---------|--------|
+| `message` | A new message | `message_id`, `room`, `author`, `nickname`, `content`, `timestamp`, `edited`, `reply_to`, `reactions` |
+| `edit` | A previously-streamed message's content changed | same as `message` (with the new `content` and `edited: true`) |
+| `delete` | A previously-streamed message was deleted | `message_id`, `room`, `author`, `nickname`, `timestamp` (no `content`) |
+
+`reply_to` is `null` for non-replies, otherwise `{ "author", "preview" }`. `edit`/`delete` events are emitted only for messages the stream actually surfaced. Bridges should key off `type` and tolerate unknown future types.
+
 ## Configuration
 
 - `--node-url <URL>`: override the Freenet node URL (default `ws://127.0.0.1:7509/...`).
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -161,6 +161,20 @@ fn monitor_seen_key(msg: &river_core::room_state::message::AuthorizedMessageV1) 
     msg.id().0 .0.to_string()
 }
 
+/// Whether a monitor stream should emit a deletion event for a now-deleted
+/// message. True only if the message was previously surfaced to the stream
+/// (`seen`) and a deletion hasn't already been emitted for it
+/// (`deleted_emitted`). The caller has already confirmed the message is
+/// deleted. Keeping this pure makes the one-shot / only-if-shown semantics
+/// unit-testable. freenet/river#323.
+fn should_emit_deletion(
+    seen: &HashMap<String, String>,
+    deleted_emitted: &HashSet<String>,
+    key: &str,
+) -> bool {
+    seen.contains_key(key) && !deleted_emitted.contains(key)
+}
+
 /// Choose the `member_info` nickname to publish when re-adding a member who
 /// was pruned for inactivity (see [`ApiClient::build_rejoin_delta`]).
 ///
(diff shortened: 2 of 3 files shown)


sanity (commented):
## Comprehensive PR Review: #324

### Summary
- **PR Title:** fix(cli): surface message deletions as events in `message stream`
- **Type:** fix (riverctl monitor) — closes #323
- **Review tier:** Full — bridge-critical monitor path + new consumer-facing JSON event
- **Reviewers run:** code-first, testing, skeptical, big-picture + external Codex (initial); skeptical + Codex re-reviews on each fix HEAD

### Findings and resolutions

| 1 | **Codex** | **Should (real bug)** | Subscribe pre-seed marked ALL pre-existing messages `seen` but only displayed the last N — so a later deletion of a never-shown message (default `initial_messages=0` shows nothing) emitted a spurious `delete`. (The three Claude reviewers traced only the already-deleted case and missed this — caught by the external pass.) | Extracted pure `deletions_to_suppress_at_start`; pre-seed `deleted_emitted` with every pre-existing message NOT in the shown window. |
| 2 | **skeptical (re-review)** | **Medium (real bug, introduced by fix #1)** | An unshown pre-existing message EDITED live (→ surfaces, emits `edit`) then deleted would NOT emit a `delete` (key still suppressed) — violating the documented "only for surfaced messages" contract. | `emit_new_and_edited` now removes a key from `deleted_emitted` when it surfaces the message; later deletion is reportable again. One-shot still holds. |
| 3 | testing | Should | Pre-seed logic untested. | Unit-tested the extracted helper (shown→not suppressed, unshown→suppressed, none-shown→all); completed `should_emit_deletion` truth table. |
| 4 | skeptical/code-first | Low | aged-out-message gap; inline deleted check; polling no-pre-seed rationale. | Added comments; switched to `is_deleted()`. |
| 5 | big-picture | Low | Undocumented bridge JSON contract. | Documented `message`/`edit`/`delete` JSONL schema in `cli/README.md`. |
| 6 | big-picture | Follow-up | Reactions not surfaced as live events. | Noted for a future issue (out of scope). |

### Verified safe (final pass on 47d53d50)
Refactored emit loop preserves New/Edited semantics exactly; one-shot delete holds (deleted message excluded from `display_messages` → never re-surfaced → never un-suppressed); no spurious delete for never-surfaced messages; borrows sequential; #173 preserved; polling path unchanged (its `seen` only ever holds shown messages).

### Testing
64 CLI lib tests pass (helper + truth-table + wiring pin extended for `emit_deletions`). End-to-end on the local node across three scenarios: (a) shown message deleted → `delete` event; (b) **unshown pre-existing message deleted → no event** (fix #1); (c) **unshown message edited-then-deleted → edit + delete** (fix #2). Skeptical noted the remove-on-surface path is loop-bound (not unit-testable) but is covered by e2e (c).

### Verdict
**State:** Ready to Merge (pending green CI on current HEAD).
**HEAD SHA reviewed:** 47d53d50595b8065e085b7a5b9877b74820cccd5

Two real bugs were found across the review rounds (one by the external model, one by the adversarial re-review) and both fixed + verified; the final re-review pass (skeptical + Codex) on 47d53d50 is clean.

[AI-assisted - Claude]

PR #303, opened 2026-05-23
feat(cli): carry private-room secrets in riverctl invitations

sanity:
## Problem

Before this change, riverctl's `Invitation` struct lacked the `room_secrets` field added to the UI's `Invitation` in #301. CBOR `#[serde(default)]` on the UI side kept old/new invitations interoperable, but the asymmetry left both directions broken for private rooms:

- A **UI-created** private-room invitation accepted by riverctl silently dropped `room_secrets` — the CLI invitee was stuck on `[Encrypted: N bytes, vN]` until the room owner's chat-delegate back-filled an `encrypted_secrets` blob.
- A **CLI-created** private-room invitation carried no `room_secrets` at all, so the invitee never gained access.

The CLI also sealed the invitee nickname with `SealedBytes::public` unconditionally, leaking plaintext nicknames into private rooms whenever an accept did succeed.

Closes #302.

## Approach

1. **Wire compatibility.** Added `Invitation::room_secrets: Vec<(u32, [u8; 32])>` with `#[serde(default)]`, byte-identical to the UI's `ui::components::members::Invitation`. Hand-written `Debug` mirrors the UI's PR-#301 redaction so `{:?}`-logging an invitation does not print secret bytes.

2. **Inviter side (`create_invitation`).** New `cli::private_room` module collects every secret the CLI holds for a private room:
   - **Owner** → derives via `river_core::key_derivation::derive_room_secret` for every version present in `state.secrets.encrypted_secrets` (plus `current_version`), matching the deterministic convergence the UI rotation path depends on.
   - **Non-owner** → decrypts the member's own `encrypted_secrets` blobs via `decrypt_secret_from_member_blob_raw`, then folds in persisted `invitation_secrets` for versions the contract has not yet provided an owner-signed blob for. Owner-signed wins on collision (mirrors `RoomData::repopulate_secrets_from_state`).

3. **Invitee side (`accept_invitation`).** Replaced the unconditional `SealedBytes::public` with the equivalent of the UI's `seal_invitee_nickname`. When neither the contract blob nor the invitation carries a secret at `current_version`, member_info is **deferred** rather than leaked as plaintext into a private room. A `tracing::warn!` surfaces the deferral; full heal parity with the UI's `build_member_info_heal` is left as a follow-up.

[...]

PR #327, opened 2026-06-02
fix(cli): refuse re-accept of invitation to prevent identity clobber

### LINKED ISSUE #308: riverctl: re-accept of invitation clobbers non-invitation_secrets fields wholesale
## Background

Skeptical-review round-3 finding M1 on PR #303.

\`accept_invitation\` calls \`Storage::add_room_with_invitation_secrets\`, which builds a fresh \`StoredRoomInfo\` and \`insert\`s it into the map — overwriting whatever was there. PR #303 fixed the field that is most user-visibly broken by this (\`invitation_secrets\`, now merged), but the other fields are still wholesale-replaced on re-accept:

- \`signing_key_bytes\` — silently replaces the stored identity with \`invitation.invitee_signing_key\`. If a user re-accepts a different invitation for the same \`room\`, their identity flips on the next CLI command.
- \`state\` — replaced with the fresh GET. Mostly benign but a partial/stale fetch could downgrade.
- \`self_authorized_member: None\`, \`invite_chain: Vec::new()\` — cleared, then restored by the next \`store_authorized_member\` call. A panic between the two leaves a half-state on disk.
- \`previous_contract_key: None\` — cleared. Re-populated by \`load_rooms\` on the next load if needed, so likely safe by accident.

\`cli/src/commands/identity.rs::import_identity\` does check \`get_room().is_some() -> Err\` to refuse silent overwrite — \`accept_invitation\` does not.

## Scope

Decide and pick one:

**Option A (recommended)**: Refuse re-accept by default. \`accept_invitation\` checks \`get_room()\` first; if Some, returns an error directing the user to \`riverctl room remove <owner>\` first or pass a \`--force\` flag. Matches \`import_identity\`'s posture and avoids silent data loss for all fields.

**Option B**: Make all fields merge-on-re-accept (mirroring the #303 invitation_secrets fix). More complex; needs per-field semantics (which signing_key wins? which state is authoritative? etc.).

## References

- PR #303 — round-3 review identified the gap
- \`cli/src/api.rs::accept_invitation\`
- \`cli/src/storage.rs::add_room_with_invitation_secrets\`
- \`cli/src/commands/identity.rs::import_identity\` — model for Option A

[AI-assisted - Claude]

sanity:
## Problem

`accept_invitation` (`cli/src/api.rs`) called `Storage::add_room_with_invitation_secrets`, which builds a fresh `StoredRoomInfo` and `insert`s it into the rooms map, overwriting whatever was there. PR #303 fixed the most user-visible field (`invitation_secrets`, now merged), but every other field was still wholesale-replaced on re-accept:

- **`signing_key_bytes`** — silently replaced with `invitation.invitee_signing_key`. Re-accepting a *different* invitation for a room the user already belonged to flipped their stored identity, so every subsequent CLI command authenticated with the wrong key (auth failures that look like server-side problems).
- **`self_authorized_member` / `invite_chain`** — cleared, then restored by the next `store_authorized_member` call; a crash between the two leaves a half-state on disk.
- **`previous_contract_key`** — cleared.
- **`self_nickname`** — cleared.

`commands::identity::import_identity` already guards against exactly this with a `get_room().is_some() -> Err` check; `accept_invitation` did not.

## Approach

**Option A** from the issue (recommended there): refuse the re-accept by default. `accept_invitation` now checks `self.storage.get_room(&room_owner_vk)?.is_some()` *before* the network GET and, if the room already exists, returns an error directing the user to `riverctl room leave <owner>` first. This mirrors `import_identity`'s posture and avoids silent data loss for all fields with no per-field merge semantics to get wrong.

The refusal message is built by a small free function `reaccept_refusal_error` so it is unit-testable without a live Freenet node (the rest of `accept_invitation` needs one) — mirroring the existing `rejoin_preferred_nickname` extraction discipline.

No `--force` flag was added: `import_identity` has none either, and a `--force` would reintroduce the silent-overwrite footgun this fixes. The deliberate-replace path is `riverctl room leave`, consistent across both entry points.

## Testing

- `storage::tests::reaccept_guard_prevents_clobber` — the substantive regression. Populates a room (identity + authorized member + invite chain + nickname), asserts the guard's trigger (`get_room(...).is_some()`) is true, then exercises the exact storage path the old `accept_invitation` used and asserts it is a full replace (documents the destructive behavior the guard prevents).
- `api::reaccept_guard_tests::accept_invitation_has_reaccept_guard` — source-grep pin that the guard exists in the `accept_invitation` body **and runs before** the network GET. Verified it FAILS when the guard is removed and PASSES with it.
- `api::reaccept_guard_tests::refusal_error_names_room_and_points_to_leave` — pins the user-facing message names the room and points at `riverctl room leave`.

Full `cargo test -p riverctl` (67 lib + integration tests) green; `cargo fmt` / `cargo clippy -p riverctl --all-targets` clean. No WASM / contract / delegate changes (CLI logic only), so no migration entry needed.

Closes #308

[AI-assisted - Claude]

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -239,6 +239,32 @@ fn rejoin_preferred_nickname(
         .unwrap_or_else(|| SealedBytes::public("Member".to_string().into_bytes()))
 }
 
+/// Error returned when `accept_invitation` is asked to join a room the CLI
+/// already has stored credentials for (issue freenet/river#308).
+///
+/// Re-accepting an invitation used to rebuild the `StoredRoomInfo` from
+/// scratch and `insert` it, wholesale-clobbering the existing room's
+/// `signing_key_bytes`, `self_authorized_member`, `invite_chain`,
+/// `previous_contract_key`, and `self_nickname`. The most severe case is a
+/// silent identity flip: re-accepting a *different* invitation for the same
+/// room replaced the stored signing key, so every subsequent CLI command
+/// authenticated with the wrong key. We refuse the re-accept instead — the
+/// same posture `commands::identity::import_identity` already takes — and
+/// point the user at `riverctl room leave` to opt into a deliberate replace.
+///
+/// Kept as a free function so the user-facing message is unit-testable
+/// without a live Freenet node (the rest of `accept_invitation` requires
+/// one). Mirrors the `rejoin_preferred_nickname` extraction discipline above.
+fn reaccept_refusal_error(room_owner_vk: &VerifyingKey) -> anyhow::Error {
+    let owner_key_str = bs58::encode(room_owner_vk.as_bytes()).into_string();
+    anyhow!(
+        "You already have an identity for room {owner_key_str}. Accepting this \
+         invitation would overwrite your stored signing key, membership, and \
+         nickname for that room. Leave it first with `riverctl room leave \
+         {owner_key_str}` if you want to replace it."
+    )
+}
+
 /// On-wire invitation artifact. **MUST stay byte-identical to the UI's
 /// `ui::components::members::Invitation`** — both clients exchange these via
 /// base58+CBOR and the encoded string is fingerprinted for processed-invite
@@ -986,6 +1012,17 @@ impl ApiClient {
 
         let room_owner_vk = invitation.room;
         let contract_key = self.owner_vk_to_contract_key(&room_owner_vk);
+
+        // Refuse to re-accept an invitation for a room we already have stored
+        // credentials for (issue freenet/river#308). Re-accepting rebuilds the
+        // `StoredRoomInfo` from scratch, silently clobbering the existing
+        // `signing_key_bytes`, `self_authorized_member`, `invite_chain`,
+        // `previous_contract_key`, and `self_nickname`. Bail out *before* the
+        // network GET — same posture as `import_identity`. The user can opt
+        // into a deliberate replace via `riverctl room leave <owner>`.
+        if self.storage.get_room(&room_owner_vk)?.is_some() {
+            return Err(reaccept_refusal_error(&room_owner_vk));
+        }
 
         info!(
             "Invitation is for room owned by: {}",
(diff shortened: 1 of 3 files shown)


sanity:
## Review summary

Self-review across four lenses (code-first, testing-gaps, skeptical/adversarial, big-picture) plus external non-Claude model reviews (Codex CLI `codex review --base main`, re-run on each new HEAD; Gemini CLI for a second opinion). All findings addressed:

- **Codex P2 — recovery path was a dead end.** The guard pointed users at `riverctl room leave <owner>`, but that command was a `// TODO` stub that printed and returned without removing anything, so re-accept stayed blocked in a loop. Fixed: added `Storage::remove_room` and wired `RoomCommands::Leave` to it (commit `fbd5bfd`). This also makes `import_identity`'s pre-existing "remove it first with `riverctl room leave`" message functional.
- **Gemini Low — orphaned outbound-DM plaintext.** `remove_room` left the room's cached DM bodies / archived-thread entries in `outbound_dms.json` (plaintext storage leak). Fixed: `remove_room` now prunes both `entries` and `hidden_threads` for the room, best-effort (commit `2442cd4`).
- **Gemini Low — fragile source-grep pin** (`accept_invitation_has_reaccept_guard`): kept. It's the established codebase pattern for guards that can't be integration-tested without a live node (`accept_invitation_calls_seal_invitee_nickname`, `rejoin_nickname_wiring_pinned`, `repopulate_secrets_call_sites_pinned`). It searches within the function body and asserts the guard runs *before* the network GET; verified it FAILS when the guard is removed.
- **Gemini Low — no file locking on load→mutate→save in `remove_room`**: out of scope. This is the project-wide storage pattern across every `Storage` mutation; not introduced here.

Final Codex re-review on HEAD `2442cd4`: no regressions or correctness issues. `cargo fmt` / `cargo clippy -p riverctl --all-targets` clean; `cargo test -p riverctl` green (69 lib tests + integration).

[AI-assisted - Claude]

PR #328, opened 2026-06-02
fix(ui): preserve paragraph breaks in chat messages

### LINKED ISSUE #158: fix: paragraphs / newlines not preserved in messages
### Problem
When message text is shown on the browser page, all paragraphs created by newlines in the msg are wrapped in `<p>` elements.

the `<p>` elements lost their original margin, due to tailwind or whatever.
Without any margin for `<p>`  there are no empty lines and no paragraph breaks in the msg text.
It's just a wall of text.

**original user agent style:**
```
p {
    display: block;
    margin-block-start: 1em;
    margin-block-end: 1em;
    margin-inline-start: 0px;
    margin-inline-end: 0px;
    unicode-bidi: isolate;
}
```

**now:**
```
*, :after, :before, ::backdrop {
    box-sizing: border-box;
    border: 0 solid;
    margin: 0;
    padding: 0;
}
```

### Solution
Wrapping the original msg text in `<p>`  tags creates unnecessary complexity.
Just show the original text as it is. With some sanitization of course, that happened **before** storing it. 
But without any unnecessary formatting acrobatics.

`messageDiv.textContent = message`

```
.message {
  white-space: pre-wrap;
  word-break: break-word;
}

```

sanity:
## Problem

Multi-paragraph messages rendered as a single wall of text. The Markdown renderer (`message_to_html`) emits a `<p>` per blank-line-separated paragraph, but Tailwind v4's Preflight reset (`@import "tailwindcss"`) zeroes the margin on every element. So the paragraphs collapsed with no visible gap between them — the "wall of text" reported in #158.

Single newlines were already fine (`message_to_html` converts `\n` to `  \n`, which becomes a `<br>`). Only true blank-line paragraph breaks were broken.

## Approach

The message body is rendered inside a `.prose` container, and `ui/assets/tailwind.css` already re-adds spacing the global reset strips for lists, code blocks, and blockquotes — it was simply missing the equivalent rule for paragraphs. Add:

```css
.prose p { margin: 0 0 0.5rem; }
.prose p:last-child { margin-bottom: 0; }
```

The `0.5rem` matches the existing `.prose ul/ol` spacing (chat density), and trimming the last paragraph's bottom margin keeps the bubble from having a dangling gap.

This keeps the existing Markdown + linkify pipeline intact. The issue suggested rewriting the render to plain `textContent`, but that would have discarded URL linkification and the `target="_blank"` / `rel` handling, so it isn't a drop-in fix. A scoped CSS rule restores the paragraph spacing with no logic change.

The room-description prose (`[&>p]:m-0 [&>p]:inline`, line 1572) is unaffected: its inline arbitrary-variant overrides have higher specificity and use a direct-child selector, so it keeps its single-line truncated rendering.

`ui/assets/styles.css` (the compiled, minified output) is a gitignored build artifact regenerated by `cargo make build` → `build-tailwind` in CI, so only the source `tailwind.css` is committed.

## Testing

- `blank_line_produces_separate_paragraphs` — pins the HTML structure (two paragraphs → two `<p>` blocks).
- `prose_paragraph_spacing_css_present` — pins the `.prose p` rule in the source stylesheet so a future Tailwind bump or CSS refactor that silently drops it fails CI. **Verified this test FAILS with the rule removed and PASSES with it restored.**
- `cargo fmt --check`, `cargo clippy -p river-ui` (no new warnings), and `cargo test -p river-ui --bins` all clean.

Closes #158

[AI-assisted - Claude]

### DIFF
--- a/ui/assets/tailwind.css
+++ b/ui/assets/tailwind.css
@@ -91,6 +91,22 @@ html, body {
   color: var(--color-text);
 }
 
+/* Paragraph spacing in prose (Markdown messages).
+   Tailwind v4's Preflight reset zeroes margins on every element, so the
+   <p> blocks the Markdown renderer emits for blank-line-separated
+   paragraphs collapse into a single wall of text (freenet/river#158).
+   Re-add vertical spacing between paragraphs, but trim the outer edges
+   so a message bubble has no dangling gap above its first paragraph or
+   below its last. Single newlines are already preserved as <br> by
+   message_to_html, so this rule only affects true paragraph breaks. */
+.prose p {
+  margin: 0 0 0.5rem;
+}
+
+.prose p:last-child {
+  margin-bottom: 0;
+}
+
 /* Link styling for prose content (Markdown messages) */
 .prose a {
   color: var(--color-accent);
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -2483,6 +2483,41 @@ mod tests {
         );
     }
 
+    /// Issue #158: a blank line between two blocks of text is a paragraph
+    /// break, and the Markdown renderer must emit a separate `<p>` for each
+    /// so the stylesheet can space them apart. This is the HTML-structure
+    /// half of the fix; the CSS half is pinned by
+    /// `prose_paragraph_spacing_css_present` below.
+    #[test]
+    fn blank_line_produces_separate_paragraphs() {
+        let html = message_to_html("First paragraph.\n\nSecond paragraph.");
+        let paragraphs = html.matches("<p>").count();
+        assert_eq!(
+            paragraphs, 2,
+            "blank-line-separated text should render as two <p> blocks: {html}"
+        );
+    }
+
+    /// Issue #158 root cause: Tailwind v4's Preflight reset zeroes the
+    /// margin on every element, so the `<p>` blocks above collapse into a
+    /// single wall of text unless the stylesheet re-adds paragraph spacing.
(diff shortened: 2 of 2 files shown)


sanity:
## Review summary (Light tier — UI-only CSS + tests, no high-risk surface)

**Self-review across four lenses:**
- **Code-first:** `.prose p` margin rule restores the spacing the Tailwind v4 Preflight reset strips, mirroring the existing `.prose ul/ol/blockquote/code` rules. `:last-child` trims the trailing gap.
- **Scoping (key concern):** `.prose` appears in 3 places. The message body (`conversation.rs:2185`) is the intended target. The room-description prose (`:1572`) carries `[&>p]:m-0 [&>p]:inline`, which compiles to a higher-specificity direct-child rule and therefore **overrides** `.prose p` — its inline-truncated rendering is unaffected. Verified.
- **Testing:** the CSS guard test was confirmed to FAIL with the rule removed and PASS with it restored, so it genuinely reproduces the regression. The common single-paragraph case is covered by `:last-child` (no dangling margin).
- **Big-picture:** no WASM / contract / delegate / wire-format touched; keeps the existing markdown+linkify pipeline (the issue's suggested `textContent` rewrite would have dropped URL linkification).

**External non-Claude review (codex `review --base main`):** "The change adds scoped paragraph spacing for rendered Markdown prose and pins the behavior with focused tests. I did not identify any introduced correctness, security, or maintainability issues that warrant an actionable finding." No findings.

[AI-assisted - Claude]

PR #329, opened 2026-06-02
fix(ui): heal "Unknown" private-room member when secret arrives via UPDATE

### LINKED ISSUE #295: Private-room invitee can stay "Unknown" for a session if their room secret arrives via UPDATE
## Problem

Follow-up from PR #294 review (Codex finding P2).

PR #294's member-info self-heal (`RoomData::build_member_info_heal`) runs on **GET** of an existing room (`handle_get_response`). For a **private** room, when an invitee accepts before the owner's encrypted-secret back-fill has reached them, the invitation-accept PUT correctly omits `member_info` (it cannot seal the nickname without the secret — see PR #294, finding M1) and relies on the self-heal to publish it later.

The room secret normally arrives through a subscription **UPDATE**, not a GET. The update-notification path repopulates secrets but never runs the self-heal. So the invitee can remain visible as "Unknown" to every other peer until the next full GET happens — i.e. until they reload the app.

## Impact

- Private rooms only. The public-room path (and the official Freenet room) is unaffected.
- Not a leak / crash / data loss — it self-resolves on the next GET (app startup or suspension-recovery refresh).
- Degraded UX: a new private-room member shows as "Unknown" to others for the rest of their session.

## Proposed fix

Run the self-heal when the secret becomes available, not only on GET. Options:
- Invoke `build_member_info_heal` from the update-notification path (`handle_update_notification`) after `repopulate_secrets_from_state` decrypts a new secret version.
- Or relocate the heal trigger into `process_rooms` / the sync loop so it covers every state-ingestion path uniformly.

The second is cleaner (one trigger site) but is an architectural change deserving its own PR + review.

[AI-assisted - Claude]

sanity:
## Problem

The member-info self-heal (`RoomData::build_member_info_heal`, PR #272/#294)
ran **only** on the GET response path (`get_response.rs`). For a **private**
room, a freshly-invited member's invitation-accept PUT correctly omits
`member_info` when the room secret is not yet available to seal the nickname
(it cannot leak a plaintext nickname — see #294 finding M1) and relies on the
self-heal to publish it later.

The room secret normally arrives via a subscription **UPDATE**, not a GET. The
UPDATE path (`handle_update_notification` → `room_synchronizer.apply_delta` /
`update_room_state`) repopulates secrets but never invoked the heal. So the
invitee remained visible as "Unknown" to every other peer for the rest of their
session — until the next full GET (app reload).

Follow-up from PR #294 review (Codex finding P2).

## Approach

Wire the heal into **both** synchronizer UPDATE paths (`apply_delta_inner` and
`update_room_state_inner`). After `repopulate_secrets_from_state` decrypts a
newly-arrived secret (`new_secrets > 0`), build the heal against the post-merge
state inside the `ROOMS.with_mut` borrow; if the member is still stranded, send
a self-signed, `member_info`-only UPDATE *after* the borrow releases — the same
idempotent delta shape the GET path already uses (factored into a shared
`send_member_info_heal_update` helper).

This is option 1 from the issue (invoke from the update path after the secret
arrives). The cleaner "single trigger site" refactor (option 2) is deliberately
left out of scope — it is an architectural change deserving its own PR.

Why it does not loop: the heal's own echoed-back UPDATE carries `member_info`
(so `build_member_info_heal` returns `None`) and no new secret (so
`new_secrets == 0` and the heal is never even attempted). UI-only change — no
delegate/contract WASM touched, so no migration required.

## Testing

- `member_info_heal_fires_when_secret_arrives_via_update` (new): reproduces the
  exact timeline — stranded private-room invitee, secret absent (heal defers),
  then the owner's blob arrives via UPDATE, `repopulate_secrets_from_state`
  decrypts it, and `build_member_info_heal` now returns a self-signed,
  Private-sealed entry carrying the invitee's chosen nickname.
- `member_info_heal_update_path_wiring_pinned` (new): source-grep pin that the
  heal is wired into both UPDATE call sites and the sender helper — verified to
  fail when a call site is removed. Mirrors the existing
  `repopulate_secrets_call_sites_pinned` pattern.
- Full `cargo test -p river-ui --bins` green (263 passed).

Closes #295

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/private-rooms.md
+++ b/.claude/rules/private-rooms.md
@@ -98,14 +98,29 @@ local-state copy — the same build-once-reuse discipline the synthesised
 join_event follows.
 
 The remediation for already-stranded members is
-`RoomData::build_member_info_heal`: on every GET of an existing room it
-detects "self in `members`, absent from `member_info`" and re-publishes a
-self-signed `member_info` (folded into the PUT for imported rooms, sent as
-a standalone UPDATE for already-subscribed rooms). A non-owner's
-`member_info` is only valid when self-signed by that member's own key, so
-this heal can ONLY run client-side, by the affected member — never
-owner-side. For a private room the heal defers (publishes nothing) until
-the room secret is available, so it never leaks a plaintext nickname.
+`RoomData::build_member_info_heal`: it detects "self in `members`, absent
+from `member_info`" and re-publishes a self-signed `member_info` (folded
+into the PUT for imported rooms, sent as a standalone UPDATE for
+already-subscribed rooms). A non-owner's `member_info` is only valid when
+self-signed by that member's own key, so this heal can ONLY run
+client-side, by the affected member — never owner-side. For a private room
+the heal defers (publishes nothing) until the room secret is available, so
+it never leaks a plaintext nickname.
+
+The heal fires from TWO trigger sites, because the secret a private-room
+invitee needs to seal their nickname can arrive on either path:
+- **GET path** (`get_response.rs`): every GET of an existing room.
+- **UPDATE path** (`room_synchronizer.rs`, issue #295): after
+  `repopulate_secrets_from_state` decrypts a newly-arrived secret
+  (`new_secrets > 0`) in either `apply_delta_inner` (delta) or
+  `update_room_state_inner` (full state). The room secret normally arrives
+  via a subscription UPDATE, not a GET, so without this trigger a new
+  private-room invitee stayed "Unknown" to every other peer for the rest of
+  their session (until the next full GET, i.e. an app reload). Both UPDATE
+  call sites are pinned by `member_info_heal_update_path_wiring_pinned`; the
+  fire-on-secret-arrival behaviour by
+  `member_info_heal_fires_when_secret_arrives_via_update` (both in
+  `ui/src/room_data.rs`).
 
 ## In-memory secret repopulation (issue #251)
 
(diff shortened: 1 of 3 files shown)


sanity:
## Multi-model review summary (Light tier — UI-only, no high-risk surface)

Risk tier: **Light** — change is confined to `ui/src/` (no consensus / crypto / wire-format / delegate-contract WASM / concurrency-primitive surface). One external pass + adversarial self-read, per the review rule.

**Codex (`codex review --base main`):** no findings — "adds the UPDATE-path member_info self-heal wiring without introducing an evident correctness issue. The helper sends an idempotent member_info-only delta after the room-state borrow is released, and the tests cover the intended trigger behavior and call-site wiring."

**Gemini:** no blocking findings. Two observations, both dismissed:
1. *`room_owner_copy` unused (room_synchronizer.rs:1101)* — false positive: it is used at lines 1389 & 1398 (`notify_new_messages` + current-room check), and is pre-existing code untouched by this PR.
2. *`WEB_API` write lock held across `.await` in the heal sender* — this is byte-for-byte the established codebase idiom for sending requests (identical to the GET-path heal send at `get_response.rs:863`, and every other send site in the synchronizer). Changing it is an out-of-scope, codebase-wide refactor; not a regression introduced here.

**Adversarial self-review (loop / re-entrancy / borrow):**
- No UPDATE loop: the heal's own echoed-back UPDATE carries `member_info` (so `build_member_info_heal` → `None`) and no new secret (so `new_secrets == 0`, heal never attempted).
- No `RefCell` panic: the heal is built inside `ROOMS.with_mut` and sent only after the borrow releases (deferred via the existing `pending_*` pattern).
- Idempotent: once the entry lands, the heal returns `None`.

[AI-assisted - Claude]

PR #331, opened 2026-06-02
fix: carry plaintext nickname in IdentityExport so a join-window export keeps it (#298)

### LINKED ISSUE #298: Identity export taken before a private-room join-heal loses the chosen nickname on re-import
## Background

PR #297 fixes the auto-generated nickname overriding a user's chosen nickname when joining a private room. It persists the chosen nickname in `RoomData.self_nickname` so the member-info self-heal and rejoin paths can restore it.

## Residual gap (out of scope for #297)

`IdentityExport` (`river-core`, `common/src/room_state/identity.rs`) carries the sealed `member_info: Option<AuthorizedMemberInfo>` but **not** the plaintext nickname.

So in this narrow window:

1. User accepts a private-room invite; the room secret has not yet arrived, so `member_info` sealing is deferred and `self_member_info` is `None`.
2. Before the GET-path self-heal runs, the user exports their identity.
3. The export's `member_info` is `None`.
4. On re-import (`ui/src/components/members.rs`), `self_nickname` is set to `None` and `self_member_info` to `None`.
5. The heal on the re-imported room mints a generated default handle — the user's chosen nickname is lost.

This is a pre-existing limitation, narrow (it requires exporting in the sub-second-to-minutes window between join and heal), and documented in an inline comment at the import site.

## Fix options

- Add a plaintext `self_nickname: Option<String>` field to `IdentityExport`. This is an `IdentityExport` wire/serialization-format change (it is armored-string serialized), so it needs `#[serde(default)]` for backward compatibility and a deliberate decision about carrying a plaintext nickname in an exported-identity blob.
- Or: accept the gap (the heal still produces a usable, non-ugly handle; only the *chosen* nickname is lost, and only in this window).

Low priority.

[AI-assisted - Claude]

sanity:
## Problem

`IdentityExport` carries the sealed `member_info` but not the plaintext nickname. In the narrow window between accepting a private-room invite and the member-info self-heal sealing the nickname, `member_info` is `None`. If the user exports their identity in that window:

1. The export's `member_info` is `None`.
2. On re-import, both `self_member_info` and `self_nickname` were set to `None`.
3. `build_member_info_heal` then mints a generated default handle — the user's chosen nickname is lost.

PR #329 made the heal fire reliably on the UPDATE path (not just GET), but that did not fix #298: the heal still had no nickname to restore because `IdentityExport` never carried one. The heal's nickname source order is `self_member_info` → `self_nickname` → generated default; with both `None`, it falls through to the default.

## Approach

Add a plaintext `self_nickname: Option<String>` field to `IdentityExport` with `#[serde(default)]` (the export is an armored CBOR blob; the attribute keeps old tokens decoding cleanly as `None`).

- **Export** (UI `members.rs`, CLI `identity.rs`): populate `self_nickname` from the existing `self_nickname` already tracked in `RoomData` / `StoredRoomInfo`.
- **Import** (UI `members.rs`): set `self_nickname: export.self_nickname` instead of hardcoded `None`, so the heal restores the chosen nickname.
- **Import** (CLI `identity.rs`): when `member_info` has no usable public nickname (absent or sealed), fall back to persisting the carried plaintext `self_nickname`, and use it for the import summary line.

Plaintext here adds no new exposure class: the export already carries the private signing key, a far more sensitive secret. The private-room heal still defers (publishes nothing) until the room secret arrives, so no plaintext nickname is ever leaked to the network.

This is a client-side armored blob exchanged between a user's own devices; `IdentityExport` is never referenced by the room-contract or chat-delegate WASM, so no migration entry is required (the committed WASMs are untouched; the CI migration check keys on committed-WASM bytes).

## Testing

- New `test_self_nickname_survives_roundtrip_without_member_info` in `common/src/room_state/identity.rs` reproduces #298: an export with `member_info: None` but `self_nickname: Some(...)` round-trips and preserves the nickname. Verified it FAILS before the fix (temporarily marking the field `#[serde(skip)]` yields `left: None, right: Some("ChosenName")`) and PASSES after.
- Strengthened `test_backward_compat_no_room_name` to assert an old token (no `self_nickname` field) decodes with `self_nickname: None`.
- Strengthened `test_roundtrip_with_invite_chain_and_member_info` to assert `self_nickname` round-trips.
- `cargo test -p river-core` (174 lib tests), `cargo test -p riverctl --lib` (69 tests, incl. the `update_self_nickname` source-grep pin), `cargo test -p river-ui --bins` (265 tests) all green.
- `cargo fmt`, `cargo clippy -p river-core -p riverctl`, and `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` all clean.

Closes #298

[AI-assisted - Claude]

### DIFF
--- a/cli/src/commands/identity.rs
+++ b/cli/src/commands/identity.rs
@@ -163,6 +163,10 @@ async fn export_identity(
         invite_chain,
         member_info,
         room_name,
+        // Carry the chosen nickname in plaintext so an export taken before
+        // the private-room join-heal sealed `member_info` doesn't lose it on
+        // re-import (freenet/river#298).
+        self_nickname: room_info.self_nickname.clone(),
     };
 
     let armored = export.to_armored_string();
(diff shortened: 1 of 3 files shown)


sanity:
## Review summary

- *Code-first*: the fix wires the existing `self_nickname` (tracked in `RoomData` / `StoredRoomInfo`) through the export/import boundary; the heal's nickname source order (`self_member_info` → `self_nickname` → generated default) does the rest. Straight-line assignment at the four sites.
- *Testing*: the wire-format round-trip is the testable core and is covered by a new regression test in `common/` that was verified to fail before the fix. UI/CLI wiring is plain assignment, pinned indirectly by existing source-grep tests (`update_self_nickname`, `member_info_heal_*`).
- *Skeptical*: no new plaintext-exposure class (the export already carries the private signing key); the private-room heal still defers until the secret arrives, so nothing leaks to the network.
- *Big-picture*: extends the same nickname-preservation pattern as #297/#321 across export/import; the stale inline comment at the import site is updated.

**External review (Codex `codex review --base main`):** one P1 finding — "change touches `common/`, may move room-contract/delegate WASM keys; add a migration entry."

*Dismissed with justification:* The CI migration check (`scripts/check-room-contract-migration.sh` / `check-delegate-migration`) keys on the **committed** WASM file bytes (base SHA vs head SHA), not a source rebuild. This PR changes no committed WASM (`ui/public/contracts/`, `cli/contracts/` are untouched vs HEAD), so the checks see the contract/delegate keys as unchanged and require no migration entry. `IdentityExport` is also never referenced by the room-contract or chat-delegate code. The migration concern applies to a future *publish* action (`cargo make publish-all`, which rebuilds + republishes the WASM), not to this code-only PR. I confirmed `git diff HEAD -- ui/public/contracts/ cli/contracts/` is empty.

[AI-assisted - Claude]

PR #332, opened 2026-06-02
fix(build): retarget test-publish webapp via base_path, not length-unsafe WASM sed

### LINKED ISSUE #257: test-publish: webapp sed-replace ID-length mismatch corrupts WASM

sanity:
## Problem

`cargo make compress-webapp-test` (the `publish-river-test` path) retargeted a built webapp from the production contract ID (`raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`, 43 chars) to a freshly-generated test contract ID with a byte-for-byte `sed` over HTML, JS, **and the WASM binary** — assuming all IDs are the same byte length.

They aren't. Base58 of a 32-byte key is 43, 44, or occasionally 45 chars depending on leading zero bytes. Production is 43; a fresh test key has a ~1/256 chance per generation of also being 43. When the test ID differs in length, the `sed` over the `.wasm` shifts every byte after the first match while the module's own length-prefix / section-length fields keep their old values, corrupting it. The browser then fails with `WebAssembly.instantiateStreaming(): illegal flag value …`. Hit twice in the #244 cycle; the workaround was looping `generate-test-keys` (~256 attempts mean). P3 — only the test-publish path is affected; production publishes use committed parameters and never run this task.

## Approach

The contract ID is baked into the UI WASM **at compile time**, in two places, and a text substitution can't preserve the WASM's length fields:
1. **`DIOXUS_ASSET_ROOT`** (the `base_path` from `Dioxus.toml`) — release web builds resolve `asset!()` URLs (stylesheets, logo) from this compile-time value, **not** from `window.location` (confirmed in `dioxus-cli-config::web_base_path`, which uses `option_env!("DIOXUS_ASSET_ROOT")` in release). So it MUST be the test ID for a test deploy, or assets load from the production contract.
2. **`FALLBACK_BASE_URL`** in `invite_member_modal.rs` — a hardcoded production contract URL.

Fix (the issue's recommended Option #1, plus the proven `scripts/local-republish.sh` mechanism):
- **Build the test webapp with the test contract ID as `base_path` from the start**, so the WASM's `DIOXUS_ASSET_ROOT` and all HTML/JS loaders bake the test ID. No post-build `sed` of any file. `compress-webapp-test` no longer depends on the prod-base_path `build-ui-with-cleanup`; it runs the UI build itself with the test base_path and restores `Dioxus.toml` afterward (via an EXIT trap).
- **Drop the hardcoded production contract ID from `FALLBACK_BASE_URL`** (now a `CONTRACT_ID` placeholder). The fallback is only reached off-browser, where the contract URL is meaningless; this just removes a baked-in prod ID and doesn't change browser behavior (invite URLs come from `window.location`).
- **Add a `web-container-tool export-parameters` subcommand** to write the contract parameters (verifying-key bytes) from the test keys without a webapp archive, breaking the chicken-and-egg where the test ID is needed before the build but `sign` needs the built archive.
- **Force a rebuild** (`touch ui/src/main.rs`) so a new test key's base_path actually takes effect, and **assert the built output contains this run's test ID and no production ID anywhere** (HTML/JS/WASM) before packaging — so a base_path that fails to take effect, or a stale dx artifact carrying a previous test ID, fails the build instead of silently shipping a mis-targeted webapp.

## Testing

- Unit test (`web-container-tool`): pins that `export-parameters` and `sign` produce byte-identical parameters (both are the verifying key) — if they diverged, the baked base_path would target a different contract than the one published.
- **End-to-end, length-mismatch case:** ran `cargo make compress-webapp-test` with a freshly-generated test key whose ID is 44/45 chars (length-mismatched vs the 43-char prod ID — the exact case that corrupted the WASM before). Verified the packaged `webapp-test.tar.xz` references the test ID in HTML/JS, contains no production ID anywhere, and the UI WASM passes `wasm-validate`.
- **End-to-end, stale-build case:** built once, regenerated test keys (new ID), rebuilt — verified the new build references the NEW test ID with zero stale-ID and zero prod-ID, WASM valid.
- `cargo test -p river-ui --bins` (265 tests) and `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` pass. `cargo fmt`/`clippy` clean.

No live publish performed. No contract/delegate WASM changes; the only WASM affected is the (non-committed, build-generated) UI app WASM, so no migration entry is needed.

Closes #257

[AI-assisted - Claude]

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -390,37 +390,102 @@ echo "Run 'cargo make sign-webapp-test' to sign with test keys"
 '''
 
 [tasks.compress-webapp-test]
-description = "Retarget and compress webapp for test contract"
-dependencies = ["build-ui-with-cleanup", "build-web-container"]
+description = "Build (retargeted to the test contract) and compress the webapp"
+# NOTE: deliberately does NOT depend on build-ui-with-cleanup. That task bakes
+# the PRODUCTION base_path into the build. For the test contract we must bake
+# the TEST contract ID as the base_path BEFORE building, so we run the build
+# ourselves here. build-chat-delegate + build-tailwind are the same UI build
+# prerequisites build-ui depends on.
+dependencies = [
+    "build-chat-delegate",
+    "build-tailwind",
+    "build-web-container",
+    "build-web-container-tool",
+    "clean-stale-assets",
+]
 script = '''
+set -e
 PROD_ID="raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv"
+TOOL="target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool"
+DIOXUS_TOML="ui/Dioxus.toml"
 
-# Compute test contract ID from test parameters
-if [ -f "target/webapp/webapp-test.parameters" ]; then
-    TEST_ID=$(fdev get-contract-id \
-        --code target/wasm32-unknown-unknown/release/web_container_contract.wasm \
-        --parameters target/webapp/webapp-test.parameters 2>/dev/null)
-elif [ -f "test-contract/contract-id.txt" ]; then
-    TEST_ID=$(cat test-contract/contract-id.txt)
-else
+if [ ! -f "test-contract/test-keys.toml" ]; then
+    echo "No test keys found. Run 'cargo make generate-test-keys' first"
+    exit 1
+fi
+
+# The test contract ID is derive(web_container_contract.wasm, parameters),
+# and parameters are just the test signing identity's verifying-key bytes.
+# We need the ID BEFORE building the UI so the correct base_path is baked into
+# the WASM as DIOXUS_ASSET_ROOT (release web builds resolve asset!() URLs from
+# the compile-time-baked base_path, NOT from window.location). Generate the
+# parameters from the test keys directly so we don't depend on a prior sign.
+mkdir -p target/webapp
+"$TOOL" export-parameters \
+    --parameters target/webapp/webapp-test.parameters \
+    --key-file test-contract/test-keys.toml
+
+TEST_ID=$(fdev get-contract-id \
+    --code target/wasm32-unknown-unknown/release/web_container_contract.wasm \
+    --parameters target/webapp/webapp-test.parameters)
+if [ -z "$TEST_ID" ]; then
     echo "ERROR: Cannot determine test contract ID"
     exit 1
 fi
 
-echo "Retargeting webapp: $PROD_ID → $TEST_ID"
+echo "Building webapp retargeted to test contract: $PROD_ID -> $TEST_ID"
 
-# Copy build output to temp dir
-TEMP_DIR=$(mktemp -d)
-cp -r target/dx/river-ui/${BUILD_PROFILE}/web/public/* "$TEMP_DIR/"
+# Set base_path to the TEST contract ID for this build, then restore the
+# committed (commented, production) value afterwards. This is the same
+# approach scripts/local-republish.sh uses and is the ONLY length-safe way to
+# retarget the contract ID — rewriting the built WASM with a different-length
+# base58 ID corrupts the module (freenet/river#257).
+restore_dioxus() {
+    perl -pi -e "s|^\\s*#?\\s*base_path\\s*=\\s*\".*\"|#base_path = \"v1/contract/web/${PROD_ID}\"|" "$DIOXUS_TOML"
+}
+trap restore_dioxus EXIT
 
-# Replace contract ID in all files (HTML, JS, WASM - all same byte length)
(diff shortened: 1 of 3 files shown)


sanity:
## Review summary

Risk tier: **Light** (build-tooling + a one-constant UI change; no high-risk surface — no contract/delegate WASM, wire format, concurrency, or auth). One external non-Claude pass (Codex) + a self four-lens read.

### External model (Codex CLI, `codex review --base main`)

Run iteratively as the approach evolved. Two substantive findings, both fixed and re-verified:

1. **[P2] WASM-baked `DIOXUS_ASSET_ROOT` left pointing at production.** The first iteration (text-only retarget, leave WASM untouched) avoided corrupting the WASM but left the release WASM's compile-time `DIOXUS_ASSET_ROOT` at the production base path, so a test deploy would fetch `asset!()` URLs (stylesheets, logo) from the production contract. **Fixed** by abandoning post-build retargeting entirely and building with the test ID as `base_path` from the start, so the WASM bakes the test ID. Verified the built WASM/HTML/JS contain only the test ID.
2. **[P2] Stale dx artifacts could ship a previous test ID.** A `base_path` edit isn't a source change dx tracks, so a rerun after regenerating test keys could reuse artifacts carrying the prior test ID; the negative prod-ID check wouldn't catch it. **Fixed** by forcing a rebuild (`touch ui/src/main.rs`) and adding a positive assertion that the build references *this run's* test ID. Verified end-to-end with a regenerate-keys-then-rebuild scenario (new ID present, zero stale ID).

Final Codex pass on the merged approach: clean — "I did not identify any discrete regression introduced by this patch."

### Self review (four lenses)

- **Code-first:** `export-parameters` and `sign` now share `write_parameters` (single source of truth for the params bytes); `Dioxus.toml` is restored via an EXIT trap even if `dx build` fails.
- **Testing:** unit test pins `export-parameters` == `sign` param bytes; the corruption + stale-build scenarios are validated by real `cargo make compress-webapp-test` runs with `wasm-validate`.
- **Skeptical:** `FALLBACK_BASE_URL` is only read off-browser; nothing parses its contract-ID segment, so the `CONTRACT_ID` placeholder is safe. No committed WASM touched.
- **Big-picture:** implements the issue's recommended Option #1 (remove the hardcoded ID + runtime/base_path derivation); no migration needed.

[AI-assisted - Claude]

PR #333, opened 2026-06-02
feat(ui): auto-resume subscription on reload mid-invitation-flow

### LINKED ISSUE #218: feat(ui): auto-resume subscription on reload mid-invitation-flow
Follow-up from #216.

After #216 lands, the UX for a mid-subscription reload is:

1. User opens \`?invitation=<code>\`, picks a nickname, clicks Accept.
2. Subscription begins (\`PENDING_INVITES\` populated, \`AcceptInvitation\` message sent to synchronizer).
3. User reloads the page before the room arrives.
4. The fingerprint guard correctly suppresses the modal — but \`PENDING_INVITES\` is reset to empty (it's in-memory only), so the user sees no \"Subscribing...\" indicator either. The synchronizer will eventually surface the room via the chat delegate when sync completes, but in the meantime there is no UI feedback.

Better UX: persist the chosen nickname alongside the pending invitation in localStorage. On reload, if both are present, automatically re-call \`accept_invitation\` with the saved nickname so the synchronizer resumes and the user sees the \"Subscribing...\" state.

This also gives a recovery path for the case where the initial \`AcceptInvitation\` send failed (e.g. synchronizer not yet started at click time), which is currently impossible without re-opening the URL.

[AI-assisted - Claude]

sanity:
## Problem

Follow-up from #216. After #216, a reload *mid-subscription* (user opened `?invitation=`, picked a nickname, clicked Accept, then reloaded before the room arrived) leaves the user with no feedback:

- The fingerprint guard correctly suppresses the nickname-prompt modal, but
- `PENDING_INVITES` is in-memory only, so it is empty on reload — no "Subscribing…" indicator either.
- If the original `AcceptInvitation` send never reached a running synchronizer, there was previously no recovery path short of re-opening the original URL.

## Approach

Persist the chosen nickname alongside the pending invitation, and auto-resume on reload.

- On Accept, save the nickname under a **sibling** localStorage key `river_pending_invitation_nickname` (no change to the existing `river_pending_invitation` format — the stored value is a raw encoded invitation, not a struct, so a sibling key avoids any stored-data migration).
- On the localStorage-recovery path in `App`, when both the invitation artifact and a saved nickname are present, **auto-resume** by re-calling `accept_invitation` with that nickname. That re-populates the in-memory `PENDING_INVITES` (status `PendingSubscription`) and re-sends `AcceptInvitation` to the synchronizer, so the modal re-renders the "Subscribing…" / "Preparing to subscribe…" progress state. This also recovers the failed-initial-send case (the unbounded synchronizer channel queues the message even if the loop is not yet up).
- The saved-nickname signal **takes precedence over** the processed-fingerprint guard from #216, because `accept_invitation` marks the invitation processed *up front* — so a mid-flight reload always sees it as processed. The invitation artifact only stays in storage until the room subscribes or the user dismisses (both now clear the nickname key too), so "nickname present" is the authoritative "join not yet finished, resume it" signal. No resume loop: completion (`render_subscribed_state`) and dismiss (`dismiss_invitation_persistently`) both clear storage.
- A reload **before** clicking Accept still shows the nickname prompt (no nickname saved), matching pre-PR behaviour and #216's "mark only on definitive action" rule.

The three-way recovery decision (Resume / Discard / Prompt) is extracted into a pure `decide_recovered_invitation` helper so it is host-testable — the localStorage reads that feed it are wasm-only.

UI-only change: no `common/`, `delegates/`, `contracts/`, or Cargo-manifest changes, so **no delegate/contract migration is required**.

## Testing

Four new host unit tests in `receive_invitation_modal.rs` pin the decision matrix:

- `recovered_invitation_with_nickname_resumes` — nickname present + already-processed ⇒ **Resume** (the #218 regression: with the naive processed-first ordering this would Discard, dropping the user). Verified this test FAILS against the buggy ordering and PASSES with the fix.
- `recovered_invitation_with_nickname_resumes_even_when_not_processed` — defensive precedence check.
- `recovered_invitation_processed_without_nickname_discards` — accepted-then-left / dismissed ⇒ Discard.
- `recovered_invitation_not_processed_without_nickname_prompts` — reloaded before deciding ⇒ Prompt.

Full `cargo test -p river-ui --bins` green (269 tests). `cargo fmt --check` clean; `cargo clippy` and `cargo check` on `wasm32-unknown-unknown` clean (no new warnings).

### Manual test plan (local Freenet node)

1. Owner publishes an invite link; joiner opens it, enters a nickname, clicks Accept.
2. Reload before the room arrives. **Expected:** modal shows "Subscribing…/Preparing to subscribe…" (auto-resumed), not a nickname re-prompt or blank UI.
3. After the room loads, reload again. **Expected:** no modal (storage cleared on subscribe).
4. Open a fresh invite link, reload **without** clicking Accept. **Expected:** nickname prompt re-appears.

Closes #218

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -18,8 +18,8 @@ use crate::components::room_list::edit_room_modal::EditRoomModal;
 use crate::components::room_list::receive_invitation_modal::{
     accept_invitation, clear_invitation_from_storage, decide_recovered_invitation,
     is_invitation_processed, load_invitation_from_storage, load_invitation_nickname_from_storage,
-    save_invitation_to_storage, ReceiveInvitationModal, RecoveredInvitationAction,
-    PRESENT_INVITATION_REQUEST,
+    save_invitation_to_storage, take_resume_once, ReceiveInvitationModal,
+    RecoveredInvitationAction, PRESENT_INVITATION_REQUEST,
 };
 use crate::invites::PendingInvites;
 use crate::room_data::{CurrentRoom, Rooms};
@@ -104,6 +104,19 @@ pub fn App() -> Element {
     crate::components::invite_click_interceptor::install_invite_click_interceptor();
 
     let mut receive_invitation = use_signal(|| None::<Invitation>);
+
+    // One-shot guard for the localStorage auto-resume path (#218). The
+    // recovery block below runs in the `App` component BODY, which re-executes
+    // on every render. The `Resume` branch has side effects (re-sends
+    // `AcceptInvitation` and resets `PENDING_INVITES` to `PendingSubscription`)
+    // and the persisted nickname stays in localStorage until the join
+    // completes — so without this guard, every re-render while the
+    // subscription is in flight would re-fire `accept_invitation`, resetting
+    // the status and looping. We resume at most once per page load; that
+    // single resume re-populates `PENDING_INVITES` and the synchronizer drives
+    // the rest. Uses the same `use_hook(Rc<Cell>)` one-shot idiom as
+    // `dm_thread_modal.rs`.
+    let invitation_resume_fired = use_hook(|| std::rc::Rc::new(std::cell::Cell::new(false)));
 
     // Bridge the click interceptor's `INTERCEPTED_INVITATION_CODE` global
     // into the local `receive_invitation` signal that drives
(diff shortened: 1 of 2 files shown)


sanity:
## Review summary

**External model review:** `codex review --base main` was run on each revision of the diff. It found two real bugs, both now fixed, and the final pass is clean:

- **[P1] Per-render auto-resume loop** — the localStorage recovery runs in the `App` component body (re-executes every render) and the saved nickname persists until the join completes, so the `Resume` branch would re-fire `accept_invitation` on every re-render (re-sending `AcceptInvitation`, resetting `PENDING_INVITES`), looping. Fixed with a one-shot `use_hook(Rc<Cell>)` guard (`take_resume_once`); pinned by `resume_fires_at_most_once_across_many_renders`.
- **[P2] Stale nickname applied to a different invitation** — the nickname was a single global key not bound to any invitation, so accepting A then opening (not accepting) B before A's subscription cleared storage could auto-accept B with A's nickname. Fixed by binding the nickname to the invitation's fingerprint (`river_pending_invitation_nickname_fp`); `load_invitation_nickname_from_storage` returns the nickname only on a fingerprint match, and `save_invitation_to_storage` (the single fresh-save chokepoint) clears any stale binding. Pinned by `nickname_binding_matches_only_its_own_invitation`.
- **Final codex pass:** "I did not identify any discrete correctness, security, or maintainability issues introduced by this diff. The added invitation resume logic appears guarded against repeated render-side retries and stale nickname reuse."

- *Code-first:* accept saves invitation+nickname+fp; recovery validates fp, decides Resume/Discard/Prompt, fires resume once.
- *Testing:* 4 decision-matrix tests + one-shot guard + fp-binding test; the `resume` regression test was verified to FAIL under the naive processed-first ordering and PASS with the fix. `cargo test -p river-ui --bins` green (271 tests). `fmt`/`clippy`/`check` on `wasm32-unknown-unknown` clean, no new warnings.
- *Skeptical:* covered resume-then-error (Dismiss clears storage, no loop), half-written fp binding (nickname dropped), and new-invite-after-resume-in-session (URL path skips recovery).
- *Big-picture:* matches the issue; UI-only, no delegate/contract migration, additive, follows the existing localStorage + fingerprint patterns.

Risk tier: Light (self-contained UI recovery logic, no high-risk surface).

[AI-assisted - Claude]

PR #334, opened 2026-06-02
ci: authenticate and cache dioxus-cli install to fix binstall rate-limit flake

### LINKED ISSUE #316: ci: ui-playwright-tests flaky on cargo binstall rate-limit
## Summary

\`ui-playwright-tests\` (and the \`build\` job inside the same workflow that installs Dioxus CLI) can fail when GitHub's unauthenticated API rate limit hits the \`cargo binstall -y dioxus-cli\` step.

Hit on PR #314 build https://github.com/freenet/river/actions/runs/26469025310/job/77937061097:

\`\`\`
INFO has_release_artifact{release=GhRelease { repo: GhRepo { owner: \"DioxusLabs\", repo: \"dioxus\" }, tag: \"v0.7.9\" }} ... Received status code 403 Forbidden, will wait for 120s and retry
WARN The package dioxus-cli v0.7.9 will be installed from source (with cargo)
...
error[E0599]: no function or associated item named \`credential_helper\` found for struct \`Cred\` in the current scope
  --> auth-git2 v0.5.8
\`\`\`

What happened:
1. \`cargo binstall\` tried to fetch the prebuilt dioxus-cli binary from the GitHub Releases API.
2. The API returned 403 (unauthenticated rate limit).
3. Binstall fell back to building from source.
4. The source build of \`dioxus-cli v0.7.9\` failed because its transitive dependency \`auth-git2 v0.5.8\` uses \`Cred::credential_helper\`, which was removed in a newer \`git2\` crate.

The first failure (rate limit) is the trigger; the second (broken source build) just removes the safety net.

## Proposed fix

Pass \`GITHUB_TOKEN\` to \`cargo binstall\` so it makes authenticated requests (5000/hr ceiling vs 60/hr unauthenticated):

\`\`\`yaml
- name: Install Dioxus CLI
  env:
    GITHUB_TOKEN: \${{ secrets.GITHUB_TOKEN }}
  run: cargo binstall -y dioxus-cli
\`\`\`

Additional belt-and-braces options if the rate limit still bites:
- Pin the Dioxus version (\`cargo binstall -y dioxus-cli@0.7.9\`) so binstall doesn't first query for the latest tag.
- Cache \`~/.cargo/bin/dx\` between runs (we already cache \`cargo-make\` similarly).

## Workaround for now

Re-run the workflow; the rate limit resets after the window expires. This is the kind of intermittent breakage we have a rule against tolerating (see \`~/.claude/rules/flaky-tests.md\`), so the fix above should land soon.

## References

- PR #314 hit this: https://github.com/freenet/river/pull/314
- Failed run: https://github.com/freenet/river/actions/runs/26469025310/job/77937061097
- \`.github/workflows/build.yml\` "Install Dioxus CLI" step

[AI-assisted - Claude]

sanity:
## Problem

`build` and `ui-playwright-tests` install dioxus-cli with `cargo binstall -y dioxus-cli`. binstall first queries the GitHub Releases API for the prebuilt binary. **Unauthenticated** that call is capped at **60 requests/hr**, and it was hitting **403 Forbidden** (rate limit). binstall then falls back to a from-source build, which fails because a transitive dep (`auth-git2 v0.5.8` → `git2`) uses an API removed in newer git2 — so the whole job dies intermittently.

The rate-limit is the trigger; the broken source build just removes the safety net. (See #316 for the failing run on #314.)

## Approach

Fix both `Install Dioxus CLI` steps:

- **Pass `GITHUB_TOKEN`** so binstall makes **authenticated** requests (5000/hr ceiling instead of 60/hr). This directly removes the 403 trigger. `secrets.GITHUB_TOKEN` is auto-provided to every run — no repo config needed.
- **Cache `~/.cargo/bin/dx`** keyed on `hashFiles('ui/Cargo.toml')` (so a dioxus version bump invalidates it rather than pinning a stale CLI forever), and **gate the install on the cache step's `cache-hit` output**. A warm cache skips the install and the API call entirely; a miss always (re)installs with `--force`.

Gating on the cache result (not file existence) matters because `freenet-default-runner` is **self-hosted**: a stale `dx` can linger across a cache-key change, and a file-existence check would wrongly skip the reinstall and use the stale CLI. On a cache hit, `actions/cache` restores the correct cached binary, overwriting any stale one.

I deliberately did **not** pin the dioxus-cli version (a third option floated in the issue) to avoid a CLI/library version-mismatch risk — the token + cache fully address the documented rate-limit trigger.

## Testing

- This is a CI-workflow-only change; verification is the CI run itself (no unit test applies to a `.yml`).
- Validated with `actionlint` (incl. its shellcheck + expression checks): no new findings — only the repo's pre-existing `freenet-default-runner` self-hosted-label and old-action-version warnings remain.
- YAML parses cleanly; the `steps.cache-dioxus-cli.outputs.cache-hit` expression and the new `id` references validate.
- External review: `codex review --base main` — first pass flagged the file-existence vs. cache-result gap (now fixed by gating on `cache-hit`); re-review on the fixed content found no issues.

Closes #316

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,8 +52,30 @@ jobs:
     - name: Install cargo-binstall
       uses: taiki-e/install-action@cargo-binstall
 
+    - name: Cache Dioxus CLI
+      id: cache-dioxus-cli
+      uses: actions/cache@v3
+      with:
+        path: ~/.cargo/bin/dx
+        # Re-key on ui/Cargo.toml so a dioxus version bump invalidates the
+        # cached `dx` rather than pinning it to a stale CLI version forever.
+        key: ${{ runner.os }}-dioxus-cli-${{ hashFiles('ui/Cargo.toml') }}
+
     - name: Install Dioxus CLI
-      run: cargo binstall -y dioxus-cli
+      # Gate on the cache *result*, not on whether the file exists:
+      # freenet-default-runner is self-hosted, so a stale ~/.cargo/bin/dx can
+      # persist from an earlier job even when this key misses (e.g. after a
+      # dioxus bump). A file-existence check would then skip the install and
+      # use the stale CLI; a cache-key miss must always (re)install.
+      #
+      # Pass GITHUB_TOKEN so binstall makes authenticated GitHub API requests
+      # (5000/hr) instead of unauthenticated ones (60/hr). The unauthenticated
+      # limit was getting hit, returning 403, forcing a from-source build that
+      # then failed on a broken transitive dep (freenet/river#316).
+      if: steps.cache-dioxus-cli.outputs.cache-hit != 'true'
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: cargo binstall -y --force dioxus-cli
 
     - name: Setup Node.js
       uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary

**Self-review (4 lenses):**
- *Code-first*: symmetric change across both jobs (cache step + cache-gated install); reads correctly.
- *Testing*: no repo-level test applies to a workflow `.yml`; verification is actionlint + the CI run. actionlint (with shellcheck/expression checks) reports only pre-existing warnings; YAML parses; the `cache-hit` expression and new step `id`s validate.
- *Skeptical*: confirmed `secrets.GITHUB_TOKEN` is auto-provided to every run; the `Cache cargo` step's paths don't clobber `~/.cargo/bin/dx`; stale-cache-after-dioxus-bump mitigated by keying on `ui/Cargo.toml`.
- *Big-picture*: matches the issue's primary (GITHUB_TOKEN) + caching recommendations; did NOT version-pin dioxus-cli to avoid a CLI/library mismatch risk.

**External (Codex `codex review --base main`):**
- Pass 1 — 1×P2: gating the install on `[ -x ~/.cargo/bin/dx ]` (file existence) could reuse a **stale** `dx` that lingers on the self-hosted runner across a cache-key change, defeating the invalidation. **Fixed**: now gated on `steps.cache-dioxus-cli.outputs.cache-hit != 'true'`, and a miss reinstalls with `--force`.
- Pass 2 (on the fixed content) — no findings.

No dismissed findings.

[AI-assisted - Claude]

PR #335, opened 2026-06-02
polish(cli): make identity-export vk-mismatch hint conditional on override

### LINKED ISSUE #285: polish(cli): make identity-export vk-mismatch error message conditional on override-set vs not
Follow-up from PR #282 skeptical round-2 review.

## Context

PR #282 added a vk-match guard in \`cli/src/commands/identity.rs::export_identity\` that errors when \`signing_key.verifying_key() != authorized_member.member.member_vk\`. The current error message says:

> "This usually happens when \`--signing-key-file\` / \`RIVER_SIGNING_KEY_FILE\` overrides the signing identity but \`rooms.json\` still holds another identity's cached membership state."

## Polish

That message assumes the override is set. The check can also fire WITHOUT an override if rooms.json gets corrupted (e.g., the chat-delegate sync writes \`signing_key_bytes\` for one identity but \`self_authorized_member\` for another — exactly the bug-class that motivated PR #282 in the first place). When the check fires without an override, the message misleads the user into re-running without an override they never set.

## Fix shape

Add \`Storage::has_signing_key_override() -> bool\` or expose the field. In \`export_identity\`'s error path, branch on it:

\`\`\`rust
let hint = if storage.has_signing_key_override() {
    "This usually happens when --signing-key-file overrides the signing identity ..."
} else {
    "rooms.json appears corrupted: the cached AuthorizedMember.member_vk does not match \\
     signing_key_bytes. Try re-accepting the invitation, or import a fresh identity token."
};
\`\`\`

## Test

Skeptical reviewer also noted: no test pins the vk-match guard. Extract \`check_export_coherence(signing_key, authorized_member)\` as a pure helper and write a unit test that constructs an AuthorizedMember signed by key A, calls the helper with signing key B as override → asserts error.

## Severity

LOW. Message clarity issue, not a correctness bug. The guard itself works correctly; only the user-facing diagnostic could mislead in the rare rooms.json-corruption case.

[AI-assisted - Claude]

sanity:
## Problem

The vk-match guard in `cli/src/commands/identity.rs::export_identity` (added in #282) refuses to export an identity when the signing key's verifying key does not match the cached `AuthorizedMember.member_vk`. Its error message unconditionally blamed `--signing-key-file` / `RIVER_SIGNING_KEY_FILE` overrides:

> This usually happens when `--signing-key-file` / `RIVER_SIGNING_KEY_FILE` overrides the signing identity but `rooms.json` still holds another identity's cached membership state.

But the guard also fires with **no override set** — when `rooms.json` is internally inconsistent (e.g. the chat-delegate sync wrote `signing_key_bytes` for one identity and `self_authorized_member` for another, the exact bug-class that motivated #282). In that case the message sends the user chasing an override they never set.

## Approach

Extract the guard into a pure helper `check_export_coherence(signing_key, authorized_member, has_signing_key_override)` that branches **only the diagnostic hint**:

- **override set** → keep the override wording (drop / re-point the override).
- **no override** → point at `rooms.json` corruption (re-accept the invitation or import a fresh identity token).

The underlying key-verification (`signing_key.verifying_key() == authorized_member.member.member_vk`) is byte-identical to before — this is a message-only change, not a validation change. Added `Storage::has_signing_key_override()` predicate (follows the existing `resolve_signing_key` pattern) so the call site can tailor the hint without exposing the key.

This also closes the test-coverage gap the skeptical reviewer flagged: the guard now has a pure, directly-testable helper.

## Testing

Three new unit tests in `cli/src/commands/identity.rs`:
- `coherence_passes_when_signing_key_matches_member` — coherent pair accepted in both override states.
- `coherence_error_mentions_override_when_override_set` — mismatch + override → references the override mechanism, not corruption.
- `coherence_error_mentions_corruption_when_no_override` — mismatch + no override → references corruption, never the override.

The no-override test is the load-bearing regression: against the old unconditional message it fails both assertions (verified). `cargo test -p riverctl --lib`: 82 passed. `cargo fmt --check` and `cargo clippy -p riverctl --all-targets` clean. External Codex review: no findings.

Closes #285

[AI-assisted - Claude]

### DIFF
--- a/cli/src/commands/identity.rs
+++ b/cli/src/commands/identity.rs
@@ -2,7 +2,7 @@ use crate::api::ApiClient;
 use crate::output::OutputFormat;
 use anyhow::{anyhow, Result};
 use clap::Subcommand;
-use ed25519_dalek::VerifyingKey;
+use ed25519_dalek::{SigningKey, VerifyingKey};
 use river_core::room_state::identity::IdentityExport;
 use river_core::room_state::member::{AuthorizedMember, Member, MemberId};
 use river_core::room_state::ChatRoomParametersV1;
@@ -137,24 +137,11 @@ async fn export_identity(
     // authorized_member.member.member_vk; otherwise importing the token
     // produces an identity whose secret key signs nothing the room
     // contract accepts.
-    //
-    // This catches the case where `--signing-key-file` overrides the
-    // signing identity but `self_authorized_member` is read from
-    // `rooms.json` (which still has the previous identity's
-    // `AuthorizedMember`). The two pieces would otherwise be packaged
-    // together with mismatched verifying keys, silently breaking the
-    // imported identity on first use.
-    if signing_key.verifying_key() != authorized_member.member.member_vk {
-        return Err(anyhow!(
-            "Refusing to export an identity with mismatched signing key and \
-             authorized member. The signing key's verifying key does not match \
-             the cached AuthorizedMember.member_vk for this room. This usually \
-             happens when `--signing-key-file` / `RIVER_SIGNING_KEY_FILE` overrides \
-             the signing identity but `rooms.json` still holds another identity's \
-             cached membership state. Re-run without the override (or with the \
-             override pointing at THIS identity) to produce a coherent token."
-        ));
-    }
+    check_export_coherence(
+        &signing_key,
+        &authorized_member,
+        api_client.storage().has_signing_key_override(),
+    )?;
 
     let export = IdentityExport {
         room_owner: room_owner_key,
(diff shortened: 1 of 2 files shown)


sanity:
## Review summary

- *Code-first*: helper returns `Ok(())` on key match (early return), else builds the error with a context-appropriate hint; the `==`/`!=` verification is unchanged from the inline guard. Call site passes `storage().has_signing_key_override()`.
- *Testing*: 3 tests cover the match (both flag values) and both hint branches; the no-override test asserts the override mechanism is NOT referenced, which is the regression assertion that fails against the old unconditional message (verified).
- *Skeptical*: override-set-but-keys-match → `Ok`, no message (correct). No stale references to the old message text anywhere (grepped). Diff touches only `cli/` source — no `common/`/`delegates/`/`contracts/`/`Cargo.*`, so no WASM/migration concern.
- *Big-picture*: message-only polish; key-verification behavior intact — not a crypto/auth change.

**External model (Codex, `codex review --base main`):** clean — "The change preserves the existing coherence check while tailoring the diagnostic based on whether a signing-key override is active. The added helper and tests do not introduce an apparent correctness issue." No findings to address.

[AI-assisted - Claude]

PR #336, opened 2026-06-02
test(ui): pin markdown sanitization against XSS regression (#315)

### LINKED ISSUE #315: audit: pin markdown::Options::gfm() never enables allow_dangerous_html
## Background

During multi-model review of PR #314 (fix for #227 stored XSS via member nickname), reviewers flagged that three `dangerous_inner_html` sinks remain in the UI tree:

- \`ui/src/components/conversation.rs:1531\` — \`desc_html\` (room description)
- \`ui/src/components/conversation.rs:2105\` — \`msg.content_html\` (message body)
- \`ui/src/components/direct_messages/dm_thread_modal.rs:1127\` — \`body_html\` (DM body)

All three are produced by \`message_to_html\` → \`markdown::to_html_with_options(_, Options::gfm())\`. The \`markdown\` crate's GFM mode defaults \`allow_dangerous_html=false\` and \`allow_dangerous_protocol=false\`, so attacker-supplied raw HTML in a message / description body is escaped at the markdown layer rather than smuggled into the DOM.

The audit confirmed those sites are safe **today**. The concern is silent regression: a future \`markdown\` crate upgrade or an \`Options::gfm()\` flip could re-enable dangerous HTML pass-through without anyone noticing, since none of the existing tests pin the sanitization behavior.

## Proposed work

1. Add a unit test in \`ui/src/components/conversation.rs::tests\` that runs \`message_to_html(\"\")\` and asserts the output contains the *literal* escaped string (\`&lt;img\` etc.), not an executable \`\` tag.
2. Add a similar test pinning the \`javascript:\` protocol sanitization for autolinks (\`message_to_html(\"<javascript:alert(1)>\")\` etc.).
3. Optionally add a Playwright spec rendering a message with a stored XSS payload and asserting \`document.querySelector('script,img[onerror]')\` is null.

## Why this is a separate issue

PR #314 fixes the actual exploitable XSS (member nickname rendered raw). The markdown sinks have a layered defense already; adding the regression pins is hygiene, not a security fix. Scope-creeping #314 to include this would dilute the focused fix.

## References

- PR #314 (this fix): https://github.com/freenet/river/pull/314
- Original XSS issue: #227
- Prior art (room header structural fix): #226

[AI-assisted - Claude]

sanity:
## Problem

River renders chat message bodies, room descriptions, and DM bodies by feeding attacker-controlled text through `markdown::to_html_with_options(_, &markdown::Options::gfm())` and injecting the result via `dangerous_inner_html` at three sites:

- `ui/src/components/conversation.rs` — room description (`desc_html`) and message body (`msg.content_html`)
- `ui/src/components/direct_messages/dm_thread_modal.rs` — DM body (`body_html`)

This is **safe today**: the `markdown` crate's GFM mode defaults `allow_dangerous_html = false` and `allow_dangerous_protocol = false`, so raw HTML is escaped and dangerous URL schemes are neutralized at the markdown layer. The audit (issue #315) confirmed this. But no test pinned that behavior, so a future `markdown` crate upgrade or an `Options::gfm()` change could silently re-enable raw-HTML pass-through and re-open the stored-XSS hole closed alongside #227 / #314.

## Approach

Audit-and-pin only — **no production code change**. Verified against `markdown` 1.0.0 that `Options::gfm()` leaves both dangerous flags at their `false` defaults (`CompileOptions::gfm()` only sets `gfm_tagfilter`; `allow_dangerous_html`/`allow_dangerous_protocol` stay at the `bool` default). Added two regression tests to the existing `conversation.rs` test module:

- **`raw_html_is_escaped_not_executed`** — ``/`<svg>` payloads must be escaped to `&lt;…`, never emitted as live tags. ``/`<svg>` are deliberately chosen because they are NOT on the GFM tagfilter's neutralization list, so they genuinely trip if `allow_dangerous_html` is ever flipped on (`<script>`/`<iframe>` are masked by the tagfilter even with dangerous HTML enabled, so they only verify escaping, not the flip).
- **`dangerous_url_schemes_are_neutralized`** — `javascript:` / `vbscript:` / `data:` autolinks and `[text](scheme:…)` links must never reach an `href`.

Both assertions were validated to **PASS on the safe `gfm()` path and FAIL when the dangerous flags are forced on**, so they are genuine tripwires, not tautologies.

The optional Playwright spec from the issue is intentionally skipped — the native Rust unit tests fully satisfy the stated goal (catch a silent regression if the markdown defaults flip).

## Testing

- `cargo test -p river-ui --bins` — 273 passed (incl. the 2 new tests).
- `cargo fmt --check` and `cargo clippy -p river-ui --bins` clean for the added code (pre-existing warnings elsewhere untouched).

Closes #315

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -2432,6 +2432,85 @@ fn MessageGroupComponent(
 mod tests {
     use super::*;
 
+    /// Issue #315 — pin that the markdown renderer never passes raw HTML
+    /// through to the DOM. `markdown_to_html` feeds attacker-controlled
+    /// message text into `markdown::to_html_with_options(_, Options::gfm())`,
+    /// whose output is then injected via `dangerous_inner_html` at three
+    /// sites (room description, message body, DM body). GFM mode defaults
+    /// `allow_dangerous_html = false`, so raw HTML is escaped rather than
+    /// emitted as live markup. This test fails loudly if a future `markdown`
+    /// upgrade or an `Options` change ever flips that on — which would
+    /// re-open the stored-XSS hole closed alongside #227 / #314.
+    #[test]
+    fn raw_html_is_escaped_not_executed() {
+        // Each payload is a classic stored-XSS vector. With
+        // allow_dangerous_html=false the leading `<` must be escaped to
+        // `&lt;`, so the markup survives as inert text instead of a live
+        // element.
+        //
+        // `<img>` and `<svg>` are deliberately chosen: they are NOT on the
+        // GFM tagfilter's neutralization list, so if `allow_dangerous_html`
+        // were ever flipped on they would pass through as live `<img …>` /
+        // `<svg …>` tags — making these the payloads that actually trip the
+        // tripwire. (`<script>`/`<iframe>` are masked by the tagfilter even
+        // with dangerous HTML enabled, so they can't distinguish the flip;
+        // they're covered below only for the escaping guarantee.)
+        let live_tag_vectors = ["<img src=x onerror=alert(1)>", "<svg onload=alert(1)>"];
+        for payload in live_tag_vectors {
+            let html = message_to_html(payload);
+            assert!(
+                html.contains("&lt;"),
+                "raw HTML payload should be HTML-escaped (expected `&lt;`): \
+                 input={payload:?} output={html:?}"
+            );
+            assert!(
+                !html.contains("<img")
+                    && !html.contains("<svg")
+                    && !html.contains("<iframe")
+                    && !html.contains("<script"),
+                "raw HTML payload must not survive as an executable tag: \
+                 input={payload:?} output={html:?}"
+            );
+        }
+
+        // `<script>`/`<iframe>` must still be escaped on the safe path.
+        for payload in ["<script>alert(1)</script>", "<iframe></iframe>"] {
+            let html = message_to_html(payload);
+            assert!(
+                html.contains("&lt;"),
+                "raw HTML should be HTML-escaped (expected `&lt;`): \
+                 input={payload:?} output={html:?}"
+            );
+        }
+    }
+
+    /// Issue #315 — pin that the markdown renderer never emits a dangerous
+    /// URL scheme in an `href`. GFM mode defaults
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary

**Self-review (4 lenses):**
- **Code-first:** Tests-only diff, no production change. Both tests call the real `message_to_html` entry point, which funnels into the shared `markdown_to_html` used by all three `dangerous_inner_html` sinks (room description at `conversation.rs:1573`, message body at `conversation.rs:2186`, DM body at `dm_thread_modal.rs:1127`).
- **Testing:** Each assertion was validated to PASS on the safe `gfm()` path and FAIL when `allow_dangerous_html`/`allow_dangerous_protocol` are forced on, so they are genuine tripwires, not tautologies. Payloads were chosen with GFM-tagfilter awareness: ``/`<svg>` (NOT tagfilter-masked) are the ones that actually catch an `allow_dangerous_html` flip; `<script>`/`<iframe>` only verify escaping.
- **Skeptical:** Confirmed empirically against markdown 1.0.0 that `<javascript:>`, `[..](javascript:)`, `[..](vbscript:)`, `[..](data:text/html,<script>)` all neutralize the href to empty.
- **Big-picture:** Audit-and-pin only — the security control is already correct (verified `CompileOptions::gfm()` leaves both dangerous flags at their `false` default). This is hygiene, NOT a security-control change.

**External model review (Codex / gpt-5.5):** No findings. Verdict: *"The change only adds regression tests around markdown HTML and URL sanitization, and I did not identify any introduced build, test, or production correctness issues."* Codex independently confirmed via the markdown crate's own `test_compile_options` / `test_options` unit tests that `Options::gfm()` keeps `allow_dangerous_html = false` and `allow_dangerous_protocol = false`.

**Note (not blocking):** `ui/Cargo.toml` declares `markdown = "1.0.0-alpha.21"` but `Cargo.lock` resolves it to the stable `1.0.0` (the caret requirement is satisfied by 1.0.0 per semver). The audit and these tests pin the *resolved* 1.0.0 behavior. No action needed for this PR.

**Tests:** `cargo test -p river-ui --bins` — 273 passed (incl. the 2 new). `cargo fmt --check` + `cargo clippy -p river-ui --bins` clean for the added code.

[AI-assisted - Claude]

PR #337, opened 2026-06-02
fix(ui): preserve private-room edits across optimistic apply_delta (#310)

### LINKED ISSUE #310: Edited private-room messages briefly revert to original when a new message is sent
## Reported by @Ivvvor in freenet/river#246

> I get some weirdness with edited private room messages. When an ID sends a new message it 'un-edits' the edited messages briefly, until the new message is displayed and sent.

## Symptom

In a private room, after editing a previous message, sending any new message causes the previously-edited message to briefly revert to its original pre-edit content (a visual flicker), then re-render with the edit applied once the new message lands.

## Likely cause

Sending a new message triggers a local optimistic state update plus a re-render. For private rooms, `RoomSynchronizer::apply_delta_inner` / `update_room_state_inner` decrypt the private action messages and `rebuild_actions_state_with_decrypted` to surface the edited content. If the send path mutates `room_state.recent_messages` in a way that resets / does not preserve the `actions_state` ordering, the conversation view briefly renders the pre-edit ciphertext placeholder (or the original message text) before the next merge re-applies the edit.

Possible code locations to start from:

- `ui/src/components/conversation.rs` — the send flow (`do_send`-style handlers, edit/delete action wiring around the "edit message" / "delete message" hover buttons).
- `ui/src/components/app/freenet_api/room_synchronizer.rs` — the `rebuild_actions_state_with_decrypted` calls inside `apply_delta_inner` and `update_room_state_inner`.
- `common/src/room_state/message.rs` — `MessagesV1::apply_delta` and how edit actions (the `set_text` action) commute with new-message inserts.

A reasonable starting hypothesis: the send path inserts the new message into local `recent_messages` without going through the same decrypt + actions-rebuild that `apply_delta_inner` does, so for a frame or two the edits don't render until the network echo / re-merge applies the actions again.

Severity: visual only, no data loss. Worth fixing because it suggests state isn't being kept consistent between local optimistic updates and post-merge rendering.

## Related

- #246 (memory blowup) — unrelated, just reported in the same thread.

[AI-assisted - Claude]

sanity:
## Problem

In a private room, an edited message briefly reverted to its original text whenever a new message was sent (reported by @Ivvvor in #246). The optimistic send/edit/delete/react handlers call `ChatRoomStateV1::apply_delta`, whose `MessagesV1` impl always ends with the **non-decrypting** `rebuild_actions_state()`. For a private room that can only decode PUBLIC action messages, so it cleared `actions_state` and the edit (carried by a PRIVATE action message) vanished until the network echo ran the decrypt-aware rebuild — a visible flicker back to the pre-edit text.

The wipe is **not** specific to message deltas: `MessagesV1::apply_delta` re-runs the public-only rebuild unconditionally, even when the delta carries no `recent_messages` at all. So sending a DM, changing a nickname, banning a member, editing the room name/description/settings, or rotating the secret all reverted private edits the same way.

## Approach

Extract the decrypt-aware rebuild — previously inlined byte-identically in three network-ingestion paths (`apply_delta_inner`, `update_room_state_inner`, GET handler) — into `RoomData::rebuild_private_actions_state` (a no-op on public rooms), and call it after **every** local optimistic `apply_delta`:

- conversation.rs: send, edit, delete, react
- direct_messages.rs / dm_thread_modal.rs: DM send, DM purge
- member_info_modal: nickname edit, ban (after ban + post-ban rotation)
- room_list: room name, description, settings, max_members, manual secret rotation

The three network paths now delegate to the same helper, removing ~100 lines of duplication.

Pure UI/state-rendering fix — **no** contract/delegate/wire-format/crypto/migration change (only `ui/` source touched).

## Testing

New `private_edit_survives_new_message_send` regression test in `room_data.rs`: authors a private message + encrypted edit, sends a new message, and asserts the edit is wiped by `apply_delta` alone and restored by the helper; also pins the broader empty-delta (DM/nickname/ban/config) class. Verified the test FAILS when the helper is neutralized and PASSES with it. Full `cargo test -p river-ui --bins` green (272 passed); `cargo fmt` clean.

Closes #310

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/freenet_api/response_handler/get_response.rs
+++ b/ui/src/components/app/freenet_api/response_handler/get_response.rs
@@ -12,7 +12,6 @@ use crate::components::app::{CURRENT_ROOM, PENDING_INVITES, ROOMS, WEB_API};
 use crate::constants::ROOM_CONTRACT_WASM;
 use crate::invites::PendingRoomStatus;
 use crate::room_data::RoomData;
-use crate::util::ecies::decrypt_with_symmetric_key;
 use crate::util::{
     from_cbor_slice, get_current_system_time, owner_vk_to_contract_key, to_cbor_vec,
     try_from_cbor_slice,
@@ -27,10 +26,8 @@ use freenet_stdlib::prelude::{
 };
 use river_core::room_state::member::MemberId;
 use river_core::room_state::member_info::{AuthorizedMemberInfo, MemberInfo};
-use river_core::room_state::message::{AuthorizedMessageV1, MessageId, MessageV1, RoomMessageBody};
-use river_core::room_state::privacy::PrivacyMode;
+use river_core::room_state::message::{AuthorizedMessageV1, MessageV1, RoomMessageBody};
 use river_core::room_state::{ChatRoomParametersV1, ChatRoomStateV1, ChatRoomStateV1Delta};
-use std::collections::HashMap;
 use std::sync::Arc;
 
 pub async fn handle_get_response(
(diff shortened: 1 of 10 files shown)


sanity:
## Review summary

**Self-review** (code-first / testing / skeptical / big-picture): the helper `RoomData::rebuild_private_actions_state` is a byte-identical extraction of the decrypt-aware rebuild previously inlined in the three network paths; it's `&mut self`, guarded by `is_private()` (no-op for public rooms), and now called after every local optimistic `apply_delta`. Pure UI/state fix — no contract/delegate/wire-format/crypto/migration surface touched (only `ui/` source). The regression test was verified to FAIL with the helper neutralized and PASS with it.

**Codex CLI (`codex review --base main`)**: two passes.
- First pass raised one **[P2]**: the wipe also affects config/secret-rotation paths (room name/description/settings/max_members, manual rotation) that call `apply_delta` without the rebuild. Addressed in commit `25a70ecc` by wiring the helper into all of those success paths (edit_room_modal.rs, room_name_field.rs).
- Second pass on the updated code: clean — *"The changes centralize the private action-state rebuild logic and invoke it after local optimistic deltas that would otherwise clear decrypted private actions. I did not find any introduced correctness issues in the diff."*

No findings dismissed; the one substantive finding was fixed.

[AI-assisted - Claude]

PR #339, opened 2026-06-02
test: cross-side fixed-vector test for Invitation wire format (CLI ↔ UI)

### LINKED ISSUE #305: Cross-side fixed-vector test for Invitation wire format (CLI ↔ UI)
## Background

PR #303 added \`Invitation::room_secrets\` to riverctl with comments enforcing "MUST stay byte-identical" to UI's \`ui::components::members::Invitation\`. The two suites each round-trip their own \`Invitation\` type independently — there is no shared, frozen byte vector that both sides MUST decode and re-encode identically.

This means a future \`#[serde(rename = ...)]\` slip, a field added to one side but not the other, or a serde-attr drift would compile-and-test-clean on both sides while silently breaking UI↔CLI invitation exchange.

## Scope

1. Define a canonical base58/CBOR fixture string (e.g. \`INVITATION_FIXED_FIXTURE_V302\`) carrying a deterministic Invitation with all fields populated, including \`room_secrets\` with at least two entries.
2. Add a test in \`cli/src/api.rs::invitation_tests\` that decodes the fixture, asserts every field, then re-encodes and asserts the bytes are identical.
3. Add an equivalent test in \`ui/src/components/members.rs\` using the SAME fixture string.
4. Either keep the literal string duplicated (with a doc-comment cross-reference) or extract to a shared constant in a non-WASM-compiled location (the same blocker as issue #302 scope item 4).

## References
- PR #301 (UI side of \`room_secrets\`)
- PR #303 (CLI side of \`room_secrets\`)
- Issue #302 scope item 4 (eventual struct unification)

[AI-assisted - Claude]

sanity:
## Problem

PR #303 added `Invitation::room_secrets` to riverctl with comments enforcing "MUST stay byte-identical" to the UI's `ui::components::members::Invitation`. But the two suites each round-tripped their *own* `Invitation` type independently — there was no shared, frozen byte vector both sides had to decode and re-encode identically.

That meant a future `#[serde(rename = …)]` slip, a field added to one side but not the other, or a serde-attr drift would compile-and-test-clean on both sides while silently breaking UI↔CLI invitation exchange. (The two structs currently agree — verified — so this is purely adding the missing guardrail, not a format change.)

## Approach

Add a frozen base58(CBOR) fixture string, `INVITATION_FIXED_FIXTURE_V302`, carrying a deterministic `Invitation` with every field populated and two non-contiguous `room_secrets` entries (versions 0 and 3). The same literal is duplicated on both sides with cross-reference doc-comments — the option the issue explicitly permits (scope item 4).

Each side adds `invitation_decodes_frozen_cross_side_fixture`, which:
1. decodes the fixture via that side's exact production wire path,
2. asserts every field individually (plus full struct equality),
3. re-encodes via the production path and asserts the bytes are **byte-identical** to the frozen string.

Because both sides pin the *same* literal, a divergence on either side fails that side's test. The fixture was generated once, deterministically — ed25519 signing is deterministic per RFC 8032, and ciborium emits CBOR map keys in struct field-declaration order — so the bytes are reproducible across both crates (same workspace-pinned `river-core` / `ciborium` / `ed25519-dalek` / `bs58`).

## Testing

- `cargo test -p riverctl --lib` — 80 passed (incl. new test).
- `cargo test -p river-ui --bins` — 272 passed (incl. new test).
- `cargo fmt --check` clean; no new clippy findings in the touched files.
- Verified the regression-detecting property: temporarily reordering the UI `Invitation` struct fields makes the re-encode byte-identity assertion FAIL, then reverted.
- `codex review --base main`: no actionable findings.

No production code changed — test-only, additive.

Closes #305

[AI-assisted - Claude]

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -3709,6 +3709,105 @@ mod invitation_tests {
         }
     }
 
+    /// Frozen cross-side wire-format fixture (issue freenet/river#302/#305).
+    ///
+    /// A base58(CBOR)-encoded [`Invitation`] with every field populated and
+    /// two `room_secrets` entries (non-contiguous versions 0 and 3). The
+    /// **same string literal** appears in the UI at
+    /// `ui/src/components/members.rs`
+    /// (`tests::INVITATION_FIXED_FIXTURE_V302`). Both sides decode it,
+    /// assert every field, then re-encode and assert the bytes are
+    /// byte-identical — so a `#[serde(rename = …)]` slip, a field reorder,
+    /// a serde-attr drift, or a field added to one side but not the other
+    /// can no longer compile-and-test-clean while silently breaking the
+    /// CLI↔UI invitation exchange.
+    ///
+    /// **Do NOT regenerate this string casually.** It pins the on-wire
+    /// format. If a future change legitimately alters the encoding, both
+    /// copies (here and in the UI) must change together and the diff must
+    /// be reviewed as a wire-format change. The string was produced once,
+    /// deterministically, from the seeds in
+    /// [`fixed_fixture_expected_invitation`] (ed25519 signing is
+    /// deterministic per RFC 8032, so the bytes are reproducible).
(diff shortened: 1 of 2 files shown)


sanity:
## External review (codex review --base main)

> The changes add frozen cross-side invitation wire-format tests without modifying production behavior. I did not identify any actionable correctness issues in the diff.

Codex specifically investigated fixture determinism (the stability risk for a frozen vector) — confirmed `sign_struct` uses deterministic ed25519 signing over CBOR, so the fixture bytes are reproducible across both crates. No findings to address.

Self-review across the four lenses (code-first / testing / skeptical / big-picture) surfaced no issues beyond what the PR body covers: the change is test-only and additive, both `Invitation` structs already agree byte-for-byte (so this is a pure guardrail, not a format change), and the regression-detecting property was verified by simulating a field reorder.

[AI-assisted - Claude]

PR #340, opened 2026-06-03
feat: @mentions (UI + riverctl) with per-room notification preference

sanity:
## Problem

River had no way to mention a specific member. Plain `@name` text doesn't link to anyone and breaks when the member renames. There was also no way to tune notifications per room — every external message in any non-active room fired a desktop notification.

## Approach

**Inline wire token, no contract change.** A mention is stored inside the existing message `text` field as `@[Display Name](rv:<lossless-hex-member-id>)`. The `rv:` reference (the member's stable `MemberId`) is authoritative; the name is a fallback snapshot. Clients re-resolve the member's *current* nickname from `member_info` by id, so a rendered chip follows renames. The codec lives in `river-core` behind a new client-only `mentions` feature (mirrors the existing `migration` feature) so it is **never compiled into the room-contract / chat-delegate WASM** — their bytes and keys are unchanged, **no delegate/contract migration**. This was verified by rebuilding both WASMs from origin/main's `common` vs this branch's `common` on the same machine: byte-identical.

- **Rendering** (`conversation.rs`): mentions are extracted *before* markdown (swapped for inert private-use sentinels, then substituted with chip HTML after markdown + anchor finalization), so rendering is independent of markdown's link grammar. The chip shows the current nickname, is HTML-escaped (freenet/river#227), highlights mentions of the local user, and opens the member-info modal on click (document-level delegation, mirroring `invite_click_interceptor`).
- **Composer** (`message_input.rs`): typing `@` opens a keyboard-navigable dropdown of room members (minus self), filtered prefix-first then substring. ↑/↓ move, Enter/Tab insert the wire token at the caret, Esc closes. Caret handling converts between UTF-16 (DOM `selectionStart`) and UTF-8 byte offsets. The candidate list is a prop computed by the parent, so keystroke isolation is preserved.
- **Per-room notifications** (`room_data.rs`, `notifications.rs`, `edit_room_modal.rs`): a `NotificationMode` (All / MentionsAndReplies / Muted) stored as a sidecar map on `Rooms` — rides the existing `rooms_data` delegate blob like `removed_rooms`, no new storage key, survives migration. `notify_new_messages` gates on the mode: Muted suppresses all; MentionsAndReplies fires only when a new message @mentions the local user or replies to one of their messages (resolved by target message id). Default `All` is unchanged behaviour. A non-owner-gated selector in Room Details sets it.
- **riverctl parity**: `message list`/`stream` render tokens as `@name` (current public nickname, snapshot fallback — the CLI cannot decrypt private-room nicknames). `send`/`reply` resolve a bare `@nickname` to a token when it unambiguously, case-insensitively matches a current member's public nickname (the CLI counterpart to the autocomplete picker). Documented in `--help`.

## Testing

- New unit tests: codec (encode/parse/round-trip, id hex, sanitization, multi-mention, non-token text, typed-mention resolution); rendering (rename-following, snapshot fallback, XSS escaping, self-highlight, no-mention fast path, multi-mention); composer helpers (caret detection, candidate filtering, UTF-16↔byte round-trip); notification-mode merge/default.
- All suites green: `river-core` + full common suite, `river-ui` native (288), `riverctl` (83).
- Contract/delegate WASM proven byte-identical (no migration).
- Verified end-to-end in a real browser (Playwright against example-data): chip renders with the resolved current name, autocomplete dropdown lists members with self excluded, keyboard select inserts the correct token.

## Design decisions (from discussion with the maintainer)

- Composer inserts the **raw token** (visible while typing) rather than a rendered pill — simplest, unambiguous, plain `<textarea>` (no contenteditable risk).
- Token carries a **name snapshot overridden on display** by the current nickname.
- Three notification modes including **Muted**.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/river-publish.md
+++ b/.claude/rules/river-publish.md
@@ -23,6 +23,16 @@ git diff HEAD~1 --name-only | grep -E '(common/|delegates/|contracts/|Cargo\.(to
 
 If ANY match, migration is needed. If ONLY `ui/` changed, skip to Step 4.
 
+**Exception — feature-gated client-only `common/` modules.** A change confined to
+a `common/` module gated behind a client-only Cargo feature (one enabled by
+`river-ui`/`riverctl` but NOT by the room-contract or chat-delegate crates —
+e.g. `migration`, `mentions`) does **not** enter the delegate/contract WASM, so
+it needs no migration entry. This holds only because those WASMs are built with
+`-p <crate>` scoping (never `--workspace`, which would unify the feature in).
+Don't add a migration entry for such a change — CI's `check-delegate-migration`
+and `check-room-contract-migration` confirm the WASM is byte-identical and will
+fail loudly if it isn't.
+
 ### Step 2: Add Migration Entry
 
 ```bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -152,13 +152,16 @@ jobs:
       # See freenet/river#241 — these tests pin the wire format and catch
       # silent drift in the ECIES derivation. The `migration` feature also
       # gates the `migration` module's unit tests (legacy room-contract key
-      # derivation, newest-first ordering — freenet/river#292).
+      # derivation, newest-first ordering — freenet/river#292). The `mentions`
+      # feature gates the `mention` module's codec tests (token encode/parse,
+      # lossless id hex, sanitization, typed-mention resolution) — without it
+      # those 12 tests are compiled out and never gate.
       #
       # `--features` takes ONE comma-separated value; passing space-separated
       # feature names makes cargo treat the extras as positional test-name
       # filters (and a second one is a hard "unexpected argument" error).
       # `ecies-randomized` enables `ecies` transitively.
-      run: cargo test -p river-core --lib --features ecies-randomized,migration
+      run: cargo test -p river-core --lib --features ecies-randomized,migration,mentions
 
     - name: Test river-ui (UI crate unit tests)
       env:
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -50,7 +50,7 @@ dialoguer = "0.11"
 atty = "0.2"
 
 # Internal dependencies
-river-core = { version = "0.1.6", path = "../common", features = ["ecies", "ecies-randomized", "migration"] }
+river-core = { version = "0.1.6", path = "../common", features = ["ecies", "ecies-randomized", "migration", "mentions"] }
 freenet-stdlib = { workspace = true, features = ["net"] }
 freenet-scaffold = "0.2.2"
 
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -4857,3 +4857,123 @@ mod monitor_tests {
         );
     }
 }
+
+#[cfg(test)]
+mod mention_cli_tests {
+    use super::*;
+    use river_core::mention::encode_mention;
+    use river_core::room_state::message::{AuthorizedMessageV1, MessageV1, RoomMessageBody};
+    use std::time::SystemTime;
+
+    fn member_id(sk: &SigningKey) -> MemberId {
+        MemberId::from(&sk.verifying_key())
+    }
+
+    /// Build a room state whose `member_info` carries the given (key, nickname)
+    /// entries.
+    fn state_with_members(members: &[(SigningKey, SealedBytes)]) -> ChatRoomStateV1 {
+        let mut state = ChatRoomStateV1::default();
+        for (i, (sk, nickname)) in members.iter().enumerate() {
+            let info = MemberInfo {
+                member_id: member_id(sk),
+                version: i as u32,
+                preferred_nickname: nickname.clone(),
+            };
+            state
+                .member_info
+                .member_info
+                .push(AuthorizedMemberInfo::new_with_member_key(info, sk));
+        }
+        state
+    }
+
+    fn msg_with_text(text: String) -> AuthorizedMessageV1 {
+        let author = SigningKey::from_bytes(&[200u8; 32]);
+        let m = MessageV1 {
+            room_owner: MemberId::from(author.verifying_key()),
+            author: member_id(&author),
+            content: RoomMessageBody::public(text),
+            time: SystemTime::UNIX_EPOCH,
+        };
+        AuthorizedMessageV1::new(m, &author)
+    }
+
+    // --- resolve_outgoing_mentions (send path) ---
+
+    #[test]
+    fn resolves_unambiguous_public_nickname() {
+        let alice = SigningKey::from_bytes(&[1u8; 32]);
+        let state = state_with_members(&[(alice.clone(), SealedBytes::public(b"alice".to_vec()))]);
+        let out = resolve_outgoing_mentions(&state, "hi @alice!");
+        assert_eq!(
+            out,
+            format!("hi {}!", encode_mention(member_id(&alice), "alice"))
+        );
+    }
+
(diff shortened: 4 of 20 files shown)


sanity (commented):
## Comprehensive PR Review: #340

### Summary
- **PR Title:** feat: @mentions (UI + riverctl) with per-room notification preference
- **Type:** feat
- **CI Status:** build, check-wasm-sync, check-delegate-migration, check-room-contract-migration all green on `c895fb9b`; ui-playwright-tests pending
- **Linked Issues:** none referenced
- **Review tier:** Full
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four) + external Codex pass

---

### Code-First Analysis
**Independent understanding matches the description closely.** Inline wire token `@[name](rv:<16-hex>)` carried in the message `text` field; `rv:` is a lossless 64-bit `MemberId` encoding (verified vs. the truncated base32 `Display`), the name is a fallback snapshot. Chip rendering re-resolves the current nickname (rename-following), sentinel-substitution keeps mentions out of the markdown grammar, escaping guards `dangerous_inner_html`. Per-room `NotificationMode` rides a `#[serde(default)]` sidecar on `Rooms`. riverctl renders tokens and resolves bare `@nickname`. **No material intent-vs-implementation gaps.**

### Testing Assessment
**Coverage Level:** good for the codec/UI rendering/caret math; gaps in CI gating and the new decision paths.

| Unit (codec) | ⚠️ | 12 tests exist & pass, but **not run in CI** (feature-gating gap — see Must Fix #1) |
| Unit (UI render/caret) | ✅ | rename-follow, XSS escape, self-highlight, no-mention fast path, UTF-16↔byte all covered |
| Unit (notification gate) | ❌ | `message_notifies_self` has no tests (Should Fix #2) |
| Unit (CLI resolve/render) | ❌ | `resolve_outgoing_mentions` ambiguity + `render_mentions_for_terminal` untested (Should Fix #3) |
| Integration / E2E | ✅ (manual) | Playwright-verified chip render + autocomplete against example-data |

### Skeptical Findings
**Risk Level:** low. The two highest-risk claims were verified by direct build/fuzz, not just diff-reading.

| XSS via chip HTML | refuted | `conversation.rs:548` `render_mention_chip_html` | name escaped for text + attr; `data-member-id` hex-only; class static. Test pins `` neutralized. |
| Sentinel smuggling | refuted | `conversation.rs:524` | pre-existing `U+E000/E001` stripped from text runs; `OPEN idx CLOSE` delimiters unambiguous; markdown passes sentinels verbatim (empirically confirmed). |
| Parser panic (UTF-8/malformed) | refuted | `common/src/mention.rs` | 2M-iteration fuzz of `parse_segments` + `resolve_typed_mentions`, zero panics; ASCII-only delimiters keep slices on char boundaries. |
| No-migration claim | verified | feature gate | byte-identical BLAKE3 for room_contract/chat_delegate with `mentions` present vs. absent; CI `check-delegate-migration` green; all WASM build targets use `-p` (audited). |
| Whitespace-only nickname in autocomplete | low | `conversation.rs:1931` | `!name.is_empty()` doesn't exclude `"   "`. Cosmetic. (Consider #4) |
| Stale-offset insertion on paste-while-open | low | `message_input.rs:140-175` | offsets re-validated for boundary/length only; rare misplacement on concurrent edit, no panic. Acceptable. |

### Big Picture Assessment
**Goal Alignment:** yes. **Anti-Patterns:** none. **Removed Code:** none (0 removed tests/assertions/guards; two `fn`→`pub(crate) fn` widenings are additive). **Scope:** focused. Backwards-compat holds (CBOR `#[serde(default)]` sidecar like `removed_rooms`; mentions degrade to a plain markdown link on old clients). Follows the `migration`-feature, `removed_rooms`-sidecar, and `invite_click_interceptor` precedents.

### Documentation
- Code docs: complete (new public items documented).
- `--help`: updated for `message send`/`reply`.
- Optional: a one-liner in `AGENTS.md` / `river-publish.md` that feature-gated client-only `common/` modules (`mention`, like `migration`) produce no contract/delegate WASM change and need no migration entry.

---

### Recommendations

#### Must Fix (Blocking)
1. **Codec tests don't run in CI.** `mention` and its 12 tests are behind `--features mentions`, which no `river-core --lib` CI step enables (`.github/workflows/build.yml`). `cargo test -p river-core --lib mention` runs 0 tests today. Add `mentions` to a `river-core --lib` feature list so the suite actually gates. (Per the repo's "a bug past CI is a bug in CI" rule.)

#### Should Fix (Important)
2. **Test `message_notifies_self`** (`notifications.rs`) — the new MentionsAndReplies decision. Mention branch is purely testable; extract the reply-authorship check into a pure helper (taking the `recent_messages` slice) and test both branches incl. the "target scrolled out → conservatively no-notify" case.
3. **Test the CLI mention helpers** (`cli/src/api.rs`) — `resolve_outgoing_mentions` (unambiguous link, shared-nickname→plain, private/no-public→plain, already-encoded token untouched) and `render_mentions_for_terminal` (current name vs snapshot fallback).

#### Consider (Suggestions)
4. Exclude whitespace-only nicknames from autocomplete (`name.trim().is_empty()`).
5. Doc-note that a removed-then-rejoined room inherits its prior `notification_modes` preference (local-wins merge).
6. The CLI silently sends plain text when a `@word` doesn't resolve (esp. private rooms — encrypted nicknames can't match). Documented in `--help`; a stderr hint is optional.

---

### Verdict
**HEAD SHA reviewed:** c895fb9b7c212f35fdd5ac90d5a8c60f6479ea04

No high/medium-severity bugs; the security-critical paths (XSS, parser, migration) all held under direct verification. The blocking item is the CI gating gap (Must Fix #1) — the codec tests must actually run. Should Fix #2/#3 close the untested new decision/resolution logic. After these land, a fresh review pass on the new HEAD is required before merge (changes exceed the trivial threshold and touch the notification path).

[AI-assisted - Claude]

sanity (commented):
## Re-Review After Fixes: #340

Re-ran an external Codex pass (full PR) and an adversarial skeptical pass on the new HEAD after the review fixes in `c66d170a`. The fix commit is overwhelmingly additive (tests + one CI feature-flag + docs); the only production-logic changes are a behavior-preserving extraction of the notification decision into a pure helper and a one-line autocomplete filter tweak.

**Findings addressed (from the prior pass):**
- **Must Fix #1 (CI codec-gating):** fixed — `mentions` added to the `river-core --lib` test step; the 12 codec tests now run (verified: 0 → 12).
- **Should Fix #2 (notification gate untested):** fixed — extracted pure `mentions_or_replies_to_self(...)` and added 6 tests (mention-of-self, mention-of-other, reply-to-own, reply-to-other, reply-target-scrolled-out, plain).
- **Should Fix #3 (CLI resolver untested):** fixed — 8 tests for `resolve_outgoing_mentions` (unambiguous/ambiguous/unknown/private/encoded) and `render_mentions_for_terminal`.
- **Consider #4/#5/#6:** whitespace-only nicknames excluded from autocomplete; rejoin-inherits-preference documented; river-publish.md notes feature-gated client-only `common/` modules need no migration entry.

**Re-review result:** Codex — no actionable regressions. Skeptical — verified the refactor is behavior-preserving across all input quadrants (incl. the ROOMS-unreadable mention path), all new tests are correct and pass, the CI fix is load-bearing, and the no-migration claim holds (no `.wasm`/migration files touched; `-p`-scoped WASM builds; CI migration checks green).

**Verdict:** Ready to Merge.
**HEAD SHA reviewed:** c66d170ad68fbe69babdcb9b7c01688bf90143a0

[AI-assisted - Claude]

PR #346, opened 2026-06-04
feat(ui): drag-and-drop room reordering; move private-room lock to the right

sanity:
## Problem

Two small Rooms-list UX requests:

1. **No way to reorder rooms.** The rail rendered rooms in raw `HashMap`
   iteration order, which is arbitrary and unstable across renders. Users
   couldn't put their most-used rooms at the top.
2. **Private-room lock icon sat to the left of the room name.** Requested move
   to the right.

## Approach

**Reorder (drag-and-drop):**
- Added `Rooms::room_order: Vec<VerifyingKey>` — a local, per-device display
  order persisted in the existing `rooms_data` delegate blob via
  `#[serde(default)]`, the same persistence model as `notification_modes` and
  `removed_rooms`. **No delegate migration needed:** `Rooms` is a `ui/`-only
  type, not compiled into delegate/contract WASM, and the field is additive +
  `serde(default)` so old blobs deserialise cleanly.
- `ordered_room_keys()` returns manually-positioned rooms first, then any
  not-yet-positioned rooms in a deterministic key-byte order. This also fixes
  the pre-existing arbitrary/unstable render order.
- `move_room(dragged, target)` drops the dragged room immediately before the
  target; `move_room_to_end(dragged)` backs a tail drop zone so the final slot
  is reachable (every row drop inserts *before* its target). `merge` keeps this
  device's order authoritative (adopts incoming-only keys, prunes to live
  rooms); `leave_room` prunes the left room.
- Each room row is `draggable`; all drag-state signal mutations route through
  `crate::util::defer()` (Dioxus WASM signal-safety), and the drag-over border
  space is reserved up front (transparent → accent) so there is **no layout
  shift**.

**Lock icon:** moved after the room-name span in the flex row (now right of the name).

## Testing

- `river-ui` native unit tests (`cargo test -p river-ui --bins`, all green):
  `ordered_room_keys` determinism + stale-entry filtering; `move_room`
  up/down/self-drop/unknown-key; `move_room_to_end`; `leave_room` pruning;
  `merge` union+prune; a populated `room_order` ciborium round-trip
  (`rooms_round_trip_preserves_room_order`); and the pre-247 backward-compat
  blob test extended to assert `room_order` defaults empty.
- `cargo clippy` — no new warnings in the touched files.

## Notes / limitations

- HTML5 drag-and-drop is desktop/pointer only; touch devices won't reorder
  (rooms still render and click-to-open still works). Acceptable for v1.
- Reordering is a local, per-device view preference (last-write-wins on the
  shared `rooms_data` blob, identical to the existing `removed_rooms` /
  `notification_modes` fields). A future `rooms_data` persistence rework (CAS /
  per-room-key split) would protect `room_order` for free since it rides the
  same blob — additive, low conflict risk.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -502,6 +502,7 @@ fn initial_rooms() -> Rooms {
         removed_rooms: std::collections::HashSet::new(),
         notification_modes: Default::default(),
         migrated_rooms: Vec::new(),
+        room_order: Vec::new(),
     }
 }
 
--- a/ui/src/components/direct_messages.rs
+++ b/ui/src/components/direct_messages.rs
@@ -639,6 +639,7 @@ mod tests {
             notification_modes: Default::default(),
             migrated_rooms: Vec::new(),
             removed_rooms: std::collections::HashSet::new(),
+            room_order: Vec::new(),
         }
     }
 
--- a/ui/src/components/room_list.rs
+++ b/ui/src/components/room_list.rs
@@ -25,6 +25,10 @@ use river_core::room_state::privacy::PrivacyMode;
 // Access the build timestamp (ISO 8601 format) environment variable set by build.rs
 const BUILD_TIMESTAMP_ISO: &str = env!("BUILD_TIMESTAMP_ISO", "Build timestamp not set");
 
+// Stable Dioxus key for the reorder tail drop zone (keys must be unique and
+// distinct from any room's `{room_key:?}`).
+const TAIL_DROP_ZONE_KEY: &str = "room-reorder-tail-drop-zone";
+
 /// Convert UTC ISO timestamp to local time string
 fn format_build_time_local() -> String {
     #[cfg(target_arch = "wasm32")]
@@ -67,9 +71,13 @@ pub fn RoomList() -> Element {
 
     // Drag-and-drop reorder state (a local view preference). `dragged_room`
     // is the row currently being dragged; `drag_over_room` is the row the
-    // cursor is hovering over, used only to draw the insertion indicator.
+    // cursor is hovering over, used to draw the insertion indicator;
+    // `drag_over_end` is the same for the tail (move-to-end) drop zone.
+    // All mutations to these are routed through `crate::util::defer()` per the
+    // Dioxus signal-safety rules (this component reads them during render).
     let mut dragged_room = use_signal(|| None::<VerifyingKey>);
     let mut drag_over_room = use_signal(|| None::<VerifyingKey>);
+    let mut drag_over_end = use_signal(|| false);
 
     // Memoize the room list to avoid reading signals during render.
     // Rooms render in the user's drag-chosen order first, then any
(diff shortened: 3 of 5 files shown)


sanity (commented):
## Comprehensive PR Review: #346

### Summary
- **PR Title:** feat(ui): drag-and-drop room reordering; move private-room lock to the right
- **Type:** feat (UI)
- **CI Status:** migrations green; build + ui-playwright-tests running at time of writing (merge gated on green)
- **Linked issues:** none closed; follow-up #348 filed (touch reorder)
- **Review tier:** Full (touches the persisted `rooms_data` delegate blob + a load-bearing always-mounted Dioxus component with documented signal-safety hazards)
- **Reviewers run:** code-first, testing, skeptical, big-picture (all four) + external Codex (gpt-5.5). Three passes total as fixes landed.

### Code-First Analysis
**Independent understanding matches stated intent.** New `Rooms::room_order` (local display order persisted in the `rooms_data` blob via `#[serde(default)]`); `ordered_room_keys()` renders positioned rooms first then a deterministic key-byte tail (also fixes the pre-existing arbitrary `HashMap` render order); `move_room`/`move_room_to_end` materialise + persist the order; drag handlers on each row + a mid-drag tail drop zone; lock icon moved right. **No gaps** — description is faithful, including the no-migration justification.

### Testing Assessment
| Unit | ✅ | `ordered_room_keys` (determinism + stale filter), `move_room` (up/down/self/unknown), `move_room_to_end`, `leave_room` prune, `merge` union/prune, populated ciborium round-trip, pre-247 backward-compat default |
| Integration | ✅ (proxy) | round-trip test exercises the real save/load ciborium shape (verified vs `chat_delegate.rs` + `response_handler.rs`) |
| Simulation | N/A | local view preference, not network/contract state |
| E2E/Playwright | ⚠️ acceptable gap | drag handlers are thin and delegate to tested logic; HTML5-DnD Playwright sim is flaky (would risk the repo's flaky-test debt) |

### Skeptical Findings
**Risk: low.** Two adversarial passes found **no high/medium/low-severity bugs**. Verified: `move_room` insert-before-target off-by-one (correct both directions), `merge` union/prune ordering + bounded growth, ghost/stale filtering (double-guarded), `move_room_to_end` idempotency/no-dupes, `peek()`+`defer()` (no stale/lost drag state), ondrop→ondragend ordering (idempotent), no re-render loop from the reactive tail zone.

### Big Picture Assessment
**Goal-aligned, no scope creep.** No tests/fixes removed (the `map.iter()`→`ordered_room_keys()` swap is functionally equivalent + `filter_map` is belt-and-suspenders). **"No delegate migration needed" independently verified:** `Rooms` is defined only in `ui/`, no delegate/contract crate depends on `river-ui`, the delegate stores the blob as opaque `Vec<u8>`, and the field is additive `serde(default)`. No skills/rules made stale.

### Documentation
- Code docs: complete (field, methods, cross-device local-wins consequence documented).
- No architecture/user docs affected.

### Recommendations
#### Must Fix (Blocking)
None outstanding. (Resolved across passes: Codex P1 — fallible `try_read` for drag-state render reads, now applied.)

#### Should Fix (Important)
All resolved: tail drop zone for the unreachable last slot (Codex P2); deferred drag-state mutations; populated `room_order` round-trip test; `move_room` doc-comment accuracy.

#### Consider (Suggestions)
- **Touch-device reorder (Codex P2):** HTML5 DnD is pointer-only; touch users can't reorder. **Dismissed for this PR with justification** — accepted v1 limitation (not a regression; click-to-open still works), the request was specifically drag-and-drop, and the persistence layer is input-agnostic so a touch path is purely additive. Tracked in **#348**.

### Verdict
**State:** Ready to Merge (once CI is green on the current HEAD).
**HEAD SHA reviewed:** 339cf6d628cf6a0a0acd2610104f101f38a47d42

Re-review after fixes: prior-pass findings (Codex 2×P2 + P1, testing/code-first round-trip gap, doc-comment) are all resolved on this HEAD; final Codex pass confirms the persistence and desktop DnD logic are sound, leaving only the accepted touch-device limitation (#348).

[AI-assisted - Claude]

PR #349, opened 2026-06-04
fix(cli): mark truncated reply_to previews with an ellipsis

sanity:
## Problem

A River user building an XMPP bridge (`ofansifkapital-xmpp`) reported that `riverctl`'s JSON output truncates the `reply_to` preview to 50 characters with **no indication** it was cut. A consumer can't tell whether the quoted message ended naturally or was clipped mid-word.

This affects both JSON-producing paths:
- `riverctl message list --format json` → `reply_to.preview`
- the monitor stream's JSONL → `reply_to.preview`

## Approach

Add a single shared `truncate_reply_preview` helper that appends `"..."` **only when characters were actually dropped** (operating on `char`s, not bytes, so multi-byte/emoji previews never panic or split a codepoint). Both `reply_context` and `reply_context_display` route through it, so the human and JSON renderings share one source of truth for the truncation marker and can't drift.

While here, the two **terminal** renderings previously hardcoded a trailing `...` on every reply prefix — even for short previews that were never truncated (the inverse of the reporter's bug). Now that the preview string carries its own ellipsis only when warranted, the hardcoded marker is dropped, so the terminal shows `...` iff the preview was actually clipped.

### Why `...` (three dots) and not `…`
The reporter offered either. Three ASCII dots match the existing terminal convention, keep the long-reply terminal output byte-identical to before, and avoid any encoding concern for downstream consumers.

### Behaviour
| preview length | before (JSON) | after (JSON) |
|---|---|---|
| ≤ 50 chars | shown whole | shown whole (no marker) |
| > 50 chars | 50 chars, silently cut | 50 chars + `...` |

## Testing

- `reply_context_extracts_author_and_truncated_preview` — clipped preview keeps 50 content chars + `...` (53 total).
- `reply_context_preview_boundaries` — short/empty/exactly-50 get **no** marker; 51-char and 60-emoji inputs get `...`; no codepoint split.
- `reply_context_display_marks_truncation` (new) — pins the display path that feeds **both** JSON outputs: marks when clipped, omits when it fits.
- Existing mention-boundary tests still pass (the marker adds no raw token syntax).
- Full `riverctl` lib suite: 107 passed. `cargo fmt` + `cargo clippy` clean.

CLI-only change — no contract/delegate/common WASM touched, so no migration needed.

[AI-assisted - Claude]

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -154,15 +154,37 @@ pub(crate) fn resolve_outgoing_mentions(room_state: &ChatRoomStateV1, text: &str
     })
 }
 
-/// If `msg` is a reply, return `(target_author_name, truncated_preview)` for
-/// display. Returns `None` for non-reply content (or content that fails to
-/// decode). Mirrors the reply rendering in `riverctl message list` so the
-/// monitor stream surfaces reply context too (it previously did not — a reply
-/// rendered as a plain message). freenet/river — Rogue Worm report.
+/// Truncate a reply preview to at most [`REPLY_PREVIEW_MAX_CHARS`] characters
+/// for display, appending `"..."` **only when characters were actually
+/// dropped**. A preview that fits is shown verbatim; a clipped one carries a
+/// visible marker so a reader (a terminal user or a JSON-consuming bridge) can
+/// tell the quoted text was cut rather than ending there. Operates on `char`s,
+/// not bytes, so a multi-byte / emoji preview never panics or splits a
+/// codepoint.
 ///
-/// Shared with `riverctl message list` (cli/src/commands/message.rs) so the two
-/// renderings can't drift.
-pub(crate) fn reply_context(
+/// Single source of truth for the truncation marker, so the human and JSON
+/// renderings of reply context can't drift on whether/how a preview is clipped.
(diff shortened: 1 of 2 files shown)


sanity:
## Review — Light tier (external model + adversarial read)

**Codex (external, non-Claude):** clean.
> The shared truncation helper correctly appends an ellipsis only when content is dropped, preserves Unicode codepoints, and updates both human and JSON reply rendering consistently. No actionable regressions were identified.

**Adversarial read of `truncate_reply_preview`:** `chars.by_ref().take(50)` advances the iterator exactly 50 times; `chars.next().is_some()` peeks the 51st to detect dropped content. Verified boundaries: empty → no marker; exactly-50 → no marker (no false positive); 51+ → marker. `char`-based, so no codepoint split. In `reply_context_display`, mentions render on the full preview *before* truncation, so the rendered (displayed) length drives the indicator — a long raw preview that renders short correctly gets no marker.

No blocking findings. CLI-only diff (no contract/delegate/common WASM), so no migration. `cargo fmt`/`clippy` clean, full riverctl lib suite green (107 passed).

[AI-assisted - Claude]

PR #352, opened 2026-06-05
fix(ui): recover rooms stranded by an interrupted delegate migration

sanity:
## Problem

Nacho reported that his **"Freenet Devs"** room disappeared from his room list after the per-room delegate update (#345) and he had to rejoin via a previous invite link. The room came back on its own after a few seconds in some cases, but it should never disappear.

**Root cause:** the legacy→per-room migration re-saves rooms one per-room key at a time. If that re-save is cut short — a per-room CAS write fails, or the tab is closed mid-migration — the current delegate is left with a **partial** per-room key set. On the next load, the code takes the `PerRoom` path, loads only the keys that made it across, calls `mark_legacy_migration_done()`, and never re-probes the legacy delegate. Any room whose key wasn't written is stranded until the user rejoins.

## Approach

A persistent `legacy_migration_in_progress` localStorage flag:

- **Set** before a migration's per-room re-save (both the legacy-delegate re-save in `hydrate_loaded_rooms` and the current-blob explosion in `migrate_current_blob_to_per_room`).
- **Cleared** only on a *full* successful re-save (alongside `mark_legacy_migration_done`). On failure it is left set.

The `PerRoom` load branch now reads the flag:

- Flag **not** set → per-room set is authoritative: `mark_legacy_migration_done()` as before (never probe legacy, per #253).
- Flag **set** (a prior migration was interrupted) → do **not** mark done; after loading + hydrating the existing keys, re-run `migrate_current_blob_to_per_room()`, which re-explodes a current `rooms_data` blob if present, else fires the legacy probe — recovering any stranded room.

Because the migration re-save uses CAS read-merge-write, re-migrating an already-present room is a no-op; only genuinely-missing rooms are added. No data is destroyed (the legacy blob and existing per-room keys are left in place).

### Why this is low-risk

- **Clean users**: migration completed → flag cleared, done flag set → `PerRoom` load marks done as before, no recovery fires. Zero behavioural change.
- **Brand-new users**: never have legacy data, never set the flag. Zero cost.
- **Self-healing across sessions**: the flag is keyed per legacy-set generation (same scoping as the existing done flag), so a later delegate bump re-runs migration cleanly anyway.
- **No probe spam**: recovery's `fire_legacy_migration_request` is bounded by the existing per-session `LEGACY_MIGRATION_ATTEMPTED` guard — at most one probe per page load.

## Testing

Source-pin tests, matching this file's existing convention (the load/migration path is too coupled to the Dioxus signal runtime + WebSocket to drive in a unit test):

- `migration_in_progress_key_distinct_from_done_key` — the in-progress and done localStorage keys must be distinct (and both per-legacy-set).
- `interrupted_migration_is_recovered_on_next_load` — pins that the legacy re-save marks in-progress before saving and clears it on success, and that the per-room load gates `mark_done` on the flag and re-runs the migration when one was interrupted.

`cargo test -p river-ui --bins` → 341 passed. `cargo fmt` clean, `cargo clippy` introduces no new warnings.

UI-only change (no `common/`/`delegates/`/`contracts/` edits) → no delegate migration entry needed.

Follow-up to #345.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/river-publish.md
+++ b/.claude/rules/river-publish.md
@@ -108,13 +108,31 @@ where stale legacy data overwrites newer state on the current delegate
 2. `LEGACY_DELEGATES` is generated at compile time from `legacy_delegates.toml`
    by `ui/build.rs`.
 3. The current delegate's `ListResponse` is classified by `plan_load_from_keys`:
-   - **Has `room:<vk>` keys** → current is authoritative: `load_rooms_per_room`
-     plain-GETs each slot (+ `rooms_meta`), reconstructs `Rooms`, hydrates, and
-     calls `mark_legacy_migration_done()`. Legacy probes never fire.
+   - **Has `room:<vk>` keys** → `load_rooms_per_room` plain-GETs each slot
+     (+ `rooms_meta`), reconstructs `Rooms`, and hydrates. What it does about
+     legacy migration is gated by `decide_per_room_load_action` on the
+     **interrupted-migration flag** (`is_legacy_migration_in_progress()`):
+     - *Not interrupted* (the common case) → the per-room set is authoritative:
+       call `mark_legacy_migration_done()`; legacy probes never fire (#253).
+     - *Interrupted* (a prior legacy→per-room migration was cut short before it
+       wrote every key — freenet/river#345 follow-up, Nacho's "Freenet Devs"
+       disappeared-after-update) → do NOT mark done; after loading the partial
+       set, re-run `migrate_current_blob_to_per_room()` once (armed via
+       `arm_legacy_migration_recovery()`, which also clears the per-session
+       attempt guard so same-session recovery isn't a no-op) to recover any
+       stranded room. The re-save is per-room CAS read-merge-write, so an
+       already-present room merges rather than being clobbered.
    - **No per-room keys, only a legacy `rooms_data` blob** →
      `migrate_current_blob_to_per_room` reads it and re-saves as per-room keys
      (non-destructive — the blob is left as a rollback fallback).
    - **Nothing** → `fire_legacy_migration_request()`.
+
+   The **interrupted-migration flag** is a per-legacy-set localStorage marker
+   (`river_legacy_migration_in_progress:<fingerprint>`, parallel to the
+   `…_done:` flag): set BEFORE any migration re-save (`hydrate_loaded_rooms`'s
+   legacy branch and the current-blob explosion alike) and cleared ONLY on a
+   FULL successful re-save. A partial/aborted re-save therefore leaves it set,
+   which is what drives the recovery above.
 4. `fire_legacy_migration_request` probes each legacy delegate TWO ways:
    (a) fixed `GetRequest` for `[ROOMS_STORAGE_KEY, OUTBOUND_DMS_STORAGE_KEY]`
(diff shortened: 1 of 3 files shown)


sanity:
## Multi-model review summary (Full tier — touches delegate migration / data-loss surface)

Ran 4 parallel `freenet:*` reviewers (code-first, skeptical, testing, big-picture) + an external Codex pass on the first commit, then a focused skeptical re-review of the fix delta. All findings addressed:

| Finding | Source | Resolution |
|---|---|---|
| **Same-session recovery is a no-op**: the initial migration this session already sets the per-session `LEGACY_MIGRATION_ATTEMPTED` guard, so the recovery's legacy probe returned immediately — stranded rooms weren't recovered until a full page reload. | Codex P2 | Added `arm_legacy_migration_recovery()`: fires recovery at most once per session **and** clears the attempt guard so the probe actually runs. Bounded by a session-static `RECOVERY_REFIRE_DONE` so reconnects don't re-probe (no error-spam regression). |
| **False-positive regression test**: the source-pin asserted a file-wide `contains("migrate_current_blob_to_per_room().await;")`, also satisfied by the unrelated `LoadPlan::MigrateCurrentBlob` arm — deleting the recovery call still passed. | testing MAJOR | Scoped the assertion to the `if action.recover` block. Verified by mutation: removing the recovery call now fails the test. |
| **Inline decision untestable off-WASM** | testing / skeptical | Extracted `decide_per_room_load_action(bool) -> PerRoomLoadAction` (pure) with a behavioral truth-table test, replacing reliance on the brittle source-pin. |
| **Stale doc**: `river-publish.md` claimed the per-room load unconditionally marks done / "legacy probes never fire". | big-picture MAJOR | Updated "How Delegate Migration Works" to describe the in-progress flag + interrupted-migration recovery branch. |
| **Overclaimed "no-op"** re same-key/different-identity rooms | skeptical M1 | Narrowed the comment — the re-save is a CRDT union for same-identity rooms; the diverged-identity case keeps the local copy per `reconcile_room_present` (pre-existing, not widened). |
| **Deserialize-error asymmetry** in `migrate_current_blob_to_per_room` | skeptical / testing L2 | Documented why the corrupt-current-blob path intentionally does nothing. |
| **Source-pin slices could panic** instead of failing cleanly | skeptical NIT | Clamped both slices with `.min(production.len())`. |

**On cross-session convergence** (reviewers flagged the in-progress flag "never clears" when the legacy delegate is empty/uninstalled): this is the **same** bounded, harmless per-session re-probe the existing empty-legacy path already performs — one fire-and-forget probe per load, no re-save, no data loss. The recovery comment now states this accurately rather than overclaiming one-shot convergence. The skeptical re-review confirmed the delta is clean: no error-spam reopened, the two non-atomic guard ops run without an `.await` between them (atomic under single-threaded WASM), and the decision truth-table is preserved.

**Known out-of-scope** (skeptical M3): a room can also be stranded by a *normal-operation* save failure (CAS exhaustion / `ConnectionClosed`) with no migration involved — most save callers log without retry. That's a distinct, pre-existing root cause not addressed here; flagging it for a separate follow-up.

Tests: `cargo test -p river-ui --bins` → 342 passed. `fmt` clean, `clippy` no new warnings. UI-only change → no delegate-migration entry needed (CI `check-delegate-migration` confirms the WASM is byte-identical).

[AI-assisted - Claude]

Issue #302, opened 2026-05-21
riverctl: support private-room secrets in invitations (CLI follow-up to #301)


sanity:
## Background

PR #301 made the River **UI** embed a private room's symmetric secrets in the `Invitation` artifact (`Invitation::room_secrets`), so a newly-invited member can decrypt the room immediately on join without waiting for the owner's chat-delegate to back-fill an `encrypted_secrets` blob.

`riverctl` was deliberately left out of scope. Its `Invitation` struct (`cli/src/api.rs`) is a separate copy and was not given the `room_secrets` field. CBOR's tolerance of a missing `#[serde(default)]` field keeps old/new invitations interoperable (neither side errors), but:

- A **UI-created** private-room invitation accepted by `riverctl` silently drops `room_secrets` — the CLI invitee gets the pre-#301 behavior (stuck until the owner back-fills).
- A **CLI-created** private-room invitation carries no `room_secrets` at all.

This is not a regression — the CLI cannot decrypt private rooms today regardless (`accept_invitation` in `cli/src/api.rs` seals nicknames as `SealedBytes::public`, and there is no CLI-side `secrets` map / decrypt path).

## Scope

Add real private-room support to `riverctl`:

1. Add `room_secrets: Vec<(u32, [u8; 32])>` (`#[serde(default)]`) to the CLI's `Invitation` struct, kept wire-identical to the UI's.
2. CLI `create_invitation`: populate `room_secrets` from the secrets the CLI holds — for a room the CLI **owns**, derive them from the owner seed via `river_core::key_derivation::derive_room_secret`.
3. CLI `accept_invitation`: decrypt/seed the secrets, seal the invitee nickname properly instead of `SealedBytes::public`, and persist the secrets so private-room content is decryptable.
4. Consider unifying the duplicated `Invitation` struct (UI + CLI) behind a shared, non-WASM-compiled type to prevent drift — note the WASM-migration constraint that kept it out of `common/` for #301.

## References
- PR #301 — "feat(ui): carry private-room secrets in the invitation"
- Flagged by the multi-model review of #301 (Codex, code-first reviewer)

[AI-assisted - Claude]

PR #353, opened 2026-06-09
fix(cli): carry invitation_secrets through IdentityExport for private-room imports

### LINKED ISSUE #306: riverctl: IdentityExport should carry invitation_secrets for non-owner private-room imports
## Background

Follow-up to PR #303 (issue #302), skeptical-review round-2 finding M2.

\`riverctl identity import\` reconstitutes a room from an \`IdentityExport\` bundle. For a non-owner of a PRIVATE room, this currently calls \`Storage::add_room\` (the no-secrets variant) — \`StoredRoomInfo.invitation_secrets\` is left empty.

If the imported network state does NOT yet contain an owner-signed \`encrypted_secrets\` blob addressed to the imported member (delegate back-fill lag, or back-fill was via a different delegate generation), the imported user has no way to recover the secret:

- They cannot decrypt private-room content (already true; CLI has no message-decrypt path).
- They cannot issue invitations carrying useful \`room_secrets\` via \`create_invitation\` — the result will be \`room_secrets: []\`, replicating the PR-#303 bug for any downstream invitee.

## Scope

1. Add a \`#[serde(default)] invitation_secrets: HashMap<u32, [u8; 32]>\` field to \`IdentityExport\` (likely in \`common/src/room_state/identity.rs\` or wherever the export type lives). \`#[serde(default)]\` keeps old exports backward-compatible.
2. \`identity export\` populates the field from \`StoredRoomInfo.invitation_secrets\` for the room.
3. \`identity import\` calls \`Storage::add_room_with_invitation_secrets\` (instead of \`add_room\`) so the persisted map is seeded.
4. Round-trip test: export → wipe storage → import → assert \`get_invitation_secrets\` returns the original map.

## Owner case is fine

\`IdentityExport::room_owner\` carries the signing key, and the owner's blob IS in \`state.secrets.encrypted_secrets\` (because owner-creation always self-addresses, confirmed at \`ui/src/room_data.rs:1232\` and \`common/src/room_state/secret.rs:450-451\`). So the gap is specifically non-owner identities.

## References

- PR #303 (CLI invitation_secrets)
- Issue #302 (parent feature)
- Issue #304 (CLI heal counterpart — natural prune hook)
- \`cli/src/commands/identity.rs:238-243\` (the offending \`add_room\` call site)

[AI-assisted - Claude]

sanity:
## Problem

`riverctl identity import` reconstitutes a room from an `IdentityExport` bundle. For a non-owner of a PRIVATE room it called `Storage::add_room` (the no-secrets variant), so `StoredRoomInfo.invitation_secrets` was left empty after import.

If the imported network state does not yet contain an owner-signed `encrypted_secrets` blob addressed to the imported member (delegate back-fill lag, or back-fill via a different delegate generation), the imported user has no copy of the room secret. The concrete new harm: their `riverctl invitation create` produces an invitation with `room_secrets: []`, silently replicating the bug PR #303 fixed for the normal accept-invitation path. This is a silent secret-loss vector on device migration / backup-restore for private rooms.

The owner case was already fine — the owner's secret is in the owner-addressed contract blob (`common/src/room_state/secret.rs:450-451`), so the gap was specifically non-owner identities.

## Approach

Add a `#[serde(default)] invitation_secrets: HashMap<u32, [u8; 32]>` field to `IdentityExport`:

1. `export_identity` populates it from `StoredRoomInfo.invitation_secrets`.
2. `import_identity` seeds storage via the existing `Storage::add_room_with_invitation_secrets` (added for #302/#303) instead of `add_room`.
3. `#[serde(default)]` keeps old tokens decoding cleanly as an empty map — the same backward-compat shape as the `self_nickname` field (#331).

Empty for public rooms and for owners, so those paths behave exactly as before.

**No delegate/contract migration required.** `IdentityExport` is never referenced by the room-contract or chat-delegate crates, so it is dead-code-eliminated from their WASM (same as the #331 precedent that added `self_nickname`). CI's `check-delegate-migration` / `check-room-contract-migration` confirm the WASM is byte-identical.

## Testing

- `test_invitation_secrets_survive_roundtrip` (common): armored export → decode preserves the secrets map byte-for-byte.
- `test_backward_compat_no_room_name` extended: a pre-#306 token (no `invitation_secrets` key in the CBOR map) decodes with the field defaulting to an empty map.
- `invitation_secrets_survive_export_armor_import_persist` (cli): full export → armor → wipe → decode → storage-seed loop, then `get_invitation_secrets` returns the originals.
- `empty_invitation_secrets_round_trip_to_empty` (cli): public/pre-#306 export persists an empty map (no panic, no invented entries).
- `import_identity_seeds_invitation_secrets_wiring_pinned` (storage): source-grep pin (placed in `storage.rs` so it can't be self-satisfied by test code in the scanned file) guarding the export-population and import-seeding call sites against a future revert to no-secrets `add_room`.

Failing-first verified: removing the export population is a compile error (the new field is a mandatory struct field, not `Option`); reverting import to `add_room` trips the pin's `export.invitation_secrets.clone()` assertion.

Local runs (all green): `cargo test -p river-core --lib` (181), `--features ecies-randomized,migration,mentions` (211), `cargo test -p riverctl --lib` (117), `cargo fmt`, `cargo clippy -p river-core -p riverctl` clean on the changed code.

Closes #306

[AI-assisted - Claude]

### DIFF
--- a/cli/src/commands/identity.rs
+++ b/cli/src/commands/identity.rs
@@ -154,6 +154,11 @@ async fn export_identity(
         // the private-room join-heal sealed `member_info` doesn't lose it on
         // re-import (freenet/river#298).
         self_nickname: room_info.self_nickname.clone(),
+        // Carry the invitation-carried room secrets so a non-owner of a
+        // private room can still forward useful `room_secrets` via
+        // `invitation create` after re-importing on another device
+        // (freenet/river#306). Empty for public rooms and for owners.
+        invitation_secrets: room_info.invitation_secrets.clone(),
     };
 
     let armored = export.to_armored_string();
@@ -224,13 +229,18 @@ async fn import_identity(
             )
         })?;
 
-    // Store the room with the imported identity
+    // Store the room with the imported identity, seeding the persisted
+    // `invitation_secrets` from the export so a non-owner of a private room
+    // keeps the secret across a device migration (freenet/river#306). For a
+    // public room or an export taken before this field existed, the map is
+    // empty and this behaves exactly like the previous `add_room` call.
     let contract_key = api_client.owner_vk_to_contract_key(&export.room_owner);
-    api_client.storage().add_room(
+    api_client.storage().add_room_with_invitation_secrets(
         &export.room_owner,
         &export.signing_key,
         room_state,
         &contract_key,
+        export.invitation_secrets.clone(),
     )?;
 
     // Store the authorized member and invite chain for rejoin support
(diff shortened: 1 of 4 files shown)


sanity:
## Multi-model review (Full tier — wire-format / serialization surface)

**Review path.** External-model attempts: \`codex review\` was down (usage limit, resets Jun 10 ~21:21) and \`gemini-2.5-pro\` returned 429 capacity-exhausted, but **\`gemini-2.5-flash\` ran successfully** — so an independent non-Claude pass did happen. Supplemented with three Claude adversarial lenses (serialization/round-trip compat, skeptical bug-hunt, code-first intent-vs-impl).

### Gemini (gemini-2.5-flash) — independent external pass
All findings INFO severity, no blocking issues. Verbatim summary: the change correctly addresses #306; backward compatibility is explicitly handled for both the armored token and \`rooms.json\`; test coverage is strong (serialization round-trip, end-to-end import/export persistence, and backward-compat). It confirmed the \`#[serde(default)]\` annotation, the export-population and import-seeding call sites, and the pinned wiring test. Nothing to fix.

### Claude adversarial lenses
- **Serialization / round-trip compat:** ciborium encodes this struct as a named map (proven by \`test_backward_compat_no_room_name\`, which omits three trailing fields yet still decodes \`member_info\` at the correct position — a positional-array encoding would misalign). So \`#[serde(default)]\` makes old tokens decode with an empty map. Matches the \`room_name\`/\`self_nickname\` precedent. Clear.
- **Skeptical bug-hunt:** import calls \`add_room_with_invitation_secrets\` **before** \`store_authorized_member\`; the latter uses \`get_mut\` and only sets \`self_authorized_member\`/\`invite_chain\`, so it does not clobber the just-seeded secrets. No drop/clobber path. Clear.
- **Code-first intent-vs-impl:** all four steps from the issue implemented as specified (field + serde default, export populates from \`StoredRoomInfo.invitation_secrets\`, import seeds via the secrets-carrying variant, round-trip test). Clear.

### No delegate/contract migration required
\`IdentityExport\` is not referenced by the room-contract or chat-delegate crates (verified by grep), so it is dead-code-eliminated from their WASM — same as the #331 \`self_nickname\` precedent, which also shipped without a migration entry. CI's \`check-delegate-migration\` / \`check-room-contract-migration\` will confirm the WASM is byte-identical.

**Outcome:** no blocking findings from any reviewer; all addressed. Awaiting CI green on HEAD before merge.

[AI-assisted - Claude]

sanity:
## Update: UI consumer fix + re-review

The first push broke CI: the UI (`ui/src/components/members.rs`) is the **other** consumer of the `IdentityExport` wire type, and the new mandatory field caused `E0063: missing field invitation_secrets` in the UI build. While fixing the construction site I found the UI import path was **also** dropping the secrets (`invitation_secrets: HashMap::new()`) — the exact #306 gap, mirrored on the UI side. Per finish-the-fix, both consumers now carry the secrets end-to-end:

- `ExportIdentityModal` populates `IdentityExport.invitation_secrets` from `RoomData.invitation_secrets`.
- `ImportIdentityModal` restores `RoomData.invitation_secrets` from the export (was an empty map); `repopulate_secrets_from_state` folds it into the runtime `secrets` map on the next sync.

**Re-review (code changed since the prior review — prior review now stale).** External models remain unavailable (Codex usage-limit; Gemini-2.5-pro 429-exhausted). The earlier Gemini-2.5-flash external pass covered the common/CLI wire change, which is unchanged. The new UI commit is additive and mirrors the already-reviewed CLI shape. Adversarial read of the UI delta (failure-mode lens = the broken consumer):
- Export: `room_data.invitation_secrets.clone()` — correct field/type.
- Import: `export.invitation_secrets` is the last move out of `export`; no use-after-move (compiles).
- Semantics: imported secrets land in `RoomData.invitation_secrets` and are folded into the `#[serde(skip)]` runtime `secrets` map by `repopulate_secrets_from_state`, matching the documented UI mechanism and the CLI behavior.

No WASM change (migration CI checks already passed and `cargo make build` left WASMs byte-identical). Verified `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` compiles.

[AI-assisted - Claude]

PR #350, opened 2026-06-05
feat(cli): send encrypted messages to private rooms

sanity:
## Problem

riverctl's `message send` always built a **public** `RoomMessageBody`, so posting to a **private** room is rejected by the room contract:

```
Error: Failed to apply message delta: "Cannot send public messages in private room"
```

This makes riverctl unusable for any private room. Concretely, it blocks setting up a CI/notification bot in the new private `freenet-dev` room (Nacho's request: CI-failure / nightly / PR-merge announcements, like the old Matrix bot).

## Approach

Mirror the UI's proven send path (`ui/src/components/conversation.rs`): for a private room, CBOR-encode the text content, seal it under the room's **current-version** secret with AES-256-GCM, and emit `RoomMessageBody::private`.

The room secret is resolved by the **same** logic riverctl already uses to seal invitee nicknames (`seal_invitee_nickname`):

1. the member's own owner-signed `encrypted_secrets` blob in contract state (authoritative), else
2. the secret carried in the `Invitation` this member joined with (`invitation_secrets`).

`collect_secrets_for_room` already folds both sources; the new `build_message_body` helper picks `current_version` from it. If no secret is available for the current version, it **errors** rather than silently sending a public body the contract would reject (and rather than sealing under a stale version other members can't read).

Wired into both send paths: `send_message` (storage identity) and `send_message_with_key` (explicit `--signing-key`, which is how a stateless CI runner posts).

**Out of scope (unchanged):** `edit` / `delete` / reactions / `reply` still emit public bodies — they use different content types and aren't needed here. They were already non-functional in private rooms (riverctl never supported private-room sending at all); a follow-up issue will track adding sealed action/reply bodies.

## Testing

New unit tests in `cli/src/private_room.rs`:
- public room → plaintext body (unchanged)
- private room, invitation-carried secret → sealed body; AES-GCM **roundtrip** decodes back to the original text
- private room, contract-blob secret → sealed body; roundtrip
- private room, no secret → error (never falls back to a public body)
- private room rotated past the held version → error (won't seal under a non-current version)

`cargo test -p riverctl --lib private_room` — 22 passed. `cargo fmt` + `cargo clippy` clean.

[...]

PR #354, opened 2026-06-09
feat(cli): seal edit/delete/reaction/reply bodies in private rooms

### LINKED ISSUE #351: riverctl: seal edit/delete/reaction/reply bodies in private rooms
## Problem

PR #350 added private-room (encrypted) sending for plain `riverctl message send` via `cli::private_room::build_message_body`. The other message paths still build **public** `RoomMessageBody` values unconditionally, so they fail in a private room with `"Cannot send public messages in private room"`:

- `message reply`  — `ApiClient::send_reply` (`cli/src/api.rs`)
- `message edit`   — `ApiClient::edit_message`
- `message delete` — `ApiClient::delete_message`
- reactions        — `ApiClient::add_reaction` / `remove_reaction`

These were never functional in private rooms (riverctl had no private-send path at all before #350), so #350 introduces no regression — but now that plain text send works, the asymmetry is a latent surprise.

## Approach

Mirror the UI's private action/reply sealing. The UI seals action and reply content with `RoomMessageBody::private_action` and the private reply path in `ui/src/components/conversation.rs`. The CLI already has all the pieces:

- `build_message_body` establishes the secret-resolution pattern (current-version secret from contract blob or invitation secret).
- `RoomMessageBody::private_action` (`common/src/room_state/message.rs`) is the action analog of `private_text`.
- `ReplyContentV1` / `CONTENT_TYPE_REPLY` for the reply body.

Factor the secret-resolution out of `build_message_body` (or add `build_action_body` / `build_reply_body` siblings) and route the five paths through it, with the same "error rather than public-fallback / over-size" guards.

## Notes

- Distinct from #304 (CLI `build_member_info_heal` / "Unknown" member rendering) — that's the member_info path, this is message action/reply bodies.
- Not needed for the freenet-dev CI-notification bot (it only posts plain text), so this is lower priority than #350 but should be tracked so the gap isn't rediscovered as a bug.

Surfaced by the multi-model review of #350.

[AI-assisted - Claude]

sanity:
## Problem

PR #350 added private-room (encrypted) sealing for plain `riverctl message send` via `cli::private_room::build_message_body`. The other message paths still built **public** `RoomMessageBody` values unconditionally, so they fail in a private room with `"Cannot send public messages in private room"`:

- `message reply`  — `ApiClient::send_reply`
- `message edit`   — `ApiClient::edit_message`
- `message delete` — `ApiClient::delete_message`
- reactions        — `ApiClient::add_reaction` / `remove_reaction`

These were never functional in private rooms, so this introduces no regression — but now that plain text send works, the asymmetry is a latent surprise (and blocks full private-room use of riverctl).

## Approach

Mirror the #350 send-sealing pattern for actions and replies, factoring the shared logic so all four message kinds (text / action / reply) make the **identical** privacy decision — a single secret-resolution site means no kind can leak as an unsealed public body.

- Extract `resolve_current_secret` (current-version secret; error rather than public fallback; never seal under a stale version) and `guard_message_size` from `build_message_body`.
- Add `build_action_body` (edit/delete/reaction/remove_reaction → CBOR-encode `ActionContentV1`, AES-256-GCM seal, `RoomMessageBody::private_action`) and `build_reply_body` (reply → `ReplyContentV1`, sealed `private` body with `CONTENT_TYPE_REPLY`), mirroring the UI's paths in `ui/src/components/conversation.rs`.
- Route the five `ApiClient` action/reply call sites through the helpers.
- Public-room behaviour is byte-identical to the dedicated `RoomMessageBody::{edit,delete,reaction,remove_reaction,reply}` constructors (pinned by tests). Private-room: error rather than public fallback, never seal under a stale version, error on over-`max_message_size` (the contract otherwise silently drops it).
- For a private reply, the target author name and content preview are sealed alongside the text (they are `ReplyContentV1` fields), so the reply context is **not** leaked in the clear.

Confined to the `cli/` (riverctl) crate plus a doc update — no `common/`/contract/delegate WASM change, so **no migration entry is required** (`git status` shows no `.wasm` changes).

## Testing

New unit tests in `cli/src/private_room.rs` (riverctl lib: **126 passed**):
- public room → byte-identical to the dedicated constructor (action + reply)
- private room → seals + AES-GCM roundtrip for all four action kinds and reply (reply test asserts author name / preview are sealed, not leaked)
- private room, contract-blob secret → seals
- private room, no secret → error (never a public body)
- private room rotated past the held version → error
- over-`max_message_size` → error (not silent drop)
- end-to-end contract `apply_delta` acceptance for the sealed action + reply

Verified **failing-first** per the project test philosophy: the seal/roundtrip/error tests fail when the action path emits a public body (simulating the pre-fix bug), and pass with the fix.

`cargo fmt` + `cargo clippy -p riverctl --all-targets` clean.

Closes #351

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/private-rooms.md
+++ b/.claude/rules/private-rooms.md
@@ -147,28 +147,35 @@ version (and prunes it from `invitation_secrets`).
 
 riverctl carries the same `Invitation::room_secrets` wire shape as the UI,
 with `cli::private_room::{collect_secrets_for_room, collect_invitation_secrets,
-seal_invitee_nickname, current_secret_from_state, build_message_body}` as the
-byte-identical CLI counterparts. `StoredRoomInfo::invitation_secrets` (a
-`HashMap<u32, [u8; 32]>` in `rooms.json`) persists the invitation-carried
-secrets across CLI invocations.
+seal_invitee_nickname, current_secret_from_state, build_message_body,
+build_action_body, build_reply_body}` as the byte-identical CLI counterparts.
+`StoredRoomInfo::invitation_secrets` (a `HashMap<u32, [u8; 32]>` in
+`rooms.json`) persists the invitation-carried secrets across CLI invocations.
 
-**Sending (issue #350).** `cli::private_room::build_message_body` is the CLI
-analog of the UI's text-send path in `ui/src/components/conversation.rs`:
-for a private room it CBOR-encodes the text, seals it under the
-current-version secret with `encrypt_with_symmetric_key`, and emits
-`RoomMessageBody::private_text`. Wired into both `ApiClient::send_message`
-(storage identity) and `send_message_with_key` (explicit `--signing-key`,
-the stateless CI-runner path), so `riverctl message send` now works in a
-private room. **Deliberate divergence from the UI:** when no secret is
-available for the current version the UI logs a warning and falls back to a
-**public** body (`conversation.rs`, which the contract then rejects);
-`build_message_body` instead **errors** — never sending a public body into a
-private room, and never sealing under a stale (non-current) version. It also
-errors on an over-`max_message_size` body, which the contract would otherwise
-silently drop. Only plain `message send` is sealed; `message reply` / `edit`
-/ `delete` / reactions still emit public bodies and therefore still fail in a
-private room (they were never functional there — tracked separately for
-sealed action/reply bodies).
+**Sending (issues #350, #351).** `cli::private_room::build_message_body`
+(text), `build_action_body` (edit / delete / reaction / remove_reaction), and
+`build_reply_body` (reply) are the CLI analogs of the UI's send paths in
+`ui/src/components/conversation.rs`: for a private room each CBOR-encodes its
+content and seals it under the current-version secret with
+`encrypt_with_symmetric_key`, emitting `RoomMessageBody::private_text` /
+`private_action` / `private` (with `CONTENT_TYPE_REPLY`) respectively. The text
+helper is wired into both `ApiClient::send_message` (storage identity) and
+`send_message_with_key` (explicit `--signing-key`, the stateless CI-runner
+path); the action/reply helpers are wired into `ApiClient::{edit_message,
+delete_message, add_reaction, remove_reaction, send_reply}` — so `riverctl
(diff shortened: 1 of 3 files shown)


sanity:
## Multi-model review (Full tier — privacy-sensitive sealing)

**External-model attempts:**
- `codex review --base main` (gpt-5.5): **usage limit hit** (resets Jun 10 ~21:21). Could not run.
- Gemini `gemini-2.5-pro`: **capacity exhausted** (429, all 10 retries failed).
- Gemini `gemini-2.5-flash`: **ran successfully** — independent non-Claude pass. Found **no issues** at any severity; specifically confirmed all four action kinds + reply are sealed in private rooms, the reply context (author name / preview) is sealed not leaked, public-room behavior is unchanged, and the test suite (incl. the end-to-end `apply_delta` acceptance test) is comprehensive.

Because the two stronger external models were down and this is a Full-tier (privacy) change, I added the **Claude-subagent fallback** with three distinct adversarial lenses on the checked-out code (run serially — this is a background agent that cannot spawn subagents):

1. **Skeptical bug-hunt** — verified the sealing math and that `guard_message_size` measures `content.content_len()` (= ciphertext length for private bodies), which **exactly matches** the contract's own drop check at `common/src/room_state/message.rs:204` (`content_len() <= max_message_size`). The guard therefore matches what the contract would silently drop. No bug.
2. **Code-first intent-vs-impl** — confirmed private actions are emitted with `CONTENT_TYPE_ACTION` (`private_action`) and private replies with `CONTENT_TYPE_REPLY` via `RoomMessageBody::private(...)`, byte-for-byte matching the UI's decode/encode paths in `ui/src/components/conversation.rs`. Public-room bodies are byte-identical to the dedicated constructors (pinned by `*_matches_dedicated_constructor` tests).
3. **The failure mode for this change (privacy leak)** — grepped every non-test `RoomMessageBody` construction in `cli/`. The only production message-body paths are the 5 action/reply sites wired here plus the 2 text-send sites #350 already wired; the DM path uses separate ECIES (`compose_direct_message`) and is unrelated. **No path still emits a public/plaintext body in a private room.**

**Outcome:** clean. No blocking findings; nothing to fix.

[AI-assisted - Claude]

PR #355, opened 2026-06-09
test(ui): tighten connection-indicator regression tests (#274)

### LINKED ISSUE #274: test(ui): tighten connection-indicator regression tests (follow-up to #273)
Follow-ups from the multi-model review of PR #273 (Bug #5: WebSocket connection indicator on no-rooms screen). All Important / Suggestion level — non-blocking, but worth picking up.

### Important

1. **Assert exactly ONE pill is visible at each viewport.** Both placements (`RoomList` always-visible + `Conversation` no-room inline) share `data-testid="connection-status-indicator"`. The current spec uses `:visible` + `.first()`, which masks the regression where `md:hidden` accidentally inverts (or the Tailwind v4 `@source "../src/**/*.rs"` directive gets dropped — see AGENTS.md). Add `await expect(page.locator(PILL_TESTID)).toHaveCount(1)` at desktop (1280px) AND mobile (375x812).

2. **State-transition coverage.** All four current tests assert "Connecting..." at initial paint. A regression that replaces `try_read()` with `peek()` (no subscription at all per Dioxus 0.7.x) would leave the pill stuck and not be caught. Either assert the dot's bg-color class so a stuck-state regression is detectable, or drive a state transition via test fixture.

### Suggestion

3. Tighten `waitForApp` selector — `expect(page.locator(\"aside, .app-root button\")).not.toHaveCount(0)` matches any button. Should wait for the rail itself.

4. Run the new spec across `webkit` and `mobile-safari` projects too — Firefox-mobile is where the `try_read()` rule originated.

5. Consider a `data-testid=\"rooms-rail\"` instead of `hasText: 'Rooms'` for the rail-membership assertion — text changes break it brittle-ly.

## Context

- Original bug report: Ivvor on Matrix 2026-05-17 — "I had no connection indicator in the bottom right to signify my setup wasn't fully connected to the node."
- Fixed in PR #273 (merged 2026-05-17).
- These follow-ups don't gate behavior, just harden the regression test surface.

[AI-assisted - Claude]

sanity:
## Problem

The connection-indicator regression suite (`ui/tests/connection-status-indicator.spec.ts`, added in PR #273 for Bug #5 — Ivvor's "no connection indicator on the no-rooms screen") had four review follow-ups (#274) that left it able to silently pass through real regressions:

1. It asserted `.first()` of the `:visible` pill set. Both pill placements share `data-testid="connection-status-indicator"`, so `.first()` tolerated an `md:hidden` inversion (or a dropped Tailwind v4 `@source "../src/**/*.rs"` directive) that left BOTH copies visible — exactly the class of regression that would re-break the new-user experience.
2. All tests only asserted the indicator was present. A regression that left the pill stuck/blank (e.g. swapping the reactive `try_read()` for a non-subscribing `peek()`, per the Dioxus 0.7.x signal-safety rule) would not be caught.
3. `waitForApp` waited on `"aside, .app-root button"`, which matches any button anywhere in the app — it could resolve before the rail that carries the indicator rendered.
4. Rail-membership assertions keyed off the brittle visible text `"Rooms"` / `"Active Members"`.

## Approach

Test-only hardening, plus two `data-testid` attributes on existing UI source for stable hooks:

- **Exact visible count (item 1):** assert `expect(VISIBLE_PILL).toHaveCount(1)` at desktop (1280px) and mobile (375x812), and `expect(PILL).toHaveCount(2)` for the DOM (both placements present, one shown one `md:hidden`). Catches both the "both visible" inversion and the "placement dropped" regression that `.first()` masked.
- **Coherent-state check (item 2):** the visible pill's dot bg-colour and label text must form a valid `SynchronizerStatus` pair (green↔Connected, yellow↔Connecting, red↔Disconnected/Error). A stuck/blank pill fails. Chosen over driving a state transition because the `no-sync` build exposes no state-injection fixture; this still detects the stuck-state regression the issue names.
- **Tighter `waitForApp` (item 3):** anchor on the new `rooms-rail` testid.
- **Cross-browser (item 4):** the spec already runs across all five Playwright projects (chromium, firefox, webkit, mobile-chrome, mobile-safari) via `testMatch: "*.spec.ts"` with no per-project filtering — confirmed by running the suite (30 tests = 6 × 5 projects, all green). No config change needed.
- **Stable testids (item 5):** add `data-testid="rooms-rail"` to the `RoomList` `aside` and `data-testid="members-rail"` to the `MemberList` `aside`; rail assertions now anchor on these instead of visible text. These are UI-WASM-only source attributes — no contract/delegate WASM is affected, so no migration is involved.

## Testing

- Built locally: `cargo make build-tailwind` + `cargo make build-ui-example-no-sync`, served the built static dir with `python3 -m http.server 8082`.
- `npx playwright test connection-status-indicator`: **30 passed** (6 tests × chromium/firefox/webkit/mobile-chrome/mobile-safari).
- Full suite (`npx playwright test`): **251 passed, 3 skipped** — the testid additions don't disturb any existing layout test. (One pre-existing `message-layout.spec.ts` mobile-safari test was flaky on a loaded machine and passed on retry; it uses the old loose `waitForApp` selector and is untouched here.)
- **Non-vacuity proof:** injected a CSS override forcing both pills visible (simulating the `md:hidden` inversion). The old `.first()` assertion still passed (the masked bug) while the new `toHaveCount(1)` assertion correctly failed — confirming the tightened assertion catches the regression class #274 describes.
- Empirically confirmed the rendered DOM: desktop/mobile each have 2 pills in the DOM, exactly 1 visible; in the `no-sync` build the state is `Error: WebSocket error...` with a `bg-red-500` dot, which the coherent-state check validates.

Closes #274

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -374,7 +374,14 @@ pub fn MemberList() -> Element {
     }
 
     rsx! {
-        aside { class: "w-full md:w-56 flex-shrink-0 bg-panel border-l border-border flex flex-col",
+        aside {
+            // Stable hook for the connection-indicator regression tests
+            // (freenet/river#274): the members rail is the PRE-FIX location
+            // of the connection pill (Bug #5). Tests assert this rail
+            // carries no indicator, anchoring on the testid instead of the
+            // brittle visible text "Active Members".
+            "data-testid": "members-rail",
+            class: "w-full md:w-56 flex-shrink-0 bg-panel border-l border-border flex flex-col",
             // Header
             div { class: "px-4 py-3 border-b border-border flex-shrink-0",
                 div { class: "flex items-center gap-2",
--- a/ui/src/components/room_list.rs
+++ b/ui/src/components/room_list.rs
@@ -139,7 +139,14 @@ pub fn RoomList() -> Element {
     let drag_over_end_now: bool = drag_over_end.try_read().map(|g| *g).unwrap_or(false);
 
     rsx! {
-        aside { class: "w-full md:w-64 flex-shrink-0 bg-panel border-r border-border flex flex-col overflow-y-auto",
+        aside {
+            // Stable hook for the connection-indicator regression tests
+            // (freenet/river#274): the rooms rail is the always-rendered
+            // left column that carries the persistent connection pill, so
+            // tests anchor on this testid instead of the brittle visible
+            // text "Rooms".
+            "data-testid": "rooms-rail",
+            class: "w-full md:w-64 flex-shrink-0 bg-panel border-r border-border flex flex-col overflow-y-auto",
             // Mobile back button (hidden on desktop)
             div { class: "md:hidden flex items-center px-3 py-2 border-b border-border flex-shrink-0",
                 button {
--- a/ui/tests/connection-status-indicator.spec.ts
+++ b/ui/tests/connection-status-indicator.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect, Page } from "@playwright/test";
+import { test, expect, Locator, Page } from "@playwright/test";
 
 // Regression tests for Bug #5 (Ivvor, Matrix 2026-05-17): the WebSocket
 // connection indicator must remain visible when the user has no rooms —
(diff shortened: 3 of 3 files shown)


sanity:
## Review (Light tier — test-only + two data-testid attributes)

**External-model pass:**
- `codex review --base main` — **unavailable** (OpenAI usage limit hit; resets Jun 10 21:21, far off).
- `gemini --approval-mode plan` — **ran successfully** (independent non-Claude pass). No blocking findings; confirmed the `.first()` → `toHaveCount(1)` change is non-vacuous, the `bg-red-500` Disconnected/Error colour-sharing is handled correctly, and the `data-testid` additions are side-effect-free. Two recommendations:
  1. *Add a selected-room case* so re-introducing the indicator into the (now-rendered) `MemberList` can't add a second desktop pill unnoticed — **addressed** in commit 5e6b72da (`still exactly one indicator when a room IS selected`).
  2. *Colour regex robustness* — noted safe for the current codebase; no change needed.

**Self-review lenses (skeptical / code-first / flakiness):**
- *Non-vacuity proof:* injected a CSS override forcing both pills visible (simulating the `md:hidden` inversion). The old `.first()` assertion still passed (the bug it masked); the new `toHaveCount(1)` correctly failed. The tightened assertion genuinely catches the regression class #274 names.
- *Code-first note:* the issue's literal snippet was `page.locator(PILL_TESTID).toHaveCount(1)`, but the DOM legitimately holds **two** placements (the rooms-rail copy + the `md:hidden` Welcome-screen copy), so a literal all-pills count of 1 would be wrong. Implemented the issue's stated intent — "exactly ONE pill *visible* at each viewport" — as `VISIBLE_PILL` count = 1, plus a `PILL` DOM count = 2 in the no-room states to guard against a dropped placement.
- *Flakiness:* hardened `expectCoherentState` to wait for label text before the non-retrying reads. Spec is green across all five Playwright projects (chromium / firefox / webkit / mobile-chrome / mobile-safari) over three consecutive runs (35/35 each).

**Item 4 (cross-browser):** no config change needed — `playwright.config.ts` already defines `webkit` + `mobile-safari` and uses `testMatch: "*.spec.ts"` with no per-project filtering, so every project runs this spec.

Local result: `npx playwright test connection-status-indicator` → **35 passed**; full suite `npx playwright test` → **251 passed, 3 skipped** (one pre-existing `message-layout.spec.ts` mobile-safari test was flaky on a loaded machine and passed on retry; it's untouched here).

[AI-assisted - Claude]

Issue #279, opened 2026-05-17
fix(invite): in-app accept re-shows full nickname flow for previously-accepted-then-left rooms


sanity:
Follow-up from PR #278 skeptical review (Important #1).

## Problem

`present_invitation()` writes to `PRESENT_INVITATION_REQUEST` and the `App` bridge effect routes it into `receive_invitation` — opening `ReceiveInvitationModal` with the full nickname-prompt accept flow. The URL-bar path and click-interceptor path both gate on `is_invitation_processed(&inv.to_encoded_string())`. The new in-app bridge does NOT.

For a previously-accepted-then-`LeaveRoom`'d room, clicking Accept on a stale invite card re-presents the full accept flow instead of recognising "you've already acted on this fingerprint." Mitigated when the room is still in `ROOMS` (button shows "Already a member" / disabled), but `LeaveRoom` removes from ROOMS while leaving the processed-hash entry.

## Suggested fix

Gate the bridge effect in `ui/src/components/app.rs:244` on `is_invitation_processed(&inv.to_encoded_string())`. If already processed, render a small toast ("you've already acted on this invite") instead of opening the modal.

## Severity

Important UX, not a correctness bug. Recovery: accept the invite again proceeds to the standard already-a-member or pending-invite path. But the experience is confusing.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-05-26T16:04:35.385089021+00:00
Issue hash: `eed869cd`

## Value Assessment

The fix described in this issue has already been implemented. The `is_invitation_processed` gate is present at line 270 of `app.rs` — exactly where the issue's suggested fix recommended — and there's even a regression test at line 508 that pins the gate so it can't be removed without failing CI. The code comments at lines 245–248 and 491–500 explicitly reference issue #279, confirming this was addressed as part of or shortly after PR #278's skeptical review.

**This issue is obsolete and should be closed.** The described bug — `PRESENT_INVITATION_REQUEST` bridge effect failing to check `is_invitation_processed`, causing the full nickname-prompt flow to re-open for previously-accepted-then-left rooms — is fixed. All three invitation entry paths (URL bar, click interceptor, and DM-thread bridge) now consistently gate on `is_invitation_processed`, and the specific regression guard test `dm_accept_bridge_gates_on_is_invitation_processed` at line 508 ensures it stays fixed.

[...]

PR #356, opened 2026-06-10
fix(ui): keep a failed invite-accept retryable in the sandboxed gateway iframe

sanity:
## Problem

In the deployed Freenet gateway the River UI runs inside an
`<iframe sandbox="allow-scripts allow-forms allow-popups">` with **no** `allow-same-origin`. In that opaque origin `window.localStorage` throws `SecurityError` (#219), which is why River records "this invitation was acted on" in the top-level URL hash (`#river-processed=...`) rather than localStorage.

`accept_invitation` marked the invitation processed **up front** — before the room-contract GET/subscribe completed. When that GET failed to complete (intermittently — triggered by **freenet-core #4345**'s large multi-fragment GET), the invite was **permanently suppressed on reload**:

- The URL-bar gate saw `is_invitation_processed == true` and skipped the modal.
- The localStorage-based auto-resume (#218) is dead in the sandboxed iframe (`SecurityError`).

Net effect: a join that was **accepted but never completed** became un-retryable forever — the user was stuck unless they hand-edited the `#river-processed=...` value out of the top-level URL. Real users hit exactly this on the official Freenet River room (reported via Matrix). The underlying GET failure is **freenet-core #4345**.

## Approach

Move `mark_invitation_processed` out of `accept_invitation` (up front) and into `render_subscribed_state` (**terminal success**). Dismiss still marks via `dismiss_invitation_persistently`. The processed-set now records only **terminal** outcomes — a successful join, or an explicit decline/cancel/close — never a mid-flight accept:

| Outcome | Marked processed? | Reload behavior |
|---|---|---|
| Accepted, **join never completed** | No (the fix) | **Re-surfaces** → user can retry |
| Accepted, **join succeeded** | Yes (`render_subscribed_state`) | Suppressed synchronously — no nickname re-prompt (#215 / #216) |
| Accepted, joined, then **left** | Yes (success mark written before leaving; leaving never clears it) | Suppressed (#279) |
| **Declined / cancelled / closed** | Yes (dismiss) | Suppressed (#279) |

The mark lands on the **synchronous, iframe-safe top-level URL hash**, so the reload decision has no dependency on the chat-delegate `ROOMS` hydration finishing first. (An earlier `ROOMS`-membership-based approach was rejected because it flashed a nickname prompt for every already-joined reload during the async hydration window — a #215/#216 regression.) The #218 dev-mode localStorage auto-resume (`decide_recovered_invitation`, which keys on a saved nickname, not on the processed flag) is unaffected.

This is a **UI-only** change — one file, `ui/src/components/room_list/receive_invitation_modal.rs`. It does **not** touch delegate/contract/common WASM, so no delegate migration is needed. No `.wasm` files are in the commit.

## Testing

- `cargo test -p river-ui --bins` → **348 pass**.
- New tests in the `receive_invitation_modal` test module cover the four required outcomes by driving the host-testable processed-set (`PROCESSED_CACHE`): `accepted_but_incomplete_join_is_not_suppressed`, `successful_join_suppresses_on_reload`, `dismissed_invite_suppresses_on_reload`, `accepted_then_left_stays_suppressed_on_reload`.
- Two source-grep **pin** tests guard the relocation so a future refactor can't silently re-introduce the gravestone: `accept_invitation_does_not_mark_processed_up_front` and `render_subscribed_state_marks_processed`. **Both pins fail on current `main`** (main marks on accept and not on success), confirming the before/after.
- `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` clean (the one remaining warning — an unused `SynchronizerMessage` import in `app.rs` — predates this change).
- `cargo fmt --check` clean.

## Review note

The external-model review pass (per the repo's multi-model-review policy) could **not** run: `codex` hit its usage limit (resets ~21:21 CT) and `gemini` returned `QUOTA_EXHAUSTED` (resets in ~2h) — both genuinely unavailable at review time. Because this PR is **not** being merged here (handed back for human review), I did not substitute the full Claude-subagent fallback; I did perform a structured multi-lens self-review (dev-mode #218/#333 resume, #279 accepted-then-left, #215/#216 already-joined reload, signal-safety, terminal-success reachability). Recommend a real external pass before merge.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -317,12 +317,13 @@ pub fn App() -> Element {
     //   1. A nickname is ALSO saved → the user already clicked Accept and the
     //      subscription was still in flight at reload. Auto-resume with that
     //      nickname (#218) so the "Subscribing…" indicator returns instead of
-    //      a blank UI / re-prompt. This MUST take precedence over the
-    //      processed-fingerprint check below, because `accept_invitation`
-    //      marks the invitation processed up front — so a mid-flight reload
-    //      always sees `is_invitation_processed == true`. Storage-still-present
-    //      + nickname-present is the authoritative "join not yet finished"
-    //      signal here.
+    //      a blank UI / re-prompt. The saved-nickname signal — not the
+    //      processed flag — is what drives this: `accept_invitation` does NOT
+    //      mark the invitation processed (the mark happens only at terminal
+    //      success in `render_subscribed_state`, or on dismiss), so a
+    //      mid-flight reload sees `is_invitation_processed == false`.
+    //      Storage-still-present + nickname-present is the authoritative
+    //      "accepted but join not yet finished" signal here.
     //   2. No nickname, but the invitation was already acted on in this
     //      browser → discard (the user accepted-then-left, or dismissed).
     //   3. No nickname, not yet processed → the user reloaded before deciding;
--- a/ui/src/components/room_list/receive_invitation_modal.rs
+++ b/ui/src/components/room_list/receive_invitation_modal.rs
@@ -5,7 +5,6 @@ use crate::invites::{PendingRoomJoin, PendingRoomStatus};
 use crate::room_data::Rooms;
 use dioxus::logger::tracing::{error, info, warn};
 use dioxus::prelude::*;
-use ed25519_dalek::VerifyingKey;
 use river_core::room_state::member::MemberId;
 use std::cell::RefCell;
 #[cfg(target_arch = "wasm32")]
@@ -529,7 +528,7 @@ fn render_invitation_content(inv: Invitation, invitation: Signal<Option<Invitati
         Some(PendingRoomStatus::Error(e)) => render_error_state(&e, &inv, invitation),
         Some(PendingRoomStatus::Subscribed) => {
             // Room subscribed and retrieved successfully, close modal
-            render_subscribed_state(&inv.room, invitation)
+            render_subscribed_state(&inv, invitation)
         }
         None => render_invitation_options(inv, invitation),
     }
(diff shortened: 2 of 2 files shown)


sanity:
## Consolidated review — multi-perspective (Claude-subagent fallback)

**Risk tier:** Light-plus. UI-only, single-subsystem, but it touches the invitation state machine, which has a long regression history (#215/#216/#219/#220/#279/#291/#333), so it got a wider perspective sweep.

### External-model availability (per the multi-model-review rule)
The independent non-Claude pass could not run: both external reviewers were genuinely down at review time —
- **Codex**: `You've hit your usage limit … try again at 9:21 PM` (CT).
- **Gemini**: `429 QUOTA_EXHAUSTED … reset after ~1h38m`.

Both were probed directly (not assumed). Since waiting wasn't practical, I used the rule's prescribed fallback: **three independent Claude subagents with distinct adversarial lenses**, each reviewing the checked-out code and blind to the others —
1. **Skeptical bug-hunt** — render-time side-effect safety, optimistic-`Subscribed` gravestone risk, missed terminal-success paths, close-before-mark race.
2. **Regression + signal-safety** — #215/#216 (nickname re-prompt), #279 (declined / accepted-then-left), #218/#333 (dev-mode resume), #291, dioxus-signal-safety.
3. **Code-first** — intent-vs-impl, signature change, fingerprint-basis consistency.

When the external models reset, a real Codex pass on the final content is worth running before the next release; flagging for the record.

### Findings & resolution
- **[fixed]** Three stale comments still claimed `accept_invitation` "marks processed up front" (removed by this PR) → corrected in `590372ff` (doc-only).
- **[verified clean]** `mark_invitation_processed` in `render_subscribed_state` touches only the thread-local `PROCESSED_CACHE` + the URL-hash postMessage (no Dioxus signal write), so it's safe in the render body; `append_fingerprint` dedups so repeated renders don't spam the hash.
- **[verified clean]** `PendingRoomStatus::Subscribed` is reached only after the room is merged into `ROOMS` (the same point the old code already cleared localStorage), so marking there is not a new optimistic gravestone.
- **[verified clean]** No regression against #215/#216/#279/#218/#333/#291; `leave_room` never touches the processed-set, so accepted-then-left stays suppressed.
- **[noted]** The four behavioral tests are somewhat tautological; the two source-grep **pin** tests are the real regression guards and were confirmed to fail on `main`.

**Verdict:** Approved. Merging once CI is green on the current HEAD.

[AI-assisted - Claude]

Issue #94, opened 2026-02-11
In msg edit dialog:  Cancel (ESC) - Save (Enter) key bindings should be active without textarea being focused


kerbo7:
Cancel (ESC) - Save (Enter) buttons in edit message box:

The key bindings for ESC and ENTER button are not active, unless you click inside the textarea.

Both keys ESC and Enter should preferably also work if textarea doesn't have focus.

Like, clicking EDIT and then immediately hitting ESC or ENTER should close the edit window.

PR #357, opened 2026-06-13
fix(ui): show @mention autocomplete when editing a message

sanity:
## Problem

The `@mention` autocomplete popup worked in the new-message composer but **not** when editing an existing message. The mention machinery (query detection, candidate filtering, caret math, dropdown rendering) lived entirely inside `MessageInput` and was hardcoded to the composer's `#message-input` textarea, so the inline edit form in `MessageGroupComponent` had none of it — typing `@` while editing did nothing.

## Approach

Extract the mention machinery into a shared `components/conversation/mention.rs` module, parameterised by the target textarea's DOM `id` so the same code drives both call sites instead of being duplicated (which would drift):

- `detect_mention_query`, `filter_mention_candidates`, `duplicate_candidate_names`, UTF‑16↔byte caret helpers — moved verbatim (logic unchanged), with their unit tests.
- `update_mention_from_input`, `handle_mention_keydown`, `apply_mention_selection` — thin wrappers that take a textarea `id`.
- `MentionDropdown` component — the dropdown markup, single‑sourced.

The composer is rewired to call the shared module; its rendered DOM is unchanged (verified — composer popup still works). The edit textarea in `MessageGroupComponent` is wrapped in a `relative` container, given a unique `edit-msg-<id>` id, and wired to the same oninput/onkeydown/dropdown path. Its mentionable‑member list is derived from the `member_names` map the component already receives (self excluded, sorted by name), so no new prop plumbing.

Keyboard parity with the composer: Escape closes the dropdown first (a second Escape cancels the edit); Enter/Tab accept a candidate **without** saving; clicking elsewhere dismisses via `onfocusout`.

**UI‑only change** — no delegate/contract/common WASM touched, so no migration entry is required.

## Testing

- Relocated mention unit tests pass (`cargo test -p river-ui --bins` → 348 passed).
- Headless Playwright run against the `example-data` build confirms:
  - composer popup still shows (**regression check**);
  - edit popup now shows, with candidates, on‑screen and unclipped (the fix);
  - mouse‑pick and keyboard (ArrowDown+Enter) both insert the correct mention token and close the dropdown;
  - Enter‑accept of a candidate does **not** save the edit;
  - Escape closes the dropdown without cancelling the edit.

Screenshot (editing a message, dropdown open above the edit box):

`@Alice Williams (Owner)` / `@Ben Martin (Member)` candidates render above the edit textarea.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -2270,6 +2270,28 @@ fn MessageGroupComponent(
                                                             let _ = el.scroll_to(ScrollBehavior::Smooth).await;
                                                         });
                                                     },
+                                                    // Global key bindings on the container (#94): Esc cancels,
+                                                    // Enter saves. Kept here (not solely on the textarea) so the
+                                                    // "(Esc)"/"(Enter)" button shortcuts work when keyboard focus
+                                                    // is on a button. The textarea's @mention handler calls
+                                                    // stop_propagation when it consumes a key, so these never
+                                                    // double-fire with mention navigation.
+                                                    onkeydown: {
+                                                        let msg_id = msg_id_for_save.clone();
+                                                        let original = original_text.clone();
+                                                        move |e: KeyboardEvent| {
+                                                            if e.key() == Key::Escape {
+                                                                editing_message.set(None);
+                                                            } else if e.key() == Key::Enter && !e.modifiers().shift() {
+                                                                e.prevent_default();
+                                                                let new_text = edit_text.read().clone();
+                                                                if !new_text.is_empty() && new_text != original {
+                                                                    on_edit.call((msg_id.clone(), new_text));
+                                                                }
+                                                                editing_message.set(None);
+                                                            }
+                                                        }
+                                                    },
                                                     // `relative` anchors the @mention autocomplete
                                                     // dropdown to the textarea.
                                                     div { class: "relative",
@@ -2306,27 +2328,16 @@ fn MessageGroupComponent(
                                                                 );
                                                             },
                                                             // @mention navigation (Arrow/Enter/Tab/Esc) takes
-                                                            // precedence while the dropdown is open; otherwise
-                                                            // Esc cancels the edit and Enter saves it (#94).
-                                                            onkeydown: {
-                                                                let msg_id = msg_id_for_save.clone();
-                                                                let original = original_text.clone();
-                                                                move |e: KeyboardEvent| {
-                                                                    if mention::handle_mention_keydown(
-                                                                        &kd_id, &e, edit_text, edit_mention, || {},
-                                                                    ) {
-                                                                        return;
-                                                                    }
-                                                                    if e.key() == Key::Escape {
-                                                                        editing_message.set(None);
-                                                                    } else if e.key() == Key::Enter && !e.modifiers().shift() {
-                                                                        e.prevent_default();
-                                                                        let new_text = edit_text.read().clone();
-                                                                        if !new_text.is_empty() && new_text != original {
-                                                                            on_edit.call((msg_id.clone(), new_text));
-                                                                        }
-                                                                        editing_message.set(None);
-                                                                    }
+                                                            // precedence while the dropdown is open. When it
+                                                            // consumes the key, stop_propagation keeps the
+                                                            // container's Esc-cancel / Enter-save (above) from
+                                                            // also firing for that same key. Non-mention keys
+                                                            // bubble up to the container handler unchanged.
+                                                            onkeydown: move |e: KeyboardEvent| {
+                                                                if mention::handle_mention_keydown(
+                                                                    &kd_id, &e, edit_text, edit_mention, || {},
+                                                                ) {
+                                                                    e.stop_propagation();
                                                                 }
                                                             },
                                                             // Dismiss the dropdown when focus leaves the textarea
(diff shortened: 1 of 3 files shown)


sanity (commented):
## Comprehensive PR Review: #357

### Summary
- **PR Title:** fix(ui): show @mention autocomplete when editing a message
- **Type:** fix (UI) + internal refactor (extract shared module)
- **CI Status:** migration checks green; `build` + `ui-playwright-tests` running on current HEAD (must be green before merge)
- **Linked Issues:** none linked (PR is self-describing)
- **Review tier:** Light (UI-only, no high-risk surface) — escalated reviewer set because the diff moves code across files and touches the working composer
- **Reviewers run:** `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer`, external **Codex** (twice — initial HEAD `de667831` and post-fix HEAD `5ee1fda0`)

---

### Code-First Analysis
**Independent understanding:** Extracts the `@mention` machinery out of `MessageInput` into a shared `conversation/mention.rs` module parameterised by a textarea DOM `id`, then wires it into the inline message-edit form so `@` autocomplete works while editing (previously composer-only).
**Stated intent:** matches.
**Alignment:** Full. Verified by two reviewers line-by-line; the moved pure-logic helpers are byte-for-byte identical to the originals (independently diffed), and the composer's rendered DOM is unchanged.
**Gaps:** none.

---

### Testing Assessment
**Coverage level:** adequate for a UI-wiring change.

| Unit | ✅ | 6 mention unit tests relocated verbatim to `mention.rs`; full `river-ui --bins` suite = 348 passed |
| Integration | N/A | — |
| Simulation | N/A | — |
| E2E | ✅ (manual) | Headless Playwright against the `example-data` build exercised both popups and all keyboard paths (see Verdict) |

CI runs `ui-playwright-tests` (layout regression) on the PR.

---

### Skeptical Findings
**Risk level:** low.

| Concern | Severity | Location | Resolution |
|---------|----------|----------|------------|
| Esc/Enter dead when focus is on the Cancel/Save buttons (handler moved container→textarea) | **Medium (regression)** | conversation.rs edit form | **Fixed** in `5ee1fda0` — restored the container-level #94 handler; textarea handler is now mention-only and `stop_propagation`s when it consumes a key (no double-fire). Verified: Esc from a focused button cancels again. |
| Sync `text.set`/`mention.set` from event handlers (defer only the caret) | P2 (Codex) | mention.rs `apply_mention_selection` | **Dismissed — see reconciliation below.** |
| `edit_mention_members` recomputed per keystroke | Low | conversation.rs | Not changed — negligible for realistic room sizes; the per-keystroke re-render is pre-existing (driven by `edit_text`). |
| `MentionDropdown` uses `.read()` not `try_read()` | Low | mention.rs | Not changed — identical to the original composer; the skeptical reviewer explicitly classified this as **not a regression**. |

**Reconciliation of the Codex P2 (sync signal mutation) vs. the skeptical reviewer (safe):**
Investigated directly. The synchronous `message_text.set` + `mention.set` (deferring only the caret reposition) is **byte-identical to the shipping composer's `apply_mention_selection`** on `main` — this PR preserves it, it does not introduce it. The edit form on `main` likewise already set `edit_text` synchronously from `oninput`. The signals involved are **local `use_signal`s** (`message_text`/`mention`/`edit_text`/`edit_mention`), not the app-wide `GlobalSignal`s (`ROOMS`, `CURRENT_ROOM`) whose broad memo fan-out drives the re-entrant-RefCell crashes the signal-safety rule targets. Deferring the `set` would also require reworking the caret-after-flush ordering (the deferred `set_caret` relies on the synchronous `set` having already scheduled the re-render) inside code shared with the proven composer — a higher-blast-radius change to working code than the finding warrants. Decision: keep parity with the shipping composer. (A future hardening pass could defer local-signal sets across *both* the composer and edit form together, but that is out of scope and not this PR's regression to fix.)

---

### Big Picture Assessment
**Goal alignment:** yes — fixes the user-visible gap by eliminating duplication (shared module) rather than copy-pasting composer logic.
**Anti-patterns:** none (no `#[ignore]`, weakened tests, commented-out code, or TODO deferrals).
**Removed-code concerns:** none — the 342-line reduction in `message_input.rs` is fully accounted for in `mention.rs`; all 6 unit tests relocated, not deleted.
**Scope:** focused — three `ui/src/` files, nothing unrelated.

---

### Documentation
- Code docs: complete (new module + public items carry doc comments).
- Architecture / rules: no stale references — the only `mention` hit in `.claude/rules/` is the unrelated `river_core::mention` Cargo feature. `check-delegate-migration` + `check-room-contract-migration` pass, confirming the "UI-only, no migration" claim.
- User docs: n/a.

---

### Recommendations

#### Must Fix (Blocking)
- None outstanding. (The Esc/Enter-from-button regression was the one blocking item; fixed in `5ee1fda0` and verified.)

#### Should Fix (Important)
- None.

#### Consider (Suggestions)
1. Optional future hardening: defer local-signal sets in `apply_mention_selection`/`update_mention_from_input` across both the composer and the edit form together (addresses the Codex P2 at the source, for both call sites). Out of scope here; risk of touching the proven composer outweighs the benefit for this fix.
2. Optional: `use_memo` the edit-form candidate list if very large rooms ever make the per-keystroke clone+sort noticeable.

---

### Verdict
**State:** Ready to Merge once `build` + `ui-playwright-tests` are green on the current HEAD.
**HEAD SHA reviewed:** `5ee1fda0f75f6c24373e5447923ec045e757f706`

Re-review note: this is the post-fix pass. The prior pass's one blocking finding (Esc/Enter from the edit buttons) is resolved and re-verified end-to-end; the remaining items are low-severity / pre-existing and dismissed with justification above.

[AI-assisted - Claude]

PR #358, opened 2026-06-16
feat(cli): implement member_info heal counterpart to UI's build_member_info_heal

### LINKED ISSUE #304: riverctl: implement member_info heal counterpart to UI's build_member_info_heal
## Background

Follow-up to #302 / PR #303. When riverctl accepts a private-room invitation but no secret is available at the room's `current_secret_version` (e.g. the room rotated between invitation creation and accept), PR #303 correctly **defers** `member_info` rather than leaking a plaintext nickname. The CLI member surfaces as **"Unknown"** to other peers and stays that way indefinitely — there is no remediation path on the CLI side.

The UI fixed the same class of problem in PR #294 via `RoomData::build_member_info_heal`: on every GET of an existing room it detects "self in `members`, absent from `member_info`" and re-publishes a self-signed `member_info` (folded into the PUT for imported rooms, sent as a standalone UPDATE for already-subscribed rooms). For a private room the heal defers (publishes nothing) until the room secret is available, so it never leaks a plaintext nickname.

## Scope

Port the heal to riverctl:

1. After every CLI state ingestion (GET response, delta-apply, room re-import), detect the "self in members but absent from member_info" condition for the current `signing_key`.
2. If the room is private and no secret is available at `current_secret_version`, defer (do nothing).
3. Otherwise, build an `AuthorizedMemberInfo` via `seal_invitee_nickname` (PR #303 already extracted the helper) and emit a standalone UPDATE delta carrying just `member_info`.
4. The nickname source: a new `StoredRoomInfo.self_nickname: Option<String>` field, mirroring UI's `RoomData.self_nickname`, captured at invitation accept-time.

## References
- PR #294 — "fix(ui): unknown-member regression after invite_accept" (UI heal)
- PR #303 — "feat(cli): carry private-room secrets in riverctl invitations" (this PR; documents the gap)
- UI source: `ui/src/room_data.rs::build_member_info_heal`

[AI-assisted - Claude]

sanity:
## Problem

When `riverctl` accepts a private-room invitation but no secret is available at the room's `current_secret_version` (the room rotated between invitation creation and accept), PR #303 correctly **defers** `member_info` rather than leaking a plaintext nickname. The CLI member then surfaces as **"Unknown"** to every other peer and stays that way indefinitely — there was no remediation path on the CLI side.

The UI fixed the same class of problem in PR #294 via `RoomData::build_member_info_heal`: on every GET of an existing room it detects "self in `members`, absent from `member_info`" and re-publishes a self-signed `member_info`. For a private room the heal defers until the room secret is available, so it never leaks a plaintext nickname.

This ports that heal to riverctl.

Closes #304

## Approach

- **`crate::private_room::build_member_info_heal`** — a pure, node-free function mirroring the UI's `build_member_info_heal`:
  - skips the room **owner** (their `member_info` is managed separately);
  - skips when self is **not in `members`** (the pruned-member case is already handled by `build_rejoin_delta`, which re-adds to `members`);
  - skips when self **already has a `member_info` entry** (not stranded);
  - otherwise builds a self-signed `AuthorizedMemberInfo`. A non-owner's `member_info` is only valid when self-signed by that member's own key, so the heal can only be produced client-side by the affected member — exactly what this returns.
  - Public rooms publish public bytes; private rooms seal under the current-version secret via the existing `seal_invitee_nickname`, **deferring (returning `None`) when no current-version secret is available** so a plaintext nickname is never leaked.
  - An over-long stored nickname is clamped to the `"Member"` placeholder (same guard as `rejoin_preferred_nickname`) so the heal UPDATE can't be rejected for exceeding `max_nickname_size`.

- **`ApiClient::heal_member_info`** — wired into the central GET path (`get_room`), so every read/send command re-attempts the heal once a secret arrives. It's **best-effort**: a heal failure only `warn!`s, never fails the command that fetched the state. It emits a standalone `member_info`-only UPDATE delta and folds the same entry into the locally-stored state.

- **Nickname source** is the existing `StoredRoomInfo.self_nickname` field (added in #321), so this introduces **no new persisted field** and no new `StoredRoomInfo` serde/backward-compat surface.

### Why the GET path (not just accept)

`accept_invitation` already publishes `member_info` when it can, and correctly defers otherwise. The missing piece was *remediation later*, once a secret arrives — which is precisely what the GET-path trigger gives: every subsequent `riverctl` command on the room re-checks and heals. This mirrors the UI, which fires the heal on every GET.

> Note: the triage hypothesis mentioned a hardcoded `"Member"` fallback in `build_rejoin_delta`; that was already fixed in PR #321 via `rejoin_preferred_nickname` (which seals for private rooms and clamps to `max_nickname_size`). No change needed there.

## Testing

`cargo test -p riverctl --lib` — 140 pass. New tests in `cli/src/private_room.rs`:

- `heal_public_room_missing_member_info_is_healed` — missing → healed (public).
- `heal_private_room_with_secret_seals_nickname` — private with secret → sealed, never plaintext.
- `heal_returns_none_when_member_info_already_present` — present → unchanged.
- `heal_returns_none_when_self_not_in_members`, `heal_returns_none_for_owner` — no-op cases.
- `heal_private_room_without_secret_defers`, `heal_private_room_defers_when_secret_lacks_current_version` — private-no-secret / stale-version → defer.
- `heal_falls_back_to_member_placeholder_when_no_nickname`, `heal_clamps_over_long_nickname_to_member_placeholder` — fallback / clamp.
- `heal_output_is_accepted_by_member_info_apply_delta` — end-to-end: the healed entry is accepted and retained by the room contract's own `MemberInfoV1::apply_delta`.

Plus a source-grep pin `storage::tests::member_info_heal_wiring_pinned` guarding the GET-path wiring against silent removal.

`cargo fmt` + `cargo clippy -p riverctl --all-targets` clean. No `common/`/`delegates/`/`contracts/` changes, so no delegate/contract migration needed.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/private-rooms.md
+++ b/.claude/rules/private-rooms.md
@@ -183,12 +183,22 @@ cannot be re-derived, so the owner-as-inviter path always decrypts the
 owner-addressed contract blob from `state.secrets.encrypted_secrets` like
 any other member.
 
-The CLI does NOT currently have a `build_member_info_heal` counterpart: a
-private-room invitee whose invitation lacks `current_version`'s secret
-will defer `member_info` and surface as **"Unknown"** to other peers
-indefinitely (filed as freenet/river#304). The CLI also does NOT prune
-superseded `invitation_secrets` entries the way the UI does — storage
-waste only; the heal path is the natural place to hook the prune.
+The CLI has a `build_member_info_heal` counterpart (issue freenet/river#304),
+in `cli/src/private_room.rs`, mirroring the UI's `RoomData::build_member_info_heal`:
+it detects "self in `members`, absent from `member_info`" and returns a
+self-signed `AuthorizedMemberInfo` to re-publish. A private-room heal seals
+the nickname under the current secret and **defers** (returns `None`) when no
+current-version secret is available, so it never leaks a plaintext nickname.
+`ApiClient::heal_member_info` drives it from the central GET path
+(`get_room`) — every read/send command re-attempts the heal once a secret
+arrives — and emits a standalone `member_info`-only UPDATE delta. The heal is
+skipped under a `--signing-key` override that selects an identity other than
+the room's stored one (the persisted nickname/secrets are not that identity's).
+Wiring is pinned by `member_info_heal_wiring_pinned` in `cli/src/storage.rs`;
+behaviour by the `heal_*` tests in `cli/src/private_room.rs`.
+
+The CLI still does NOT prune superseded `invitation_secrets` entries the way
+the UI does — storage waste only.
 
 ### Rejoin nickname restoration (`StoredRoomInfo.self_nickname`)
 
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -840,7 +840,88 @@ impl ApiClient {
             self.subscribe_to_contract(found_id).await?;
         }
 
+        // Self-heal the "Unknown member" condition (issue freenet/river#304):
+        // if this member is in `members` but absent from `member_info`, they
+        // render as "Unknown" to every other peer. This is the CLI counterpart
+        // of the UI's GET-path `build_member_info_heal` trigger. Best-effort —
+        // a heal failure must not fail the read/send command that fetched the
+        // state, so we only warn.
+        if let Err(e) = self.heal_member_info(room_owner_key, &room_state).await {
+            warn!("member_info self-heal (issue #304) did not complete: {e}");
+        }
+
         Ok(room_state)
+    }
+
+    /// Detect and remediate the "Unknown member" condition for the current
+    /// member (issue freenet/river#304): self present in `state.members` but
+    /// absent from `state.member_info`. When detected, publish a standalone
+    /// `member_info`-only UPDATE so other peers stop rendering this member as
+    /// "Unknown", and fold the same entry into the locally-stored state.
+    ///
+    /// The CLI counterpart of the UI's `RoomData::build_member_info_heal`
+    /// (`ui/src/room_data.rs`), driven from the same place: every GET of an
+    /// existing room (see [`Self::get_room`]). The heal-decision logic lives in
+    /// the pure [`crate::private_room::build_member_info_heal`] so it is
+    /// unit-testable without a node; this method owns the storage lookup and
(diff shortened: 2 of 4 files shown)


sanity:
## Review summary

**External model:** Codex (`codex review --base origin/main`), run iteratively — 5 review rounds. Found 4 actionable P2 findings, all fixed; final round clean ("did not identify any discrete, actionable regressions").

| # | Codex finding (P2) | Resolution |
|---|---|---|
| 1 | Heal reuses stored identity's nickname/secrets under a `--signing-key` override that selects a *different* identity | Skip heal when resolved key ≠ stored `signing_key_bytes` |
| 2 | `get_room` returned the pre-heal state, so callers wrote stale state back, dropping the heal locally | `heal_member_info` now returns the healed state; `get_room` rebinds to it |
| 3 | Local full-state `apply_delta` runs `post_apply_cleanup`, which could prune the unanchored member being healed | Fold the entry into the `member_info` sub-state directly, no full-state apply |
| 4 | `"Member"` fallback (6 bytes) exceeds a `max_nickname_size < 6` → contract rejects, CLI persists invalid entry | Defer the heal when even the placeholder doesn't fit |
| 5 | The *network* member_info-only UPDATE also triggers `post_apply_cleanup` → prunes an unanchored member instead of healing | `build_member_info_heal` simulates `post_apply_cleanup` and defers unless self survives it |

**Claude self-review (4 lenses, serial):** code-first / testing / skeptical / big-picture — no further findings. Confirmed: only `member_info` is published (no `MembersDelta`), the heal is best-effort (a publish failure only warns), the expensive cleanup-simulation only runs for genuinely stranded members (after the `has_member_info` short-circuit), and the private-rooms.md doc was updated to reflect that the heal now exists.

**Risk tier:** Light (CLI-only logic, no protocol/wire/migration change — migration CI checks pass), with identity/private-room care applied per the backward-compat note. No `common/`/`delegates/`/`contracts/` changes.

Tests: `cargo test -p riverctl --lib` — 142 pass (11 new heal tests + 3 pins). `cargo fmt` + `cargo clippy` clean.

[AI-assisted - Claude]

PR #359, opened 2026-06-16
fix(cli): lock and atomically write riverctl storage to prevent lost updates

### LINKED ISSUE #307: riverctl: rooms.json has no file locking — concurrent invocations may lose updates

sanity:
## Problem

`cli/src/storage.rs` persisted `rooms.json` / `outbound_dms.json` with a bare `fs::write` and **no advisory lock**. Every mutating operation is a `load → mutate → save` sequence, so two concurrent `riverctl` invocations (e.g. `accept` + `dm send` + `invite create` from a script or cron) could both load the same base snapshot and the later `save` would silently clobber the earlier writer's update — a lost-update race (issue #307). The non-atomic in-place write also let a concurrent reader observe a half-written, unparseable blob. PR #303's `invitation_secrets` persistence widened the loss surface to cryptographic material.

```
T1: load_rooms() -> {0: A, 1: B}
T2: load_rooms() -> {0: A, 1: B}
T2: merge -> {0,1,2}; save   # disk now {0,1,2}
T1: merge -> {0: A', 1: B}; save   # disk now {0,1} -- writer 2's room LOST
```

## Approach

Two defenses, both contained in the CLI storage layer:

1. **Cross-process advisory lock (`fs2`)** on a dedicated `.river.lock` file, held across the whole `load → mutate → save` critical section so concurrent invocations serialize rather than interleave. A new `with_lock` helper owns acquire/release; each mutating method calls a private `*_unlocked` body, so the lock is taken exactly once and never re-entered.
   - **Reentrancy hazard (called out in triage):** `load_rooms` regenerates contract keys and saves them back. `fs2` locks are per-open-file-description, so a nested lock on a fresh handle — even same-process — blocks forever. The internal save now goes through `save_rooms_unlocked`, so it cannot self-deadlock under the outer lock. `remove_room`'s outbound-DM-cache prune is likewise folded into one lock via an `*_unlocked` prune helper.
2. **Atomic writes** via temp-file + `rename(2)` (`atomic_write`), so a reader (or a crash mid-write) never sees a partial blob.

Two `mutate_rooms` / `mutate_outbound_dms` critical-section primitives let the external `api.rs` (`clear_previous_contract_key`) and `dm.rs` (`persist_outbound_plaintext`, `prune_outbound_cache_for_room`) callers do their domain mutation under **one** lock, instead of a racy `load()` … `save()` pair spanning two separate lock acquisitions.

The single-writer/locking model and the no-nesting rule are documented in the `Storage` type-level doc comment.

## Testing

New concurrency tests in `storage.rs`, plus a source-scrape pin guarding the lock + atomic-write + reentrancy-guard invariants against a future refactor:

- `concurrent_add_room_does_not_lose_updates` — 16 threads (barrier-aligned) each insert a distinct room; all 16 must survive.
- `concurrent_outbound_dm_appends_do_not_lose_updates` — same for the outbound-DM cache.
- `concurrent_reads_never_observe_a_torn_write` — a reader hammering `load_rooms` while writers run must never hit a parse error.

**TDD discipline:** verified all three fail reliably against the pre-fix behavior (lost inserts + `serde_json` EOF parse errors mid-write) and pass with the fix. Full `riverctl` lib suite (133 tests) green; `cargo fmt` + `cargo clippy --all-targets` clean.

No WASM / contract / delegate changes — pure CLI file-I/O, so no migration entry needed.

Closes #307

[AI-assisted - Claude]

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -78,6 +78,11 @@ chrono = { version = "0.4", features = ["serde"] }
 toml = "0.8"
 directories = "5.0"
 
+# Advisory file locking — serializes concurrent riverctl invocations'
+# load→mutate→save sequences against rooms.json / outbound_dms.json so a
+# racing writer can't clobber another's freshly-merged update (issue #307).
+fs2 = "0.4"
+
 [dev-dependencies]
 freenet-test-network = "0.1.23"
 tempfile = "3"
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -1802,13 +1802,15 @@ impl ApiClient {
 
     /// Clear the previous_contract_key after successful migration.
     fn clear_previous_contract_key(&self, owner_vk: &VerifyingKey) -> Result<()> {
-        let mut storage = self.storage.load_rooms()?;
+        // Single locked load→mutate→save so a concurrent invocation can't
+        // clobber this clear (issue freenet/river#307).
         let owner_key_str = bs58::encode(owner_vk.as_bytes()).into_string();
-        if let Some(room_info) = storage.rooms.get_mut(&owner_key_str) {
-            room_info.previous_contract_key = None;
-            self.storage.save_rooms(&storage)?;
-        }
-        Ok(())
+        self.storage.mutate_rooms(|storage| {
+            if let Some(room_info) = storage.rooms.get_mut(&owner_key_str) {
+                room_info.previous_contract_key = None;
+            }
+            Ok(())
+        })
     }
 
     /// Migrate a room to a new contract by PUTting the state
--- a/cli/src/commands/dm.rs
+++ b/cli/src/commands/dm.rs
@@ -1046,20 +1046,21 @@ fn persist_outbound_plaintext(
     timestamp: u64,
     plaintext: String,
 ) -> Result<()> {
-    let mut store = api.storage().load_outbound_dms()?;
+    // Single locked load→mutate→save so a concurrent DM-send / room-leave can't
+    // clobber this append (issue freenet/river#307).
     let room_bytes = room_owner_vk.to_bytes();
-    store.entries.push(OutboundDmEntry {
-        room_owner_vk: room_bytes,
-        sender,
-        recipient,
-        purge_token,
-        timestamp,
-        plaintext,
-    });
-
-    apply_per_pair_cap(&mut store);
-    api.storage().save_outbound_dms(&store)?;
-    Ok(())
+    api.storage().mutate_outbound_dms(|store| {
+        store.entries.push(OutboundDmEntry {
+            room_owner_vk: room_bytes,
+            sender,
+            recipient,
+            purge_token,
+            timestamp,
+            plaintext,
+        });
+        apply_per_pair_cap(store);
+        Ok(())
+    })
 }
 
 /// Pure helper: drop the oldest entries from each `(room, sender,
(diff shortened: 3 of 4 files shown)


sanity:
## Review summary (serial path — dispatched subagent, no parallel reviewers)

Per `~/.claude/rules/multi-model-review.md`: ran one external non-Claude pass (Codex) + four Claude lenses serially (code-first, testing, skeptical, big-picture). Light-tier change (local CLI file I/O), but treated the concurrency surface carefully.

### External: Codex (`codex review --base origin/main`)

One finding:

- **[P2] "Windows `std::fs::rename` fails when destination exists" → DISMISSED (false positive).** Codex claimed `atomic_write`'s `fs::rename(tmp, target)` would error on Windows once `rooms.json` exists. This is incorrect for the current Rust std library. The official docs state `std::fs::rename` *"Renames a file or directory to a new name, replacing the original file if `to` already exists"* ([std::fs::rename](https://doc.rust-lang.org/std/fs/fn.rename.html)); the Windows impl uses `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`, so a file→file rename overwrites the destination (this is exactly our case: `rooms.json.tmp.<pid>` → `rooms.json`, both regular files in the same dir). The platform caveat in the docs is about directory-vs-file type mismatches, which never applies here. No change needed. (riverctl's primary targets are Linux/macOS regardless.)

### Claude lenses

- **Code-first:** implementation matches issue #307 scope (doc the model, atomic write, advisory lock around load→mutate→save, same treatment for `outbound_dms.json`, concurrency test). The `mutate_rooms`/`mutate_outbound_dms` primitives always re-save even on a no-op mutation — an idempotent identical-bytes rewrite, documented; `prune_outbound_cache_for_room` still short-circuits before the lock when there's nothing to purge.
- **Testing:** three new concurrency tests proven to fail pre-fix (lost inserts + `serde_json` EOF mid-write) and pass post-fix, plus a source-scrape pin for the lock/atomic-write/reentrancy invariants. The reentrancy guard is exercised by existing tests that would hang on a self-deadlock (`test_load_rooms_sets_previous_contract_key_on_mismatch` drives the `load_rooms → save_rooms_unlocked` path under the outer lock).
- **Skeptical:** the triage-flagged reentrancy hazard is handled — every `with_lock` body calls only private `*_unlocked` helpers, never a re-locking method (`fs2` locks are per-open-file-description, so a nested fresh handle would self-deadlock). Stale `.tmp.<pid>` on a hard crash is best-effort-cleaned only on the error path; benign (next same-PID run overwrites). `load_rooms` takes the exclusive lock for a read because it can write during key regeneration — correct over-conservatism for a one-shot CLI.
- **Big-picture:** no tests removed, no stale doc/skill references to the old `save_rooms` shape, no WASM/contract/delegate changes (no migration entry needed).

All findings addressed. Proceeding to CI gate.

[AI-assisted - Claude]

PR #360, opened 2026-06-16
fix(ui): bound restored-room sync retries and surface a terminal error

### LINKED ISSUE #290: Restored/imported room spins forever if its contract is absent from the network
## Summary

Follow-up to #287 / PR #289. PR #289 fixes the bare-`Subscribe`-rejection bug that hung sync for restored rooms. After that fix, a room loaded from delegate storage is synced via `process_rooms()` — and an *imported* room (default placeholder state, `is_awaiting_initial_sync() == true`) takes the `GET` path with `return_contract_code = true`.

- If the room's contract **still exists on the network**, the GET returns state, the WASM is cached, and the room syncs — #287 resolved.
- If the room's contract **no longer exists on the network** (a long-abandoned room whose contract has been dropped, or one never PUT to the post-WASM-change contract key), the GET fails.

In the failure case River never reaches a terminal state:

- For a "contract not found" error, the `freenet_synchronizer.rs` error handler resets the room to `Disconnected` and schedules a retry `ProcessRooms` (originally added for the freenet-core#1470 contract-creation race) — so River re-GETs indefinitely; or
- if the GET simply times out, the room sits in `Subscribing` and is re-driven by later `ProcessRooms` cycles.

Either way the room never settles, and because the "Syncing room state from the network…" spinner (`room_list.rs` / `conversation.rs`) is gated on `is_awaiting_initial_sync()` — which stays `true` while the room holds placeholder state — **the spinner spins forever**.

## Why this is separate from #287

Distinct, pre-existing bug with a different precondition: contract *genuinely absent from the network*, vs. #287's *contract present but WASM not cached locally*. PR #289 fixes the evidenced #287 symptom; this issue is the remaining never-ending-spinner case when the room data is genuinely unrecoverable.

## Proposed fix

1. **Bound the retry.** After N failed GET attempts for an imported room whose contract cannot be found, stop retrying and move the room to a terminal `RoomSyncStatus::Error`. Be careful not to regress the freenet-core#1470 race handling — keep the bound generous.
2. **Surface it in the UI.** The spinner is gated on `is_awaiting_initial_sync()`, not on `RoomSyncStatus`, so a terminal `Error` status alone will not clear it. `room_list.rs` and `conversation.rs` need a terminal/error state ("This room could not be found on the network — it may have been removed") instead of the perpetual spinner.

## Owner reconstruction (possible enhancement)

For a room the user **owns**, the exported `IdentityExport` carries the owner signing key, `member_info`, `invite_chain` and `authorized_member`. River could in principle reconstruct a minimal valid state, sign a fresh `Configuration` with the owner key, and `PUT` it — re-creating the room rather than failing. Larger enhancement; scope separately if desired.

## References

- Original report: #287
- Fix for the bare-Subscribe half: #289

[AI-assisted - Claude]

sanity:
## Problem

Follow-up to #287 / PR #289. After #289, a room loaded from delegate storage syncs via `process_rooms()`; an *imported* room (placeholder state, `is_awaiting_initial_sync() == true`) takes the `GET` path with `return_contract_code = true`.

If the room's contract is **genuinely absent from the network** (a long-abandoned room whose contract was dropped, or one never PUT to the current contract key), the sync never settles:

- a **"contract not found"** error resets the room to `Disconnected` and schedules a retry `ProcessRooms` **forever** (the freenet-core#1470 race handler, originally meant for a transient miss), or
- the GET **times out** and the room is reset from `Subscribing` back to `Disconnected` and re-driven on the next `ProcessRooms` cycle, indefinitely.

Either way the room keeps its placeholder state, so `is_awaiting_initial_sync()` stays `true`, and the "Syncing room state from the network…" spinner (`room_list.rs` / `conversation.rs`) **spins forever** (freenet/river#290).

## Approach

Bound the retries and give the loop a terminal state:

- **`RoomSyncInfo` gains a `failed_sync_attempts` counter.** A new `SyncInfo::record_failed_sync_attempt()` increments it and, until the generous bound `MAX_SYNC_ATTEMPTS_BEFORE_ERROR` (10) is reached, resets the room to `Disconnected` (retry); at the bound it promotes the room to terminal `RoomSyncStatus::Error` and returns "stop retrying". The bound is deliberately generous so it does **not** regress the freenet-core#1470 contract-creation race (which resolves within a couple of retries). A successful `Subscribed` clears the counter so a later transient outage starts from zero rather than inheriting a near-exhausted budget.

- **Both failure paths route through the bound:** the contract-not-found error handler in `freenet_synchronizer.rs` (now only schedules a retry while at least one matching room is still within budget), and the `Subscribing`-timeout sweep in `rooms_awaiting_subscription()` (a timed-out room is no longer re-added to the awaiting set once it has been given up on).

- **The spinner sites stop spinning on the terminal state.** `room_list.rs` shows a warning marker instead of the perpetual spinner; `conversation.rs` shows *"This room could not be found on the network — it may have been removed."* Both are scoped to placeholder-state rooms, so a fully-synced room that hits some other transient `Error` (e.g. a PUT failure) is unaffected and keeps its existing rendering.

Why not the owner-reconstruction enhancement from the issue: that's explicitly scoped as a larger, separate enhancement. This PR fixes the never-ending-spinner symptom.

## Testing

New unit tests in `sync_info.rs`:
- `record_failed_sync_attempt_bounds_retries_then_errors` — retries up to the bound (status stays `Disconnected`), then promotes to terminal `Error` and stops. **Verified it FAILS against the pre-fix unbounded behavior** (the room never reaches `Error`, panicking on "reaching the bound must stop retrying") and passes with the fix.
- `successful_subscription_resets_failure_budget` — a `Subscribed` clears the counter so the next failure starts over rather than tipping straight to `Error`.
- `record_failed_sync_attempt_on_unknown_room_is_noop`.

Full `cargo test -p river-ui --bins`: **351 passed, 0 failed.** `cargo fmt` clean; `cargo clippy -p river-ui --bins` introduces no new warnings (the 11 reported are pre-existing and outside the changed regions).

Closes #290

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/freenet_api/freenet_synchronizer.rs
+++ b/ui/src/components/app/freenet_api/freenet_synchronizer.rs
@@ -486,9 +486,20 @@ impl FreenetSynchronizer {
                                             // promoted to a terminal Error and we stop retrying it.
                                             let mut any_room_still_retrying = false;
                                             for room_key in &matching_rooms {
-                                                let should_retry = SYNC_INFO
-                                                    .write()
-                                                    .record_failed_sync_attempt(room_key);
+                                                // The retry bound only applies to a room still
+                                                // awaiting its initial sync (placeholder state —
+                                                // the #290 case). A room with valid synced state is
+                                                // never given up on, so a transient contract-not-found
+                                                // can't strand it.
+                                                let awaiting_initial_sync =
+                                                    ROOMS.read().map.get(room_key).is_some_and(
+                                                        |rd| rd.is_awaiting_initial_sync(),
+                                                    );
+                                                let should_retry =
+                                                    SYNC_INFO.write().record_failed_sync_attempt(
+                                                        room_key,
+                                                        awaiting_initial_sync,
+                                                    );
                                                 if should_retry {
                                                     any_room_still_retrying = true;
                                                     info!(
--- a/ui/src/components/app/freenet_api/freenet_synchronizer.rs
+++ b/ui/src/components/app/freenet_api/freenet_synchronizer.rs
@@ -5,7 +5,7 @@ use super::error::SynchronizerError;
 use super::response_handler::ResponseHandler;
 use super::room_synchronizer::RoomSynchronizer;
 use crate::components::app::chat_delegate::{mark_legacy_migration_done, set_up_chat_delegate};
-use crate::components::app::sync_info::{RoomSyncStatus, SYNC_INFO};
+use crate::components::app::sync_info::SYNC_INFO;
 use crate::components::app::{ROOMS, SYNC_STATUS, WEB_API};
 use crate::util::{owner_vk_to_contract_key, sleep};
 use dioxus::logger::tracing::{debug, error, info, warn};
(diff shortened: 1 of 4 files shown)


sanity:
## Review summary

**External model (codex)** ran twice on this PR and found two real issues, both fixed:

1. **[P2, fixed in 089929f1]** `record_failed_sync_attempt` bounded retries for *all* rooms, so a fully-synced room hitting 10 transient timeouts/contract-not-found errors would be promoted to terminal `Error` and silently stop being re-PUT/re-synced. Now the bound applies **only** to rooms still awaiting their initial sync (placeholder state — the #290 case); a synced room is never given up on. Added test `record_failed_sync_attempt_never_errors_a_synced_room`.
2. **[P2, fixed in 74422a03]** The UI matched any `RoomSyncStatus::Error(_)` on a placeholder room and hardcoded "could not be found on the network", misreporting WebSocket/API send failures (which also set `Error`) as the room being removed. Both spinner sites now surface the **stored** error string; the #290 terminal case stores the contract-not-found message, other failures show their own.

**External re-review on the latest content:** codex hit its usage limit (resets ~1:29 PM CT) and gemini was unresponsive/timed out on this run. Per the multi-model-review rule I fell back to a diverse-Claude-lens review for the current HEAD (the latest commit only adds the "surface stored error" UI change on top of the twice-codex-reviewed logic):

- **code-first:** implementation matches intent; counter resets on `Subscribed`; `is_awaiting_initial_sync()` scoping is consistent across both call sites.
- **skeptical:** the contract-not-found handler and the `Subscribing`-timeout sweep don't double-count within a cycle (the error handler clears `subscribing_since` → the sweep can't also fire that cycle); `ROOMS.read()`/`SYNC_INFO.write()` in the synchronizer async task is the same pattern the pre-fix code used there (no render-time re-entrancy); a terminal-Error placeholder room stays terminal (no further GET), so the error keeps showing.
- **change-specific (spinner reactivity):** the `room_items` memo and the conversation render body both read `SYNC_INFO` on the success path, so they re-render when the status flips to `Error`.

No blocking findings from the lenses. `cargo test -p river-ui --bins`: 352 passed, 0 failed. `cargo fmt` clean; `cargo clippy -p river-ui --bins` introduces no new warnings.

[AI-assisted - Claude]

PR #361, opened 2026-06-16
feat(ui): Enter commits and closes the edit-room dialog (#21)

### LINKED ISSUE #21: Hitting Enter should close edit room and user dialogs (useful for rename)

sanity:
## Problem

In the edit-room dialog, pressing Enter in the room-name field did nothing useful: the field auto-saves on change, but the dialog stayed open, so renaming a room always required a separate click on the close (✕) button or the backdrop. Issue #21 asks for Enter to commit the change and close the dialog, the way it works elsewhere.

The member-info / user dialog **already** closes on Enter: its backdrop has an `onkeydown` handler that closes on Enter/Escape, and `NicknameField` saves on Enter, so that surface needs no change. The only gap was the edit-room dialog, which had no Enter handling at all.

## Approach

Add an `onkeydown` handler to `RoomNameField`'s single-line input that, on Enter, commits the current value and closes the dialog via `EDIT_ROOM_MODAL`.

- Refactor the save logic from an event-typed closure into `save_room_name(String)` so the same path is driven by both `onchange` (blur / native commit) and the new Enter handler. The handler commits the live value **before** deferring the modal close, because closing unmounts the input and a pending `onchange` would otherwise be lost.
- The owner gate is unchanged: `save_room_name` returns early for non-owners, and the private-room privacy guard / revert behavior (#299) is preserved untouched.
- The room-**description** field is a multiline `textarea` and is deliberately left alone, so Enter there still inserts a newline (per the issue's "avoid submitting on Enter in a multiline field").
- **IME safety:** browsers report `Key::Process` (not `Key::Enter`) for the Enter keystroke that confirms an IME composition, so the handler does not fire mid-composition. Captured by the pure `enter_commits_and_closes` predicate.

## Testing

- New native unit tests on the extracted `enter_commits_and_closes` predicate (gated by CI's `cargo test -p river-ui --bins`): Enter closes; IME `Process` does not; Escape/Tab/Backspace/char/arrow do not.
- `cargo test -p river-ui --bins room_name_field` — 3 passed.
- `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` clean.
- `cargo clippy -p river-ui --bins` — no new warnings in this file.
- `cargo fmt` clean.

**Manual verification** (river UI has no automated harness for modal interaction): open a room's edit dialog as owner, edit the name, press Enter — the new name is saved and the dialog closes. As a non-owner the field is disabled; Enter just dismisses the dialog.

Closes #21

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/room_list/room_name_field.rs
+++ b/ui/src/components/room_list/room_name_field.rs
@@ -1,13 +1,24 @@
-use crate::components::app::{CURRENT_ROOM, ROOMS};
+use crate::components::app::{CURRENT_ROOM, EDIT_ROOM_MODAL, ROOMS};
 use crate::util::ecies::{seal_for_room, unseal_bytes_with_secrets};
 use dioxus::logger::tracing::*;
 use dioxus::prelude::*;
-use dioxus_core::Event;
 use freenet_scaffold::ComposableState;
 use river_core::room_state::configuration::{AuthorizedConfigurationV1, Configuration};
 use river_core::room_state::privacy::RoomDisplayMetadata;
 use river_core::room_state::{ChatRoomParametersV1, ChatRoomStateV1Delta};
 use wasm_bindgen_futures::spawn_local;
+
+/// Whether a keydown in the room-name input should commit the value and close
+/// the edit-room dialog (freenet/river#21).
+///
+/// Only the plain `Enter` key triggers it. In particular this returns `false`
+/// for `Key::Process`, which is what browsers report for the `Enter` keystroke
+/// that *confirms an IME composition* (e.g. selecting a CJK candidate) — we must
+/// not close the dialog on that keystroke. Extracted as a pure function so the
+/// IME-vs-submit decision is unit-testable without a Dioxus runtime.
+fn enter_commits_and_closes(key: &Key) -> bool {
+    key == &Key::Enter
+}
 
 #[component]
 pub fn RoomNameField(config: Configuration, is_owner: bool) -> Element {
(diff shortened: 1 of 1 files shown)


sanity:
## Multi-model review (Light tier — UI interaction)

Reviewed the checked-out PR code through an external non-Claude model plus four Claude lenses. **No blocking findings.**

### External model pass
- **Codex (`codex review --base origin/main`)**: unavailable — hit OpenAI usage limit (resets 1:29 PM). Tried first as the preferred external reviewer.
- **Gemini (`gemini`, fallback per multi-model-review rule)**: completed a full adversarial review, **no issues found**. Explicitly verified: (1) closure cloning / ownership across the `onchange`/`onkeydown` split, (2) signal-safety — GlobalSignal mutations (`ROOMS`, `EDIT_ROOM_MODAL`) are deferred via `crate::util::defer`, local `room_name.set()` follows the established in-handler pattern, (3) commit-before-close timing avoids the lost-edit race on unmount, (4) IME composition correctly handled (`Key::Process` ≠ `Key::Enter`), (5) edge cases (non-owner quick-dismiss, empty-name no-save).

### Claude lenses
- **Code-first**: implementation matches stated intent; member-info dialog already closed on Enter (backdrop `onkeydown`) so correctly left untouched — gap was only the edit-room dialog.
- **Testing**: pure `enter_commits_and_closes` predicate unit-tested (Enter / IME `Process` / Escape/Tab/etc.), gated by CI `cargo test -p river-ui --bins`. Behavioral wiring documented as manual verification (no Dioxus-runtime UI harness — consistent with `dm_rail_section` precedent).
- **Skeptical**: no re-entrant borrow (read guard on `room_name` dropped before `save_room_name`); the `save_room_name` synchronous local `.set()` matches the pre-existing pattern in `NumericConfigField`/`MaxMembersField`/`RoomDescriptionField`; spawn_local save survives component unmount.
- **Big-picture**: no tests removed; no WASM/contract/delegate/common changes → no delegate migration required (confirmed via `git status`); `bug-prevention-patterns.md` concurrency/wire-format patterns don't apply to this UI change.

### Minor (non-blocking, no change)
Enter (onkeydown) and a subsequent blur (onchange) can both call `save_room_name`. Harmless: `ComposableState::apply_delta` is idempotent for the same `configuration_version`, and the second call no-ops at worst.

[AI-assisted - Claude]

PR #364, opened 2026-06-16
fix(delegate-sub): re-arm EnsureRoomSubscription dedup on transport reconnect

### LINKED ISSUE #277: fix(delegate-sub): retry EnsureRoomSubscription on transient transport recovery
Follow-up from PR #276 (Bug #6 fix) skeptical review.

## Context

PR #276 added a `Skip-on-Failed` branch: when `migrate_signing_key` returns `MigrationResult::Failed`, the chained spawn skips `ensure_room_subscription_once` entirely. This was a correct trade-off vs. the round-1 design (which fired `EnsureRoomSubscription` anyway and let the dedup self-clear on Err) — it eliminates the "noisy retry loop on persistent delegate failure" anti-pattern.

## Gap

The trade-off lost the "transient transport blip + recovery" path. If `migrate_signing_key` returns `Failed` because the WebSocket was briefly down, and then the connection comes back mid-session, delegate-driven secret rotation stays disabled until the user reloads the tab. `ensure_room_subscription_once` is only called once per room from the load-rooms path — no retry loop, no signal-driven re-fire.

## Suggested fix

Hook an `on_reconnect` / `on_synchronizer_status_change` listener that, when transitioning from Disconnected/Error → Connected, retriggers `ensure_room_subscription_once` for all owned private rooms (with the existing per-session dedup intact — only fires if the previous attempt failed).

Alternatively: detect via the response handler when a subsequent contract notification arrives without the corresponding subscription record on the delegate side, and re-trigger then.

## Severity

Medium. The fallback ("user reloads the tab") works in practice. This is a polish item — visible to users who experience network blips during the load-rooms window but is recoverable without code changes.

[AI-assisted - Claude]

sanity:
## Problem

PR #276's Skip-on-Failed branch closed the "noisy retry on persistent delegate failure" loop, but lost the "transient transport blip + recovery" path (#277).

The per-session `ENSURE_SUBSCRIPTION_SENT` dedup set is process-global and survives a WebSocket reconnect, but the node-level contract subscriptions it tracks do **not** — a transport drop tears them down. A room whose `EnsureRoomSubscription` succeeded before a blip stays marked "done" in the dedup set, so the reconnect's load-rooms pass returns `Ok(false)` (dedup hit) and the delegate is **never re-subscribed**. Delegate-driven secret rotation (back-filling `encrypted_secrets` for newly-invited members) then stays silently disabled for the rest of the session until the user reloads the tab.

The same gap also covers the issue's stated case where a transient blip made `migrate_signing_key` return `Failed` and the chained `EnsureRoomSubscription` was skipped.

## Approach

Clear the whole dedup set in the synchronizer's `ConnectionLost` handler (the issue's suggested `on_reconnect` hook). On the subsequent `Connect`, `set_up_chat_delegate()` re-fires the load-rooms pass, which re-runs `ensure_room_subscription_once` for every owned private room.

This does **not** re-introduce the retry storm PR #276 fixed: clearing only re-arms the dedup; the actual re-subscribe still flows through the existing once-per-reconnect load-rooms path (gated on a non-`Failed` signing-key migration), not a tight retry loop. The `#[cfg(test)]`-only `reset_ensure_subscription_dedup` is promoted to a production helper with the reconnect rationale documented.

UI-only change — no `common/`, `delegates/`, `contracts/`, or Cargo dependency changes, so **no delegate/contract WASM migration is required**.

## Testing

- `ensure_subscription_dedup_reset_on_connection_lost_allows_resubscribe` — models the bug: a key marked "sent" stays a dedup hit without the clear (the #277 symptom), and becomes re-insertable after the clear.
- `connection_lost_resets_ensure_subscription_dedup_wiring_pinned` — source-pins the `ConnectionLost` arm's call to `reset_ensure_subscription_dedup()` (verified fails-before / passes-after by temporarily removing the call). Guards against silent regression, mirroring the existing `rejoin_flag_wiring_pinned` precedent.
- Full `cargo test -p river-ui --bins`: 357 passed, 0 failed.

Closes #277

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -405,16 +405,56 @@ pub async fn set_up_chat_delegate() -> Result<(), String> {
     }
 }
 
-/// Per-session dedup set for `EnsureRoomSubscription` calls. The UI may
+/// Per-session dedup map for `EnsureRoomSubscription` calls. The UI may
 /// re-fire its load-rooms path on every `rooms_data` reload, but we only
 /// need to ask the delegate to (re-)subscribe each room once per session —
 /// the delegate's secret store keeps the sub_index across `process()`
 /// invocations.
-static ENSURE_SUBSCRIPTION_SENT: std::sync::LazyLock<Mutex<HashSet<RoomKey>>> =
-    std::sync::LazyLock::new(|| Mutex::new(HashSet::new()));
+///
+/// The value is the `request_id` of the call that currently owns the slot.
+/// It makes the failure-path clear (`clear_ensure_subscription_sent`)
+/// epoch-aware: a clear only removes the entry if it still belongs to the
+/// caller doing the clearing. Without this, a slow in-flight call that
+/// times out AFTER a reconnect re-armed the dedup (freenet/river#277) and a
+/// newer call re-claimed and succeeded would wrongly erase the newer call's
+/// success marker, causing avoidable repeat `EnsureRoomSubscription` calls
+/// on later reloads (codex review on PR #364). Same epoch-correlation shape
+/// as the `request_id` mix-in already used for `PENDING_REQUESTS`.
+static ENSURE_SUBSCRIPTION_SENT: std::sync::LazyLock<Mutex<HashMap<RoomKey, RequestId>>> =
+    std::sync::LazyLock::new(|| Mutex::new(HashMap::new()));
 
-/// Reset the per-session dedup set. Used in tests; not called in production.
-#[cfg(test)]
+/// Clear the entire per-session ensure-subscription dedup set so the next
+/// load-rooms pass re-fires `ensure_room_subscription_once` for every owned
+/// private room.
+///
+/// Called from the synchronizer before each load-rooms pass on a
+/// reconnect/wake (freenet/river#277), at the two entry points that re-run
+/// `set_up_chat_delegate()`:
+///  - the `Connect` handler (a real transport drop, or the sleep/wake
+///    `PageBecameVisible` check finding the connection dead);
+///  - the `RefreshAllRooms` handler (the sleep/wake `PageBecameVisible` check
+///    with an apparently-live connection, which refreshes instead of
+///    reconnecting — yet the node-side subscription may have died while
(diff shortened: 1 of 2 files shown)


sanity:
## Review summary (FULL tier — delegate subscription)

**External model (codex review, OpenAI) — ran across 3 rounds as the code evolved; final HEAD is clean.**

- Round 1 (clear in `ConnectionLost`): no findings.
- Round 2 (clear moved to `Connect`): **P3** — failure-path clear could race an in-flight call and erase a newer call's success marker after a reconnect re-armed the dedup.
- Round 3 (current HEAD): no findings — *"The changes correctly re-arm the EnsureRoomSubscription deduplication on reconnect/wake paths and make failure cleanup epoch-aware. I did not find any discrete regressions introduced by this patch."*

A separate round also surfaced a **P2**: the sleep/wake `PageBecameVisible` path that finds the connection apparently-live routes to `RefreshAllRooms` (not `Connect`), so the original `Connect`-only clear missed it.

**Findings addressed:**

- **P2 (RefreshAllRooms wake path):** re-arm + re-run the load-rooms pass in the `RefreshAllRooms` handler too, since it does not route through `Connect`. Pinned by `reconnect_resets_ensure_subscription_dedup_before_load_rooms_wiring_pinned` (asserts BOTH arms call the reset before `set_up_chat_delegate()`).
- **P3 (epoch race):** dedup changed from `HashSet<RoomKey>` to `HashMap<RoomKey, RequestId>`; the failure-path clear now removes the entry only when the slot is still owned by the clearing call. Pinned by `clear_ensure_subscription_sent_is_epoch_scoped`.

**Claude-lens adversarial pass (supplementing the external review for this high-risk surface):** checked Dioxus signal-safety (the dedup is a plain `Mutex`, not a signal — no `defer()` needed), `RefreshAllRooms` idempotency, clear-before-trigger ordering on both arms, epoch correctness under whole-map reset, and the first-connection no-op. No issues.

All three regression tests verified fails-before / passes-after by temporarily reverting each mechanism. `cargo test -p river-ui --bins`: 358 passed, 0 failed.

[AI-assisted - Claude]

PR #369, opened 2026-06-21
fix(ui): close public→private TOCTOU window in nickname save (#318)

### LINKED ISSUE #318: Nickname save race: public→private mid-edit can still leak plaintext member_info
## Problem (surfaced during PR #317 review)

PR #317 fixed the freenet/river#299 leak where a private-room nickname edit could publish `SealedBytes::public` if the room secret hadn't arrived yet. A narrower follow-up race remains:

1. User opens the member-info modal in a **public** room and starts editing their nickname.
2. `NicknameField::save_changes` runs synchronously: reads `is_private = false`, no secret, calls `seal_for_room(false, None, ...)` → returns `Some(SealedBytes::public(...))` (correct for a public room).
3. The closure builds the delta and schedules a `crate::util::defer(...)` (`setTimeout(0)`) closure to apply it.
4. Between steps 3 and 4, a network message arrives changing the room to **private** mode and `ROOMS.with_mut` applies that config update.
5. Our deferred closure fires — `ROOMS.with_mut` now sees a private room, but our pre-built `delta` carries a plaintext `member_info` entry. `apply_delta` accepts it (`common/src/room_state/member_info.rs` validates declared length and signature only — no privacy check). `mark_needs_sync` then publishes the plaintext into the now-private room.

Window is tight (single event-loop tick, typically ~4ms) but real.

## Why this only affects nickname (not room name / description)

`ui/src/components/room_list/room_name_field.rs` and `edit_room_modal.rs::RoomDescriptionField` publish into `Configuration`. `common/src/room_state/configuration.rs::apply_delta` (around L112–115) rejects a `SealedBytes::Public` display-metadata write to a private room, so the bad delta would fail to apply locally and `mark_needs_sync` wouldn't fire. The leak is bounded to **member_info** for now.

## Fix direction

Move the seal+delta build into the deferred closure for `NicknameField`, so `is_private` / `current_secret_opt` are read at the same moment as `apply_delta`. With both reads in the same `ROOMS.with_mut` block the race window collapses to zero. If `seal_for_room` returns `None` inside the deferred block, surface the revert via `temp_nickname.set(...)` from inside the closure (a deferred-context signal write is safe).

A stronger defense-in-depth fix would be to add a privacy guard to `common/src/room_state/member_info.rs::apply_delta` mirroring `configuration.rs`: reject a `SealedBytes::Public` `preferred_nickname` write when the room is `PrivacyMode::Private`. That would also force a delegate/contract WASM migration but would close this race at the contract level for all current and future UI surfaces.

## Scope notes

- Pre-existing pattern, surfaced (not introduced) by PR #317.
- Narrow window in practice — the room owner has to flip privacy mode while a member is mid-edit.
- The matching `Configuration` apply-delta backstop should also be cross-referenced if/when the member_info guard is added so they stay consistent.

[AI-assisted - Claude]

sanity:
## Problem

PR #317 closed the wide freenet/river#299 leak, but a residual **TOCTOU race** remained in `NicknameField::save_changes` (`ui/src/components/members/member_info_modal/nickname_field.rs`):

1. The privacy read (`is_private` / `get_secret`), the `seal_for_room` call, and the `ChatRoomStateV1Delta` build all ran **synchronously**.
2. `apply_delta` + `mark_needs_sync` ran **later**, inside a `crate::util::defer` (`setTimeout(0)`) closure.
3. If a network message flipped the room **public→private** in that one-tick gap, the pre-built plaintext `SealedBytes::public` nickname was applied to the now-private room and synced out. `MemberInfoV1::apply_delta` has no contract-level privacy guard (it checks only declared length + signature + membership), so it accepted the leak.

The window is tight (~4ms, one event-loop tick) but real, and it leaks a member's plaintext nickname into a nominally-encrypted room.

## Approach

Move the privacy read, the seal, the delta build, **and** `apply_delta` all into the **single deferred `ROOMS.with_mut` block**, reading `is_private` / `get_secret` from the *same* `room_data` borrow the delta is applied to. The read and the apply are now atomic with respect to the event loop, so the race window collapses to zero — a public→private flip either lands fully before the closure (it seals public for a still-public room, correct) or fully after (the room is private and `seal_for_room` encrypts or defers).

The freenet/river#299 no-secret guard is preserved: in a private room with no local secret, `seal_for_room` returns `None`, the edit is dropped, and the input reverts to the on-network value via `temp_nickname.set(...)` — now a deferred signal write, which is a sanctioned pattern (cf. `room_list.rs` drag handlers).

### Why the client, not the contract

The issue notes a "stronger" defense-in-depth option: add a `PrivacyMode` guard to `MemberInfoV1::apply_delta` mirroring `configuration.rs`. **This PR deliberately does NOT take that route**, because it is **not backward-compatible**:

- `member_info.rs::apply_delta` is compiled **into the room-contract WASM**. Changing it moves the contract key (`BLAKE3(wasm, params)`) for every owner → requires a `legacy_room_contracts.toml` migration entry.
- A `return Err` guard would, via the `#[composable]` macro's `?`-propagation, abort the **entire** delta/merge — not just the offending entry. Any existing private room that already carries a public nickname (deployed WASM accepts these today) would have its state **rejected** by new-WASM nodes, diverging from old-WASM nodes. That violates the Backwards Compatibility Rule in `AGENTS.md`.

So the robust *and* backward-compatible enforcement point is the UI, which this PR fixes. The contract-level guard, if ever wanted, belongs in a separate migration-gated change.

## Testing

- New contract-level characterization test `issue_318_member_info_apply_delta_has_no_contract_privacy_guard` in `common/tests/private_room_test.rs` pins both halves of the invariant this fix relies on:
  1. `MemberInfoV1::apply_delta` **accepts** a public nickname in a private room today (no contract guard — the UI must be the gate). If someone adds the guard, this fails first and forces treating it as the coordinated WASM migration it is.
  2. The analogous `Configuration` write **is rejected** at the contract level (the existing guard this fix mirrors in spirit; why the leak only ever affected `member_info`).
- `cargo test -p river-core --test private_room_test` → 12 passed (11 pre-existing + new).
- `cargo test -p river-core --lib` → 181 passed.
- `cargo fmt` clean; `nickname_field.rs` is clippy-clean on the wasm target (zero warnings from the change).
- **No contract WASM change** (`ui/public/contracts/room_contract.wasm` untouched) → `check-room-contract-migration` passes, confirming backward compatibility.

## Compatibility

**Fully backward-compatible.** Client-only change; no contract/delegate WASM modification; no migration entry required.

Closes #318

[AI-assisted - Claude]

### DIFF
--- a/common/tests/private_room_test.rs
+++ b/common/tests/private_room_test.rs
@@ -1017,11 +1017,20 @@ fn issue_318_member_info_apply_delta_has_no_contract_privacy_guard() {
     let mut state_for_config = state.clone();
     let old = state_for_config.clone();
     let config_result = state_for_config.apply_delta(&old, &params, &Some(config_delta));
+    // Assert the SPECIFIC rejection reason, not just `is_err()`: the version
+    // bump, signature, and non-zero `max_*` all pass, so the privacy guard at
+    // `configuration.rs` is the only thing that should reject this delta. If
+    // some unrelated future validation starts erroring first, a bare
+    // `is_err()` would keep passing while silently no longer testing the
+    // privacy guard. Pinning the message keeps the test honest.
+    let config_err = config_result.expect_err(
+        "Configuration::apply_delta MUST reject public display metadata in a private room",
+    );
     assert!(
-        config_result.is_err(),
-        "Configuration::apply_delta MUST reject public display metadata in a \
-         private room (the existing contract-level guard #318's UI fix mirrors). \
-         If this no longer errors, the configuration privacy guard regressed.",
+        config_err.contains("encrypted display metadata"),
+        "expected the configuration privacy guard to reject this delta, but got \
+         a different error ({config_err:?}). The existing contract-level guard \
+         (#318's UI fix mirrors it) may have regressed or moved.",
     );
 }
 
--- a/common/tests/private_room_test.rs
+++ b/common/tests/private_room_test.rs
@@ -894,6 +894,137 @@ fn delegate_driven_rotation_round_trip() {
     }
 }
 
+/// freenet/river#318 contract-level invariant pin.
+///
+/// The UI nickname-save flow (`ui/.../nickname_field.rs`) is the privacy gate
+/// that keeps a plaintext `SealedBytes::Public` nickname out of a private
+/// room: it reads `is_private` and seals the nickname atomically with
+/// `apply_delta` so a public→private reconfiguration can't slip a stale
+/// plaintext delta through. That UI guard is load-bearing *because the
+/// contract does NOT enforce the same rule for `member_info`*.
+///
+/// This test pins both halves of that assumption so a future change can't
+/// silently invalidate it:
+///
+/// 1. `MemberInfoV1::apply_delta` ACCEPTS a `SealedBytes::Public`
+///    `preferred_nickname` even when the room is `PrivacyMode::Private`
+///    (there is no contract-level privacy guard for member_info — only
+///    nickname length + signature + membership are checked). If someone
+///    adds such a guard here, this assertion fails first and forces them
+///    to treat it as the coordinated room-contract WASM migration it is
+///    (the guard would move the contract key AND, because composed
+///    `apply_delta` propagates a sub-delta `Err` via `?`, reject the WHOLE
+///    delta from any existing room that already carries a public nickname —
+///    a CRDT-divergence / backwards-compat break per AGENTS.md). The robust
+///    place to enforce this without a migration is the UI, which #318 does.
+///
+/// 2. The analogous `Configuration` write IS rejected at the contract level
+///    (`configuration.rs::apply_delta`), which is why the leak window only
+///    ever affected member_info — room name / description writes fail to
+///    apply locally and never reach `mark_needs_sync`. This is the existing
+///    guard #318's UI fix mirrors in spirit.
+#[test]
+fn issue_318_member_info_apply_delta_has_no_contract_privacy_guard() {
+    use river_core::room_state::member::MembersDelta;
+    use river_core::room_state::member_info::{AuthorizedMemberInfo, MemberInfo, MemberInfoV1};
+    use river_core::room_state::ChatRoomStateV1Delta;
+
+    let owner_sk = SigningKey::generate(&mut OsRng);
+    let owner_vk = owner_sk.verifying_key();
+    let owner_id = MemberId::from(&owner_vk);
+    let params = ChatRoomParametersV1 { owner: owner_vk };
+
+    // A private room with one non-owner member.
+    let member_sk = SigningKey::generate(&mut OsRng);
+    let member_vk = member_sk.verifying_key();
+    let member_id = MemberId::from(&member_vk);
+
+    let mut state = ChatRoomStateV1 {
+        configuration: AuthorizedConfigurationV1::new(
+            Configuration {
+                privacy_mode: PrivacyMode::Private,
+                owner_member_id: owner_id,
+                ..Default::default()
+            },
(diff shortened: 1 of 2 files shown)


sanity:
## FULL-tier multi-model review — consolidated result

Treated as **Full / high-risk** tier (privacy/security surface + a contract-state question). All reviewers read the actual checked-out code, not just the diff.

### Reviewers run
- **External model — Codex (`codex review --base origin/main`): CLEAN.** "The changes keep nickname sealing and application within the same deferred room-state mutation, and the added regression test accurately documents the current contract behavior. I did not identify any introduced correctness, security, or compatibility issues in the diff."
- **External model — Gemini: UNAVAILABLE.** `gemini` CLI returned `IneligibleTierError` (free-tier client deprecated, "migrate to Antigravity"). Codex (the primary external non-Claude pass) ran successfully, so the independent-model requirement is met; the Claude lenses below add the extra perspective diversity for this high-risk change.
- **Five Claude lenses** (code-first, skeptical, testing, big-picture, dedicated privacy/migration-compatibility) — all **APPROVE, no blockers.**

### Verified claims (cross-checked by multiple reviewers against source)
- **Race is closed.** `is_private` / `get_secret` / `seal_for_room` / `apply_delta` are now atomic inside one synchronous `ROOMS.with_mut` borrow; WASM is single-threaded with no `.await` in the closure, so a public→private flip cannot interleave. `room_data.is_private()` reads the same `room_state` the delta mutates.
- **No new panic / RefCell re-entrancy.** Nothing inside `with_mut` re-borrows `ROOMS`/`CURRENT_ROOM`; `mark_needs_sync` and the `temp_nickname.set` revert run *after* the borrow is released (matches the established `let outcome = ROOMS.with_mut(...)` pattern).
- **#299 protections intact** (no-secret → `seal_for_room` returns `None` → edit dropped + input reverted). The `seal_for_room_call_sites_pinned` test still passes.
- **Backward-compatible / no migration.** Diff touches only `ui/.../nickname_field.rs` + `common/tests/private_room_test.rs`. `ui/public/contracts/room_contract.wasm` is byte-identical → `check-room-contract-migration` passes (and is green on this PR).
- **Declining the contract-level guard is the correct call.** Independently verified: `member_info.rs::apply_delta` is compiled into the room-contract WASM; the `#[composable]` macro `?`-propagates a sub-delta `Err`, aborting the *entire* merge → a new-WASM node would reject any existing private room already carrying a public nickname, diverging from old-WASM peers (a backwards-compat break per AGENTS.md). The new characterization test is a tripwire that forces a future contract guard to be done as a coordinated migration.

### Findings addressed
- **(fixed)** Testing lens — Part 2 of the new test asserted only `is_err()`; tightened to assert the specific `"encrypted display metadata"` error so it fails loudly if the config privacy guard regresses/moves. (commit e0825ba1)
- **(filed #370)** Privacy + testing lenses — `room_name_field.rs` / `edit_room_modal.rs` share the same build-outside/apply-inside TOCTOU shape, but are **bounded** (not fixed) by the `Configuration::apply_delta` guard, so they do not leak. Both reviewers agreed this is out of #318's scope (which fixes the only *leaking* surface). Tracked for defense-in-depth in #370.
- **(filed #370)** Skeptical lens — a pre-existing nickname double-save version-drop / cached-self-state divergence, unchanged by this PR. Tracked in #370.

### Scope boundary (deliberate)
This PR fixes the one surface that actually leaked (`member_info`, which has no contract-level privacy guard). The sibling Configuration surfaces are protected by the existing display-metadata guard; mirroring the atomic pattern there is defense-in-depth, tracked in #370.

No blocking findings remain. Merge gated on CI green at the current HEAD.

[AI-assisted - Claude]

PR #377, opened 2026-07-03
ci: tag-triggered riverctl release (crates.io + GitHub release)

sanity:
## Problem

The GitHub releases page on this repo is stale at **v0.1.22 (Dec 25, 2025)** while `riverctl` has kept shipping to crates.io (**0.1.71**, Jun 2026). The cause: publishing is a laptop-only `cargo make publish-all` that runs `cargo publish` but never creates a git tag or a GitHub release. So the releases page shows a misleading "Latest", offers no prebuilt binaries, and the crates.io publish happens from an unreviewed local checkout.

## Approach

Add `.github/workflows/release-riverctl.yml`, triggered by pushing a `riverctl-v<version>` tag. The tag is both the trigger and the version source of truth.

Jobs:
1. **verify** — refuses to run unless the tag version == `cli/Cargo.toml` version **and** the tagged commit is on `origin/main` (publish-from-main gate).
2. **publish-crates** — publishes `river-core` (only if that version isn't already on crates.io, waiting for sparse-index propagation) then `riverctl`. Idempotent/skip-if-exists, so a CLI-only release skips the unchanged `river-core`. Versions resolved via `cargo metadata` because `common/` (river-core) inherits `[workspace.package]` version.
3. **build-binaries** — cross-builds `riverctl` for x86_64 Linux, x86_64 + aarch64 macOS, x86_64 Windows (GitHub-hosted runners; repo is public so these are free).
4. **github-release** — creates the GitHub release with those binaries + `--generate-notes`.

**Scoped to riverctl only.** The River UI (Dioxus → Freenet, contract key `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`) keeps releasing independently via `cargo make publish-river` — they are not 1:1 and must stay decoupled.

Docs updated so future work uses the tag flow: a "Releasing riverctl" section + a "two independent release surfaces" table in `.claude/rules/river-publish.md`, a pointer in `AGENTS.md`, and a deprecation note on the local `publish-all` task.

## Required before this can publish

Add the crates.io token as a repo secret (not currently present in repo or org secrets):
```bash
gh secret set CARGO_REGISTRY_TOKEN --repo freenet/river
```
A crates.io **publish-update** token scoped to `river-core` + `riverctl`.

## Testing

- YAML validated; version extraction and `cargo metadata` resolution checked locally (river-core 0.1.12, riverctl 0.1.71).
- The full publish path is only exercised on a real tag push (needs the secret). First real release will supersede v0.1.22 as "Latest".
- No WASM files touched (verified per the repo's wasm-safety rule).

## Notes / open questions

- Existing stale releases (v0.1.22 etc.) are left as history; the first tagged release becomes "Latest". Say if you'd rather I delete the old ones.
- Binary target set is the standard four — easy to trim/extend (e.g. add linux musl static).
- I left `cargo make publish-all` functional but documented it as legacy/fallback rather than ripping `cargo publish` out of it, to avoid surprising anything that depends on it. Happy to remove the crates publish from it in a follow-up so there's a single publish path.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/river-publish.md
+++ b/.claude/rules/river-publish.md
@@ -80,13 +80,68 @@ git push origin main
 
 **Version policy:** the canonical source is `published-contract/contract-version.txt`, which `cargo make sign-webapp` increments and writes back each publish. NEVER base the version on wall-clock time (`date +%s / 60`) — the previous scheme bit us 2026-05-16 when the on-network version had drifted ahead of the timestamp-derived value. The counter file makes the version a strict monotonic local invariant; gaps are fine (the contract enforces monotonicity, not contiguity).
 
-## Publishing River UI + riverctl Together
+## Two independent release surfaces
 
-**CRITICAL:** When room contract WASM changes, both UI and riverctl must be republished — they embed the WASM and derive the contract key from it.
+The river repo publishes **two things on independent cadences** — do not conflate them:
+
+| Artifact | Where it goes | How it's released |
+|----------|---------------|-------------------|
+| **riverctl** (the CLI) | crates.io (`river-core` + `riverctl`) + GitHub release with prebuilt binaries | **tag-triggered CI** — see below |
+| **River UI** (Dioxus webapp) | Freenet (contract key `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`) | `cargo make publish-river` (the checklist above) |
+
+They are NOT 1:1 — a CLI-only fix ships riverctl without touching the UI, and a UI-only change publishes to Freenet without a new riverctl. The one coupling is room-contract WASM: when it changes, BOTH must be republished (they embed the WASM and derive the contract key from it).
+
+## Releasing riverctl (crates.io + GitHub release) — tag-triggered
+
+**Canonical path since 2026-07.** riverctl is released by pushing a git tag; the
+`.github/workflows/release-riverctl.yml` workflow then publishes to crates.io
+AND cuts the GitHub release (with prebuilt Linux/macOS/Windows binaries and
+auto-generated notes). This is why the GitHub releases page previously rotted at
+v0.1.22 while crates.io marched on — publishing used to be a laptop-only
+`cargo publish` that never touched GitHub.
 
 ```bash
-cargo make publish-all
+# 1. Bump the version in cli/Cargo.toml (via a normal PR), and — if the
+#    common/ crate (river-core) also changed — bump [workspace.package] version
+#    and riverctl's `river-core = { version = "..." }` requirement to match.
+#    Merge to main. (CI's check-cli-wasm guards the embedded room_contract.wasm.)
+
+# 2. From main, tag with the riverctl version and push the tag:
+git checkout main && git pull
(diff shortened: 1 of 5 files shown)


sanity:
## Multi-model review (Full tier — CI/deploy-config + supply-chain surface)

Ran an external **Codex** pass plus **3 Claude reviewers** with distinct lenses (skeptical bug-hunt, supply-chain/security, release-correctness+docs). Strong convergence. All findings fixed in `e71e6f72`; nothing dismissed silently.

### Fixed
| Sev | Finding | Fix |
|-----|---------|-----|
| major | crates.io "already published?" check used the JSON API via `curl \| grep -q` — pipefail/SIGPIPE-fragile (can misreport an already-published crate as missing → failed republish), and the JSON API is rate-limit/crawler-blocked per repo history | Read the **sparse index** (`index.crates.io`, what `cargo publish` actually resolves against) into a variable first; `grep -F`; `curl --retry`. The propagation wait now polls the index too. |
| major | crates.io publish ran in parallel with the fallible binary builds → a failed target could leave crates.io published with no GitHub release | `publish-crates` now `needs: build-binaries` — build first, publish (irreversible) last |
| major | `gh release create` not idempotent → re-run after any transient failure aborts with "already exists" | `gh release create … \|\| gh release upload … --clobber` |
| major | actions pinned to floating `@v4` tags (retarget risk in a token-bearing publish workflow) | SHA-pinned: checkout v4.3.1, upload-artifact v4.6.2, download-artifact v4.3.0 |
| minor | no top-level `permissions:` floor | `permissions: contents: read` top-level; `contents: write` only on `github-release` |
| minor | `rustup show` no longer installs the pinned toolchain on rustup ≥1.28 (GH runners) | `rustup toolchain install` (provisions 1.94.1 from `rust-toolchain.toml` explicitly) |
| minor | `${{ }}` version interpolated into `run:` shell (defense-in-depth) | pass `version`/`tag` via `env:` vars; explicit `origin/main` refspec fetch |
| minor | header comment misdescribed the crates.io token scope | corrected + least-privilege note |

### Confirmed sound (not bugs)
- Tag→shell flow is **injection-safe** — uses quoted `$GITHUB_REF_NAME`, not `${{ github.ref_name }}` template interpolation; `riverctl-v*` filter prevents leading-dash tags.
- The **from-main gate** (`git merge-base --is-ancestor $GITHUB_SHA origin/main`) is real and **fail-closed**; a tag on an unmerged commit is rejected.
- Version extraction matches only the package `version` line (not the `river-core` dep line); all four matrix targets are native per-runner (no cross-linker).

### Surfaced, not auto-fixed (your call — see PR thread)
- **Publish behind a GitHub Environment with required reviewers / tag-protection ruleset** — would block the "push tag → auto-publish" flow, so it's a policy choice.
- **Commit `Cargo.lock` + `--locked`** — conflicts with this repo's deliberate no-lockfile policy; a repo-wide decision, not this PR.
- **Scope the crates.io token to just `river-core`+`riverctl` with an expiry** — reduces blast radius if it leaks; currently an all-crates token by choice.
- **A CI-green-on-tagged-commit gate in the workflow** — deferred; main is gated by PR review and the release is only triggered from main, and I verify CI green on the exact commit before tagging.

[AI-assisted - Claude]

PR #380, opened 2026-07-06
fix(ui): keep self message bubbles inside the viewport on mobile

sanity:
## Problem

On narrow mobile viewports, **self (right-aligned) message bubbles spill off both edges of the screen**, clipping the message text on the left and right. This is the "text fields going off the edge of the screen" report.

Audited against a real populated peer at iPhone SE (375), iPhone 12/13 (390), small Android (360), and iPhone XR (414), portrait. The self bubbles measured **528px wide inside a 375px viewport** (`left: -169, right: 359`), i.e. right-aligned content clipped off the left edge. Received (gray, left-aligned) bubbles, the room list, the members list, and the create-room / edit-room / invite-member / member-info modals all measured clean at every viewport — the bug is specific to self message bubbles.

Before (real peer): a blue self message reads `org/build/manual/further-reading/…` with the leading `https://freenet.` clipped off the left, and the tail clipped off the right.

## Root cause

The self bubbles wrapper is `flex flex-col items-end`, so each per-message wrapper is a **non-stretched flex item that sizes to its content**. When a self bubble reaches its `max-w-prose` (65ch, ~528px) cap it **escapes the `max-w-[75%]` column**, and because it is right-aligned it overflows off the left edge (clipping on the right). The trigger in the wild is a self **reply** whose reply-strip preview holds a long unbreakable URL: `.reply-strip` renders `white-space: nowrap`, driving the bubble to its full max width. Received (left-aligned) bubbles were unaffected because their wrapper is a plain block that already fills the column.

This does **not** cause document-level horizontal scroll (the bubble is clipped), which is why the existing `#212` doc-overflow assertion never caught it.

## Approach

Add `min-w-0 max-w-full` to the per-message wrapper (`ui/src/components/conversation.rs`) so it can never exceed its column. `min-w-0` lets the flex item shrink below its content's min-size so `max-w-full` (`max-width: 100%`) actually binds. Minimal, no redesign.

Desktop is unchanged: on wide screens the column is wider than `max-w-prose`, so bubbles still cap at the 65ch readable width (verified: widest bubble still 528px at 1280px, no overflow).

## Testing

- **New Playwright regression test** (`ui/tests/message-layout.spec.ts`) at 375px: asserts no `.max-w-prose` bubble overflows either viewport edge. Runs on all 5 CI projects (chromium, firefox, webkit, mobile-chrome, mobile-safari).
- **Reproducing example data**: a self reply to the long-URL message (`ui/src/example_data.rs`). Without the fix that bubble measures `left -169 / right 359 / width 528` at 375px (the test's assertion is non-empty → fails); with the fix it is clamped to the column (`width 257`, no overflow → passes).
- Full existing layout suite green across all 5 projects (**112 passed, 3 skipped**), including `#205/#206/#207/#210/#212/#221` and the responsive-layout suite — no regression.
- After-fix verification across 360 / 375 / 390 / 414 / 768 / 1280: no bubble overflow at any width; desktop unchanged.
- `cargo test -p river-ui --bins`: 374 passed. `cargo fmt` clean.

## Migration

**UI-only change.** Only `ui/src/` and `ui/tests/` are touched — no `common/`, `delegates/`, `contracts/`, or `Cargo.toml`/`Cargo.lock` changes, and no `.wasm` files modified. **No delegate/contract migration required.**

Screenshots (saved locally for review): `before-*` / `repro-nofix-*` show the 528px overflow; `after-fix-*` / `fixed2-*` show it clamped.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -2294,10 +2294,20 @@ fn MessageGroupComponent(
                         let reply_target_id_val = msg.reply_to_message_id.clone();
 
                         rsx! {
+                            // `min-w-0 max-w-full` clamps this per-message wrapper to
+                            // its column (`max-w-[75%]`) width. For a SELF message the
+                            // enclosing bubbles wrapper is `flex flex-col items-end`, so
+                            // without this the wrapper is a non-stretched flex item that
+                            // sizes to the bubble's `max-w-prose` (65ch) content width and
+                            // escapes the column. A self reply whose nowrap reply-strip
+                            // preview holds a long URL then overflows both edges of a
+                            // narrow mobile viewport (clipped, text cut off). `min-w-0`
+                            // lets the flex item shrink below its content's min-size so
+                            // `max-w-full` can actually take effect.
                             div {
                                 key: "{msg.id}",
                                 id: "msg-{msg.id}",
-                                class: "flex flex-col group",
+                                class: "flex flex-col group min-w-0 max-w-full",
                                 // Container for message bubble + hover actions
                                 div {
                                     class: "relative",
--- a/ui/src/example_data.rs
+++ b/ui/src/example_data.rs
@@ -403,6 +403,49 @@ fn add_example_messages(
         messages.messages.push(long_url_msg);
     }
 
+    // A SELF (owner-authored) reply whose quoted preview is the long URL above.
+    // The reply strip renders its preview with `white-space: nowrap`, so an
+    // unbreakable long URL in the preview drives the bubble to its full
+    // `max-w-prose` width. On a self (right-aligned) message that used to
+    // overflow the viewport on narrow mobile screens, clipping the bubble off
+    // both edges. Kept in example data so the mobile-overflow Playwright
+    // regression test has a self bubble that WOULD overflow without the
+    // per-message width clamp.
+    if let Some(long_url_msg) = messages.messages.last().cloned() {
+        let target_id = long_url_msg.id();
+        let target_author_id = long_url_msg.message.author;
+        let target_author_name = room_state
+            .member_info
+            .member_info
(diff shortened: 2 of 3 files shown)


sanity:
## Multi-model review (Light tier)

UI-only CSS/markup change, small and self-contained (not a high-risk surface), so Light tier: one external (non-Claude) model pass plus an adversarial read.

**External — Codex (`codex review --base main`):** clean. Verdict: *"narrowly scoped to constraining message bubble wrappers and adding regression coverage/example data for the mobile overflow case. I did not identify any introduced functional regressions or test-breaking issues."*

**Adversarial read (Claude):** no issues found.
- Received (left-aligned) bubbles use a plain block wrapper, so `max-w-full` is a no-op there and `min-w-0` is harmless — confirmed by `#206`/`#212` still passing.
- The inline edit container (`width:100%; max-width:550px`) now also clamps to the column on narrow screens, which is strictly better; `#205` (edit-box-fits-viewport) still passes.
- The example-data self reply verifies and renders in all 112 Playwright runs, and does not disturb `#212`'s `.first()` selector (the received long-URL bubble precedes it in DOM order).
- Desktop bubbles still cap at `max-w-prose` (~528px) — no desktop regression (verified at 1280px).

**Tests:** full layout suite green on all 5 projects (chromium/firefox/webkit/mobile-chrome/mobile-safari) — 112 passed, 3 skipped (hover test, by design); new 375px regression test passes on all 5; `cargo test -p river-ui --bins` 374 passed; `cargo fmt` clean. The new test was validated against the pre-fix build (the reproducer bubble measured `left -169 / right 359 / width 528` at 375px, failing the assertion).

Waiting on `build` + `ui-playwright-tests` CI before merge.

[AI-assisted - Claude]

Issue #382, opened 2026-07-06
River UI live-update stream can silently stall (frozen tab, no auto-reconnect)


sanity:
## Symptom

Viewing a River room through **try.freenet.org**, the room appeared **frozen** at the page-load time (~7:13am CT 2026-07-06): no new messages arrived in that tab, while the **same room on another peer showed current messages**. A hard reload fixed it.

## Root cause (diagnosed)

The **node's copy was fresh the entire time** — this is NOT a hosting/staleness bug at the node. During the freeze the hosted node kept applying the room's updates and re-broadcast every one to ~28-30 downstream peers (fan-out only happens after a successful local store write, so this proves the local copy was current). Node/framework state is byte-identical now.

The freeze is at the **client / WebSocket-notification layer**: the browser tab's live-update stream stopped delivering/applying `UpdateNotification`s while the node kept sending them. Two likely mechanisms:

1. **Half-open / silently-dead WebSocket** — the socket died without a TCP reset, so the node logged no disconnect, kept the client in its notifier map, and kept `try_send`-ing into a channel the browser was no longer reading/rendering.
2. **Subscription/notifier registration silently didn't take effect for that page load** — the GET snapshot rendered, but the live-notification registration never became active, pinning the tab to the initial snapshot.

Either way, the UI **sits frozen with no recovery** and the user has no indication their view is stale — especially bad for hosted (try.freenet.org) users who leave a tab open.

## Proposed fix

Make the River UI WS/sync client detect a dead or stalled subscription and **auto-recover**:
- **Liveness detection**: a heartbeat / watchdog on the WS connection (e.g. no traffic within N seconds triggers a health check), and/or detect a closed/errored socket that isn't currently being retried.
- **Auto-reconnect + re-GET**: on detecting a dead/stale connection, reconnect the WebSocket, re-establish the room subscription(s), and re-GET current state so the tab catches up — instead of sitting on the stale snapshot.
- Surface a subtle "reconnecting…" indicator so a stalled view is visible rather than silently wrong.

Investigate the existing reconnect logic in the UI's WS/sync layer first; part of this may already exist but not cover the half-open case.

## Notes

- UI-only (WS/sync client + reactive state); no contract/delegate WASM change expected → no room migration.
- Follows the Dioxus signal-safety rules (this is reactive WASM code).
[...]

PR #383, opened 2026-07-06
fix(ui): auto-recover stalled live-update WebSocket (#382)

sanity:
## Problem

Viewing a River room through try.freenet.org, a tab can silently **freeze**: no
new messages arrive even though the same room on another peer is current, and a
hard reload is the only fix (freenet/river#382).

Root cause is at the client / WebSocket layer, not the node. The stdlib `WebApi`
wires `onclose` and `onerror`, both of which River already maps to
`ConnectionLost` → reconnect + re-subscribe + re-GET. That covers a socket that
closes or errors cleanly. A **half-open** socket does not: when the underlying
TCP connection dies without a FIN/RST (NAT/idle timeout, a silent network drop),
the browser fires **neither** event and `WebSocket.readyState` stays `OPEN`.
River keeps believing it is `Connected`, the node keeps sending
`UpdateNotification`s into a channel nothing reads, and the tab sits frozen with
no recovery and no indication. The one existing trigger that could help —
`PageBecameVisible` — never fires for a tab simply left open and visible.

## Approach

Add a **liveness watchdog** for the established WebSocket
(`connection_watchdog.rs`):

- **Track inbound activity.** Every inbound WS message (success or error) and
  every connection-open bumps a `LAST_WS_ACTIVITY_MS` timestamp and a monotonic
  `WS_ACTIVITY_SEQ` counter. Plain atomics, not Dioxus signals.
- **Probe on silence.** After `LIVENESS_IDLE_PROBE_MS` (60s) with no inbound
  traffic, send a `ContractRequest::Get` of the room the user is viewing
  (`subscribe: false`). The node is already subscribed to that room, so it
  answers from its **local** fresh copy — the round-trip reflects WS/node
  health, not network routing. Silence alone is not proof of death (an idle room
  legitimately receives nothing), so the probe is the discriminator.
- **Escalate on no reply.** If no message arrives within
  `LIVENESS_PROBE_TIMEOUT_MS` (20s) of the probe, treat the socket as dead and
  raise `ConnectionLost`, driving the existing reconnect + re-subscribe + re-GET
  path.

**The GET probe is deliberate double-duty:** its response is merged idempotently
by `handle_get_response` (the same CRDT merge every GET uses), so beyond proving
liveness it also **resyncs** the room. That covers the *second* #382 mechanism —
a live socket that silently stopped delivering notifications — un-freezing the
view within one probe interval even when no reconnect is needed.

### Anti-storm / anti-false-positive design

- Dormant unless `SYNC_STATUS == Connected` — while disconnected/reconnecting the
  existing backoff owns recovery, so the watchdog can never add reconnect
  pressure. Declaring death raises `ConnectionLost` exactly once; status then
  flips to `Disconnected` and the watchdog goes dormant until reconnected.
- **At most one probe outstanding** per idle period; any inbound message (the
  probe reply or a real update) clears it. A healthy idle tab costs one local
  GET per minute.
- If there is **no room to probe**, the watchdog never declares death (nothing
  was sent, so silence proves nothing) — no false reconnect.
- The 20s timeout is far larger than a local GET's real latency, so a
  momentarily slow node cannot trigger a false reconnect. A false positive, if
  it ever happened, is one recoverable, backoff-bounded reconnect.

### Visible indicator

Recovery flows through the **existing** `ConnectionStatusIndicator` pill: on a
watchdog-detected stall it visibly cycles Connected → Disconnected → Connecting
→ Connected, so a stalled view becomes visible (and self-corrects) rather than
silently wrong. No new UI element was added.

### Signal-safety compliance

- No new Dioxus **signal mutations**. The watchdog only stamps atomics, reads
  `SYNC_STATUS`/`CURRENT_ROOM`/`ROOMS` via `try_read()`, sends over the existing
  mpsc channel, and sends the probe via `WEB_API` — the same
  `WEB_API.write().as_mut()` + `send().await` pattern
  `follow_upgrade_pointer_if_needed` already uses (safe because the browser
  `WebApi::send` is `async` but contains no `.await`, so it never yields to let
  another task re-borrow `WEB_API`).
- The loop is launched via `safe_spawn_local` (setTimeout-deferred), never raw
  `spawn_local` from inside the polled message-loop future.

## Review findings addressed (Codex)

- **[P3] NodeQueries blocked for contract web apps.** The first draft probed with
  `NodeQuery::ConnectedPeers`. freenet-core blocks `NodeQueries` from contract
  origins (the published UI is a contract web app), which would have spammed
  node/client error logs every interval. Switched the probe to a
  contract-app-allowed room `Get` (which also gives the resync bonus above).
- **[P2] Same-millisecond / coarse-clock probe reply.** The first draft compared
  timestamps (`last_activity_ms > probe_sent_ms`), which could miss a reply
  delivered within the same coarse `Date.now()` tick (Firefox RFP / Tor round to
  100ms) and falsely reconnect. Replaced with a monotonic inbound-message
  **sequence counter** that is immune to clock resolution.

## Testing

- **Unit tests (deterministic):** 8 tests pin the pure `watchdog_tick` state
  machine — dormant-while-disconnected, recent-activity-waits,
  idle-threshold-probes, coarse-clock-safe reply-recovery, within-timeout-waits,
  timeout-reconnects, reply-preferred-over-timeout-same-tick, and a full
  idle→probe→death cycle. `cargo test -p river-ui --bins`: 382 passed. `cargo
  clippy -p river-ui --all-targets -- -D warnings`: no new findings vs main.
- **Headless smoke test:** built `example-data,no-sync`, served, loaded in
  headless Chromium — WASM instantiates, UI renders all rooms + the connection
  indicator, **zero panics / instantiate errors**.
- **Playwright suite:** `ui/tests` across Chromium/Firefox/WebKit/mobile-Chrome/
  mobile-Safari — 272 passed. The only non-passing UI-behaviour test locally,
  `message-layout.spec.ts:499` (a #380 self-message-bubble-overflow assertion),
  **reproduces identically on an unmodified `f80b2e9e` (this branch's parent)
  build with the same local browser**, so it is a pre-existing / local-browser
  issue, not a regression from this sync-only diff. (The sandboxed-iframe specs
  hardcode `src=http://localhost:8082` and pass once served on that exact port.)
  CI runs the suite on pinned browsers as the authoritative gate.

### What is NOT deterministically tested (honest scope)

The true half-open-WS stall cannot be reproduced in the headless harness (it
requires a TCP connection that dies without a FIN/RST). The unit tests prove the
**decision logic** (when to probe, when to declare dead, when to recover); they
do not exercise a real dead socket. End-to-end confirmation needs live/manual
validation: leave a try.freenet.org tab open until its socket goes half-open and
confirm it auto-recovers within ~90s (per the issue's diagnostic tip:
`riverctl message list` against the hosted node vs another peer while the tab is
frozen). The probe/timeout constants are conservative and easy to tune from
production behaviour.

## UI-only — no migration

Only `ui/src/` changed. Contract and delegate WASM (`room_contract.wasm`,
`chat_delegate.wasm`) are byte-identical to `origin/main` (`git diff` clean), so
there is no delegate/room migration and no republish coupling.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/freenet_api.rs
+++ b/ui/src/components/app/freenet_api.rs
@@ -5,6 +5,7 @@
 
 pub mod backward_probe;
 pub mod connection_manager;
+pub mod connection_watchdog;
 pub mod constants;
 pub mod error;
 pub mod freenet_synchronizer;
--- a/ui/src/components/app/freenet_api/connection_manager.rs
+++ b/ui/src/components/app/freenet_api/connection_manager.rs
@@ -79,6 +79,11 @@ mod imp {
             let web_api = WebApi::start(
                 websocket.clone(),
                 move |result: Result<HostResponse, ClientError>| {
+                    // Every inbound message — success OR error — proves the
+                    // socket delivered bytes, so it feeds the liveness watchdog
+                    // (freenet/river#382). Recorded before any early return.
+                    crate::components::app::freenet_api::connection_watchdog::record_ws_activity();
+
                     // Check for AUTH_TOKEN_INVALID error - this means the node was restarted
                     // and we need to refresh the page to get a new valid token
                     if let Err(ref e) = result {
@@ -195,6 +200,10 @@ mod imp {
                 {
                     move || {
                         info!("WebSocket connected successfully");
+                        // Reset the liveness clock on (re)connect so the
+                        // watchdog doesn't treat a freshly-opened, still-idle
+                        // socket as long-silent (freenet/river#382).
+                        crate::components::app::freenet_api::connection_watchdog::record_ws_activity();
                         // Clear the reload-loop guard so a future node restart can
                         // trigger a fresh reload without being falsely detected as a loop.
                         if let Some(window) = web_sys::window() {
--- a/ui/src/components/app/freenet_api/connection_watchdog.rs
+++ b/ui/src/components/app/freenet_api/connection_watchdog.rs
@@ -1,0 +1,561 @@
+// Much of this module is compiled but unused on native (the watchdog only runs
+// against a real browser WebSocket; its call sites are `#[cfg(wasm32)]`). The
+// pure state machine is still exercised by the native unit tests below. Mirror
+// the sibling freenet_api modules (`constants.rs`, `freenet_synchronizer.rs`),
+// which blanket-allow dead_code for the same wasm-only reason.
+#![allow(dead_code)]
+
+//! WebSocket liveness watchdog (freenet/river#382).
+//!
+//! ## The bug
+//!
+//! River opens a WebSocket to the local Freenet node and receives live room
+//! updates as `UpdateNotification`s. The stdlib `WebApi` wires `onclose` and
+//! `onerror`, both of which River maps to a `ConnectionLost` message that
+//! drives reconnect + re-subscribe + re-GET. That covers a socket that closes
+//! or errors cleanly.
+//!
+//! A **half-open** socket does not. When the underlying TCP connection dies
+//! without a FIN/RST (NAT/idle timeout, a silent network drop), the browser
+//! fires **neither** `onclose` nor `onerror` — `WebSocket.readyState` stays
+//! `OPEN`. River keeps believing it is `Connected`, the node keeps
+//! `try_send`-ing notifications into a channel nothing reads, and the tab sits
+//! frozen on its last snapshot with no recovery and no indication to the user.
+//! The only recovery trigger that could help — `PageBecameVisible` — never
+//! fires for a tab that is simply left open and visible.
+//!
+//! ## The fix
+//!
+//! Track the timestamp/sequence of the last inbound WS message (proof the
+//! socket is alive). After a period of silence, send a cheap probe: a
+//! `ContractRequest::Get` of the room the user is viewing (`subscribe: false`).
+//! The node is already subscribed to that room, so it answers from its **local**
+//! fresh copy — the round-trip reflects WS/node health, not network routing
+//! (unlike a probe that has to route toward a far key). The probe is deliberate
+//! double-duty:
+//!   1. **Liveness** — a reply proves the socket is alive; no reply within a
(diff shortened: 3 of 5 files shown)


sanity:
## Review summary (external model: Codex)

Ran `codex review --base main` twice (initial + re-review after the fixes below). Findings and resolutions:

**Round 1**
- **[P3] `NodeQueries` blocked for contract web apps** — the first draft probed with `NodeQuery::ConnectedPeers`, but freenet-core blocks `NodeQueries` from contract origins (`websocket.rs`: *"NodeQueries is not available to contract web applications"*), which the published gateway-served UI is. That would have spammed node/client error logs every interval. **Fixed:** switched the probe to a `ContractRequest::Get` of the current room (contract-app-allowed), which also gives a resync bonus.
- **[P2] same-millisecond / coarse-clock probe reply** — the first draft compared `Date.now()` timestamps, which can miss a reply delivered within the same coarse tick (Firefox `resistFingerprinting` / Tor round to 100ms) and falsely reconnect. **Fixed:** probe-reply detection now uses a monotonic inbound-message **sequence counter**, immune to clock resolution.

**Round 2 (on the fixed content)**
- **[P2] "close the watchdog-expired socket before reconnecting"** — Codex worried the `ConnectionLost` handler only `take()`s `WEB_API` without calling `WebSocket.close()`, leaking the still-OPEN half-open socket. **Not applicable (verified):** the stdlib `WebApi` has a `Drop` impl that calls `conn.close_with_code(1000)`, so `WEB_API.write().take();` drops the taken `WebApi` synchronously and **closes the underlying socket** — including the still-OPEN half-open case. A closed socket delivers no further messages, so there is no duplicate-live-socket / stale-callback scenario; the watchdog deliberately reuses the same `ConnectionLost` path every other reconnect trigger uses. (See `freenet-stdlib .../client_api/browser.rs` `impl Drop for WebApi`.)

Plus my own adversarial read (reconnect-storm, signal-safety, the `WEB_API.write()`-across-`await` borrow, dropping fresh state on re-GET) — no further issues; details in the PR description.

[AI-assisted - Claude]

PR #384, opened 2026-07-06
fix(ui): guard watchdog against duplicate ConnectionLost on message-loop lag

sanity:
## Problem

The #382 liveness watchdog (merged in #383) can raise `ConnectionLost` **more than once for a single stall episode**. After the `Reconnect` arm fires `ConnectionLost`, `watchdog_tick` reset state to `Monitoring` but nothing prevented it re-arming. The watchdog only becomes dormant once `SYNC_STATUS` flips to `Disconnected`, which happens **asynchronously** in the synchronizer message loop (`freenet_synchronizer.rs:239-244`). If that loop is busy longer than a probe tick plus the probe timeout (~30s), the watchdog re-probes and fires a **second** `ConnectionLost`; each one independently calls `schedule_reconnect` → increments `consecutive_failures` → schedules another `Connect`. Result: overlapping reconnects and inflated backoff, and the module's "declaring death raises `ConnectionLost` exactly once" doc-claim was merely *usually* true.

Found as a MEDIUM by independent review of #383.

## Approach

Add a `WatchdogState::ReconnectRequested { activity_seq_at_reconnect }` latch to the pure `watchdog_tick` state machine (the reviewer's option #2, hardened). On declaring death the machine latches instead of returning to `Monitoring`, and refuses to fire again until it **observes the reconnect actually happen**:

- **fresh inbound WS activity** — `activity_seq` advances past the snapshot (a reconnect's `onopen` calls `record_ws_activity`, as does any inbound message), **or**
- **the connection leaving `Connected`** — the pre-existing `!connected` dormant branch at the top of `watchdog_tick`.

Because the latch releases on **both** paths, it can never permanently suppress recovery — not even under a fast reconnect the watchdog never samples as disconnected (the `onopen` activity bump releases it). Invariant guaranteed: **at most one `ConnectionLost` per stall episode**.

The fix stays entirely inside the pure, unit-testable state machine — **no new atomics, no new signals, no new signal writes** (the wasm loop still only does atomics + `try_read()` + channel send, per `dioxus-signal-safety.md`). **UI-only**: no `delegates/`, `contracts/`, or `common/` code touched, so no migration; contract/delegate WASMs are byte-identical to `origin/main`.

Also adds a brief code comment at the liveness-decision site acknowledging the known **LOW**: inbound decode is async (`onmessage → FileReader → onloadend → result callback → record_ws_activity`), so a main-thread stall longer than the probe timeout while `Probing` could yield a spurious reconnect. That reconnect is recoverable and CRDT-safe (the probe GET resyncs idempotently), so behavior is intentionally unchanged — the comment records it as understood, not overlooked.

## Testing

- **`at_most_one_connection_lost_per_stall_then_recovers`** (regression test): drives the machine across a lagged window (SYNC_STATUS still `Connected`, no inbound activity) and asserts **exactly one** reconnect; then simulates a real reconnect (activity bump) and a **second independent stall**, asserting recovery fires again — proving the latch is not a permanent gag. Verified it **FAILS without the guard** (pre-fix fires 3 reconnects in the window) and **passes with it**.
- Three focused unit tests pin the latch: holds-and-does-not-refire, releases-on-fresh-activity, releases-when-disconnected.
- The existing 8 `watchdog_tick` tests updated for the new post-death state (`ReconnectRequested` instead of bare `Monitoring`).
- `cargo test -p river-ui --bins` green (386 passed); `cargo fmt --check` clean; `cargo clippy` introduces no new warnings on the file; wasm32 build compiles.

Refs #382, #383

https://claude.ai/code/session_012jmCd1tvDH1EsVTtaDVnxJ

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/freenet_api/connection_manager.rs
+++ b/ui/src/components/app/freenet_api/connection_manager.rs
@@ -200,10 +200,15 @@ mod imp {
                 {
                     move || {
                         info!("WebSocket connected successfully");
-                        // Reset the liveness clock on (re)connect so the
-                        // watchdog doesn't treat a freshly-opened, still-idle
-                        // socket as long-silent (freenet/river#382).
-                        crate::components::app::freenet_api::connection_watchdog::record_ws_activity();
+                        // Record a new socket generation on (re)connect. This
+                        // both resets the liveness clock (so the watchdog doesn't
+                        // treat a freshly-opened, still-idle socket as
+                        // long-silent) AND advances the connection-generation
+                        // counter that releases the watchdog's ReconnectRequested
+                        // latch — proof that a reconnect actually landed, as
+                        // opposed to a straggler message on the old condemned
+                        // socket (freenet/river#382; Codex #384 review).
+                        crate::components::app::freenet_api::connection_watchdog::record_ws_connected();
                         // Clear the reload-loop guard so a future node restart can
                         // trigger a fresh reload without being falsely detected as a loop.
                         if let Some(window) = web_sys::window() {
--- a/ui/src/components/app/freenet_api/connection_watchdog.rs
+++ b/ui/src/components/app/freenet_api/connection_watchdog.rs
@@ -56,8 +56,19 @@
 //! * At most **one** probe is outstanding per idle period. A healthy idle room
 //!   costs one local GET per [`LIVENESS_IDLE_PROBE_MS`]; any inbound traffic
 //!   (the probe's own reply, or a real update) clears the probe.
-//! * Declaring death raises `ConnectionLost` exactly once; `SYNC_STATUS` then
-//!   flips to `Disconnected` and the watchdog goes dormant until reconnected.
+//! * Declaring death raises `ConnectionLost` **exactly once per stall episode**.
+//!   On declaring death the watchdog latches into
+//!   [`WatchdogState::ReconnectRequested`] and refuses to raise a second
+//!   `ConnectionLost` until it observes the reconnect actually happen — either
+//!   `SYNC_STATUS` leaving `Connected`, or fresh inbound WS activity (a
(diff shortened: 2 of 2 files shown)


sanity:
## Review round 1 (Codex) — addressed

Codex flagged a **P2** on the first revision: the `ReconnectRequested` latch released on *any* inbound activity (`activity_seq` moving), so a straggler message delivered late on the **old, condemned socket** could spring the latch mid-lag, letting the watchdog re-probe and fire a second `ConnectionLost` — reopening the very duplicate-reconnect gap the latch closes.

**Fix (commit `7e880a70`):** added a distinct connection-generation counter `WS_CONNECT_SEQ`, bumped **only** by the `onopen` callback (`record_ws_connected`, wired at the single onopen site). The latch now releases when a genuinely **new socket opens** (`connect_seq` advances) or when the connection **leaves `Connected`** — never on old-socket message traffic. This still covers the fast-reconnect case (onopen bumps `connect_seq` even if the watchdog never samples the brief `Disconnected` window), so recovery is never permanently suppressed.

Two regression tests pin it: `reconnect_requested_ignores_old_socket_straggler` (single tick) and `straggler_on_old_socket_during_lag_never_refires` (driven window). Both **fail** under the old activity-keyed release (the driven one fires 6 reconnects) and pass with the connection-generation key.

Re-running the review on the updated content per the per-content review rule.

[AI-assisted - Claude]

sanity:
## Review round 2 (Codex) — clean

Re-ran `codex review --base main` on the updated content (commit `7e880a70`, the connection-generation fix). Result: **no discrete regressions found** in the modified logic. The round-1 P2 is resolved and pinned by the two straggler regression tests.

[AI-assisted - Claude]

PR #385, opened 2026-07-07
feat(ui): expose portable invite codes in the River UI

### LINKED ISSUE #381: Suggestion: Allow use of portable invite codes in River UI
Currently, a new user trying to get into the River network, using the standard UI, needs to either get an invite link from a peer or have one generated by the Freenet quickstart invite generator. The problem in both cases is that assumptions are made about the host used in that link URI which may not be the correct assumption in some user use cases. Instead, I suggest using a portable invite code that can be copy/pasted into the River UI in a similar fashion to the import of backed up River IDs now.

A standalone code is already available for riverctl usage that should be sufficient for this.

A good example of where this would be useful is in the case of a user trying out Freenet via try.freenet.org. Currently, to join River UI through that mechanism requires manually editing a generated invite link to get a usable River invite. This is not user friendly for potential first time Freenet users.

sanity:
## Problem

Every River invite link bakes in the current host (e.g. a localhost node or the production gateway) via `window.location`. A user on a different host — most importantly a first-time user trying Freenet through **try.freenet.org** — has to hand-edit the host out of the link before it works. That is a bad first impression and the exact friction reported in #381.

A host-independent portable code already exists (`Invitation::to_encoded_string()`), used by `riverctl` and by the in-app click interceptor. This PR just **exposes** that code on the create side and **accepts** a pasted code on the receive side — it builds no new codec.

## Solution

**Create side** (`ui/src/components/members/invite_member_modal.rs`): the invite-member modal now shows a "Portable invite code" field (with its own copy button and a one-line "works on any host or peer" hint) *in addition to* the existing host-baked link. Nothing about the current link UX changes.

**Receive side** (`ui/src/components/room_list/join_with_code_modal.rs`, new): a new "Enter Invite Code" button in the room list bottom actions (next to "Import ID") opens a paste modal that mirrors the Import-ID UX. It decodes via `Invitation::from_encoded_string`, surfaces unparseable input inline, and on success hands the invitation to `present_invitation` — the same public entry point the DM invite-card "Accept" button uses. So the entire accept flow (re-accept guard #365, `room_secrets` handling, processed-fingerprint gating) is reused unchanged rather than reimplemented.

Follows the Dioxus signal-safety rules: reactive reads via the existing helpers, and the only global-signal write is inside `present_invitation`, which already defers it. Local scratch signals are set directly in event handlers, matching the sibling `ImportIdentityModal`.

## Migration safety

**UI-only change.** The committed delegate/room-contract WASMs (`ui/public/contracts/`, `cli/contracts/`) are byte-identical to `main` — verified with `git hash-object` and a `sha256` of the HEAD blobs vs. the working tree. Invite-code parsing is UI Rust → UI WASM, not contract/delegate WASM, so **no migration entry is needed**.

## Testing

- New `ui/tests/portable-invite-code.spec.ts`:
  - the invite modal exposes the portable code with "Copied!" feedback,
  - a code generated on the create side, pasted into "Enter Invite Code", opens the receive-invitation modal (full round-trip),
  - unparseable input shows an inline error and opens no modal.
- Full chromium Playwright suite green (**61 passed**, including the 3 new tests).
- `cargo test -p river-ui --bins`: **388 passed**.
- `cargo fmt` clean; UI builds clean (`cargo make build-ui-example-no-sync`).

Closes #381

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/members/invite_member_modal.rs
+++ b/ui/src/components/members/invite_member_modal.rs
@@ -197,6 +197,7 @@ pub fn InviteMemberModal(is_active: Signal<bool>) -> Element {
                                 InvitationContent {
                                     invitation_text: default_msg,
                                     invitation_url: invite_url.clone(),
+                                    invitation_code: invite_code.clone(),
                                     invitation: Rc::new(invitation.clone()),
                                     is_active: is_active,
                                     regenerate_trigger: regenerate_trigger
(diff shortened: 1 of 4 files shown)


sanity:
## Review: external Codex pass + adversarial read

**Codex (external, non-Claude model)** raised one finding:

- **[P2] Defer invite-code modal signal writes** — `join_with_code_modal.rs`: synchronous Dioxus signal mutations in the new event handlers violate the project's signal-safety rule (Firefox mobile can fire subscriber notifications synchronously during the write guard's Drop → `RefCell`/scope-stack panic).

**Resolution (commit e0ffae8d):** Accepted and applied — the button/close handlers now wrap their `is_active` / `code_input` / `error_msg` writes in `crate::util::defer()`, matching the local-signal defer convention already used in `room_list.rs` (the module this modal lives in). The one deliberate exception is the controlled `<textarea>`'s `oninput` write, kept synchronous: deferring a controlled input's bound value lags the DOM and drops keystrokes, which is exactly why every text-input handler in the codebase (`ImportIdentityModal`, `receive_invitation_modal`) sets its value signal synchronously. That reasoning is documented in a comment at the top of the handler block.

**Adversarial read (mine)** found no additional issues:
- Decode is panic-safe on attacker-controlled paste — `Invitation::from_encoded_string` is base58 + `ciborium` deserialization returning `Err` (no ECIES decrypt of untrusted bytes), so malformed input surfaces inline rather than aborting the WASM.
- The portable code is byte-identical to the credential the existing link already carries in `?invitation=`, so exposing it adds no new exposure; the existing "share privately with one person" warning covers it.
- The only global-signal write (`present_invitation` → `PRESENT_INVITATION_REQUEST`) is already deferred inside `present_invitation`.

Re-verified after the fix: chromium Playwright suite green (portable-invite-code + copy-clipboard + welcome specs, 8 passed), UI builds clean, `cargo fmt` clean.

[AI-assisted - Claude]

PR #401, opened 2026-07-12
feat(ui): room list loading/migrating state (#397)

### LINKED ISSUE #397: River UI: room list shows blank (no loading/migrating state) during startup load & delegate migration — reads as "all my rooms are gone"
## Summary

During startup load — and especially during a delegate / room-contract **migration** — the River room list renders **blank with no loading or "migrating" indicator**, which reads to users as "all my rooms are gone." User-reported after this morning's stdlib-0.8 River republish (delegate + room-contract re-key, coinciding with freenet-core v0.2.97).

## User report (Ivvor, Matrix, 2026-07-12)

> When River updated I assume it was doing some kind of delegate migration because my rooms list kept changing. On another node I was left with an empty Rooms list for a few minutes. If possible it might be a good idea to give the user some kind of warning when migrations are happening so they don't just think all their rooms have gone.

This morning River republished web-container `30000337` with the freenet-stdlib 0.6→0.8 upgrade (commit `47be34ea`), which re-keyed both `chat_delegate.wasm` and `room_contract.wasm`. Existing users' rooms migrate automatically on refresh (backward-probe + delegate migration) — deliberately transparent, no room recreation. The migration itself works; the problem is purely that it's **invisible** and transiently empties the list.

## Root cause (code)

- `ui/src/components/room_list.rs` — `RoomList()` builds its items straight from the `ROOMS` signal via `ordered_room_keys()`. When the map is empty the `<ul data-testid="room-list">` renders **nothing**: there is no empty-state text, no "loading rooms," no "migrating" indicator. (`grep` for loading/empty strings in this file finds only a comment.)
- Rooms populate **one at a time** as each per-room slot GET returns — `ui/src/room_data.rs:1560` (`self.map.insert(owner_vk, room_data)`). Hence "rooms list kept changing."
- On a re-keyed (empty) current delegate, legacy migration fires and probes old delegate(s); the room-contract side walks legacy generations newest-first in `ui/src/components/app/freenet_api/backward_probe.rs` with a **12s watchdog per hop** (`PROBE_GET_TIMEOUT`). On a slow node this stretches the empty window to the "few minutes" Ivvor saw.
- The only existing "Syncing…" copy is **per-room** (`ui/src/components/conversation.rs` "Syncing room state from the network…") and requires the room key to already be in the map — it does not cover the room-list-level pre-population window.

Net: the list is empty/churning with zero indication it's still working. This ambiguity ("loading" vs "you have no rooms") hits on **every cold start / refresh**, not only migration mornings — the migration just makes the window longer and more alarming.

## Proposed fix

Add a **room-list loading / empty state** that distinguishes "still loading/migrating your rooms…" from a genuine "no rooms yet":

1. While rooms are still being loaded/hydrated (before any room key is in the map), show a "Loading your rooms…" state instead of a blank list.
2. When a migration is in flight, show a "Migrating your rooms — this is a one-time step after an update" banner. This can key off the **existing** `is_legacy_migration_in_progress()` flag (`ui/src/components/app/chat_delegate.rs:3878`, localStorage-backed).
3. Only fall through to a true "No rooms yet — create or join one" empty state once loading/migration has completed and the map is genuinely empty.

This directly answers Ivvor's request and also improves the everyday cold-start experience.

## Note: the second half of Ivvor's report is a separate freenet-core regression (now tracked)

Ivvor also noted the node "doesn't seem to be using cached contracts unless they've been re-fetched at least once." That is **freenet-core** behavior, not River. On closer analysis (credit Ivvor) it is a **regression** introduced by the serve-DURING change (#4748, commit `33a024e4`): startup restores the hosting cache but not the in-memory `InterestManager`, so the serve gate routes restored copies to the network after restart (every GET, not just the first). Filed as **freenet-core#4780** (Full-tier hosting fix); not part of this River issue.

[AI-assisted - Claude]

sanity:
## Problem

During startup load — and especially during a delegate / room-contract **migration** — the River room list rendered **blank with no loading or "migrating" indicator**, which reads to users as "all my rooms are gone." Reported by Ivvor after the stdlib-0.8 River republish (delegate + room-contract re-key).

The `ROOMS` map populates **one room at a time** as each per-room slot GET returns, and on a re-keyed (empty) current delegate the rooms only arrive later via legacy migration. So an empty map does **not** mean "no rooms" — it usually means "still loading." Today the empty map fell straight through to a blank list (or, worse, could read as the empty state), with no way to tell "loading" from "you have no rooms."

Closes #397.

## Approach

Give the room list a **tri-state** (`ui/src/components/room_list.rs`):

- **loading / migrating** → a quiet inline spinner + `Loading your rooms…` (initial load) or `Restoring your rooms…` (migration, driven by the existing `is_legacy_migration_in_progress()` localStorage flag);
- **load resolved AND genuinely empty** → the existing `No rooms yet` create/join call-to-action;
- **populated** → the list (unchanged).

The genuinely-empty CTA now only shows once the **initial load has actually resolved**, tracked by a new `INITIAL_ROOMS_LOADED` signal (`ui/src/components/app.rs`) set at the terminal points of `load_rooms_per_room` (`response_handler.rs`). Because the legacy-migration probe is fire-and-forget (a fresh user with no legacy delegate gets no response at all), the empty-delegate branch waits a **bounded settle** (`LEGACY_PROBE_SETTLE_MS = 3s`, legacy delegates are node-local) before declaring "loaded & empty" — so a migrating user never flashes the empty state during the probe gap, while a fresh user gets the CTA after a short, one-time wait.

The signal only ever transitions `false → true`, so a WebSocket reconnect never flashes the list back to "loading."

**The tri-state decision lives in a pure `room_list_display(has_rooms, migrating, initial_loaded)` fn** (unit-tested), consumed inline in `RoomList()`.

**UI-only — no contract/delegate WASM change, no re-key.** No republish in this PR; publishing is a deliberate follow-up after the look is approved.

## Aesthetics — please review the look

Restrained, near-monochrome, theme-aware; the spinner reuses the existing per-room `border-text-muted` style (no accent, no banner). Screenshots (light + dark) of the migrating and empty states:

- Migrating (light): quiet spinner + "Restoring your rooms…"
- Empty (light): muted comments icon + "No rooms yet" / "Create a room or join one with an invite code."

(Screenshots attached in the PR thread / provided to the reviewer. The red WebSocket pill visible at the bottom of the dev screenshots is the pre-existing `ConnectionStatusIndicator` — this is a no-sync dev build with no node; it is unrelated to this change.)

## Testing

- New pure-fn unit tests (`room_list_display`) pin the crux: an empty map during migration or before load resolves is **never** the empty CTA; only a resolved-and-empty map is (`cargo test -p river-ui --bins`, 393 passed).
- `cargo clippy` clean (wasm, `no-sync` and `example-data,no-sync`); `cargo fmt` clean.
- Verified visually via a dev-only `?room_list_state=migrating|loading|empty|populated` URL override (compiled out of production — `example-data`/wasm-gated), screenshotted in light + dark.

[AI-assisted - Claude]

https://claude.ai/code/session_01CD8VBzb8qsDvdcPdF6TNQG

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -56,6 +56,32 @@ pub static SYNCHRONIZER: GlobalSignal<FreenetSynchronizer> = Global::new(Freenet
 pub static WEB_API: GlobalSignal<Option<WebApi>> = Global::new(|| None);
 pub static AUTH_TOKEN: GlobalSignal<Option<String>> = Global::new(|| None);
 
+/// Whether the initial room-list load has resolved (freenet/river#397).
+///
+/// Starts `false` and flips to `true` once the current delegate's first
+/// `ListResponse` has been processed AND — for an empty current delegate — the
+/// legacy-migration probe has been given a bounded window to surface migrated
+/// rooms. The room list reads this to distinguish "still loading / migrating
+/// your rooms" (show a quiet indicator) from "you genuinely have no rooms yet"
+/// (show the create/join empty state), so an in-progress load never renders as
+/// data loss. It only ever transitions false → true, so a WebSocket reconnect
+/// that re-fires the load never flashes the list back to a loading state.
+pub static INITIAL_ROOMS_LOADED: GlobalSignal<bool> = Global::new(|| false);
+
+/// Mark the initial room-list load as resolved (freenet/river#397).
+///
+/// Called from the delegate response-handling tasks (which run without a Dioxus
+/// scope), so the signal write is routed through `defer()` per the signal-safety
+/// rules. Idempotent — the flag never transitions back to `false`.
+pub fn mark_initial_rooms_loaded() {
+    crate::util::defer(move || {
+        let already = *INITIAL_ROOMS_LOADED.peek();
+        if !already {
+            *INITIAL_ROOMS_LOADED.write() = true;
+        }
+    });
+}
+
 // Tracks which rooms need to be synced due to USER actions (not network updates)
 // This prevents infinite loops where network responses trigger more syncs
 pub static NEEDS_SYNC: GlobalSignal<std::collections::HashSet<VerifyingKey>> =
--- a/ui/src/components/app/freenet_api/response_handler.rs
+++ b/ui/src/components/app/freenet_api/response_handler.rs
@@ -587,6 +587,18 @@ impl ResponseHandler {
     }
 }
 
+/// How long to keep the room list in its "loading" state after firing the
+/// legacy-migration probe for an empty current delegate (freenet/river#397).
+///
+/// The probe is fire-and-forget: a migrating user's rooms arrive on a legacy
+/// delegate a moment later, while a fresh user with no legacy delegate gets no
+/// response at all. This bounded settle lets a migration surface its rooms (or
+/// set the in-progress flag) before the list falls through to the genuinely-empty
+/// "no rooms yet" state. Legacy delegates are LOCAL to the node, so this window
+/// only has to cover a local round-trip — a fresh user waits this long for the
+/// empty-state CTA, which is a small, one-time cost.
+const LEGACY_PROBE_SETTLE_MS: u64 = 3000;
+
 /// Per-room room load (freenet/river#345 / #65). Reconstructs the in-memory
 /// `Rooms` from the current delegate's `room:<vk>` slot keys (+ `rooms_meta`)
 /// discovered by a `ListRequest`, then hands the assembled value to the shared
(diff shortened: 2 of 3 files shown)


sanity:
Superseded: #399 merged 2026-07-13 with the same room-list loading/migrating/empty states (plus a LoadFailed error+retry state this branch lacks) and has been live since web-container 30000338. Issue #397 is closed and the fix re-shipped in today's 30000349 publish. Ian visually re-confirmed the banner from this branch's screenshots on 2026-07-21 before we spotted the supersession. Closing.

[AI-assisted - Claude]

PR #407, opened 2026-07-14
fix(ui): stop endless WebSocket reconnect loop after Android resume

### LINKED ISSUE #406: Mobile: WebSocket reconnect loop cycles Connected/Disconnected/Error forever after background→resume (never backs off)
## Symptom (reported by Ivvor, Android Chrome)

Open River on Android with a room open, let the device sleep/inactive ~1 min, then reactivate. The connection indicator enters a **mad reconnect cycle** — rapidly toggling Connected → Disconnected → Error → Connected → … that **never stops on its own**; only a full page refresh clears it. It happens even without the room list visible (you just can't see the indicator). While backgrounded, logs show WS polling with DNS errors; on resume the DNS errors stop but it "connects very briefly, then connection closed, then retries very quickly." **iPad Safari does not suffer** — it force-reloads the page after inactivity.

## Reproduction (deterministic, Playwright + mock WebSocket)

Serve a sync-enabled UI (`cargo make build-ui-example`) and mock the node WebSocket so each connection **opens then closes shortly after** (simulating the flapping socket after Android resume). Result over 40s:

```
socket opens in 40000ms: 12
intervals between reconnects (ms): 3914, 3171, 3839, 3912, 3487, 3203, 3909, 3674, 3521, 3816, 3255
backoff GREW across attempts: false   ← BUG: constant ~3.5s, no exponential backoff
```

Console (repeats forever):
```
WebSocket connected successfully → Connection established successfully   (Connected)
WebSocket connection closed → error {"error":"connection closed","source":"close"}
Connection closed, triggering reconnection
Connection failed (attempt 1), reconnecting in 2702ms                    ← always "attempt 1"
```

## Root cause (3 compounding gaps, all in `ui/src/components/app/freenet_api/`)

1. **Backoff resets on socket OPEN, not on a stable connection → non-termination (root).** `freenet_synchronizer.rs` sets `consecutive_failures = 0` the instant `initialize_connection` returns `Ok` (i.e. as soon as `onopen` fires within the 5s timeout), before the connection has proven stable or carried traffic. A socket that opens then dies immediately resets the counter every cycle, so the exponential backoff (`reconnect_delay_ms`, 3s→60s) never grows — the app retries at the ~3s floor forever. This is the "never stops."

2. **`Connect` has no dedup and `initialize_connection` overwrites `WEB_API` without closing the old socket first → self-manufactured disconnects.** Overwriting `*WEB_API.write() = Some(new)` drops the old `WebApi`, whose `Drop` closes its socket; the old socket's `onclose` (forget()'d, holds an Rc to River's error handler, `freenet-stdlib browser.rs`) fires `{"error":"connection closed"}` → mapped to `ConnectionLost` → another reconnect. So each redundant `Connect` manufactures its own disconnect — the loop self-sustains even against a healthy node.

3. **No debounce on resume + frozen-timer backlog + no-delay Connect paths.** `visibilitychange`→`PageBecameVisible` has no debounce (Android fires it several times on resume); background-throttled `setTimeout` reconnect timers all fire at once on resume; `ProcessRooms`/`RefreshAllRooms` send `Connect` with no delay when not connected. Together these produce the *rapid* churn on resume that seeds gaps #1/#2.

**Why Android but not iPad Safari:** Safari discards/reloads the tab after inactivity (fresh WASM instance, fresh state). Android Chrome resumes the same frozen WASM instance and lands in the loop with no fresh-start escape.

## Proposed fix

- Reset `consecutive_failures` only after the connection has been **stable for a dwell period** (or received real inbound traffic), not merely on `onopen`.
- Guard the `Connect` handler against connecting while already `Connecting`/`Connected`, and coalesce to a single outstanding reconnect.
- At the top of `initialize_connection`, deterministically take/close any existing `WEB_API` so an orphaned socket's `onclose` can't inject a spurious `ConnectionLost`.
- Debounce `PageBecameVisible`.

This is the `#382`/`#384` liveness-watchdog area; the watchdog itself is only the slow backstop (~80s cycle) that first surfaces the half-open socket — the rapid non-terminating loop is the reconnect path above.

[AI-assisted - Claude]

sanity:
## Problem

On Android Chrome with a room open, after the device goes inactive for ~a minute and is reactivated, River's connection indicator cycles **Connected → Disconnected → error** forever. Only a full page refresh stops it. Reported by Ivvor (Matrix):

> Go to River on an Android phone and have a room open. Then let the device go inactive. Leave for a minute or so then reactivate and the mad reconnect cycle begins.

The socket "connects very briefly, then connection closed, then retry very quickly" — a tight loop, not a backing-off retry. (iPad Safari is unaffected because it force-refreshes the page after inactivity.)

## Root cause

Three compounding bugs in the synchronizer reconnect logic (`freenet_synchronizer.rs`):

1. **Backoff reset on mere socket `onopen`.** A socket that opens then immediately dies (the Android resume symptom) reset the failure counter every cycle, so the reconnect delay never grew past the ~3s floor — the "mad cycle."
2. **Redundant `Connect` tore down a live connection.** A repeated `PageBecameVisible`, a frozen background-timer backlog firing all at once on resume, or the `ProcessRooms`/`RefreshAllRooms` no-delay Connect paths could re-enter `initialize_connection`, which overwrote a live `WEB_API` and dropped its socket. That socket's orphaned `onclose` then injected a spurious `ConnectionLost`, self-sustaining the loop.
3. **Duplicate `ConnectionLost` events stacked reconnects.** The dropped socket's own `onclose` plus the liveness watchdog reporting the same death each scheduled a reconnect, inflating backoff and multiplying timers.

## Approach

- **Reset backoff only on proven stability.** After a `Connect` succeeds, arm a dwell check (`CONNECTION_STABLE_DWELL_MS = 15s`) tagged with the connection generation (`ws_connect_seq`). The backoff resets only if *that same* connection is still up after the dwell — so an open-then-die socket keeps backing off (3s→6s→12s→…→60s cap) instead of pinning at 3s.
- **Dedup `Connect`:** skip `initialize_connection` when already connected, so a redundant Connect can't drop a live socket.
- **Coalesce `ConnectionLost`** via a single-pending-reconnect latch, so duplicate losses don't stack.
- **Stop eagerly zeroing backoff on `PageBecameVisible`** (it still sends an immediate Connect; the dedup + dwell handle correctness).

Backoff math + latch extracted into a `ReconnectState` struct so the behaviour is unit-testable off-wasm.

## Testing

- **3 new unit tests** (`ReconnectState`): flap-backs-off (`vec![3000,6000,12000,24000,48000]`), stable-resets-but-stale-dwell-doesn't, duplicate-loss-coalesced. Full suite green (431 tests).
- **Reproduced end-to-end with Playwright** — a mock node socket that opens then dies after 600ms (simulating the Android flapping socket):

  | | Before (main) | After (this PR) |
  |---|---|---|
  | socket opens / 45s | **12** | **4** |
  | reconnect intervals | constant ~3.5s | **4.1s → 5.8s → 12.6s** |
  | backoff grows | ❌ no | ✅ yes |

Closes #406

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/freenet_api/connection_watchdog.rs
+++ b/ui/src/components/app/freenet_api/connection_watchdog.rs
@@ -70,9 +70,9 @@
 //!   *asynchronously* in the synchronizer message loop: if that loop lags
 //!   longer than a probe tick plus [`LIVENESS_PROBE_TIMEOUT_MS`], a naive
 //!   watchdog would re-probe and fire a SECOND `ConnectionLost`, and each one
-//!   independently schedules a reconnect and bumps `consecutive_failures`,
-//!   producing overlapping reconnects and inflated backoff. The latch makes the
-//!   "exactly once" guarantee hold even under message-loop lag.
+//!   independently schedules a reconnect and bumps the reconnect backoff
+//!   counter, producing overlapping reconnects and inflated backoff. The latch
+//!   makes the "exactly once" guarantee hold even under message-loop lag.
 //! * The probe timeout ([`LIVENESS_PROBE_TIMEOUT_MS`]) is far larger than a
 //!   local GET's real latency, so a momentarily slow node does not trigger a
 //!   false reconnect. If no room exists to probe, the watchdog never declares
--- a/ui/src/components/app/freenet_api/constants.rs
+++ b/ui/src/components/app/freenet_api/constants.rs
@@ -41,6 +41,14 @@ pub const RECONNECT_INITIAL_MS: u64 = 3000;
 /// Maximum retry interval for reconnection attempts in milliseconds
 pub const RECONNECT_MAX_MS: u64 = 60_000;
 
+/// How long (ms) a connection must stay up before the reconnect backoff counter
+/// is reset. Resetting merely on socket `onopen` let a socket that opens then
+/// immediately dies reset the counter every cycle, defeating the exponential
+/// backoff and producing an endless ~3s reconnect loop after an Android
+/// background→resume (freenet/river#406). A connection must prove itself stable
+/// for this long before we treat it as a fresh, healthy baseline.
+pub const CONNECTION_STABLE_DWELL_MS: u64 = 15_000;
+
 /// Maximum number of retries for API requests
 pub const MAX_REQUEST_RETRIES: u8 = 3;
 
--- a/ui/src/components/app/freenet_api/constants.rs
+++ b/ui/src/components/app/freenet_api/constants.rs
@@ -47,6 +47,21 @@ pub const RECONNECT_MAX_MS: u64 = 60_000;
 /// backoff and producing an endless ~3s reconnect loop after an Android
 /// background→resume (freenet/river#406). A connection must prove itself stable
 /// for this long before we treat it as a fresh, healthy baseline.
+///
+/// Chosen at 15s deliberately, relative to the neighbouring timers:
+/// - Above connection setup: the dwell starts only *after* `initialize_connection`
+///   returns Ok (i.e. after `onopen`), and setup is bounded by
+///   `CONNECTION_TIMEOUT_MS` (5s), so a slow-but-real connect never eats the dwell.
+/// - Below the watchdog's death window: a half-open socket isn't declared dead
+///   until roughly `LIVENESS_IDLE_PROBE_MS` + `LIVENESS_PROBE_TIMEOUT_MS` (~80s),
+///   so a dwell-reset can't fire on a socket the watchdog is about to kill.
+///
+/// Tradeoff (accepted): a link that repeatedly opens and survives just under 15s
+/// before dying never reaches the dwell, so its backoff climbs toward the 60s cap
+/// even though each attempt partly succeeds. That is the desired behaviour — don't
+/// hammer a genuinely flaky link — and it self-corrects the instant one connection
+/// survives 15s. Ivvor's report is "connects very briefly" (sub-second), which
+/// backs off correctly well before this edge matters.
 pub const CONNECTION_STABLE_DWELL_MS: u64 = 15_000;
 
 /// Maximum number of retries for API requests
--- a/ui/src/components/app/freenet_api/freenet_synchronizer.rs
+++ b/ui/src/components/app/freenet_api/freenet_synchronizer.rs
@@ -49,12 +49,78 @@ fn reconnect_delay_ms(consecutive_failures: u32) -> u64 {
     (capped - jitter_range + jitter).min(RECONNECT_MAX_MS)
 }
 
+/// Reconnect bookkeeping: the exponential-backoff failure counter plus a
+/// single-outstanding-reconnect latch. Extracted from the message loop so the
+/// two behaviours behind the endless Android reconnect loop (freenet/river#406)
+/// are unit-testable: (1) the backoff must GROW across an open-then-die flap
+/// and reset only after a connection proves stable, and (2) a duplicate
+/// ConnectionLost (a dropped socket's orphaned `onclose`, or the watchdog and
+/// transport reporting the same death) must be coalesced, not stack another
+/// reconnect.
+#[derive(Default)]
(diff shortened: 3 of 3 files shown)


sanity:
## Multi-model review — Full tier (connection/async is a high-risk surface)

Ran an external **Codex** pass plus **four adversarial Claude lenses** (state-machine bug-hunt, async/concurrency lifecycle, fails-to-recover failure mode, tests + big-picture), each blind to the others and told not to rubber-stamp. Reviewed the checked-out code, not just the diff.

### Findings addressed (fixed in 6c6c5817)

| # | Severity | Finding | Reviewers | Disposition |
|---|----------|---------|-----------|-------------|
| 1 | P2 | Reconnect-timer stacking: the coalesce latch is a bare bool cleared by *any* Connect (incl. PageBecameVisible / ProcessRooms / RefreshAllRooms / initial), so an immediate Connect or a thawed background-timer backlog could leave an already-scheduled backoff timer running in parallel with a freshly-armed one. | Codex, lens 1, lens 4 | **Fixed** — generation-tag each reconnect timer; a timer sends `ScheduledReconnect{generation}` and only drives a reconnect if that generation is still current, so a superseded timer drops itself. Safe by construction (a timer is only stale because a newer one is already scheduled). New unit test `stale_reconnect_timer_is_dropped_but_current_one_fires`. |
| 2 | P1 | New `ConnectionStable` dwell spawn used raw `spawn_local` from inside the polled message-loop future → can re-enter wasm-bindgen's task scheduler and panic on Firefox mobile. | Codex, lens 1 | **Fixed** — switched the dwell spawn and the reconnect timer to `crate::util::safe_spawn_local` (defers via `setTimeout(0)`), matching the liveness watchdog's own convention. |
| 3 | P2 | Connect dedup keyed only on `is_connected()` (SYNC_STATUS); a late `onopen` can leave that true while `WEB_API` is still None. | lens 3 | **Fixed** — dedup now gates on `is_connected() && WEB_API.read().is_some()`, keeping the brake self-contained. |
| 4 | P3 | Inaccurate `PageBecameVisible` comment ("dedup makes a redundant attempt a no-op" — false when a reconnect is pending, since we're disconnected then). | lens 1 | **Fixed** — comment rewritten to describe the generation mechanism. |
| 5 | P3 | Stale `consecutive_failures` reference in watchdog rustdoc; duplicate test comment; undocumented 15s dwell rationale. | lens 4 | **Fixed** — corrected the rustdoc, dropped the dup, documented the 15s choice relative to the setup timeout and the ~80s watchdog death window, plus the accepted sub-dwell-flap tradeoff. |

### Documented, deliberately not fixed here (justified)

- **Superseded-socket `ConnectionLost` blip** (lens 2): after a deliberate `WEB_API.take()`, the dropped socket's stdlib-leaked `onclose` fires a spurious `ConnectionLost`; if a non-reconnect Connect cleared the latch meanwhile, it can cause **one** extra reconnect cycle before full recovery. Self-limiting, not the endless loop. The robust fix (thread a per-socket generation through `connection_manager` + the watchdog and suppress losses from old sockets) risks the opposite, worse failure — a bookkeeping slip that suppresses a *genuine* loss would strand the connection. On a recovery-critical surface I chose not to take that risk to eliminate a self-healing blip; documented in code + commit.
- **Sub-dwell-flap over-braking** (lens 3, lens 4): a link that survives just under 15s repeatedly backs off toward the 60s cap. That's the desired "don't hammer a flaky link" behaviour and self-corrects once one connection survives 15s. Documented on the constant.

### Reassuring clean probes (all four lenses)

No reachable **stuck-disconnected / never-recovers** state: `reconnect_pending` is always cleared unconditionally at the top of the Connect handler (before the dedup `continue`), and every `arm()` is paired with a scheduled reconnect. `connect_seq` capture timing is correct (monotonic, bumped in `onopen` before the Ok arm reads it), and the coalescing can't swallow a legitimate watchdog reconnect. All four fails-to-recover scenarios provably recover.

[AI-assisted - Claude]

Issue #408, opened 2026-07-14
Browser notifications never fire in the gateway deployment (sandboxed iframe = opaque origin)


sanity:
## Summary

River's browser (Web) notifications don't fire for users on the deployed gateway — reported by Ivvor, and confirmed by me not seeing them either for a while. Investigation shows this is **not a code regression**: the notification code is intact and correctly wired. The cause is environmental — River UI runs inside the Freenet gateway's **sandboxed iframe with an opaque (null) origin**, and browsers block the Web Notifications API from opaque-origin iframes.

## The code is intact (no regression)

The feature lives in `ui/src/components/app/notifications.rs` and is fully wired:
- Permission is requested after the user's first successful send — `request_permission_on_first_message()` (`notifications.rs:151-183`), called from `ui/src/components/conversation.rs:1820-1821`.
- New messages trigger `notify_new_messages()` (`notifications.rs:358-498`) from **both** receive paths: the delta path (`ui/src/components/app/freenet_api/room_synchronizer.rs:374-381`) and the full-state path (`room_synchronizer.rs:1335-1345`).
- The visibility gate (`notifications.rs:381-392`) correctly lets a **backgrounded** tab through (only suppresses when the tab is visible AND it's the current room).
- The per-room `NotificationMode` default is `All` (`ui/src/room_data.rs`), so external messages are not filtered out by default.

`git log -S notify_new_messages` shows only the original wiring plus signal-safety hardening — nothing ever detached the fire path. So notifications would only ever have been *seen* in a directly-served dev build (`dx serve`), not in the gateway iframe.

## Root cause: opaque-origin iframe

The gateway serves River with `sandbox="allow-scripts allow-forms allow-popups allow-downloads allow-modals"` — **no `allow-same-origin`** (freenet-core `crates/core/src/server/path_handlers.rs`, documented in `client_api.rs`: "The iframe has an opaque (null) origin because the sandbox attribute omits `allow-same-origin`"). River's own code acknowledges the opaque origin (`ui/src/components/app.rs:193-196`).

Browsers block `Notification.requestPermission()` and notification display from cross-origin/opaque-origin iframes (Chrome since v61, Firefox similarly). So permission can never reach `Granted`; `show_notification()` bails at the permission check (`notifications.rs:205-208`) and `Notification::new_with_options(...)` (`notifications.rs:249`) is never reached.

[...]

PR #409, opened 2026-07-15
feat(ui): show new-message notifications via the gateway shell bridge

sanity:
## Problem

Browser notifications for new messages silently stopped working when River moved into the gateway's **sandboxed iframe** (freenet-core#3254, Feb 2026): the iframe has an **opaque (null) origin**, and browsers block the Web Notifications API from opaque-origin iframes, so permission can never reach `granted` and notifications never fire (freenet/river#408). They worked before that, as a top-level page (Dec 2025 → Feb 2026). Reported by Ivvor.

## Approach

Route notifications through the **shell page** (same-origin with the node, a real origin) over the existing `__freenet_shell__` postMessage bridge — the same channel that already carries title/favicon updates. Paired with **freenet-core#4801**, which adds the shell-side handlers that actually show the notification.

- **Detect the iframe context** (`window.parent !== window`). Served directly (dev / `no-sync`, a real origin), the existing direct Notifications API path is used **unchanged**.
- In the iframe, `request_permission_on_first_message` asks the shell to offer notifications (the shell owns the real-origin permission prompt / affordance), and `show_notification` postMessages the notification instead of calling the API directly.
- The room's owner key is encoded as the notification `tag`; the shell posts a `notification_click` reply carrying it back, and a window `message` listener (installed once from `App`, with the signal write deferred per the Dioxus signal-safety rules) routes `CURRENT_ROOM` to the clicked room.

## Deployment / ordering

This is a no-op until a gateway serving the freenet-core#4801 shell is deployed, and neither side's deployment breaks the other (an old shell ignores the unknown message types; old River still calls the API directly, which no-ops in the iframe as it does today). The try.freenet.org gateway is a manual rebuild, so the feature lights up there once #4801 merges and that node is rebuilt.

## Testing

- Unit tests: notification `tag` round-trips the room key, and malformed tags are rejected (the click-routing correctness).
- Native + `wasm32` compile clean.
- End-to-end verification requires the paired #4801 shell; documented in freenet/river#408.

Refs freenet/river#408

### DIFF
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -44,6 +44,7 @@ web-sys = { workspace = true, features = [
     "HtmlAnchorElement",
     "HtmlElement",
     "HtmlTextAreaElement",
+    "MessageEvent",
     "MouseEvent",
     "IntersectionObserver",
     "IntersectionObserverEntry",
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -96,6 +96,11 @@ pub fn App() -> Element {
     // Must be called from within a Dioxus component where the runtime is active.
     crate::util::capture_runtime();
 
+    // In the gateway iframe, notifications are shown by the shell page; listen
+    // for its `notification_click` replies to route to the clicked room
+    // (freenet/river#408). No-op when served as a top-level page. Installs once.
+    crate::components::app::notifications::install_shell_notification_listener();
+
     // Install the document-level click interceptor that catches
     // in-page invite-URL anchor clicks and routes them through the
     // in-app receive-invitation flow instead of letting the browser
--- a/ui/src/components/app/notifications.rs
+++ b/ui/src/components/app/notifications.rs
@@ -16,9 +16,12 @@ use river_core::room_state::member::MemberId;
 use river_core::room_state::member_info::MemberInfoV1;
 use river_core::room_state::message::{AuthorizedMessageV1, RoomMessageBody};
 use std::collections::HashSet;
+use std::sync::atomic::{AtomicBool, Ordering};
 use wasm_bindgen::prelude::*;
 use wasm_bindgen_futures::JsFuture;
-use web_sys::{Notification, NotificationOptions, NotificationPermission, VisibilityState};
+use web_sys::{
+    MessageEvent, Notification, NotificationOptions, NotificationPermission, VisibilityState,
+};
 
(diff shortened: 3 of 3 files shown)


sanity:
## Multi-model review — Codex + a signal-safety/correctness lens

**Verdict:** no P1/P2 signal-safety or correctness bugs. The lens confirmed the deferred `CURRENT_ROOM` write is correct and *required* (raw JS `message` closure → no Dioxus scope → a synchronous write would hit the empty-scope panic), the iframe detection is sound (no cross-origin SecurityError), the listener is installed exactly once, and the tag round-trip / non-object-payload paths are panic-free. The dev/top-level path is unchanged.

### Fixed in the latest commit
| Finding | Reviewer | Fix |
|---|---|---|
| Notification click set `CURRENT_ROOM` but not `MOBILE_VIEW` → on mobile the user stays on the Rooms/Members panel instead of seeing the conversation (and it may be marked read without being shown) | Codex P2 | Set `MOBILE_VIEW = Chat` in the deferred callback, mirroring normal room selection |
| `web_sys::MessageEvent` only enabled transitively (via `dioxus-web/devtools`) — fragile | lens P3 | Declared explicitly in `ui/Cargo.toml` web-sys features |
| `notification_click` listener had no sender check | lens P3 | Added `event.source() == window.parent()` gate + corrected the two overclaiming comments |

### Documented, not changed (justified)
- **Message preview posted to `parent` with target origin `"*"`** (lens P3, inherent): in the intended deployment the parent is the same-origin gateway shell — inside the user's trust boundary and it *needs* the text to render the notification. `"*"` is unavoidable because the opaque-origin iframe can't name the parent's real origin, and this matches the existing title/favicon bridge (`document_title.rs`), which posts to `parent` the same way. Acceptable.
- The dev-only **direct (top-level) path** has the same pre-existing `MOBILE_VIEW` gap and a non-deferred `CURRENT_ROOM` write; left as-is to keep this PR scoped to the production (iframe) path, since that path only runs when River is served directly (dev / `no-sync`).

### Aside (separate cleanup, not this PR)
The lens noted `ui/src/components/app/tab_detection.rs` appears to be orphaned dead code (no `mod tab_detection;` anywhere). Pre-existing and unrelated; worth a follow-up.

Paired with freenet-core#4801 (the shell-side handlers); this half is a harmless no-op until that ships and the gateway is rebuilt.

[AI-assisted - Claude]

Issue #371, opened 2026-06-27
Alternative room membership mechanisms (open rooms, PoW join, delegated moderation)


sanity:
## Summary

River's only room-access model is the invitation chain. That works well for private and semi-private groups, but it has no answer for "I just want an open room anyone can join." Users are already hacking around the gap in unsafe ways, so we should design first-class alternatives.

This issue captures the problem, the hard constraints, the design space, a proposed architecture, and the known pitfalls. It is a design issue, not an implementation plan. No code should land until the direction is signed off, because this is the consensus-critical room contract.

## Motivating incident

A user wanted a general off-topic room (a low-stakes "anyone welcome" space). The invite model had no fit, so he built a Freenet website that handed out pre-signed keypairs to visitors. The site exposed every keypair to every visitor, so all the credentials were effectively leaked at once. A user reduced to publishing bearer credentials in public is a clear signal that the invite-only model is ruling out legitimate use cases.

## Why this is hard: the contract is the only referee

The invite chain is not just UX, it is the sybil/spam defense. Every member traces back to the owner by signature, so spam is bounded by "who will an existing member vouch for." Remove invites and we have to answer: what stops thousands of throwaway identities from joining and flooding the room?

And we have to answer it under the constraints of a Freenet contract, where the only referee is `validate_state`, a pure function of state + parameters + delta:

- **No trusted clock.** Timestamps in state are self-asserted by whoever signed them. So "max N joins per hour," join cooldowns, and time-decaying difficulty are all not contract-enforceable.
- **No global view, no oracle.** The contract cannot see IPs, run a captcha, or rate-limit across peers. It only checks what is cryptographically present in the delta.
- **No randomness, no I/O.**

So every admission mechanism reduces to: a proof the joiner carries in the delta that the contract can verify alone. That is a small menu, and there is no clean winner, because the core principle is unavoidable:

> **Admission cost must exceed the value an attacker extracts before removal.**

## The admission menu, honestly assessed

| Mechanism | What the contract checks | Real weakness |
|---|---|---|
| **PoW to join** | `H(room_id ‖ member_vk ‖ nonce) < target` | One-time, parallelizable cost. Weak against a motivated attacker; taxes mobile users. |
| **Passphrase gate** | `H(secret)` or a signature over a known challenge | The secret is a bearer credential. It leaks, and we are back to the motivating incident with extra steps. |
[...]

PR #411, opened 2026-07-16
feat: member deputies for ban authority (invite-subtree moderation)

### LINKED ISSUE #410: feat: member deputies for ban authority (invite-subtree moderation)
## Summary

Let a member **deputize** another member — grant them ban authority scoped to the deputizing member's invite-subtree. The owner deputizing is the whole-room case (the owner's subtree is everyone), so this covers "appoint a global moderator" without a separate mechanism. This is the **subtree-scoped implementation of the delegated-moderation section of #371** (that issue calls it a "moderator grant"; we use **deputy** as the term because "delegate" is already an overloaded first-class Freenet concept — the WASM delegates — and "deputy" also reads more precisely for a subtree-scoped authority than "moderator").

## Problem / motivation

Right now the only authority that can ban across a room is the room owner (a member can ban only their own invite-subtree, and the owner sits at the root). During a spam wave that makes the owner a single point of failure: every ban needs the owner's key at the keyboard. On 2026-07-16 the Official room was spammed and each removal required the owner key + a stale-tooling scramble, with no way to hand a trusted member the ability to help.

Deputy ban authority removes that bottleneck: a member (usually the owner, but any member with a subtree) can deputize a trusted member to ban within their subtree, and revoke it later.

## Model (contract)

**Deputy list — reuse `MemberInfo`, don't add a parallel signed record.** Add `deputies: Vec<MemberId>` to `MemberInfo` (`common/src/room_state/member_info.rs`, `MemberInfo` at :201, `version: u32` at :203). `MemberInfo` is already signed by the member and merges as **monotonic-version last-writer-wins on the whole record** (delta only for newer `version` at `member_info.rs:72`; apply replaces only if newer at `:128`; the older-version-ignored test at `:478`). That gives us two things for free:
- **Authentication**: A's deputy list lives in A's own signed `MemberInfo`, so A deputizing B is A's authenticated statement — no new signing path, no `Authorized{Grant}` wrapper.
- **Safe revoke propagation**: deputize = A republishes `MemberInfo` at `version+1` with B added; revoke = A republishes at `version+1` with B removed. Because the whole record is LWW-by-version, a stale re-broadcast of the old (B-present) record has a lower version and loses. This defuses the single nastiest CRDT trap (a stale add resurrecting a revoked grant) with machinery that already exists — no bespoke tombstone set.

**Authorization predicate.** Extend the ban auth check (today `BansV1::validate_invite_chain` at `common/src/room_state/ban.rs:159`, reached via `get_invalid_bans` :79 → `validate_single_ban` :106, which walks up the invite tree from the target and accepts iff the banner is an ancestor). New predicate: a ban by B of target T is authorized iff
`B == owner  OR  B is an ancestor of T  OR  some current ancestor A of T lists B in A.deputies`.
The subtree relation is already computable via `get_downstream_members` (`member.rs:223`) / the `invited_by` parent link.

**A deputy can be any member.** There is no requirement that the deputy be in the deputizing member's subtree. A deputization grants authority *over* the deputizer's subtree *to* an arbitrary member — e.g. the owner deputizing a trusted community member who happens to have been invited by someone else. Mechanically nothing special is needed: `deputies` is just a list of `MemberId`s and the predicate only asks whether *an ancestor of the target* lists the banner; where the banner sits in the tree is irrelevant. Consequence to note: a deputy whose scope covers their own ancestors can ban those ancestors (an owner-deputized global moderator can ban the very member who invited them). "Cannot ban the member who deputized you" protects the *deputizer*; it does not protect the deputy's mere inviter — inherent to appointing someone who outranks the invite tree. Flag if we also want a "cannot ban your own ancestors" guard (probably not for owner-deputized global mods, since that is the point).

**Revocation is retroactive, computed in `post_apply_cleanup`.** A deputy-authorized ban enforces iff, in the *current converged* state, its banner is still authorized by the predicate above. So revoking A→B causes B's bans over A's subtree to stop enforcing and those users can rejoin. This is the desired behavior: you revoke a deputy because they misused it, so undoing their bans is the point, not a side effect (Ian's call, 2026-07-16). It also fully defangs a malicious revoked deputy: every ban they made stops enforcing at once and they cannot mint new enforceable bans. (Their targets' old messages don't return — those are pruned — they simply become able to rejoin and post again.)

The one hard constraint that makes this convergent: **the enforcement recompute must live in `post_apply_cleanup` (`common/src/room_state.rs:93`), NOT in `verify`.** River's DM module already moved ban enforcement there for exactly this reason — flipping ban validity inside `verify` makes `verify` non-stable across ban-state changes and breaks convergence. The member cascade (`remove_banned_members`, `member.rs:212`) is a pure function of converged (members + deputies + bans) state, so every peer converges to the same member set regardless of delta order. **This is the single thing to verify line-by-line against `post_apply_cleanup` + the DM precedent before implementing.**

**Guardrails.**
- **Cannot ban the member who deputized you.** B may not ban anyone who currently lists B in their `deputies`. Needed beyond the strict-descendant rule: if B is a deputy of both A1 and A2 where A2 is inside A1's subtree, A1's scope would otherwise let B ban A2, a fellow deputizer. Cheap lookup here.
- **Cannot ban the owner** (already true; keep it).
- **No transitive re-deputization in v1.** A deputy cannot deputize others using authority they only hold as a deputy. Ancestors can always deputize directly; this keeps the merge reasoning bounded.
- **Bounded `deputies` length** to avoid state-bloat abuse.

## riverctl surface

- Add `Deputize { room_id, member_id }` and `RevokeDeputy { room_id, member_id }` to `MemberCommands` (`cli/src/commands/member.rs:8`; handler dispatch near `:139`).
- Implement alongside `ban_member` (`cli/src/api.rs:3950`): build the change by republishing the caller's own `MemberInfo` at `version+1` with the target added/removed from `deputies`, signed with the stored key, sent as `ChatRoomStateV1Delta { member_info: Some(...) }`. The existing client-side subtree check at `api.rs:~4016-4045` is the spot to also accept "I am a deputy" for `ban_member`.

## UI surface

Design principle: this is a **specialized, rarely-used** feature. It must be **coherent** with existing member management (not a bolted-on panel) but **low-prominence** (no new top-level nav, no always-visible controls).

- **Home it in the member-info modal**, next to Ban (`ui/src/components/members/member_info_modal/ban_button.rs`; modal at `ui/src/components/members/member_info_modal.rs`). Add a **"Deputize"** action, and a **"Revoke deputy"** action on a member who is currently your deputy. Signing goes through a new `sign_deputize` alongside `sign_ban` (`ui/src/signing.rs:91`, `sign_ban_with_fallback` :372).
- **Gating is on the VIEWER, not the target — different from Ban.** Ban is gated on the *target* being downstream of the viewer (`is_downstream`, `member_info_modal.rs:123`). Deputizing is the opposite: the deputy can be **any** member, so the "Deputize" action shows in *any* member's modal, gated instead on whether the **viewer** has authority to give — the owner, or a member whose own invite subtree is **non-empty**. (Exclude self and the owner as targets: deputizing yourself or the owner is a no-op.)
- **Anti-confusion gating (important).** A member with no invitees has an empty subtree and can ban nobody, so offering them "Deputize" advertises power they don't have — hide the action for them entirely. (The contract stays permissive: deputizing from an empty subtree is a harmless no-op that activates if they later invite people; the UI is what prevents the misconception.)
- **Frame by concrete scope, not abstract power.** Copy should read like "Let [member] help moderate people you invited," not "delegate banning authority" — separating the deputy (any member) from the scope (your subtree).
- **Subtle "Deputy" badge** on the member list (`ui/src/components/members.rs`) so it's visible who holds deputy authority, without a dedicated moderators screen (which would over-emphasize a rare feature).

## Open questions

1. **Subtree-root bans cascade.** `remove_banned_members` (`member.rs:212`) drops the banned member **and their entire downstream branch**. A deputy banning an internal subtree root wipes everything under it. Decide: can deputies ban internal roots, or should deputy bans be restricted (leaves only, or a confirmation)?
2. **`max_user_bans` eviction can un-ban.** Bans are the durable tombstone; evicting the oldest ban (`ban.rs:206-232`) lets that user rejoin. Deputy moderation multiplies ban volume, so an attacker could flood bans to evict their own. Pre-existing, but amplified — worth deciding whether the cap/eviction policy needs adjusting alongside this.
3. **Managing your own deputies in the UI.** v1 proposal is per-member (deputize/revoke on the specific member's modal + a list badge). Enough, or do we want a small "who are my deputies" view? (Leaning per-member to keep prominence low.)

## Testing (simulation + unit)

- Deputize A→B, then B bans a member in A's subtree → accepted.
- B bans a member **outside** A's subtree → rejected.
- B bans A (their deputizer) and the multi-deputizer A2 case → rejected.
- Revoke A→B → B's prior bans stop enforcing; banned users can rejoin; **all peers converge to the same member set regardless of delta order** (apply revoke and ban in both orders).
- Stale replay of the old (B-present) `MemberInfo` after revoke → loses to the higher-version revoke (LWW), does not resurrect authority.
- Subtree-root ban cascade behaves per the decision in open question 1.
- `max_user_bans` interaction per open question 2.

## Migration / compatibility (data-loss critical)

**⚠ The one thing that silently strands every existing room if done wrong.** `MemberInfo` is *individually signed* (`AuthorizedMemberInfo::new` → `sign_struct` over the ciborium bytes of the whole struct; `verify_signature` → `verify_struct` re-serializes and checks — `member_info.rs:158,188`, `util.rs:6-20`). `MemberInfo` is 3 fields today (`member_id`, `version`, `preferred_nickname`, `member_info.rs:201-205`). Adding `deputies` with a **plain** `#[serde(default)]` makes the new WASM re-serialize a 4-field struct and every existing member's signature FAILS → `MemberInfoV1::verify` errors → `validate_state` rejects → the permissionless migration PUT is refused → existing rooms migrate to empty. This passes all unit tests AND the WASM-hash migration CI (which never re-verifies signatures). **Required fix:**
```rust
#[serde(default, skip_serializing_if = "Vec::is_empty")]
pub deputies: Vec<MemberId>,   // LAST field of MemberInfo; never reorder the existing 3
```
`skip_serializing_if` makes an empty vec omit from serialization → byte-identical to the old 3-field record → old signatures still verify. Only a genuinely non-empty deputy list changes the bytes, and those records are freshly signed under the new WASM. There is no `skip_serializing_if` precedent in `common/` (prior `#[serde(default)]` cases were on UNSIGNED containers, so AGENTS.md's "just use serde(default)" backward-compat rule is incomplete for signed sub-structs). **Gate it with a regression test** that signs an old-shape (3-field) `MemberInfo` and asserts it still `verify_signature`s against the new struct with `deputies` empty (i.e. byte-identical serialization) — the test must fail if `skip_serializing_if` is dropped.

- Adding `deputies` changes `room_contract.wasm` → room contract key changes → **auto-migration** (permissionless, self-authorized state); clients re-derive the new key, probe the old, re-PUT forward (#292). **Register the outgoing hash BEFORE the WASM changes**: `cargo make add-room-contract-migration` (`common/legacy_room_contracts.toml`).
- A `common/` change also rebuilds the **chat-delegate** WASM (the delegate deserializes/re-serializes `ChatRoomStateV1` → `MemberInfo`, `delegates/chat-delegate/src/subscription.rs:299`), so **register the outgoing delegate hash too**: `cargo make add-migration` (`legacy_delegates.toml`). Both use BLAKE3, never SHA256.
- `MemberInfo` does NOT derive `Default`; it's built with ~30 explicit struct literals, so the new field is a compiler-caught break at each (no silent drop). `ChatRoomStateV1Delta { member_info: .. }` sites are unaffected (delta field type unchanged).
- Republish: UI to Freenet + riverctl to crates.io per `river-publish`. Bump `cli/Cargo.toml`, and since `river-core` (common) changed, bump `[workspace.package]` version + riverctl's `river-core` requirement. The Official room re-keys transparently (auto-migration) and does **not** need recreation — but ONLY if the `skip_serializing_if` fix above is in place.

## Implementation map (verified anchors)

- Deputy list / LWW merge: `common/src/room_state/member_info.rs` (`MemberInfo` :201, `version` :203, delta :72, apply :128).
- Ban authorization: `common/src/room_state/ban.rs` (`validate_invite_chain` :159, `validate_single_ban` :106, `get_invalid_bans` :79).
- Subtree / parent link / cascade: `common/src/room_state/member.rs` (`get_downstream_members` :223, `invited_by`, `remove_banned_members` :212).
- Convergent enforcement: `common/src/room_state.rs::post_apply_cleanup` :93.
- riverctl: `cli/src/commands/member.rs:8`, `cli/src/api.rs:3950`.
- UI: `ui/src/components/members/member_info_modal.rs:123`, `.../ban_button.rs`, `ui/src/signing.rs:91`, `ui/src/components/members.rs`.

Refs #371.

[AI-assisted - Claude]

sanity:
## Problem

The room owner is the only authority that can ban across a whole room (a member can ban only their own invite-subtree). During a spam wave that makes the owner a single point of failure — every removal needs the owner's key. On 2026-07-16 the Official room was spammed and each removal required the owner key with no way to hand a trusted member the ability to help.

## Approach

Let a member **deputize** another member, granting them ban authority scoped to the deputizing member's invite subtree (the owner deputizing is the whole-room "global moderator" case). This is the subtree-scoped delegated-moderation piece of #371.

- **Deputy list reuses `MemberInfo`** (no parallel signed record). `MemberInfo` gains `deputies: Vec<MemberId>`, authenticated by the member's own existing signature and merged monotonic-version LWW-by-record. Deputize = republish `MemberInfo` at `version+1` with the target added; revoke = republish with it removed. A stale re-broadcast loses on version (defuses the CRDT resurrection trap for free).
- **Authorization predicate** (`MembersV1::is_ban_authorized`): a ban by B of T is authorized iff `B == owner OR B is an ancestor of T OR some ancestor A of T lists B in A.deputies`. Guardrail: a non-owner can never ban a member who currently deputizes them ("can't ban your deputizer"). No transitive re-deputization (a deputy's own `deputies` only bind their genuine subtree).
- **Enforcement is retroactive, in `post_apply_cleanup` (NOT `verify`).** The member cascade is recomputed from the converged (members + `member_info.deputies` + bans) state, so revoking a deputy makes their bans stop enforcing and the affected users can rejoin. `verify` never rejects an inert ban (a revoked-deputy ban that removes nobody), which keeps it stable across deputy-state changes and preserves convergence. Bans stay an add-only CRDT (never pruned); the member set is the ground-truth join under anti-entropy.
- **riverctl**: `member deputize` / `member revoke-deputy` republish the caller's own `MemberInfo`; the client-side ban check now shares `is_ban_authorized` with the contract.
- **UI**: a "Deputize" / "Revoke deputy" action in the member-info modal (gated on the *viewer* holding authority, hidden for an empty subtree), and a subtle 🛡 "Deputy" badge in the member list.

### Data-loss safety (the load-bearing bit)

`MemberInfo` is *individually signed* over its ciborium bytes, so adding a field with a plain `#[serde(default)]` would re-serialize a 4-field struct and break every existing member's signature → `validate_state` rejects the permissionless migration PUT → **every existing room migrates to empty**. The field is therefore:

```rust
#[serde(default, skip_serializing_if = "Vec::is_empty")]
pub deputies: Vec<MemberId>,
```

so an empty list serializes byte-identically to the old 3-field record. Gated by `empty_deputies_serializes_identically_to_legacy_member_info`, which fails if `skip_serializing_if` is dropped (verified: it does fail — the CBOR map goes 3→4 entries). Migration entries for the room-contract and chat-delegate WASMs are registered (V26).

## Testing

- **New** `common/tests/deputy_ban_test.rs` (9 tests): deputy ban within/outside subtree, can't-ban-deputizer + multi-deputizer guardrail, owner-deputized global mod (incl. banning own inviter), subtree-root cascade, revoke → un-enforce + stale-replay LWW, ban-vs-revoke convergence via anti-entropy merge, `MAX_DEPUTIES` bound.
- **New** load-bearing serialization regression test in `member_info.rs`.
- Full `river-core` suite green (182 lib + 9 deputy + migration + private-room + DM + convergence).
- `cargo make sync-wasm`, `check-migration`, `check-room-contract-migration`, `cargo test --test migration_test/room_contract_migration_test` all green.
- `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` green; UI compiles with `example-data`.
- `cargo fmt` clean; `cargo make clippy` clean for changed code (one unrelated pre-existing `dm_body.rs` test lint remains; clippy CI is disabled).

A real bug found & fixed along the way: a deputy-banned member's messages orphaned in `recent_messages` (removed in `post_apply_cleanup`, which runs *after* the `recent_messages` field-apply) would fail `MessagesV1::verify`. Added a message sweep in `post_apply_cleanup`.

Closes #410. Refs #371.

[AI-assisted - Claude]


sanity:
## Multi-model review — consolidated (deputy ban authority, #410)

Full-tier review (contract / CRDT-convergence + authorization surfaces). Reviewers: two adversarial Claude lenses (convergence, security) + **three external Codex passes**, each re-run on the changed content per round. Migration safety was verified against the **real live Official-room state** (194,995 bytes: 102 member_info, 200 messages, 9 bans) at every contract commit — the new WASM's `verify` accepts it and re-serializes **byte-identically** (194,995 → 194,995), so no room is stranded on publish.

### Findings fixed

**Round 7 (contract P1s):**
- `max_user_bans` cap-eviction ran *after* ban enforcement + the banner-prune exemption → `cleanup(S) != cleanup(cleanup(S))` (idempotence) and a one-shot member removal by a ban the cap then evicts. Fixed by moving the cap to the front of `post_apply_cleanup`.
- Duplicate `member_info` records: `deputies_of` read first-match while `summarize` kept last → enforcement could diverge from the advertised anti-entropy summary.
- P2s: stale public reaction cache after banned-author message sweep; self-ban rejoin flap; `integration-test`-feature build break.

**Round 8 (root-cause consistency):**
- Canonicalizing only `deputies_of`/`summarize` was insufficient — the UI `MemberList` unioned deputies across duplicates and the nickname/deputy **writers** used first-match, so a later republish could resurrect revoked authority. Fixed with a single `MemberInfoV1::canonical()` selector routed through **every** by-id reader/writer (common + UI + CLI), plus `member_info` dedup in `post_apply_cleanup` and on the `apply_delta` write path.
- Self-ban check now walks the full cached invite chain; deputy/nickname republish version derives from `max(state, cache)` so a stale client can't silently no-op a revoke.

**Round 9:** `DeputyButton` keeps Revoke available when the viewer's invite subtree is emptied by their own deputy's ban (otherwise the appointer could never un-deputize them).

### Confirmed sound
Deputy authorization model (owner never bannable, subtree scoping, self-immunization guardrail), revoked-deputy resolution (a higher-version revoke always outranks and cannot be forged), cap-before-enforcement idempotence, and canonical selection (equal `(version, signature)` rank ⟹ byte-identical record ⟹ identical deputies).

### Dismissed (with justification)
Codex suggested rejecting bans that lack authority **at insertion**. Dismissed: ban authority is checked at **enforcement** (`is_ban_authorized` on converged state), not insertion — deliberately, so `verify` stays stable for convergence + permissionless migration and retroactive revocation works. It grants **no unauthorized power**: a ban is enforced only while its banner is *currently* authorized over the target, so pre-seeding / re-grant only shifts *timing*, never lets anyone remove a target they aren't currently authorized to ban.

### Deferred → #413
The ban cap-eviction has two pre-existing limitations (both strictly worse on `main`'s FIFO cap): a transient cross-delta-order divergence that self-heals via anti-entropy (River's eventual-consistency model — a merge-convergence regression test is included, `cap_eviction_reconciles_via_merge`), and an un-ban DoS via enforcing-absent-target flooding. Both are documented in #413 with mitigation options; the substantive hardening is deferred there per maintainer decision.

[AI-assisted - Claude]

PR #419, opened 2026-07-19
perf(ui): load rooms concurrently on startup instead of N serial delegate round-trips

### LINKED ISSUE #417: perf: startup room load does N serial delegate round-trips (collapse to one batched/parallel load)
## Problem

Loading River on a hosted gateway (e.g. `https://try.freenet.org/v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/`) is slow to show rooms even when the same UI was open moments earlier and the gateway already hosts all the relevant state. A measured breakdown (Playwright + console timeline, nova → try.freenet.org, i.e. best-case latency) shows the "Loading your rooms…" phase costs roughly 2–5s, and it is re-paid on **every** page load.

Root cause: on boot the client rebuilds `ROOMS` from scratch and loads rooms **one at a time**. The sequence is:

1. WS connect → `RegisterDelegate`
2. `fire_list_rooms_request` → `ListResponse` (the set of room keys)
3. a **serial** loop that issues a separate delegate `GetRequest` per room, awaiting each before the next (`response_handler.rs` ~`load_rooms_per_room`, the `for vk in room_vks { … .await }` around lines 706–740), plus a `rooms_meta` GET
4. then a per-room network subscribe/GET

Steps 2–3 are an N+1 round-trip pattern. Each delegate `GetRequest` is a full client↔gateway WebSocket round-trip, and they do not overlap, so wall-clock scales linearly with room count × RTT. The delegate storage on the gateway is node-local and cheap to read; the cost is almost entirely the serial round-trips, amplified by the user's real network latency. (The gateway already persists room state in the chat delegate — the data isn't missing, it's just fetched slowly.)

This is distinct from #386 (room-*join* latency) and #65 (delta vs full-state delegate *writes*); this is about *startup/reload read* fan-out.

## Approach (to be finalized in the PR)

Collapse the N serial per-room reads into a single logical round-trip. Two candidate implementations:

- **(a) Parallelize:** fire all per-room `GetRequest`s concurrently (`FuturesUnordered` / `join_all`) instead of awaiting each in turn. Least invasive; requires that multiple delegate requests can be in flight concurrently (the request/response correlation must not serialize them).
- **(b) Batch:** add a delegate operation that returns all stored room states in one call, so `ListResponse` (or a new response) carries the states directly, eliminating the N follow-up GETs entirely.

Wall-clock target: N×RTT → ~1×RTT. Preference between (a) and (b) depends on whether the delegate request path already allows concurrent in-flight requests and on how invasive a new delegate op is; the PR will justify the choice.

## Testing

- A regression test asserting room load issues a single batched request (or N concurrent requests, not N serial) — i.e. that adding a room does not add a serial round-trip to startup.
- Preserve per-room error isolation: one room failing to load must not abort the others.
- Verify `ROOMS` update ordering / `ROOMS_LOAD_STATE` gating is unchanged (no flicker, no lost rooms).

Works for both hosted and local-node deployments; needs no browser storage.

[AI-assisted - Claude]

sanity:
## Problem

On startup/reload the UI rebuilds `ROOMS` from the chat delegate by issuing one delegate `GetRequest` per room and **awaiting each before firing the next** (the `for vk in room_vks { send_delegate_request(...).await }` loop in `load_rooms_per_room`). Each `GetRequest` is a full client↔node WebSocket round-trip, so the "Loading your rooms…" phase scales as **N × RTT** and is re-paid on every page load. This dominates perceived load time on hosted gateways (e.g. try.freenet.org) once the WASM is browser-cached — the delegate data is already there on the node, it's just fetched over N serial round-trips.

Measured (Playwright, nova → try.freenet.org, best-case latency): the room-loading phase is ~2–5s and grows with room count; on a real user's higher-latency link it's a larger share of the ~10s load.

## Approach

Split the delegate request into an **enqueue** half (register the response waiter + do the synchronous WS send — the only `WEB_API` borrow) and an **`await_delegate_response`** half (the 10s-timeout wait, no borrow). The room load now **enqueues every per-room GET sequentially** (cheap, non-overlapping borrows) then **`join_all`s the awaits**, collapsing N serial round-trips into ~1.

**Parallelize, not batch.** This keeps the change pure UI-side: no delegate WASM change, no delegate-key migration, no wire-format change. A batch `GetMany` delegate op (one message instead of N) would be marginally faster but requires a key migration — filed as a possible follow-up on #417 if profiling shows per-message overhead still matters.

### Correctness (single-threaded WASM)

The one real landmine: a concurrent second `WEB_API.write()` while one is held across `api.send().await` would be a `RefCell` double-borrow panic. This is avoided **by construction** — enqueues are awaited sequentially so the borrow never overlaps; concurrency is confined to the borrow-free await phase. (Previously the serial loop relied on the incidental fact that `WebApi::send` doesn't yield for small payloads; this makes it robust regardless.)

- Enqueue (send) failures funnel through the **same per-room `Err` fold arm** as response failures, so error semantics and the source-scrape pin tests are unchanged (still exactly 3 `had_fetch_error`/`mark_fetch_failure` arms; `value: None` stays a clean skip).
- `send_delegate_request` keeps its exact signature/behavior — all 24 callers unaffected.
- The trailing single `rooms_meta` GET is left as-is (one extra round-trip, not N).
- `LoadWorkerGuard`, `per_room_terminal`, and the deferred single `ROOMS.merge` are unchanged.

## Testing

- **New** pin test `per_room_load_fans_out_concurrently`: asserts the per-room region enqueues + `join_all`s and does **not** use the serial `send_delegate_request` (the trailing meta GET, outside the region, still may).
- Existing pins green: `per_room_load_uses_terminal_and_marks_fetch_failure`, `all_awaited_load_workers_hold_a_guard`.
- 448 `river-ui` native unit tests pass; `wasm32-unknown-unknown` target compiles clean; no new clippy warnings in the changed files.

**Not yet verified end-to-end against a live node** — the load path only exercises against a real gateway, and publishing to the shared River contract must happen from `main` after merge. Plan: after merge, republish from `main` and confirm the room-loading phase is faster on try.freenet.org.

Closes #417

https://claude.ai/code/session_01Dyu4S7t9218XntqizSxAzB

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -4278,11 +4278,47 @@ pub async fn send_delegate_request_to(
     send_delegate_request_inner(delegate_key, request, key_bytes).await
 }
 
-async fn send_delegate_request_inner(
+/// A delegate request that has been SENT and is awaiting its response.
+///
+/// Splitting "send" from "await response" (vs. the combined
+/// [`send_delegate_request`]) lets a caller fire many requests and then await
+/// them concurrently. This is what collapses the startup room load's N serial
+/// round-trips into ~1 (freenet/river#417): the per-room `GetRequest`s are
+/// ENQUEUED sequentially — each enqueue does only the synchronous WS send, so
+/// the `WEB_API` borrow never overlaps another (a concurrent second borrow
+/// would be a `RefCell` double-borrow panic in single-threaded WASM) — and then
+/// their responses are awaited together via `join_all`.
+pub(crate) struct PendingDelegateRequest {
+    receiver: oneshot::Receiver<ChatDelegateResponseMsg>,
+    /// The `PENDING_REQUESTS` correlation key, needed to evict the waiter on a
+    /// timeout.
+    key_bytes: Vec<u8>,
+}
+
+/// Enqueue a request to the CURRENT delegate (bare correlation), returning a
+/// [`PendingDelegateRequest`] to await. The send/await split of
+/// [`send_delegate_request`], for concurrent fan-out (freenet/river#417).
+pub(crate) async fn enqueue_delegate_request(
+    request: ChatDelegateRequestMsg,
+) -> Result<PendingDelegateRequest, String> {
+    let key_bytes = get_request_key(&request);
+    enqueue_delegate_request_inner(CHAT_DELEGATE_KEY.clone(), request, key_bytes).await
+}
+
+/// Register the response waiter and SEND the request, WITHOUT awaiting the
+/// response. Returns a [`PendingDelegateRequest`] the caller awaits (once, or
+/// many concurrently via [`await_delegate_response`]).
+///
+/// The `WEB_API` write borrow is confined to the synchronous `api.send()` here.
+/// `WebApi::send` does not yield for sub-`CHUNK_THRESHOLD` payloads (all delegate
+/// request payloads are small keys), so awaiting each enqueue sequentially keeps
+/// the borrows strictly non-overlapping — the concurrency happens only in the
+/// subsequent `await_delegate_response` phase, which holds no `WEB_API` borrow.
+async fn enqueue_delegate_request_inner(
     delegate_key: DelegateKey,
     request: ChatDelegateRequestMsg,
     key_bytes: Vec<u8>,
-) -> Result<ChatDelegateResponseMsg, String> {
+) -> Result<PendingDelegateRequest, String> {
     debug!("Sending delegate request: {:?}", request);
 
     // Create a oneshot channel to receive the response
@@ -4312,7 +4348,9 @@ async fn send_delegate_request_inner(
         inbound: vec![freenet_stdlib::prelude::InboundDelegateMsg::ApplicationMessage(app_msg)],
     });
 
-    // Get the API and send the request, releasing the lock before awaiting
+    // Get the API and send the request, releasing the lock before returning.
+    // The borrow is confined to this synchronous send (no yield for small
+    // payloads), so sequential enqueue calls never overlap the borrow.
     let api_result = {
         let mut web_api = WEB_API.write();
         if let Some(api) = web_api.as_mut() {
@@ -4332,6 +4370,22 @@ async fn send_delegate_request_inner(
     }
 
     info!("Request sent, waiting for response...");
+    Ok(PendingDelegateRequest {
+        receiver,
+        key_bytes,
+    })
+}
+
+/// Await the response for a request already sent via `enqueue_delegate_request*`.
+/// Many of these can be awaited concurrently (e.g. `join_all`) because each holds
+/// no `WEB_API` borrow — only its own oneshot receiver and timeout.
+pub(crate) async fn await_delegate_response(
+    pending: PendingDelegateRequest,
+) -> Result<ChatDelegateResponseMsg, String> {
+    let PendingDelegateRequest {
+        receiver,
+        key_bytes,
+    } = pending;
 
     // Wait for the response with a timeout
     // Use WASM-compatible sleep from util module
@@ -4356,6 +4410,15 @@ async fn send_delegate_request_inner(
             Err("Timeout waiting for delegate response".to_string())
         }
     }
+}
+
+async fn send_delegate_request_inner(
+    delegate_key: DelegateKey,
+    request: ChatDelegateRequestMsg,
+    key_bytes: Vec<u8>,
+) -> Result<ChatDelegateResponseMsg, String> {
+    let pending = enqueue_delegate_request_inner(delegate_key, request, key_bytes).await?;
+    await_delegate_response(pending).await
 }
 
 // =============================================================================
--- a/ui/src/components/app/freenet_api/response_handler.rs
+++ b/ui/src/components/app/freenet_api/response_handler.rs
@@ -9,16 +9,17 @@ mod update_response;
 use super::error::SynchronizerError;
 use super::room_synchronizer::RoomSynchronizer;
 use crate::components::app::chat_delegate::{
-    arm_legacy_migration_recovery, cas_store_correlation_key, clear_legacy_migration_in_progress,
-    complete_pending_public_key_request, complete_pending_request, complete_pending_sign_request,
-    complete_pending_signing_key_request, decide_legacy_migration_action,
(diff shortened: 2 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #419

### Summary
- **PR Title:** perf(ui): load rooms concurrently on startup instead of N serial delegate round-trips
- **Type:** perf (UI-only; no `common/`/`delegates/`/`contracts/`/Cargo change → no delegate/contract migration)
- **Review tier:** Full (concurrency/async = high-risk surface)
- **Reviewers run:** two rounds. Round 1 (HEAD `7a49abf6`): four Freenet reviewers (code-first, testing, skeptical, big-picture) + external Codex (gpt-5.6). Round 2 (HEAD `a340103a`, after fixes): skeptical + testing re-review + external Codex re-review. HEAD is now `2a73393c` (adds only a compile-time `assert!(ROOM_LOAD_CONCURRENCY > 0)` — trivially additive, no re-review needed).

---

### What the change does
Collapses the startup room load from N serial per-room delegate GETs into a bounded concurrent fan-out. `send_delegate_request_inner` is split into an **enqueue** half (register the oneshot + synchronous WS send — the only `WEB_API` borrow) and an **`await_delegate_response`** half (the 10s-timeout wait, no borrow); `load_rooms_per_room` enqueues each wave sequentially then `join_all`s it, in **waves of `ROOM_LOAD_CONCURRENCY` (16)**.

---

### Findings and resolution

**Round 1 — Codex P1 (real, FIXED):** the original version fired all N GETs at once, which could overflow the node's freenet-core `FairEventQueue` `Default`-queue cap (`MAX_QUEUED_PER_CONTRACT` = 100) for a large account or a busy queue → rejected rooms → partial/`LoadFailed` (a regression the serial loop couldn't hit). **Fixed** by bounding the fan-out to waves of 16 (well under the cap; still collapses N round-trips into ~⌈N/16⌉).

**Round 1 — skeptical Low (dedup, FIXED):** the concurrent fan-out registers all waiters under their bare key in the single-waiter `PENDING_REQUESTS` map, so a duplicate `vk` would clobber its own sibling and spuriously mark a fetch failure. The delegate key index is unique by construction, but the serial loop was robust to a dup and the fan-out wasn't — **restored** with an order-preserving dedup.

**Round 1 — testing Should-Fix (pin, FIXED):** the concurrency pin asserted the fan-out markers were present but didn't guard the borrow-safety ordering (a single-loop `join_all(map(|vk| enqueue().await ...))` refactor would pass while re-introducing the RefCell double-borrow). **Strengthened** to assert `enqueue` precedes `join_all` and no `enqueue` appears inside the fan-out (the round-2 testing lens verified this now fails CI on three distinct unsafe refactors).

**Round 1 — skeptical Medium (FALSE POSITIVE, verified):** a claimed transient "Couldn't load your rooms" `LoadFailed` flash on WS-reconnect-during-load. Traced and refuted: a reconnect goes through `set_up_chat_delegate()` → `begin_load_attempt()`, which bumps `LOAD_ATTEMPT_GEN` **before** firing the new `ListRequest`, so the superseded pass's `worker.set_load_state`/`mark_fetch_failure` no-op via `load_attempt_is_current`. Confirmed by the code-first lens and re-confirmed in round 2. No change needed; the overlap comment now documents why.

**Round 2 — Codex P1-a (queue rejection under contention) — DISMISSED, documented:** a 16-wave is less gentle on the shared `Default` delegate queue than serial's one-at-a-time. Assessed as a rare edge, not blocking: (1) that queue drains at sub-ms (`get_secret` is a synchronous host call), so sustained 84+ depth requires a synchronized thundering herd; (2) only delegate traffic shares the `Default` queue — contract GET/SUBSCRIBE carry a contract identity and use per-contract queues (`QueueKey::Contract`), and the room-load wave is the only concurrent delegate fan-out in River (every other delegate path is serial send-then-await), so peak concurrent delegate depth ≈ 16 + a small constant; (3) under a genuine herd the aggregate across clients dominates and naive per-client auto-retry would *amplify* it (an anti-pattern the repo's own retry rules warn against); (4) failures already recover gracefully — `LoadFailed` renders a Retry control and a reconnect re-fires the load under a fresh attempt generation. Both Claude re-reviewers independently concluded 16 is safe. The serial-retry pass Codex suggests is a valid future robustness enhancement but re-opens the heavily-pinned #397 correctness path for a rare edge that already self-recovers; deferred. (Owner @sanity chose to merge as-is with this documented.)

**Round 2 — Codex P1-b (shared wave timeout) — DISMISSED, documented:** the delegate WASM is compiled *before* the room-load wave runs (`RegisterDelegate` + the `ListRequest` that drives this very load both complete first — the wave is spawned from the `ListResponse`), so the wave never pays first-compile. Each `get_secret` is sub-millisecond, so a 16-wave finishes in ~ms, far under the shared 10s window. The concern only manifests on a pathologically overloaded node where the serial loop would also fail — not a practical regression.

**Round 2 — testing nice-to-have (ADDED):** compile-time `assert!(ROOM_LOAD_CONCURRENCY > 0)` so a future edit to 0 fails to build rather than panicking in `slice::chunks(0)`.

---

### Testing
| Unit / pin | ✅ | 448 river-ui native tests pass. New `per_room_load_fans_out_concurrently` pins the bounded-wave + borrow-safety ordering; existing `per_room_load_uses_terminal_and_marks_fetch_failure`, `all_awaited_load_workers_hold_a_guard`, `worker_global_writes_are_attempt_scoped` all still green. |
| Integration/sim/E2E | N/A | No behavioral delegate-load harness exists in-repo (pre-existing gap, not introduced here); source-scrape pins are the repo convention for this path. |

`wasm32-unknown-unknown` compiles clean; no new clippy findings in the changed files. Not yet exercised end-to-end against a live node — the actual latency win will be verified after merge by republishing from `main` and measuring on try.freenet.org (publishing to the shared River contract only from `main`).

---

### Verdict
**State:** Ready to Merge. Both Claude re-review lenses returned merge-ready with no blocking findings; the one real regression (Codex round-1 P1) is fixed; the two round-2 Codex P1s are dismissed with specific, mechanism-based justification (owner-approved). Re-review was run on the fixed HEAD as required for a high-risk (concurrency) change.

**HEAD SHA reviewed:** fixes reviewed at `a340103a`; current HEAD `2a73393c` adds only the trivial compile-time guard.

[AI-assisted - Claude]

PR #421, opened 2026-07-19
perf(ui): cut new-user cold-load spinner from ~4.5s to ~1.2s (LOAD_IDLE_MS recalibration)

sanity:
## Problem

Measured the cold first-click load of River on try.freenet.org for a brand-new user (empty delegate). The app **shell is interactive at ~240ms** (Create Room / Enter Invite / Import ID clickable, "Connected"), but the **room rail shows "Loading your rooms…" for ~4 seconds** before resolving to "No rooms yet".

That ~4s is a single client-side constant: **`LOAD_IDLE_MS`** — the quiescence window the empty-load resolver waits after the last worker settles before declaring Empty. It was **`4_000`** and measured as **~89% of a new user's cold-load-to-settled time**. It hits every new user on every cold load (in the hosted null-origin iframe the localStorage "migration-done" flag can't persist, so the empty-delegate ProbeLegacy path re-runs each load). The constant's own comment said it should be "short (sub-second)" — 4_000ms contradicted that.

| Milestone (cold, brand-new user, nova best-case) | ms |
|---|---|
| Shell interactive (Create Room / Enter Invite clickable) | ~240 |
| Legacy probe burst settles (72 fire-and-forget probes) | ~490 |
| **Room rail → "No rooms yet" (gated on `LOAD_IDLE_MS`)** | **~4,511** |

## Approach

Recalibrate `LOAD_IDLE_MS` **4_000 → 1_500ms**, sized to the legacy-probe response burst plus a margin for real-world latency. The ~72 fire-and-forget legacy probes (24 delegate generations × 3) come back within ~1 RTT of dispatch (measured ~130ms spread, all in by ~490ms on a fast link), so 1_500ms keeps a **~3x margin** over that burst (covering ~1s of extra browser↔node RTT for a distant user) while cutting the new-user cold-load-to-"No rooms yet" from **~4.5s to ~2s**.

### Why 1_500 and not lower (the review's key finding)
The **anti-flash protection is timing-dependent, not structural.** The one cohort that can flash is a returning user whose **current** delegate is empty but a **legacy** delegate still has data (first load after a delegate-WASM key bump — the #397 case). Their data-bearing legacy response spawns a migrate worker that cancels the timer — but only once the response *arrives*. If the window elapses first, the rail briefly shows "No rooms yet" then self-corrects to the room list (cosmetic, no data loss). A fast-link-calibrated 750ms left only ~1.5x margin over the ~490ms burst and could flash that cohort on slow links; **1_500ms restores a comfortable ~3x margin.** Brand-new users (the optimization target) get no legacy response at all, so they resolve Empty with zero flash risk regardless of the value.

No `LoadFailed` is masked by the shorter window: transport failures on the probe path are recorded synchronously *before* the timer arms, and a late blob-parse failure self-corrects via worker settlement (`settle_terminal → LoadFailed`). `load_state_after_probe_legacy(true)` still returns `Loading` — the resolution logic and its pins are unchanged; only the debounce window is shorter.

Not addressed here (clean follow-ups): the 72-probe legacy burst itself (wasteful for new users, floods the delegate queue), and the ~1.5MB WASM download (the real-world time-to-interactive lever for distant users).

## Testing

- **Two-directional guard test** `load_idle_window_stays_calibrated` pins `1000..=2500ms`, so a future edit can regress **neither** the perf goal (too large → ~4s pad) **nor** the #397 anti-flash margin (too small → flash on slow links).
- All existing #397 behavior pins stay green: `probe_legacy_fired_stays_loading`, `new_user_resolves_to_empty_never_stuck_never_failed`, `idle_resolution_cancelled_by_any_activity`, room_list's `new_user_resolves_to_empty_not_stuck_loading`.
- 449 river-ui native tests pass; wasm32 compiles clean; no new clippy findings.
- Drive-by: corrected stale "30s" backstop comments (`LOAD_HARD_MAX_MS` is `60_000`).

UI-only (no `common/`/delegate/contract change) → no migration, no coupled riverctl publish.

https://claude.ai/code/session_01Dyu4S7t9218XntqizSxAzB

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -1262,21 +1262,23 @@ mod tests {
         );
     }
 
-    /// freenet/river#417 follow-up: the empty-load quiescence window is
-    /// deliberately short — calibrated to the ~1-RTT legacy-probe response burst
-    /// (all in by ~490 ms on a fast link), because it was measured as ~89% of a
-    /// brand-new user's cold-load time-to-"No rooms yet" on try.freenet.org.
-    /// Guard against a silent regression back to the old multi-second value
-    /// (was 4_000 ms). Safety of a short window: a legacy delegate that actually
-    /// HAS data spawns a migrate worker (`pending > 0` in `idle_should_apply`),
-    /// which keeps the state non-Empty, so a short grace can't miss a real
-    /// migration — it only speeds up the genuinely-empty new-user case.
+    /// freenet/river#417 follow-up: the empty-load quiescence window is calibrated
+    /// to the ~1-RTT legacy-probe response burst (all in by ~490 ms on a fast
+    /// link), because it was measured as ~89% of a brand-new user's cold-load
+    /// time-to-"No rooms yet" on try.freenet.org. Guard BOTH directions:
+    /// - too LARGE re-introduces the old ~4 s new-user cold-load pad (was 4_000 ms);
+    /// - too SMALL re-opens the #397 anti-flash margin — the anti-flash protection
+    ///   is timing-dependent (a delegate-bump migrator's late legacy response must
+    ///   arrive before this window elapses, else the rail flashes "No rooms yet"
+    ///   then self-corrects), so the window must keep a comfortable margin over
+    ///   browser↔node RTT for that cohort.
     #[test]
-    fn load_idle_window_stays_short() {
+    fn load_idle_window_stays_calibrated() {
         assert!(
-            LOAD_IDLE_MS <= 1_000,
-            "LOAD_IDLE_MS ({LOAD_IDLE_MS}ms) gates the new-user cold-load; keep it \
-             calibrated to the ~1-RTT legacy-probe burst, not a multi-second pad"
+            (1_000..=2_500).contains(&LOAD_IDLE_MS),
+            "LOAD_IDLE_MS ({LOAD_IDLE_MS}ms) must stay calibrated to the legacy-probe \
+             burst: small enough not to pad the new-user cold-load, large enough to \
+             keep the #397 anti-flash margin over RTT for delegate-bump migrators"
         );
         assert!(
             LOAD_IDLE_MS < LOAD_HARD_MAX_MS,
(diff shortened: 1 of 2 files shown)


sanity (commented):
## PR Review: #421 (Light tier — external Codex + skeptical read)

**Change:** `LOAD_IDLE_MS` recalibration (`4_000 → 1_500ms`) + guard test + comment fixes. UI-only; no migration.

### Round 1 findings — all addressed
- **[Codex P2 / skeptical Medium] Fast-link-calibrated flash.** The initial 750ms was sized to the ~490ms nova-local burst (~1.5x margin). Both reviewers noted the anti-flash protection is **timing-dependent, not structural**: a returning legacy-only user (empty current delegate, data on a pre-key-bump legacy delegate — the #397 cohort) could get a self-correcting "No rooms yet" flash if their data-bearing legacy response arrives after the window elapses on a slow link. **Fixed:** widened to 1_500ms (~3x margin over the burst, ~1s extra RTT headroom); still cuts new-user cold-load ~4.5s→~2s. Brand-new users get no legacy response, so zero flash risk regardless.
- **[skeptical Low] Inaccurate PR-body reasoning** ("fetch failures only occur on the PerRoom path"). **Fixed:** corrected — failures *are* recorded on the ProbeLegacy path, but no `LoadFailed` is masked because transport failures are recorded synchronously before the timer arms and late blob-parse failures self-correct via worker settlement.
- **[skeptical Low] One-directional guard test.** **Fixed:** now `load_idle_window_stays_calibrated` pins `1000..=2500ms` — guards against regression in both directions (perf pad / anti-flash margin).
- **[skeptical Low] Stale "30s" backstop comments** (`LOAD_HARD_MAX_MS` is `60_000`). **Fixed** (drive-by, both files).

### Verified safe (skeptical)
Attempt-scoping (#397) intact and independent of the window length; hard-max backstop unchanged; PerRoom (has-rooms) path unaffected (timer never arms); no `LoadFailed` masking; microtask/macrotask ordering means the only deciding factor is whether the WS response arrives before the window elapses.

### Round 2 (Codex, on the 1_500ms HEAD) — clean
> "The timeout recalibration is internally consistent with the documented performance tradeoff, and the accompanying test constrains it appropriately. The remaining changes only correct stale comments."

### Verdict
**Ready to merge.** 449 river-ui native tests pass; wasm32 compiles clean; no clippy findings. HEAD reviewed: `4679aca1`.

[AI-assisted - Claude]

PR #429, opened 2026-07-21
feat(ui): per-room notification bell + dedicated modal

sanity:
## Problem

River already has per-room notification preferences (All / Mentions & replies / Muted) — the enum, the notification-filtering logic, and delegate-backed persistence all shipped in #340. But the control was buried in a `<select>` inside the room-details (edit-room) modal, so it was undiscoverable, and its help text said "applies on this device only", which is misleading since the setting is stored in the `rooms_meta` delegate blob and soft-syncs across a user's devices.

## Approach

Surface the setting behind a **bell icon in the conversation header** (right edge, cleanly separated from the room-name/`(i)` "room details" target) that opens a **compact dedicated modal**:

- The bell reflects state — `FaBell` when notifying, `FaBellSlash` when muted — with a `title` tooltip naming the exact mode.
- The modal lists the three modes with icons (bell / @ / bell-slash), the current one highlighted with a check; selecting one applies immediately and closes (one-click pick). Backdrop / ✕ dismiss without change.
- The bell is now the **single home** for the setting, so the dropdown is removed from the room-details modal (one place, no drift).

**Storage is unchanged** — kept in the `rooms_meta` delegate blob (local-wins cross-device merge), same model as `room_order`. This was a deliberate call over localStorage: "mute this room" is the dominant case and users expect mute to follow them across devices, and it keeps notification prefs consistent with the other delegate-backed view preferences. Fixed the misleading copy → "Applies to this room and syncs across your devices."

No delegate/contract/common WASM change (UI-only), so **no migration needed**.

## Testing

- `cargo test -p river-ui --bins` — 479 passed.
- `cargo clippy` / `cargo fmt` — clean (no new warnings in changed code).
- New `ui/tests/notification-bell.spec.ts` (Playwright): header presence + default, opening the modal (3 modes), select-applies-and-closes + bell/tooltip updates + reopened selection state, close-leaves-unchanged, and that the setting no longer appears in room-details. **25/25 green** across chromium / firefox / webkit / mobile-chrome / mobile-safari.

## Screenshot

Header bell → dedicated modal:

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -15,6 +15,7 @@ use crate::components::members::member_info_modal::MemberInfoModal;
 use crate::components::members::Invitation;
 use crate::components::room_list::create_room_modal::CreateRoomModal;
 use crate::components::room_list::edit_room_modal::EditRoomModal;
+use crate::components::room_list::notification_modal::NotificationModal;
 use crate::components::room_list::receive_invitation_modal::{
     accept_invitation, clear_invitation_from_storage, decide_recovered_invitation,
     is_invitation_processed, load_invitation_from_storage, load_invitation_nickname_from_storage,
@@ -47,6 +48,8 @@ pub static MEMBER_INFO_MODAL: GlobalSignal<MemberInfoModalSignal> =
     Global::new(|| MemberInfoModalSignal { member: None });
 pub static EDIT_ROOM_MODAL: GlobalSignal<EditRoomModalSignal> =
     Global::new(|| EditRoomModalSignal { room: None });
+pub static NOTIFICATION_MODAL: GlobalSignal<NotificationModalSignal> =
+    Global::new(|| NotificationModalSignal { room: None });
 pub static CREATE_ROOM_MODAL: GlobalSignal<CreateRoomModalSignal> =
     Global::new(|| CreateRoomModalSignal { show: false });
 pub static PENDING_INVITES: GlobalSignal<PendingInvites> = Global::new(PendingInvites::new);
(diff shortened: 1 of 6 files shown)


sanity:
## Review record (Light tier)

**Scope**: UI-only (no delegate/contract/common WASM change — CI migration checks confirm byte-identical WASMs, no migration entry needed). Branch rebased onto current `origin/main` (58609f40); one conflict in `conversation.rs` resolved (main's `panel_unread` memo + this PR's `current_notification_mode` memo are both kept — purely additive), so the review below ran on the post-rebase content.

**External model pass (Codex)** on the rebased diff vs origin/main:
- **[P1] Direct signal writes in modal close handlers** (backdrop + ✕) — valid; the documented Firefox-mobile RefCell re-entrancy crash path. **Fixed** in 4b1a08ca (both wrapped in `crate::util::defer`, matching the mode-select path). The identical pre-existing pattern in `EditRoomModal` is tracked as #464.
- No other findings.

**Adversarial read** (signal-safety focus):
- `current_notification_mode` memo reads `CURRENT_ROOM.read()` (infallible) before `ROOMS.try_read()` — holds a backup subscription, per the documented try_read-Err-registers-nothing trap.
- New modal computes state inline in the render body (no `use_memo` over non-signal values) per the always-mounted-modal rule; the only mode writer is the modal itself, which closes on select.
- All signal mutations routed through `crate::util::defer`.

**Verification on the rebased head** (4b1a08ca):
- `cargo test -p river-ui --bins`: 504 passed.
- `cargo check` (wasm32, no-sync): clean; remaining warnings pre-exist on main.
- Full Playwright suite vs the built example UI: 414 passed / 15 skipped / 1 flaky (mobile-safari `mobile-touch-ux` scroll-to-latest — unrelated to this diff, passed on retry, run started against a just-regenerated build dir).
- notification-bell spec: 25/25 across chromium/firefox/webkit/mobile-chrome/mobile-safari, zero retries.

[AI-assisted - Claude]

PR #431, opened 2026-07-21
fix(ui): gate message send/edit on encoded content size, not raw text bytes

### LINKED ISSUE #430: Messages near the size limit are silently lost: UI gate counts raw text bytes but the contract enforces encoded size
## Problem

Reported by HostFat on Matrix (2026-07): a message under 1000 characters was rejected after send with the console showing

```
WARN ui/src/components/conversation.rs:1720 Message too long: 1006 encoded bytes, max 1000 bytes
```

and the message was **silently lost** — no error shown, draft cleared. Earlier the same session: "a message was lost" and "a very long message was't sent :(".

## Root cause

Three checks measure different things:

1. **Input gate + byte counter** (`message_input.rs`): raw `text.len()` ≤ `max_message_size`.
2. **Client safety net** (`conversation.rs` `handle_send_message`): `RoomMessageBody::content_len()` — the **encoded** content. CBOR framing adds ~9 bytes for plain text; replies embed the quoted author name + ~100-char preview (~150 bytes); private rooms add the 16-byte AES-GCM tag.
3. **Contract validation** (`message.rs`): prunes messages with `content_len() > max_message_size` — the authoritative limit is encoded bytes.

Any message in the gap between (1) and (2) passes the UI, has its draft cleared by `send_message`, then is dropped by the safety net with only a console warning. For replies the gap is ~150 bytes, so even ~850-char replies vanish. Multi-byte UTF-8 widens the perceived gap further (users count characters; the limit counts bytes).

**Edits have no size gate at all**: an over-limit edit action (`ActionContentV1` overhead on top of the text) is signed, sent, and silently pruned by contract validation on every peer.

riverctl is not affected — it builds the real body and errors visibly (`guard_message_size`).

## Fix

Measure the exact encoded size in the UI using the same construction code path the senders use (`RoomMessageBody::measure_text/measure_reply/measure_edit` in river-core, pinned equal to real built bodies by tests), gate Send/Save and the byte counter on that, and keep the draft in place when over the limit.

[AI-assisted - Claude]

sanity:
## Problem

HostFat (Matrix, 2026-07): messages near the 1000-byte limit are **silently lost**. The console shows `Message too long: 1006 encoded bytes, max 1000 bytes` for a message under 1000 characters; the draft is already cleared, so the text is gone with no visible error.

The input gate and byte counter compare raw `text.len()` against `max_message_size`, but the contract enforces the **encoded** content size (`RoomMessageBody::content_len`): CBOR framing adds ~9 bytes for plain text, replies embed the quoted author + ~100-char preview (~150 bytes), private rooms add the 16-byte AES-GCM tag. Any message in that gap passes the UI gate, gets its draft cleared, then is dropped by the client safety net (or, failing that, silently pruned by contract validation on every peer). Edits had **no size gate at all** and could be silently pruned the same way.

Closes #430

## Approach

Single source of truth for size: river-core gains `RoomMessageBody::measure_text / measure_reply / measure_edit`, which build the same content the send paths build (identical construction code, so they cannot drift) and add `ENCRYPTION_TAG_OVERHEAD` (16) for private rooms. The UI send gate, byte counter, Send button, and the (new) edit-form gate all use these measures, so the number the user sees is the number the contract enforces. Over-limit drafts stay in the input instead of being lost.

The old silent safety nets remain but now log an error explicitly marked as a measure-drift bug.

riverctl already did this correctly (`guard_message_size` on the built body) and is unchanged.

Alternatives considered: raising `max_message_size` doesn't fix the class (any limit has an encoding gap); subtracting a fixed overhead from the text gate undercounts replies/unicode and drifts.

## Testing

- `measure_tests` in river-core pin `measure_* == content_len()` of actually-built bodies across CBOR length-prefix boundaries, multi-byte UTF-8, reply metadata variants, and — under `ecies-randomized` — real AES-256-GCM private bodies.
- `raw_text_gate_undercounts_encoded_size` pins the bug class itself (998 raw chars → 1007 encoded > 1000; 900-char reply → over).
- New Playwright spec `message-size-gate.spec.ts` pins the user-visible fix: 998 raw chars disables Send, shows `1007/1000`, and **Enter keeps the draft**; 990 chars shows `999/1000` and Send enabled; 499 two-byte chars (998 bytes) disables Send.
- `cargo test -p river-core` (all features): green. `cargo check/clippy -p river-ui` (wasm32): clean.

[AI-assisted - Claude]

### DIFF
--- a/common/src/room_state/message.rs
+++ b/common/src/room_state/message.rs
@@ -12,6 +12,13 @@ use std::collections::HashMap;
 use std::fmt;
 use std::time::SystemTime;
 
+/// Ciphertext overhead added by AES-256-GCM (`encrypt_with_symmetric_key`):
+/// the 16-byte authentication tag appended to the plaintext. The nonce lives
+/// in a separate field of [`RoomMessageBody::Private`] and does not count
+/// toward [`RoomMessageBody::content_len`]. Pinned against real encryption
+/// by the `measure_*_matches_private_*` tests (feature `ecies-randomized`).
+pub const ENCRYPTION_TAG_OVERHEAD: usize = 16;
+
 /// Computed state for message actions (edits, deletes, reactions)
 /// This is rebuilt from action messages and not serialized
 #[derive(Clone, PartialEq, Debug, Default)]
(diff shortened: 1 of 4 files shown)


sanity:
## Consolidated multi-perspective review (Full tier)

**External model substitution notice:** Codex (`codex review --base main`, gpt-5.6-sol) was invoked and failed with a usage-limit error (quota resets Jul 25); Gemini CLI remains unavailable (deprecated free tier, `IneligibleTierError`). Per the review policy fallback, model diversity was replaced with widened perspective diversity: **four independent Claude reviewers**, each blind to the others, with distinct lenses — *skeptical bug-hunt*, *code-first (code before description)*, *testing/coverage*, and *change-specific failure mode* (measurement-drift + consumer audit). Reviewed at 4cb44ca4; fixes below re-verified at b1844531.

**Verdict: no blockers, no majors from any lens.** Key verifications:

- **Drift-proof by construction**: `measure_text/reply/edit` call the same `XContentV1::new(...).encode()` path as the body constructors; the one constant (+16 AES-256-GCM tag, no AAD, nonce excluded from `content_len`) is pinned against real encryption by CI-run tests (`ecies-randomized` builds in build.yml).
- **Boundary parity**: gate `<= max` / contract `retain(content_len <= max)` — exactly-max messages sendable and retained; pinned by a Playwright test.
- **Traced the bug flow**: 998 ASCII chars → measured 1007 > 1000 → Send disabled, Enter is a no-op, draft retained. Before: draft cleared, message silently dropped.
- **All edit save paths gated** (Enter, Save button, handler safety net); kebab/up-arrow are entry points that funnel into the gated form.
- **Consumer audit**: riverctl already correct (`guard_message_size` on built bodies, all action types); DM modal is a separate subsystem with ~256-byte headroom vs ~60-byte sealing envelope — safe by design, untouched; no other send site in the workspace uses a raw-text gate.
- **Deliberate scoping**: reaction/delete actions remain ungated — emoji comes only from the fixed picker set (worst case ~80 encoded bytes vs 1000 limit), payloads bounded by construction.
- **Semver**: river-core additions are purely additive → minor bump whenever it's next published; no coordination needed for this PR.

**Findings → resolution** (every finding fixed or dismissed with justification):
| Finding | Resolution |
|---|---|
| Test comment said 998 chars → 1006 (actual 1007) — flagged by 3 of 4 reviewers | Fixed (b1844531) |
| Edit-form gate + `measure_draft` reply/private plumbing had zero test coverage (revert-to-`text.len()` passed green) | Fixed: native `measure_draft` unit test + edit-form E2E + exact-limit E2E (9df382a1); 25/25 locally across 5 browsers |
| Save-button onclick lacked the guard Enter-save has (render-lag race → silent safety-net drop) | Fixed (b1844531) |
| No-secret private fallback over-measures by 16 bytes | Dismissed: safe direction (blocks ~16 bytes early on an already-broken room), pre-existing, never loses a message |
| `usize::MAX` fallback for edit gate when no room loaded | Dismissed: unreachable (no room ⇒ no edit form); overflow-safe arithmetic added anyway |

[AI-assisted - Claude]

PR #433, opened 2026-07-21
feat(dm): portable sent DMs — seal a sender copy into each DM (#432)

### LINKED ISSUE #432: Can't read your own sent DMs on a new node (sender plaintext is device-local only)
## Report

From Ivvor: "you can't read your own sent DMs any more on the new node." Received DMs are fine; the user's OWN sent DMs render as `sent — ciphertext only`. Also observed: *some* sent DMs visible, others missing.

## Root cause

An in-room DM (`common/src/room_state/direct_messages.rs`) is stored as ECIES ciphertext encrypted **only to the recipient's** `member_vk`. The sender cannot decrypt their own sent DM from contract state. The sender's plaintext is kept in a **device-local** chat-delegate cache (`outbound_dms`, #256) that never travels the network.

So a sent DM is readable only on the node whose local cache holds its plaintext:
- Sent from *this* node → cached here → readable.
- Sent from another node, or before a delegate-key change stranded the cache (the migration gate only probes `outbound_dms` when the current delegate has no rooms) → plaintext absent → placeholder.

This is a design limitation of the local-only cache, not a regression from a specific PR. Recent delegate-key churn (#394, #411, #416) turned existing caches "legacy" and made it surface.

## Fix

Seal a second ECIES envelope of each DM body to the sender's own key, carried in the message (`DirectMessage.sender_ciphertext`), so any device with the sender's signing key reads its own sent DMs from contract state — like received DMs already work. Forward-looking (does not recover DMs sent before the change). Room-contract + chat-delegate wire change → both re-key (migration).

[AI-assisted - Claude]

sanity:
## Problem

An in-room DM is stored in the room contract as ECIES ciphertext encrypted **only to the recipient**. The sender cannot decrypt their own sent DM from contract state, so historically the sender's plaintext lived only in a device-local chat-delegate cache (`outbound_dms`, #256). That cache never travels the network, so on a new node — or after a delegate-key change strands it (the migration gate only probes `outbound_dms` when the current delegate has no rooms) — the user sees `sent — ciphertext only` instead of their own sent messages.

Reported by Ivvor: *"you can't read your own sent DMs any more on the new node."* Closes #432.

## Approach

`compose_direct_message` now seals a **second** ECIES envelope of the same body to the sender's own `member_vk`, carried in a new `DirectMessage.sender_ciphertext: Option<Vec<u8>>`. Any device holding the sender's signing key recovers its own sent DMs from contract state via `open_own_direct_message`, exactly the way received DMs already work. The UI (`dm_thread_modal`) and CLI (`riverctl dm list`) fall back to this sender copy on an outbound-cache miss before showing the placeholder.

Why a struct field and not a combined envelope inside the opaque `ciphertext`: `unseal_dm_from_sender` consumes everything after byte 44 as the AES-GCM ciphertext, so appending a second envelope would break old recipients. A separate signed field is the only backwards-compatible carrier — at the cost of a room-contract wire change.

**Wire-format discipline**
- `Option`, `#[serde(default, skip_serializing_if = "Option::is_none")]`, placed last, so a pre-change message serializes byte-identically and old state still decodes / re-PUTs (permissionless migration intact).
- Covered by the sender signature: appended to `build_direct_message_signed_bytes` **only when `Some`**, so a legacy (`None`) message signs over the exact pre-change bytes and still verifies, and the copy can't be stripped or grafted without invalidating the signature.
- Size-capped to `MAX_DM_CIPHERTEXT_BYTES` in both `verify` and `apply_delta`.

**Forward-looking:** recovers sent DMs authored *after* this change on any device; it does **not** resurrect DMs sent before it (they carry no sender copy). Historical recovery on the same node would need the separate outbound-cache migration-gate fix.

## Migration

Adding the field changes both the room-contract **and** chat-delegate WASM (both compile `ChatRoomStateV1`), so both re-key. V28 registry entries added to `legacy_delegates.toml` and `common/legacy_room_contracts.toml` with the pre-change hashes; WASMs rebuilt via `sync-wasm`. river-core 0.1.15 → 0.1.16, riverctl 0.1.79 → 0.1.80.

## Testing

New tests in `common/tests/direct_messages_test.rs`:
- signature back-compat (None layout == legacy prefix), signature coverage (strip/graft both fail), verify + apply_delta size caps
- end-to-end sender-copy round-trips: sender reads own DM with only their key; recipient still reads; third party reads neither; legacy message → `open_own` errs

Also closes a **pre-existing CI gap**: `common/tests/direct_messages_test.rs` never ran in CI (only `--lib` + specific `--test` targets did), so the whole DM verify/apply_delta/compose surface was ungated. Added a `build.yml` step running it with `--features ecies-randomized`.

Green locally: 62 DM integration tests, 229 river-core lib, 479 river-ui bin; migration + delegate-key + wasm-sync + wbindgen checks; `fmt`; `clippy` (no new warnings).

## Not done here (deploy-time, needs your call)
- **Draft** pending multi-model review (high-risk wire/crypto/contract surface — Full tier).
- Republishing the River UI (new contract key + delegate migration ritual) and releasing riverctl are deploy steps, not in this PR.

[AI-assisted - Claude]

https://claude.ai/code/session_01RKvh1NuNCx4ucxsXPPRNZP


sanity:
## Multi-model review (Full tier)

**External-model substitution (recorded per policy):** Codex was invoked (`codex review --base main`) but **hit its usage limit** (`You've hit your usage limit … try again Jul 25`); the Gemini CLI tier is deprecated/unavailable. With both external models down and the reset several days off, I used the **diverse-Claude-subagent fallback**: three independent adversarial reviewers, each blind to the others, with distinct lenses — (1) crypto + wire-compat, (2) code-first correctness, (3) big-picture / test-coverage. Re-running the real external pass before merge is advisable (Codex resets Jul 25).

### Consensus
All three independently confirmed the **crypto, signature coverage, serde back-compat, and migration are sound**. Two reviewers reproduced the V28 migration hashes by hand and confirmed the committed wire-format snapshot (`direct_messages_wire_format.hex`) is **byte-unchanged vs main** (legacy `None` messages sign/serialize identically). CI's `check-delegate-migration`, `check-room-contract-migration`, and `check-wasm-sync` agree.

### Findings and disposition (all addressed in e785c9b)
| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| F1 | Med | `riverctl dm list` "you have an invitation" tip counted OUTBOUND invites (sender-copy fallback decodes them), but `dm accept` is inbound-only → tip led nowhere | tip now `is_invite && !outgoing` |
| F2/4a | Med | Outbound invite self-copies rendered as full Accept **cards** (shared inbound closure), breaking the `BodyKind::Invite`-inbound-only invariant + showing the sender an enabled Accept button on their own invite | outbound invites now render as the `[Invitation] …` summary via the pure, unit-tested `resolve_outbound_dm_body`; inbound reverted to its own `render_inbound_body` |
| 6 | Low | Three `#431` refs (unrelated PR) in `dm_thread_modal.rs` | → `#432` |
| 4b | Low | Refactor dropped the `can_participate()` rationale comment | restored |
| F3 | Nit | Module threat-model doc didn't note the doubled ciphertext | documented |
| Cov | Med | Render fallback untested at crate level | added 4 river-ui unit tests for `resolve_outbound_dm_body` |

### Deferred (tracked, not fixed here)
- **Historical recovery** (finish-the-fix): this PR is forward-only — it can't resurrect DMs sent before it. The same-device recoverable subset (outbound-cache migration gate) is filed as #435. New-device historical DMs are unrecoverable by any means.
- **CI gap:** most `common/tests/*.rs` suites still don't run in CI (this PR adds `direct_messages_test`); enumerated in #435.

No blocking findings remain. Draft pending CI (build + playwright) and your call on deploy (republish + delegate migration + riverctl release are deploy-time steps).

[AI-assisted - Claude]

sanity:
Closing this without merging. The work is sound and I do not want it lost, so here is exactly where it stands.

**Branch: `worktree-feat-portable-sent-dms`, head `e785c9bc`.** Not deleted. Everything below is on it.

### What is done

- `DirectMessage.sender_ciphertext: Option<Vec<u8>>` — a second ECIES envelope of the same body sealed to the sender's own `member_vk`, so any device holding the sender's signing key recovers its own sent DMs from contract state instead of relying on the device-local `outbound_dms` cache.
- Wire-format discipline worked through carefully: `Option` + `skip_serializing_if` + placed last, so a pre-change message serialises byte-identically; the field joins `build_direct_message_signed_bytes` **only when `Some`**, so a legacy message signs over the exact pre-change bytes and still verifies, and the copy cannot be stripped or grafted without invalidating the signature; size-capped in both `verify` and `apply_delta`.
- UI (`dm_thread_modal`) and CLI (`riverctl dm list`) fall back to the sender copy on an outbound-cache miss.
- Tests: signature back-compat, strip/graft rejection, size caps, and end-to-end round-trips (sender reads own; recipient still reads; third party reads neither; legacy → `open_own` errs).
- It also closes a genuine CI gap: `common/tests/direct_messages_test.rs` never ran in CI, so the whole DM verify/apply_delta/compose surface was ungated. **That part is independently valuable and worth lifting out on its own.**

### Why it is not merging as-is

1. **It is a coordinated deploy, not a merge.** Adding the field re-keys BOTH the room-contract and chat-delegate WASM, so landing it commits us to a delegate-migration ritual and a UI republish. That is the maintainer's call and the maintainer's publish — it is not something to land while clearing out drafts.
2. **Its version bumps are stale in a way that now collides.** The branch bumps riverctl 0.1.79 → 0.1.80 and river-core 0.1.15 → 0.1.16. riverctl is at 0.2.5 with a release train in flight, so those hunks have to be redone against current `main` rather than replayed.
3. **The PR asked for a Full-tier review it never got** — its own description says "Draft pending multi-model review (high-risk wire/crypto/contract surface)". A signed wire-format change to a live contract deserves that review at the moment it is going to ship, against the code that is going to ship, not months earlier.
4. **The base is from 2026-07-21** and `main` has moved a long way since (#494, #517, #518, #520, #521 among others).

### To pick it back up

`git checkout worktree-feat-portable-sent-dms`, rebase onto `main`, redo the two version bumps against current versions, re-run `sync-wasm` so the V28 registry entries in `legacy_delegates.toml` / `common/legacy_room_contracts.toml` carry the *current* pre-change hashes, then Full-tier review, then the publish ritual.

Worth doing first and separately: the `build.yml` step that runs `common/tests/direct_messages_test.rs` with `--features ecies-randomized`. That is a CI hole today regardless of whether the sender-copy feature lands.

#432 stays open and I have added the state of play there.

[AI-assisted - Claude]

Issue #398, opened 2026-07-12
Adopt the freenet-migrate crate in River (replace hand-rolled backward-probe machinery)


sanity:
## Background

River's contract/delegate migration runs on its own hand-rolled backward-probe machinery: `common/src/migration.rs`, `ui/src/components/app/freenet_api/backward_probe.rs`, `chat_delegate.rs`, the `common/build.rs` / `ui/build.rs` legacy-hash codegen, and the `legacy_room_contracts.toml` / `legacy_delegates.toml` registries. This is shipped and proven — the stdlib 0.6→0.8 republish (2026-07-12, commit 47be34ea) migrated rooms and delegate secrets transparently on the live network.

The reusable **`freenet-migrate`** crate (v0.1.0 on crates.io; repo `freenet/freenet-migrate`) is the *generalization of exactly this pattern*, extracted from River and Delta. River does not use it yet — verified, zero references in River's `Cargo.toml` / `Cargo.lock`. So the crate's own origin app doesn't dogfood it.

## Goal

Refactor River onto `freenet-migrate` / `freenet-migrate-build`, deleting the duplicated hand-rolled machinery. This dogfoods the crate on its origin app and removes a maintenance-burden copy of the pattern.

## Tasks (incremental, behavior-preserving — see `freenet-migrate/docs/design.md` §5)

- [ ] Point `common/build.rs` + `ui/build.rs` at `freenet_migrate_build::codegen()`, aliasing the generated `LEGACY_*` consts to River's existing names so call sites compile unchanged.
- [ ] Replace the internals of `common/src/migration.rs` + `backward_probe.rs` with `freenet-migrate` crate calls (a `LegacyExportAdapter` plugs River's existing per-room export for pre-crate versions).
- [ ] Swap the CI "WASM hash changed ⇒ old hash must be registered" guard to the crate's checker.
- [ ] Delete the now-duplicated hand-rolled code.
- [ ] Re-verify the full migration path (room + delegate, backward-probe) against a live 0.8 node — the same end-to-end checks the 0.6→0.8 republish passed (zero data loss, zero #204).

## Sequencing

Do this **after Delta adopts `freenet-migrate`**. Delta is the first external adopter; its adoption will shake out any crate rough edges, and River benefits from going second. **Not urgent** — River's current machinery works and is proven; this is de-duplication + dogfooding cleanup, not a functional gap.

## References

- Umbrella: freenet-core#2776 ("Tracking: Graceful upgrades for delegates and contracts")
[...]

PR #434, opened 2026-07-21
refactor: adopt freenet-migrate-build codegen for both legacy registries (#398 phase 1)

sanity:
## Problem

River's two legacy registries (`legacy_delegates.toml`, `common/legacy_room_contracts.toml`) are turned into Rust consts by two hand-rolled build scripts — a duplicated copy of the exact pattern the `freenet-migrate` crate generalizes, with no validation beyond hex-length panics. The crate's own origin app doesn't dogfood it (#398).

## Approach

Phase 1 of #398 (deliberately minimal + behavior-preserving): swap only **who generates the consts**. River keeps running its own migration machinery unchanged.

- `common/build.rs` + `ui/build.rs` now call `freenet_migrate_build::codegen()` (0.2.0, freenet/freenet-migrate#4) reading the **existing** `[[entry]]` TOMLs — same const names, types, `(delegate_key, code_hash)` tuple order, and byte values. −233/+89 lines.
- Every build now **validates** the registries: hash decode errors are build failures, and every delegate row's key is cross-checked against `BLAKE3(code_hash ‖ params)` — restoring at build time the compile check the ecosystem lost (the Feb 2026 wrong-derivation incident class).
- `ui/build.rs` behaviors preserved: `.rerun_if_changed(false)` keeps the `BUILD_TIMESTAMP_ISO` re-run heuristic alive; `.allow_missing_registry(true)` keeps the docs.rs missing-TOML fallback. `common/build.rs` keeps its rerun directives + fallback.

**Registry data finding** (from running the crate's validation over the full 24-row registry): **V1** is the only row whose recorded delegate key predates the standard derivation — it now carries `irregular_key = true` (recorded key trusted as-is; it's the key the old delegate actually had on the network). **V2's key derives correctly**, so `migration_test.rs`'s historical "V1 | V2" exemption was over-broad and is tightened to V1 only — V2 now gets full validation it never had.

## Testing

- **New pin:** `legacy_set_fingerprint_is_stable_across_codegen_changes` locks the legacy-set fingerprint (`741360a5d34a3a8c`) that keys the per-user migration-done localStorage flag. If a codegen change altered `LEGACY_DELEGATES` bytes **or order**, every user would silently re-run legacy migration — this test makes that loud. It passing on this PR is the proof the swap is byte-identical.
- 480 river-ui native tests pass; `migration_test` + `room_contract_migration_test` pass; `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` clean.
- No WASM files touched; `check-delegate-migration` / `check-room-contract-migration` / `check-cli-wasm` CI confirm byte-identical WASMs.

The crate side merged as freenet/freenet-migrate#4 (Full-tier multi-perspective review; its fixtures include River's complete registries verbatim, so the exact data this PR feeds it is pinned upstream too).

Next phases of #398 (separate PRs, after this): contract-probe internals on the crate's decision driver, then the delegate path in shadow mode.

Part of #398 / freenet-core#2776.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -702,7 +702,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2293,6 +2293,18 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "freenet-migrate-build"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c1f8c31bbf8b6902fa6b1afb866cdba9234578ac8923d6a75f3e6043551db39f"
+dependencies = [
+ "blake3",
+ "bs58",
+ "serde",
+ "toml 0.8.23",
 ]
 
 [[package]]
@@ -4645,6 +4657,7 @@ dependencies = [
  "curve25519-dalek",
  "data-encoding",
  "ed25519-dalek",
+ "freenet-migrate-build",
  "freenet-scaffold",
  "freenet-scaffold-macro",
  "freenet-stdlib 0.8.3",
@@ -4670,6 +4683,7 @@ dependencies = [
  "dioxus",
  "dioxus-free-icons",
  "ed25519-dalek",
+ "freenet-migrate-build",
  "freenet-scaffold",
  "freenet-stdlib 0.8.3",
  "futures",
@@ -4684,7 +4698,6 @@ dependencies = [
  "serde",
  "sha2",
  "thiserror 2.0.18",
- "toml 0.8.23",
  "tracing",
  "wasm-bindgen",
  "wasm-bindgen-futures",
@@ -6773,7 +6786,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -72,10 +72,9 @@ migration = []
 mentions = []
 
 [build-dependencies]
-# Used by build.rs to parse legacy_room_contracts.toml and generate the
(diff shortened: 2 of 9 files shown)


sanity:
## Consolidated multi-perspective review

**External-model pass — substitution record.** Codex remains usage-limited (reset 2026-07-25 07:53; verified failure earlier today on the crate-side PR); Gemini CLI unusable (`IneligibleTierError`). Per the review policy fallback, three independent Claude reviewers ran with distinct lenses, blind to each other: **skeptical bug-hunt**, **consumers/CI** (every surface that doesn't get the change), and **migration-failure-modes** (the data-loss classes this repo has hit before). The crate side of this change already had a Full-tier review on freenet/freenet-migrate#4, including a concrete walkthrough of this exact swap.

### Verdict: clean. Independently verified by multiple reviewers:
- **Fingerprint stability** (the one silent data-loss-class risk): `741360a5d34a3a8c` reproduced from the TOML by two reviewers independently; main's hand-rolled output and the crate's output hash identically, so no user re-runs legacy migration. The pin test runs in CI (`build.yml` → `cargo test -p river-ui --bins`), so CI-green on this PR is byte-identity proof **against the published 0.2.0 crate** — closing the only methodology caveat raised.
- **WASM identity**: neither generated file reaches either WASM (`migration` is a non-default, client-only feature; both WASMs build `-p`-scoped; `LEGACY_DELEGATES` lives in river-ui, not the delegate crate) → no re-key, no migration entry, no republish. Confirmed empirically by `check-delegate-migration` / `check-room-contract-migration` passing.
- **All 24 delegate rows recomputed via b3sum**: V1 alone fails the standard derivation (correctly flagged `irregular_key`); V2–V27 derive — validating the migration_test tightening.
- **Every TOML consumer checked**: scripts (`^code_hash`/`^version` greps), `add-migration.sh` append format, pre-commit hook, both direct-parse test files (no `deny_unknown_fields`), Makefile version detection — none affected by the new field.
- **Build-script semantics preserved**: same const names/types/order/filenames/include! sites; ui emits zero rerun-if-changed directives (BUILD_TIMESTAMP heuristic intact); common keeps its two; crates.io/docs.rs packaging intact (TOML ships inside river-core; lockfile checksum matches the published crate).

### Findings addressed
- Room-contract registry had no value-pin (asymmetry with the delegate side) → added `registry_values_and_order_are_stable_across_codegen_changes` (`d931340e569e9c74`, 27 entries) — `7cc1d042`
- Version-range comment nit → fixed — `cdda0a78`

### Dismissed with justification
- *Missing-registry fallback narrowed* (old: empty list on ANY read error; new: only on file-not-found, hard-fail on present-but-unreadable). Intentionally stricter — a silently-empty `LEGACY_DELEGATES` from a transient read error would disable every legacy probe, which is the worse failure. The docs.rs genuinely-absent case is unchanged.

Merging after CI is green on the current head.

[AI-assisted - Claude]

Issue #427, opened 2026-07-20
riverctl read follows a stale backward upgrade-pointer and REPLACES (not merges) → stale/empty message list on multi-generation rooms


sanity:
## Problem

A room shows all messages on the River UI but a stale/empty subset on `riverctl` (0.1.77+), even when both compute the same current contract key. Sending via riverctl works (the message lands on the key the UI reads) but riverctl's own `message list` shows stale state. Debug shows riverctl "grabbing a new contract key for an upgrade pointer." Reported by Ivvor 2026-07-20.

**Not data loss.** The full state (e.g. 186 messages / 106 members on the official room's `c53ded28` generation) is intact; the write path lands correctly; only riverctl's *display* is stale. Fully recoverable once fixed. Not caused by riverctl 0.1.78 / PR #415 — the read/migration/upgrade-pointer code is byte-identical to 0.1.77 (`git show <415-merge> -- cli/src/api.rs` is empty); the `c53ded28` WASM shipped in 0.1.77 (#416), a day earlier.

## Root cause: read-vs-write key split via a stale FORWARD upgrade pointer, made worse by replace-vs-merge

- **Write** (`api.rs::send_message`) always PUTs to the current bundled-WASM key (`owner_vk_to_contract_key`). The UI reads that key directly → sees the message. ✓
- **Read** (`get_room` → `ensure_room_migrated` → `follow_upgrade_chain`, `api.rs:~1475`): if the current-key state carries an `OptionalUpgradeV1` forward pointer, riverctl GETs the target and **REPLACES** its state with it (`state = next_state; id = next`), returning the target generation's state.
- The pointer on the current gen points **backward** to an intermediate generation (e.g. `c53ded28 → 5L5ZAB`), so `follow_upgrade_chain` walks `c53 → 5L`, replaces c53's fresh state with 5L's older state, hits 5L's `→c53` pointer, sees c53 already visited (cycle guard), stops, and returns **5L's stale state.**

### How the backward cross-pointer gets onto the current generation

[...]

PR #436, opened 2026-07-21
refactor(ui): drive the backward probe with freenet-migrate's decision driver (#398 phase 2)

sanity:
## Problem

The UI backward-probe recovery (freenet/river#292) is the data-loss-critical
path that recovers a room stranded under an older room-contract generation
(each room-contract WASM upgrade moves the contract key, stranding dormant
rooms under an older key). It carried its own hand-rolled probe state machine:
a `ProbeState` struct, a `BACKWARD_PROBES` side-table, and a `PROBE_EPOCH`
counter whose only job was to make late GET responses / stale watchdogs
single-shot. That is exactly the decision logic freenet-migrate 0.3.0's
sans-IO `ProbeDriver` now owns and tests exhaustively.

## Approach

Phase 2 of #398: swap the probe INTERNALS onto `freenet_migrate::ProbeDriver`,
behavior-preserving. River supplies its app semantics through a `ProbeStateOps`
impl (`RiverProbeOps`):

- `decode` = the existing defensive CBOR decode (`try_from_cbor_slice`, `None`
  on failure → a miss, never a panic).
- `is_real` = `configuration.verify_signature(&owner_vk).is_ok()` (the existing
  predicate; a default/placeholder state is a miss).
- `merge_with_local` = the existing `merge_room_states` CRDT merge (kept
  `pub(crate)` and reused verbatim, including keep-primary-on-merge-failure).
- `prepare_forward` is left as the driver's **identity** default on purpose —
  River adopts the UNSTRIPPED merge locally and strips the upgrade pointer only
  in `put_state_to_current_key` (freenet/river#427). Overriding it would strip
  too early and change what is adopted locally. A unit test pins this.

The hand-rolled machinery becomes two plain-`Mutex` maps (a route table
instance-id → `(owner, per-fire token)`, and one `ProbeDriver` per owner) plus a
`pump_probe` loop over `next_action()`. The terminal `Outcome` runs the same
recover-or-seed side effects, extracted into
`get_response::{adopt_recovered_probe_state, seed_current_key_with_local}` with
the same log strings, ROOMS placeholder-vs-merge adoption, PUT-forward, and
SYNC_INFO / notification updates. The 64-hop cap (now the driver default,
pinned by a compile-time assert) and the 12s watchdog (with its assert vs
`REPUT_DELAY_MS`) are preserved. Per-candidate observability (the pre-driver
"undeserializable" / "empty, advancing" logs) is restored by classifying the
response with the same `RiverProbeOps` predicates for the LOG only — the
decision stays the driver's.

### Correlation: where the driver is enough, and where it is not

`PROBE_EPOCH` is deleted, but its job is split, not simply subsumed:

- **Within a single probe**, the driver's per-candidate single-shot correlation
  (gated by single-shot route removal — first of {response, watchdog} owns the
  hop) covers late responses and stale timeouts.
- **Across sequential probes for the same owner**, the legacy ids are
  deterministic and get reused, so a completed probe's still-armed watchdog can
  fire against a *later* probe that re-registered the same id. The driver cannot
  see that. Each probe GET therefore carries a monotonic **per-fire token** (the
  `PROBE_ROUTES` value); a watchdog advances only if the route still carries its
  own token (`claim_route_if_current`), so a stale watchdog no-ops instead of
  rolling the later probe back to an older generation. This is the guard the old
  per-fire epoch counter provided.

Timing — the two paths that can start a second same-owner probe:

- The **subscription-timeout sweep** is serialized out of the race by
  construction: a probe re-stamps `subscribing_since` every hop and
  `PROBE_GET_TIMEOUT` (12s) is below `REPUT_DELAY_MS` (20s), so the sweep never
  reclaims a room mid-probe (pinned by a compile-time assert).
- The **reconnect path** is NOT serialized: on a WebSocket reconnect
  `room_synchronizer` sets the room `Disconnected` and clears
  `subscribing_since`, and `rooms_awaiting_subscription` re-GETs with no
  throttle — so a second probe can start well inside the first's 12s window.
  This is the unthrottled path the per-fire token exists to cover.

(The cross-probe token is the fix for the Codex/review finding on this PR — see
the `fix(ui): guard the backward probe against a stale cross-probe watchdog` and
`test(ui): exercise the route/watchdog layer …` commits.)

This mapping follows the verified adoptability review in freenet-migrate#5.

## Testing

- `cargo test -p river-ui --bins` — **485 pass**. Driver-integration + route-layer
  tests at the River level: RiverProbeOps classification (real / empty /
  undecodable), `prepare_forward` keeps the upgrade pointer (pins the byte-exact
  #427 behavior), merge preserves the recovered config, newest-generation-wins
  (anti-rollback), exhaustion-seeds-local, single-shot late-response, and the
  cross-probe stale-watchdog regression
  (`stale_watchdog_does_not_advance_a_later_same_owner_probe`) — which drives the
  real PROBE_ROUTES / per-fire token / `claim_route_if_current` gate against a
  live driver in PROBE_DRIVERS, asserting a stale watchdog neither removes the
  later probe's route nor advances its driver. The existing #292 gate +
  import-recovery source-grep pins pass unchanged.
- `cargo test -p river-core` — pass.
- `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync`
  — freenet-migrate 0.3 compiles for wasm32 (its stdlib dep unifies on the
  workspace's 0.8.3, so `ContractInstanceId` types match — no conflict).
- `cargo fmt` clean; clippy clean for the changed files; **no `.wasm` changes**.

Refs #398 (phase 2). Crate review lineage: freenet-migrate#5.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2296,6 +2296,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "freenet-migrate"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "80de2864555343441300c6fa1aced20f989ec16657d651be3bb30bff1d1224c8"
+dependencies = [
+ "blake3",
+ "bs58",
+ "ciborium",
+ "ed25519-dalek",
+ "freenet-scaffold",
+ "freenet-stdlib 0.8.3",
+ "serde",
+ "serde_with",
+]
+
+[[package]]
 name = "freenet-migrate-build"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4683,6 +4699,7 @@ dependencies = [
  "dioxus",
  "dioxus-free-icons",
  "ed25519-dalek",
+ "freenet-migrate",
  "freenet-migrate-build",
  "freenet-scaffold",
  "freenet-stdlib 0.8.3",
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -86,6 +86,10 @@ river-core = { workspace = true, features = ["ecies", "ecies-randomized", "migra
 # Freenet dependencies
 freenet-scaffold.workspace = true
 freenet-stdlib = { workspace = true, features = ["net"] }
+# Sans-IO backward-probe decision driver (freenet/river#398 phase 2). The `ui`
+# default profile exposes the probe/driver surface; its stdlib dep unifies with
+# the workspace's 0.8.x.
+freenet-migrate = "0.3"
 
 futures = "0.3.30"
 futures-timer = "3.0.3"
--- a/ui/src/components/app/freenet_api/backward_probe.rs
+++ b/ui/src/components/app/freenet_api/backward_probe.rs
@@ -18,18 +18,33 @@
 //! * Neither current nor any legacy key has state → only then is the device's
 //!   local snapshot PUT to seed the current key.
 //!
+//! ## The decision driver (freenet/river#398 phase 2)
+//!
+//! The probe *decisions* — candidate order (newest-first), what counts as a
+//! hit, when to advance, when to stop, what to adopt — live in
+//! `freenet_migrate`'s sans-IO [`ProbeDriver`]. This module is the thin I/O
+//! adapter: it pumps GETs and timeouts through the driver and runs the app's
+//! recover-or-seed side effects on the terminal [`Outcome`]. River's
+//! app-specific semantics (CBOR decode, "is this real state", CRDT-merge with
+//! the local snapshot) are supplied by [`RiverProbeOps`].
+//!
+//! The driver's per-candidate correlation is single-shot by construction, so
+//! there is no cross-probe epoch machinery to get wrong (the previous
+//! hand-rolled state machine carried one; it is gone).
+//!
 //! ## Routing & bookkeeping
 //!
 //! A GET response for a *legacy* contract key cannot be resolved back to an
 //! `owner_vk` via `SYNC_INFO` (keyed by the current contract id) or via
-//! `RoomData::contract_key` (also the current id). [`BACKWARD_PROBES`] is the
-//! side-table that maps a probe's legacy `ContractInstanceId` back to the probe
-//! it belongs to.
+//! `RoomData::contract_key` (also the current id). [`PROBE_ROUTES`] is the
+//! side-table that maps the outstanding legacy `ContractInstanceId` back to the
+//! owner whose probe it belongs to; [`PROBE_DRIVERS`] holds one in-flight
+//! [`ProbeDriver`] per owner.
 //!
-//! It is a plain `Mutex`-guarded map, NOT a Dioxus signal: it is internal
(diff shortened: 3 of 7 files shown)


sanity:
## Consolidated multi-perspective review (Full tier)

**External-model pass — substitution record.** Codex remains usage-limited (reset 2026-07-25 07:53; verified failure earlier today); Gemini CLI unusable (`IneligibleTierError`). Fallback per policy: three independent Claude reviewers, blind to each other — **skeptical diff bug-hunt**, **behavioral parity** (line-level checklist against the pre-swap machinery, derived from a full behavioral spec of the shipped probe), **Dioxus signal-safety / async re-entrancy** (River's #1 crash class). The crate side (freenet-migrate 0.3.0's driver) carried its own Full-tier review on freenet/freenet-migrate#5.

### The load-bearing finding — found independently by ALL THREE reviewers (fixed in `0937fb5f`)
Deleting the `PROBE_EPOCH` machinery was NOT fully subsumed by the driver's single-shot correlation: for **sequential probes of the same owner**, a completed probe's still-pending 12s watchdog could claim the *next* probe's route for the same deterministic newest legacy id — advancing it past its newest candidate (anti-rollback violation) or eating the route before the real response landed (dropped hit). Reachable via the reconnect path, which re-GETs `Disconnected` rooms with no throttle. Fix: per-fire monotonic token in the route value; the watchdog advances only via `claim_route_if_current` (owner AND token must match) — restoring exactly the old epoch guarantee in the new structure. `deliver_probe_response` deliberately ignores the token (any response for a routed id is for the current fire; commented at both sites). Pinned by `stale_watchdog_does_not_advance_a_later_same_owner_probe`, which exercises the route/guard layer directly (the reviewers' point that the original 6 tests only covered raw drivers). Module docs and PR body now state the correlation split honestly (driver = within-probe; token = cross-probe) and the timing relationships (`REPUT_DELAY_MS` 20s > 12s serializes the sweep path; the token covers the unthrottled reconnect path).

### Also addressed
- Observability nit: per-candidate "undeserializable" / "empty, advancing" logs restored (classification for logging only; the driver still owns the decision) — `8e2b7584`

### Verified preserved by the parity + skeptic lenses (line-level, old vs new)
Trigger gate and no-seed-on-probe-start; newest-first order and first-real-hit-wins; decode/is-real predicates; 12s watchdog=advance-never-abort; per-hop deferred `touch_subscribing_since`; send-failure→watchdog; hop cap 64 (compile-asserted); hit side effects statement-for-statement (unstripped local adoption + stripped forward PUT per #427, pinned by a new test; same log strings, register/Subscribing/mark-initial-sync/mark-needs-sync); exhaustion seed path; duplicate-start dedup; poisoned-mutex recovery; all pre-existing pin tests untouched; no WASM changes (CI-confirmed).

### Signal-safety lens: clean
All signal mutations deferred as before; no new bare reads; no mutex held across an await; `safe_spawn_local` throughout; the recursive pump is hop-cap-bounded; defer-ordering caveat respected. (Two pre-existing patterns — bare `ROOMS.read()` at the trigger site and `WEB_API.write()` across send — are unchanged from main and out of scope.)

CI green on the final head → merging per the review-is-the-gate policy.

[AI-assisted - Claude]

PR #437, opened 2026-07-21
refactor(cli): drive legacy room recovery with freenet-migrate's decision driver (#398 phase 2b)

sanity:
## Problem

riverctl's `fetch_room_state_with_recovery` (cli/src/api.rs) recovers a room
stranded under an older room-contract generation (each room-contract WASM
upgrade moves the contract key) with a hand-rolled newest-first legacy loop —
the same decision logic freenet-migrate 0.3.0's sans-IO `ProbeDriver` owns and
tests exhaustively. Phase 2b swaps the legacy loop onto the driver, matching
what phase 2 (#436) did for the UI probe.

## Approach

Behavior-preserving. The legacy backward search runs on a `ProbeDriver` with
`RiverCliProbeOps` supplying the CLI's semantics:

- `decode` = the existing `try_get_state` deserialization (ciborium +
  `rebuild_actions_state`; `None` on failure → a miss).
- `is_real` = `configuration.verify_signature(&owner_vk).is_ok()`. Defensively
  redundant in this integration (see fidelity note) but kept as the classifier
  of record.
- `merge_with_local` = **pass-through** returning the recovered state verbatim —
  the CLI does NOT merge device-local state (unlike the UI). The local snapshot
  is an empty default, so exhaustion (`SeedLocal`) / `NoLegacy` map to the
  existing not-found error; the CLI **never seeds**.
- `prepare_forward` = **identity** — `put_room_state` already strips the upgrade
  pointer on the forward PUT (#427).

Kept OUTSIDE the driver, exactly as today: the current-key fast path
(`CURRENT_GET_TIMEOUT` + upgrade-chain walk) and the per-candidate forward
upgrade-chain walk. The chain-walk stays in the I/O adapter `probe_legacy_bytes`,
which resolves a legacy GET (`LEGACY_PROBE_TIMEOUT`), walks the forward chain,
and hands the driver the resolved **raw bytes** (via the shared
`encode_probe_state`) so the driver owns the final decode + is_real
classification.

The outcome→result mapping is a pure, I/O-free `resolve_legacy_recovery_outcome`
(unit-tested on both arms): `Recovered` → the state under the current id plus a
forward-PUT decision; every non-`Recovered` outcome → the not-found `Err`. The
async forward PUT stays in `fetch_room_state_with_recovery`, gated on that
decision.

### Hop cap sized to the registry (with_max_hops)

The pre-driver loop probed EVERY legacy key unbounded; the driver's default
64-hop cap would silently strand generations past 64. The registry is 27 today
(no behavior change), but to keep that true for a future >64-generation
registry the driver is built with
`.with_max_hops(legacy_count.max(freenet_migrate::DEFAULT_MAX_PROBE_HOPS))`, so
the cap only ever guards a runaway registry and never truncates a real one.
Pinned by the source-grep test.

### Round-trip fidelity of the re-encode

`probe_legacy_bytes` re-serializes the chain-walk-resolved state and the driver
re-decodes it. This is byte-faithful: `actions_state` is the ONLY
`#[serde(skip)]` field in the whole room state and is deterministically rebuilt
in `decode`; every other field (including the individually-signed
`configuration` and the `#[serde(default, skip_serializing_if)]` `deputies`)
round-trips losslessly through ciborium, and re-serializing the config cannot
change its signature. Behavior is identical for the three cases: current-key
hit, legacy hit with forward-PUT, and exhaustion error.

### Why the raw driver, not `migrate_contract`

The crate's pumped entry point `migrate_contract` consumes a
`ContractLineageEntry` lineage, but riverctl only has newest-first `ContractKey`s
from `river_core::migration`. Feeding the raw `ProbeDriver` +
`NewestFirst::assume_ordered` + a trivial `while let Step::Get` pump avoids
fabricating lineage values just to satisfy that signature; the pump is the same
loop `migrate_contract` runs.

### Not in scope (follow-up)

`get_migration_state` / `ensure_room_migrated` (the separate migration path)
still use the hand-rolled legacy loop; swapping them is a later PR.

## Testing

- `cargo test -p riverctl` — pass (196 in the lib suite; 22 in
  `migration_recovery_tests`). New tests:
  - `RiverCliProbeOps` classification + never-merges; `prepare_forward` identity
    keeps the upgrade pointer.
  - `river_cli_probe_ops_roundtrips_fully_populated_state` — a fully-populated,
    owner-signed fixture (every field non-default; recent_messages with an edit +
    reaction + delete so `actions_state` rebuild is exercised) round-tripped
    through the SAME production `encode_probe_state` + `decode`, asserting
    equality across every field.
  - `cli_driver_recovers_fully_populated_room_verbatim` — a full `ProbeDriver`
    run recovers that fixture verbatim (`Outcome::Recovered.merged` == original).
  - `cli_driver_adopts_newest_real_generation` (anti-rollback),
    `cli_driver_exhausts_to_seed_local`.
  - `resolve_legacy_recovery_outcome` both arms:
    `recovery_outcome_recovered_returns_state_and_reports_forward_put` and
    `recovery_outcome_seed_local_and_no_legacy_map_to_not_found_err` (the
    not-found Err text pinned verbatim as a stable caller contract).
  - `legacy_recovery_routes_through_decision_driver` — source-grep pin:
    `NewestFirstWins` + `with_max_hops` + the pure resolver, and exactly one
    `put_room_state` call (CLI-never-seeds).
  - Existing #427 forward-PUT-strip pin and chain-walk tests pass unchanged.
- `cargo fmt` clean; clippy clean for `cli/src/api.rs`; **no `.wasm` changes**
  (staged by name).

Refs #398 (phase 2b). Crate review lineage: freenet-migrate#5; UI counterpart: #436.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4739,6 +4739,7 @@ dependencies = [
  "dialoguer",
  "directories 5.0.1",
  "ed25519-dalek",
+ "freenet-migrate",
  "freenet-scaffold",
  "freenet-stdlib 0.8.3",
  "freenet-test-network",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -53,6 +53,10 @@ atty = "0.2"
 river-core = { version = "0.1.15", path = "../common", features = ["ecies", "ecies-randomized", "migration", "mentions"] }
 freenet-stdlib = { workspace = true, features = ["net"] }
 freenet-scaffold = "0.2.2"
+# Sans-IO backward-probe decision driver (freenet/river#398 phase 2b): drives
+# the legacy-generation recovery search (newest-first, first real generation
+# wins). Its stdlib dep unifies with the workspace's 0.8.x.
+freenet-migrate = "0.3"
 
 # Serialization (for contract state)
 ciborium = "0.2.2"
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -240,6 +240,40 @@ impl ProbeStateOps for RiverCliProbeOps {
     // to the caller. Do NOT add a prepare_forward override.
 }
 
+/// Map a completed backward-probe [`Outcome`] to the recovery result. Pure and
+/// I/O-free so both arms are unit-testable without a live node — the async
+/// forward migration stays in `fetch_room_state_with_recovery`.
+///
+/// * `Ok((state, current_id, migrate_forward))` — a generation was recovered;
+///   `migrate_forward` reports whether the caller should PUT it onto the current
+///   contract.
+/// * `Err(..)` — nothing recovered (`SeedLocal` / `NoLegacy` / an already-taken
+///   outcome). riverctl NEVER seeds device-local state, so every non-`Recovered`
+///   outcome is the not-found error. The error TEXT is a stable contract callers
+///   and scripts may match on — keep it verbatim (pinned by
+///   `recovery_outcome_seed_local_and_no_legacy_map_to_not_found_err`).
+fn resolve_legacy_recovery_outcome(
+    outcome: Option<Outcome<ChatRoomStateV1>>,
+    current_id: ContractInstanceId,
+    legacy_count: usize,
+) -> Result<(ChatRoomStateV1, ContractInstanceId, bool)> {
+    match outcome {
+        Some(Outcome::Recovered { merged, source, .. }) => {
+            // `source` is a legacy id, never the current key (the legacy
+            // registry excludes the current WASM — pinned by
+            // `current_wasm_is_not_in_legacy_registry`), so this guard is
+            // unreachable-true today. Kept as defence: if a future change ever
+            // probed the current key, it prevents a pointless self-PUT.
+            Ok((merged, current_id, source != current_id))
+        }
+        _ => Err(anyhow!(
+            "Room not found on the current contract or any of the {legacy_count} known \
+             previous contract generations. The room may never have existed, or its \
+             state may have been garbage-collected from the network."
+        )),
+    }
+}
+
 /// Compute the contract key for a room from its owner verifying key.
 /// This uses the current bundled WASM to ensure consistency.
 pub fn compute_contract_key(owner_vk: &VerifyingKey) -> ContractKey {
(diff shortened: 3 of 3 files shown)


sanity:
## Consolidated multi-perspective review

**External-model pass — substitution record.** Codex remains usage-limited (reset 2026-07-25 07:53); Gemini CLI unusable (`IneligibleTierError`). Fallback per policy: three independent Claude reviewers, blind to each other — **skeptical bug-hunt**, **behavioral parity** (statement-level, old vs new), **test quality**. The driver itself carries its own Full-tier review (freenet/freenet-migrate#5), as does the UI-side swap (#436).

### Verdicts
- **Parity: fully preserved across all 12 checklist items** — current-key fast path byte-identical; same trigger set; same newest-first order (traced to `legacy_contract_keys_for_owner`'s `.rev()`); same per-candidate 8s timeout + chain-walk; same hit/forward-PUT semantics (strip-on-PUT intact, PUT best-effort); exhaustion error byte-identical; subscribe-only-to-current gate untouched; `get_migration_state`/chain-walk helpers untouched; +5 tests purely additive; timeout constants unchanged; logging equivalent.
- **The load-bearing fidelity claim verified independently by two lenses**: the new re-encode→re-decode surface cannot lose data or flip a decision for River's model — `actions_state` is the only `#[serde(skip)]` field (deterministically rebuilt in decode), `deputies`' skip-if-empty is lossless, and signature checks re-serialize rather than trusting wire bytes, so re-encoding can't break `is_real`. The added `is_real` gate on the chain-walk-resolved state is strictly safer, never looser.

### Findings addressed
- Round-trip fidelity untested on a populated state (top test gap) → `river_cli_probe_ops_roundtrips_fully_populated_state` (every field populated incl. edit/reaction/delete exercising `rebuild_actions_state`) + `cli_driver_recovers_fully_populated_room_verbatim`, both through the **extracted production encoder** `encode_probe_state` (no parallel test copy) — `d493e68e`/`8560de0a`
- Never-seeds grep pin defeatable by an added second PUT → now asserts `put_room_state` appears exactly once; policy anchor robust to `use` restructuring — `d493e68e`
- Async glue only grep-pinned → outcome→result mapping extracted into pure `resolve_legacy_recovery_outcome` with both arms unit-tested and the not-found error text pinned — `d493e68e`
- Driver's default 64-hop cap silently bounded the previously-unbounded walk (inert at 27 gens, future foot-gun) → `.with_max_hops(legacy_count.max(default))` + comment + pin — `8560de0a`
- `is_real` redundancy in this integration documented — `8560de0a`

### Dismissed with justification
- `source != current_id` guard is unreachable-true today (legacy registry excludes the current WASM, pinned) — kept as documented defence rather than deleted, preserving self-PUT protection if a future change probes the current key.
- Fresh `rebuild_actions_state` on the new path (vs whatever the fold left) — wire-invisible; consumers rebuild anyway; at worst more correct.

CI green on the final head (`8560de0a`) → merging per the review-is-the-gate policy. Follow-up (tracked in the PR body): `get_migration_state`/`ensure_room_migrated` remain on the hand-rolled loop.

[AI-assisted - Claude]

PR #439, opened 2026-07-21
feat(cli): riverctl identity whoami — report your own MemberId (#438)

### LINKED ISSUE #438: riverctl: no way to learn your own MemberId (the `author` on your own messages)
## Problem

There is no way to ask riverctl "what is my own member ID in this room?"

`author` in `riverctl message list/stream --format json` is a `MemberId`, but
nothing in the CLI reports the local user's own. Reported by a user building an
XMPP bridge:

> I am having problems with riverctl in identifying my own user when I send a
> message to a river room. [...] without a way to identify my own MemberId, I am
> having to use convoluted methods to identify which of the incoming messages
> are the ones I just sent from riverctl in my XMPP client. This causes somewhat
> annoying issue of the messages I send getting duplicated in my XMPP client's
> group chat buffer.
>
> I need to be able to do this precisely and without first waiting for incoming
> messages from the room and parse the stdout of riverctl for incoming messages
> or something.

The workarounds available today are all indirect:

- `riverctl member list` returns `{member_id, nickname}` for **every** member
  with no marker for which one is you, and it needs a network GET.
- `riverctl identity export` contains the key but is an armored **private key**
  blob, not something a bridge should be shelling out to on a hot path.
- Matching on your own nickname is ambiguous (nicknames are not unique).

## Proposed solution

The reporter suggested either shape; both are cheap and complementary:

> Something like a `riverctl identity whoami <room>` or including
> `self_member_id` in `riverctl room list --format json` command would be nice.

Requirements that fall out of the use case:

- Must resolve **before** any message is observed, and **without** a running
  node — it is a pure `rooms.json` read.
- Must report the *effective* signing identity, so `--signing-key-file` /
  `RIVER_SIGNING_KEY_FILE` selects what is reported. Reporting the persisted
  identity while the override signs would hand a bridge an ID matching none of
  its own messages, which is the exact bug being fixed.
- Per-room, not global: River identities are per-room, so there is no single
  global member ID.

[AI-assisted - Claude]

sanity:
## Problem

There was no way to ask riverctl "what is my own member ID in this room?"

`author` in `riverctl message list/stream --format json` is a `MemberId`, but
nothing in the CLI reported the local user's own — so a bridge could not tell
which streamed messages were its own echo. Reported by an XMPP-bridge author
whose sent messages were duplicating in their client's group chat buffer:

> without a way to identify my own MemberId, I am having to use convoluted
> methods to identify which of the incoming messages are the ones I just sent
> from riverctl [...] I need to be able to do this precisely and without first
> waiting for incoming messages from the room and parse the stdout of riverctl.

Everything available today is indirect: `member list` returns every member with
no self marker (and needs a network GET), `identity export` is an armored
private-key blob, and matching on your own nickname is ambiguous.

## Approach

Both shapes the reporter suggested:

```console
$ riverctl identity whoami <room> --format json
{
  "is_owner": true,
  "member_id": "JYFX4AGJ",
  "nickname": "bridge-bot",
  "room": "CEGwowqUhwrfevhNKDyUTvH7YwgdsYcnTzc4iuQ1nxyU",
  "signing_key_source": "stored",
  "verifying_key": "CEGwowqUhwrfevhNKDyUTvH7YwgdsYcnTzc4iuQ1nxyU"
}

$ riverctl room list --format json   # now carries self_member_id + signing_key_source
```

`whoami` with no room reports every room (River identities are per-room, so
there is no single global member ID). A single named room emits one object, not
a one-element array, so `| jq -r .member_id` works without an index.

Three decisions worth review attention:

**The member ID must equal `author`, by construction.** `api::author_member_id`
is the single source of truth for `MessageV1::author`, called by all eight
production send sites *and* by `whoami`'s derivation. The first version of this
PR instead had each site inline the derivation with a source-scrape test holding
them together — review showed that scrape silently missed the two paths that
send actual chat messages (one spelled it fully-qualified, one used `.into()`),
so the guarantee is now structural rather than textual.

**Offline, before the client exists.** `ApiClient::new*` opens a WebSocket
eagerly, so routing whoami through it made a pure local read fail outright when
the node was down — exactly when a bridge most needs its own ID. `main`
short-circuits to `identity::whoami(&storage, ..)` before the client is built
(~8ms, works with the node stopped). `whoami` takes `&Storage` rather than
`&ApiClient` so it cannot acquire a connection by accident.

**It reports the identity that will actually sign — all three overrides.**
riverctl has three: `rooms.json`, `--signing-key-file`/`RIVER_SIGNING_KEY_FILE`,
and `message send --signing-key`/`RIVER_SIGNING_KEY`. whoami initially honoured
only the first two; review found that a bridge exporting `RIVER_SIGNING_KEY`
(the normal way to run one) would send as key B while whoami reported key A —
silently recreating the very bug #438 exists to fix. whoami now takes
`--signing-key` with the same env var and the same precedence `message send`
uses (inline > file > stored), disclosed as `signing_key_source`. With an inline
key the room need not be in local storage, matching `send_message_with_key`'s
documented storage-free bot workflow.

Nickname comes from the cached room state's canonical `member_info` (what other
members see), falling back to the locally-stored nickname *only when the
effective key is the persisted one* — otherwise an override would be labelled
with another account's name (Codex finding). Private-room nicknames decrypt when
the secret is available and fall back rather than surfacing the `[Encrypted: …]`
placeholder.

`Storage::list_rooms` / `ApiClient::list_rooms` now return a `RoomListing`
struct instead of a `(VerifyingKey, String, String)` tuple. Both are `pub` on
the `riverctl` lib crate, so this is a breaking change — hence **0.2.0**, not a
patch bump (for `0.x`, Cargo treats the minor as the breaking axis, so `0.1.81`
would land inside `^0.1.80`). riverctl has zero reverse dependencies on
crates.io, so practical impact is nil.

## Testing

**215 lib tests + 7 new integration tests, all passing.**

`cli/tests/whoami_offline.rs` runs the real binary against a dead node URL —
the only way to prove the offline promise, and it catches what a source scrape
structurally cannot (routing back through `ApiClient`, dropping the `else`).
It needs no Freenet node, so it runs in normal CI rather than being `#[ignore]`d.
Covers: inline key with no storage and no node, `RIVER_SIGNING_KEY` parity with
the flag, determinism, distinct keys → distinct IDs, the single-object JSON
shape, empty-store `[]`, unknown-room error without a connection attempt, and
malformed-key rejection.

Unit tests cover the derivation (`author_member_id` sharing, pinned), the JSON
field names and null-vs-empty nickname, ownership, the override matrix, the
nickname matrix (room-state preferred, local fallback, private decrypt,
undecryptable falls back, empty falls back), and that `whoami <room>` and
`whoami` agree — they run through two independent paths.

Verified end-to-end against a live node: created a room, ran `whoami`
(`JYFX4AGJ`), sent a message, confirmed `message list --format json` reported
`"author": "JYFX4AGJ"`; and that `RIVER_SIGNING_KEY=<key> whoami` reports the
same ID that key signs with.

No WASM touched, so no delegate/contract migration is needed.

Review: Codex plus three independent Claude reviewers (skeptical bug-hunt,
code-first intent, testing/API-compatibility). All findings fixed — see the
[review comment](https://github.com/freenet/river/pull/439#issuecomment-5038299941).

Closes #438

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4724,7 +4724,7 @@ dependencies = [
 
 [[package]]
 name = "riverctl"
-version = "0.1.80"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ Commands include:
 - `riverctl message` - Send and receive messages
 - `riverctl member` - Member management
 - `riverctl invite` - Create and accept invitations
+- `riverctl identity whoami` - Your own member ID in a room (matches the `author` on your messages)
 
 Use `--format json` for machine-readable output. Run `riverctl --help` for full documentation.
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "riverctl"
-version = "0.1.80"
+version = "0.2.0"
 edition = "2021"
 authors = ["Freenet Project"]
 description = "Command-line interface for River decentralized chat on Freenet"
--- a/cli/README.md
+++ b/cli/README.md
@@ -58,21 +58,48 @@ riverctl identity whoami <room-owner-vk>         # Your member ID in one room.
 riverctl identity whoami                         # Every room you're in.
 ```
 
-The `member_id` it reports is exactly the `author` value your own messages carry
-in `message list` / `message stream --format json`, which is what a bridge needs
-to filter out its own echo:
+The `member_id` it reports is exactly the top-level `author` value your own
+messages carry in `message list` / `message stream --format json`, which is what
+a bridge needs to filter out its own echo:
 
 ```bash
 me=$(riverctl identity whoami <room-owner-vk> --format json | jq -r .member_id)
-riverctl message stream <room-owner-vk> --format json |
+riverctl message stream <room-owner-vk> --format json --no-version-check |
   jq -c --arg me "$me" 'select(.author != $me)'
 ```
 
-`whoami` reads only local storage — it works with the node stopped, and resolves
-before any message has arrived. Under `--signing-key-file` it reports the
-override identity (the one that will actually sign), and says so via
-`signing_key_source`. `room list --format json` carries the same value as
-`self_member_id`, so one call covers every room.
+(The `author` inside `reply_to` is a display nickname, not a member ID — only
+the top-level `author` is comparable.)
+
+`whoami` needs no node: it resolves from local storage, so it works with the
+peer stopped and before any message has arrived. It does need a *writable*
+config dir, since the shared storage loader may rewrite `rooms.json` when the
+bundled contract WASM has changed. Pass `--no-version-check` (or set
+`RIVERCTL_NO_VERSION_CHECK`) if you poll it, so it never makes the once-a-day
+crates.io version request.
+
+It reports the identity that will actually **sign**, across all three override
+mechanisms, using the same precedence `message send` does — inline
+`--signing-key` / `RIVER_SIGNING_KEY` beats `--signing-key-file` /
+`RIVER_SIGNING_KEY_FILE`, which beats the per-room key in `rooms.json`. The
+winner is reported as `signing_key_source` (`inline` / `override` / `stored`).
+With `--signing-key` the room need not be in local storage at all, matching
+`message send --signing-key`:
+
+```bash
+RIVER_SIGNING_KEY=<base64-key> riverctl identity whoami <room-owner-vk> --format json
+```
+
+`room list --format json` carries the same `self_member_id` (plus
+`signing_key_source`), so one call covers every room. Note that a
+`--signing-key-file` override applies to *every* room, including ones that
+identity is not a member of.
+
+**On comparing IDs.** A `MemberId` renders as 8 base32 characters — a truncated,
+non-cryptographic hash of the verifying key. That is fine for recognising your
+own messages, which is what it is for. Do not treat it as a security boundary:
+it is short enough to collide deliberately, so don't grant trust based on an
+`author` match alone.
 
 Export your identity to move between machines or back it up:
 
@@ -115,8 +142,9 @@ Run `riverctl <group> --help` or `riverctl <group> <cmd> --help` for full flags.
 
 `reply_to` is `null` for non-replies, otherwise `{ "author", "preview" }`. `edit`/`delete` events are emitted only for messages the stream actually surfaced. Bridges should key off `type` and tolerate unknown future types.
 
-`author` is your correspondent's member ID; get your own with `riverctl identity
-whoami <room-owner-vk>` and compare against it to recognise your own messages.
+The top-level `author` is the sending member's ID; get your own with `riverctl
+identity whoami <room-owner-vk>` and compare against it to recognise your own
+messages. (`reply_to.author` is a display nickname, not an ID.)
 
 ## Configuration
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -50,7 +50,31 @@ riverctl invite accept <invite-code>             # On the recipient's machine.
 
 ## Managing your identity
 
-Each room uses a separate signing key. Export it to move between machines or back it up:
+Each room uses a separate signing key, so there is no single global member ID —
+`whoami` is per-room:
+
+```bash
+riverctl identity whoami <room-owner-vk>         # Your member ID in one room.
+riverctl identity whoami                         # Every room you're in.
+```
+
+The `member_id` it reports is exactly the `author` value your own messages carry
+in `message list` / `message stream --format json`, which is what a bridge needs
+to filter out its own echo:
+
+```bash
+me=$(riverctl identity whoami <room-owner-vk> --format json | jq -r .member_id)
+riverctl message stream <room-owner-vk> --format json |
+  jq -c --arg me "$me" 'select(.author != $me)'
+```
+
+`whoami` reads only local storage — it works with the node stopped, and resolves
+before any message has arrived. Under `--signing-key-file` it reports the
+override identity (the one that will actually sign), and says so via
+`signing_key_source`. `room list --format json` carries the same value as
+`self_member_id`, so one call covers every room.
+
+Export your identity to move between machines or back it up:
 
 ```bash
 riverctl identity export <room-owner-vk> > my-identity.token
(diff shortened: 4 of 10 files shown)


sanity:
## Review round 1 — findings and resolution

Reviewed by Codex (external) plus three independent Claude reviewers with distinct lenses (skeptical bug-hunt, code-first intent, testing/API-compatibility). Every finding is fixed below; none were dismissed.

### HIGH — `RIVER_SIGNING_KEY` was a third override that whoami ignored

Found independently by two reviewers, and the most important one: riverctl has **three** identity override mechanisms, and whoami honoured only two.

| mechanism | before | now |
|---|---|---|
| `rooms.json` `signing_key_bytes` | ✅ | ✅ |
| `--signing-key-file` / `RIVER_SIGNING_KEY_FILE` | ✅ | ✅ |
| `message send --signing-key` / **`RIVER_SIGNING_KEY`** | ❌ | ✅ |

`RIVER_SIGNING_KEY` is a clap `env` var, so it gets exported in a shell profile or systemd unit — exactly what an automation bridge does. With `RIVER_SIGNING_KEY=<B>` set and `rooms.json` holding `<A>`, `message send` routes to `send_message_with_key` and stamps `author: B`, while whoami reported `A` with `signing_key_source: "stored"`. The bridge's echo filter then matches nothing and it re-ingests everything it sent — silently reproducing the exact bug #438 exists to fix. It also falsified this PR's own README claim that whoami reports "the one that will actually sign".

`whoami` now takes `--signing-key` with the same `env = "RIVER_SIGNING_KEY"` and the same precedence `message send` uses (inline > file override > `rooms.json`), disclosed as `signing_key_source` (`inline`/`override`/`stored`). Since `send_message_with_key` is explicitly documented as working *without* local storage, whoami with an inline key no longer requires the room to be in `rooms.json` either — otherwise that bot workflow was unserved.

Verified end-to-end against a live node: `RIVER_SIGNING_KEY=<owner key> riverctl identity whoami <room>` reports `JYFX4AGJ`, matching the `author` on messages that key signs.

### HIGH — the author-derivation pin test did not pin what it claimed

Two reviewers caught this. The test scraped for `author: MemberId::from(&signing_key.verifying_key())` and its docstring claimed to cover "every send path". It did not: that literal occurs only at the five **action** sites (edit/delete/react/unreact/reply). The two paths that send actual chat messages were invisible to it — `send_message` spelled it fully-qualified, and `send_message_with_key` derived it via `(&sender_vk).into()`. So the exact regression the test existed to catch could land on the primary path with CI green.

Rather than enumerate spellings, `api::author_member_id` is now the single source of truth for `MessageV1::author`, called by all eight production send sites **and** by `self_identity_from`. whoami's promise now holds by construction rather than by two derivations staying in step. This is the fix shape the repo's own bug-prevention rules prescribe for mirrored-derivation drift.

### HIGH — `whoami` had no direct test

Every test targeted the storage layer beneath it or scraped `main.rs`. Nothing executed the command, so the documented `jq -r .member_id` contract, the single-object-vs-array JSON shape, `signing_key_source`, and the offline claim were all unpinned.

Added `cli/tests/whoami_offline.rs` — 7 tests running the real binary against a dead node URL. These catch what a source scrape structurally cannot (routing back through `ApiClient`, or dropping the `else` so the short-circuit falls through), and need no Freenet node, so they run in normal CI rather than being `#[ignore]`d. Also extracted the JSON payload into a pure `whoami_json` and unit-tested the field names.

The pin tests' fragility was not hypothetical: they **broke on this very refactor** when an argument was added. Both are re-anchored on short code anchors, never on comments.

### Codex — the nickname fallback could report a different identity's name

`self_nickname` belongs to the *persisted* identity. Under `--signing-key-file` selecting a different member, with no cached `member_info` for the override, whoami printed the override's member ID next to the persisted account's name. Now the fallback applies only when the effective key **is** the persisted key — same reasoning as the CLI's `member_info` heal, which is likewise skipped under a mismatched override.

### Also fixed

- **Empty cached nickname** shadowed the fallback and printed blank instead of `(none)`.
- **`list_rooms` collected room secrets twice** per private room (and double-logged every decrypt failure); now collected once and shared.
- **`room list --format json` gained `signing_key_source`** — a file override replaces the reported identity for *every* room, including ones that identity is not a member of, and nothing disclosed that.
- **Docs**: the top-level `author` vs `reply_to.author` (which is a display nickname, not an ID — a bridge comparing against it would never match); `MemberId` is 8 base32 chars of a truncated non-cryptographic hash, fine for recognising your own messages but not a security boundary; `--no-version-check` for pollers; whoami needs a *writable* config dir since the shared loader may rewrite `rooms.json`.
- **`0.1.80` → `0.2.0`**: this breaks the published lib's `list_rooms` signature, and for `0.x` Cargo treats the minor as the breaking axis, so `0.1.80` would land inside `^0.1.79`. (riverctl has zero reverse dependencies on crates.io, so the practical impact is nil — but the version should be honest.)

### Considered and not done

- **An `is_member` field** (flagged as MEDIUM: under an override, every room reports the same id, wrong for rooms that identity isn't in). Membership is only knowable from *cached* state, so a cold or stale cache would report `is_member: false` for a room you are in — a footgun for anyone gating on it. Disclosing `signing_key_source` on both surfaces addresses the actual complaint (nothing signalled the override) without inventing a field that can lie.
- **Emitting a stronger author identifier** in message JSON to go with the truncation note. That changes the streaming output schema on a hot path and belongs in its own change; documented the limitation instead.

Tests: 215 lib + 7 new integration, all passing. Rebased onto current `main` (which rewrote ~729 lines of `api.rs`); the rebase was clean and the suite is green on the rebased head.

[AI-assisted - Claude]

sanity:
## Review round 2 — Codex, on the updated code

Re-ran the external review after round 1 (the code had changed substantially, so the earlier pass was stale). Two findings, **both consequences of the round-1 fix itself** — adding `--signing-key` to `whoami` without carrying it through:

**`room list` did not honour the inline key.** With `RIVER_SIGNING_KEY` set — precisely the case round 1 existed to fix — `room list --format json` still resolved `self_member_id` from the stored/file key, so it reported a *different* ID than `whoami`, while the README advertises it as the equivalent way to get the same value. `room list` now takes the same `--signing-key` (env `RIVER_SIGNING_KEY`) with the same precedence. Room *names* still resolve via each room's stored key, since a signing identity does not necessarily hold the room secret.

**Inline identity's nickname in a private room.** `list_rooms_as` collects the room secrets with the STORED key (it needs them for the room name) and handed that map to the nickname unsealing. For an inline identity whose secret is reachable only through its own owner-signed blob, the stored key cannot open it — so the all-rooms form reported a null nickname while the single-room form resolved it. The map is now reused only when the effective key **is** the stored key.

### A note on the regression test for the second one

The first version of that test **passed without the fix**, i.e. it was worthless. It supplied the secret via `invitation_secrets`, which the stored key can read, so the nickname resolved either way. It now places the secret only in an owner-signed blob addressed to the inline member — the real distribution path — and is verified to FAIL without the fix and pass with it.

Flagging it because it is the same failure mode as the round-1 pin test that scraped a literal matching only five of seven send paths: a test that looks like it covers something, and doesn't.

### Round 3

Re-ran Codex on the result: **no actionable correctness issues.** Verified live that `room list` and `identity whoami` now report the same `member_id` and `signing_key_source` under `RIVER_SIGNING_KEY`.

216 lib + 7 integration tests passing; clippy clean on all changed files.

[AI-assisted - Claude]

Issue #442, opened 2026-07-21
riverctl: drive get_migration_state / ensure_room_migrated with the freenet-migrate decision driver


sanity:
#437 moved `fetch_room_state_with_recovery` (the read/recovery path) onto `freenet-migrate`'s decision driver, but the ACTIVE-migration path — `ensure_room_migrated` → `check_and_migrate` → `get_migration_state` (the stored-`previous_contract_key` fast path + deep legacy sweep + CRDT merge with local, ~cli/src/api.rs:2271+) — still runs the hand-rolled loop.

Swap it onto the same `RiverCliProbeOps`/driver machinery (it differs from the read path: it DOES merge with local via `merge_room_states`, so `merge_with_local` is the real merge there, not a pass-through; candidates = stored previous key first, then the legacy registry — expressible as a NewestFirst list with the stored key prepended). Same review bar as #437: behavior parity old-vs-new, round-trip fidelity through the shared `encode_probe_state`, pins for the merge-with-local semantics.

Part of #398 / freenet-core#2776 (Gap 1).

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-07-24T12:41:11.824926633+00:00
Issue hash: `c7bbc690`

## Value Assessment

This is clearly already implemented. `get_migration_state` at cli/src/api.rs:2711-2773 now constructs a `ProbeDriver` with `RiverCliMigrationProbeOps` (wrapping `RiverCliProbeOps`), builds candidates via `migration_candidate_ids` (stored `previous_contract_key` prepended, then legacy registry, expressed as `NewestFirst`), and drives it with `NewestFirstWins` selection — exactly the swap the issue asks for. The comments explicitly reference "freenet/river#398 phase 2b, #437" and even cite "#442 review finding" inline (line 2721), and there's a dedicated `RiverCliMigrationProbeOps` whose `merge_with_local` performs the real CRDT merge via `merge_room_states`, distinguishing it from the read-path's pass-through — precisely the divergence the issue calls out. There's also an extensive block of pinning tests (source-grep pins for `get_migration_state`, wiring checks, round-trip/lossless decode tests, an end-to-end `ProbeDriver` recovery test) starting around line 6441.

[...]

PR #444, opened 2026-07-21
refactor(cli): drive active room migration with the decision driver (#442)

sanity:
## Problem

#437 moved riverctl's read/recovery path (`fetch_room_state_with_recovery`) onto freenet-migrate 0.3.0's sans-IO `ProbeDriver`, but the ACTIVE-migration path — `ensure_room_migrated` → `get_migration_state` (stored-`previous_contract_key` fast path + deep legacy sweep + CRDT merge with local) — still ran a hand-rolled newest-first loop. That is the same decision logic (newest-first selection, first-real-generation-wins, exhaustion handling) the driver owns and tests exhaustively. This completes what #437 did, for the active path.

## Approach

Behavior-preserving. The search in `get_migration_state` now runs on the same `ProbeDriver`, with the ACTIVE path's two differences from the read path supplied by a sibling ops struct + resolver:

- **`RiverCliMigrationProbeOps`** delegates `decode` / `is_real` to an inner `RiverCliProbeOps` (one classifier of record for both paths) and overrides ONLY `merge_with_local` to be the **REAL CRDT merge** — it folds the recovered network state with the device-local cache (`merged.merge(&net, &params, local)`, keep-network-on-`Err`), exactly the pre-driver semantics. (The read path's `merge_with_local` is a pass-through; this is the real merge — the key distinction, pinned by an asymmetric-state test.)
- **`migration_candidate_ids`** builds the candidate list newest-first: the stored `previous_contract_key` FIRST (the fast path), then the legacy registry (`legacy_contract_keys_for_owner`, already newest-first), with the stored key de-duplicated out of the registry tail so it is probed once. Under `NewestFirstWins` the stored key short-circuits the deep scan on a hit, preserving the fast-path-then-deep-path order.
- **`resolve_migration_state_outcome`** (pure) maps the outcome: `Recovered` ⇒ the merged state; `SeedLocal` / `NoLegacy` ⇒ the local cache migrated forward — preserving the pre-driver `Ok(local_state.clone())` fallback (the active path FALLS BACK to local, it does NOT error like the read path).

Kept exact: `LEGACY_PROBE_TIMEOUT` per probe; `ensure_room_migrated`'s surrounding logic (the swap is confined to `get_migration_state`); the plain-GET probe with **no upgrade-chain walk** (`probe_migration_bytes` uses `try_get_state`, matching the pre-driver active path — unlike the read path's `probe_legacy_bytes`, which walks the chain). The hop cap is sized to the candidate set (`with_max_hops(count.max(DEFAULT_MAX_PROBE_HOPS))`) so the default 64-hop cap can never truncate a >64-generation registry (the pre-driver deep sweep was unbounded), mirroring #437. `prepare_forward` stays identity — `migrate_room_to_new_contract` strips the upgrade pointer at PUT time (#427). The re-encode round-trip goes through the shared `encode_probe_state`, byte-faithful for the reason #437 pinned.

## Testing

`cargo test -p riverctl` — pass (223 lib incl. 7 new). New tests:
- `RiverCliMigrationProbeOps` classification via the shared classifier
- `merge_with_local` is the REAL merge — an asymmetric-state test (disjoint owner-authored messages, the merge must contain BOTH), the distinguishing active-path behavior
- `migration_candidate_ids` stored-key-first + dedupe (new key / dup key / none / unparseable)
- a full `ProbeDriver` run adopting the newest generation (anti-rollback) AND folding local in
- an exhaustion run falling back to the local cache (never an error)
- the pure `resolve_migration_state_outcome` mapping
- a source-grep pin that `get_migration_state` routes through the driver with the merge ops, the sized hop cap, the candidate helper, and the pure resolver

`cargo fmt` clean; clippy clean for `cli/src/api.rs`; no `.wasm` changes (single file, staged by name).

Refs #398 (phase 2b, active path). Read-path counterpart: #437.

[AI-assisted - Claude]

---

## Corrections from review

**Candidate dedup (was latitude choice 3) — P1, fixed in `e4ad1e4`.** The two reviewers split on this: the skeptic lens called the dedup "strictly better", while Codex flagged it a P1. Codex is right, on the transient-failure analysis. The #442 issue text suggested de-duplicating the stored key out of the registry, but the **retry semantics override that**: the pre-driver deep sweep re-GET the stored key after a fast-path miss, and that re-GET was a genuine independent retry — a "miss" covers a transient send failure / 8s `LEGACY_PROBE_TIMEOUT`, not only a true absence, so the retry could HIT where the first attempt transiently failed. De-duping dropped that retry, so a single transient timeout on the newest generation would advance to an OLDER one, migrate its stale state forward, and clear `previous_contract_key` — stranding newer network-only state.

Fixed by **keeping the duplicate**: candidates = `[stored, ...registry-newest-first]` verbatim, no dedup. `NewestFirstWins` never issues the second GET after a fast-path HIT, so the duplicate is free on the happy path and only ever costs a retry after a miss — exactly the pre-driver semantics. The candidate test now pins the DUPLICATE (stored key present twice when it is also a registry generation), plus a new driver test where a transient timeout on the newest generation is retried and recovers the newest state rather than advancing to an older one. (224 lib tests.)

**Fast-path-wins semantics (skeptic LOW) — documented, no code change, `211a0d0d`.** Because the stored key is probed FIRST under `NewestFirstWins`, a stored key that carries state is adopted even over a NEWER registry generation that also has state. In the normal case this is moot (the stored `previous_contract_key` IS the immediately-previous / newest legacy generation), but the behavior is intentional — it preserves the pre-driver fast-path-wins semantics (whose fast path returned the stored-key GET before the deep sweep ran). Documented on `migration_candidate_ids`.

The skeptic lens otherwise verified the swap preserves behavior exactly: merge argument order / params / `Err`-side, exhaustion-never-errors, no chain-walk with the same 8s timeout, forward path untouched, and a lossless round-trip.

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -349,34 +349,41 @@ fn resolve_legacy_recovery_outcome(
 /// The active-migration probe candidates, newest-first: the stored
 /// `previous_contract_key` (the immediately-previous generation) FIRST, then
 /// every known legacy generation newest-first
-/// (`legacy_contract_keys_for_owner` reverses the oldest-first registry), with
-/// the stored key de-duplicated out of the registry tail so it is probed once.
-/// This preserves the pre-driver fast-path-then-deep-path order; under
+/// (`legacy_contract_keys_for_owner` reverses the oldest-first registry). This
+/// preserves the pre-driver fast-path-then-deep-path order; under
 /// `NewestFirstWins` the stored key short-circuits the deep scan on a hit.
-/// Pure (no I/O) so the ordering + dedup are unit-testable.
+/// Pure (no I/O) so the ordering is unit-testable.
+///
+/// The stored key is deliberately NOT de-duplicated against the registry: when
+/// it equals a registry generation it appears TWICE, and that duplicate is
+/// load-bearing. A probe "miss" is not only an absent state — it also covers a
+/// transient send failure or an 8s `LEGACY_PROBE_TIMEOUT`. The pre-driver code
+/// re-GET the stored key in the deep sweep after a fast-path miss, giving it a
+/// second attempt; de-duping would drop that retry, so a single transient
+/// timeout on the newest generation would advance to an OLDER one, migrate its
+/// stale state forward, and clear `previous_contract_key` — stranding newer
+/// network-only state on the skipped generation. `NewestFirstWins` never issues
+/// the second GET after a fast-path HIT (the probe ends there), so the duplicate
+/// only ever costs a retry after a miss — exactly the pre-driver semantics
+/// (freenet/river#442 review finding).
 fn migration_candidate_ids(
     previous_contract_key_str: &Option<String>,
     room_owner_key: &VerifyingKey,
 ) -> Vec<ContractInstanceId> {
     let mut ids = Vec::new();
-    let mut seen: HashSet<ContractInstanceId> = HashSet::new();
     // Fast path: the immediately-previous contract key recorded in storage.
     if let Some(prev_key_str) = previous_contract_key_str {
         match prev_key_str.parse::<ContractInstanceId>() {
-            Ok(prev_id) => {
-                if seen.insert(prev_id) {
-                    ids.push(prev_id);
-                }
-            }
+            Ok(prev_id) => ids.push(prev_id),
             Err(e) => warn!("Stored previous_contract_key is not a valid contract id: {e}"),
         }
     }
-    // Deep path: every known previous contract generation, newest-first.
+    // Deep path: every known previous contract generation, newest-first. The
+    // stored key is re-listed here when it is a registry generation (NOT
+    // de-duped) so a transient miss on the fast-path attempt is retried before
+    // the probe advances to an older generation (see the doc comment).
     for legacy_key in river_core::migration::legacy_contract_keys_for_owner(room_owner_key) {
-        let id = *legacy_key.id();
-        if seen.insert(id) {
-            ids.push(id);
-        }
+        ids.push(*legacy_key.id());
     }
     ids
 }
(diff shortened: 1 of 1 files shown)


sanity:
## Consolidated review

Reviewers: **skeptical** and **behavioral-parity** Claude lenses plus a **Codex** external pass (available again).

### The one real finding — an instructive split verdict
Codex flagged the candidate dedup as P1: the old deep loop's re-GET of the stored `previous_contract_key` was a genuine independent **retry** — a transient 8s timeout on the fast path used to get a second chance before the probe advanced to an older generation; the dedup removed it, so one network flake on a **one-shot** migration path (which clears `previous_contract_key` afterward) could migrate stale state. Both Claude lenses initially judged the dedup fine ('deterministically missed the 2nd time' / 'not behavior worth preserving') — adjudicated for Codex: the dissents conceded the transient-case divergence exists and only discounted its likelihood, and on this network transient GET timeouts are a documented common failure mode. Fixed in `e4ad1e4`: the duplicate candidate is kept verbatim (NewestFirstWins never issues the second GET after a hit, so it costs nothing on the happy path), pinned by a retry-recovers-newest driver test and a duplicate-presence test.

### Verified preserved (all three reviewers, line-level)
Merge argument order/primacy/params/Err-fallback byte-equal to the old `get_migration_state`; stored-key-first fast-path semantics (including the documented intentional case where an older stored generation with state wins — pre-driver parity); exhaustion → local-cache clone, never an error (type-impossible now, and pinned); no chain-walk + the same 8s per-candidate timeout; `migrate_room_to_new_contract` + pointer-strip untouched and fed identically; the #437 read path completely untouched; hop cap sized to never truncate; round-trip through the shared `encode_probe_state` lossless.

CI green on the final head (`211a0d0d` — docs-only over the re-verified fix hunk). Merging.

[AI-assisted - Claude]

PR #445, opened 2026-07-22
feat(ui): unread-messages badge on the mobile hamburger button

sanity:
## Problem

On mobile, a room fills the whole screen. New messages arriving in **other** rooms were invisible — there was no indicator anywhere until the user happened to open the room list. Feature request from The Torist (2026-07-22): "It needs some kind of indicator on the top left hamburger button when there are new messages in other rooms available."

## Approach

Both mobile hamburger buttons (the room-header one and the welcome-screen one) now carry a numeric badge counting unread messages in rooms **other than the current one**, plus unread DMs — the DM rail lives behind the same button, so anything new in that panel is included.

- New pure helper `count_unread_excluding_room` + signal-reading wrapper `count_unread_behind_rooms_panel` in `document_title.rs`, reusing the existing `count_unread_in_room_data` / `count_unread_dms` machinery that already drives the room-list and DM-rail badges. All three surfaces stay consistent by construction.
- The current room is excluded for the same reason the room-list badge hides it (`!is_current`): its messages are on screen and marked read on open, so counting them would only make the badge flicker.
- A `use_memo` in `Conversation` follows the existing room-list memo pattern (`try_read()` per the Dioxus signal-safety rules); it re-runs when `ROOMS` mutates (message arrival, read-marker advance) or `CURRENT_ROOM` changes.

| In a room | Welcome screen |
|---|---|
| badge on header hamburger | badge on welcome hamburger |

## Testing

- Unit test `excluding_the_current_room_omits_its_unread` covers the exclusion logic (current room omitted, `None` counts all, unknown key is a no-op).
- New Playwright spec `hamburger-unread-badge.spec.ts`: badge visible with total on the welcome screen; opening a room drops the total (current room excluded) but keeps it non-zero; visiting every room clears the badge entirely. Passes on chromium + mobile-chrome.
- Full existing Playwright suite passes on chromium (72 passed, 6 pre-existing skips), including the responsive-layout / mobile-touch-ux suites that locate the hamburger by CSS class.
- `cargo test -p river-ui --bins`: 486 passed.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/document_title.rs
+++ b/ui/src/components/app/document_title.rs
@@ -238,6 +238,37 @@ fn count_unread_dms(rooms: &crate::room_data::Rooms) -> usize {
     total
 }
 
+/// Sum unread messages across every room in `map` except `exclude`.
+///
+/// Pure (no signal reads) so the exclusion logic is unit-testable; the
+/// signal-reading wrapper is [`count_unread_behind_rooms_panel`].
+pub fn count_unread_excluding_room(
+    map: &std::collections::HashMap<ed25519_dalek::VerifyingKey, crate::room_data::RoomData>,
+    exclude: Option<&ed25519_dalek::VerifyingKey>,
+) -> usize {
+    map.iter()
+        .filter(|(owner_key, _)| Some(*owner_key) != exclude)
+        .map(|(_, room_data)| count_unread_in_room_data(room_data))
+        .sum()
+}
+
+/// Count unread messages waiting *behind* the mobile rooms panel: rooms
+/// OTHER than the currently-open one, plus inbound direct messages (the
+/// DM rail lives in that same panel).
+///
+/// Drives the badge on the mobile hamburger buttons in the conversation
+/// header. The current room is excluded because its messages are on
+/// screen and marked read on open — counting them would only make the
+/// badge flicker (mirrors the `!is_current` guard on the room-list
+/// badge in `room_list.rs`).
+pub fn count_unread_behind_rooms_panel() -> usize {
+    let Ok(rooms) = ROOMS.try_read() else {
+        return 0;
+    };
+    let current = CURRENT_ROOM.read().owner_key;
+    count_unread_excluding_room(&rooms.map, current.as_ref()) + count_unread_dms(&rooms)
+}
+
 /// Update the document title based on current state
 pub fn update_document_title() {
     let is_visible = *DOCUMENT_VISIBLE.read();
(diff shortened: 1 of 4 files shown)


sanity:
## Multi-model review record

Reviewers: **Codex CLI** (external, `codex review --base origin/main`) + **adversarial Claude subagent** (independent, blind to Codex). Both reviewed the checked-out code, not just the diff. Dispositions, all addressed in c2d763ae:

| Finding | Source | Severity | Disposition |
|---|---|---|---|
| `count_unread_behind_rooms_panel` did `ROOMS.try_read()` (early-return) **before** `CURRENT_ROOM.read()`; an Err poll registers zero subscriptions → memo permanently frozen, badge silently stuck | adversarial | blocking | **Fixed**: infallible `CURRENT_ROOM.read()` hoisted first (backup-subscription pattern from `current_room_label`). Same pre-existing inverted order in `room_list.rs` `room_items` memo also fixed. |
| Make the `CURRENT_ROOM` read fallible (`try_read`) | Codex P1 | P1 | **Dismissed as stated, root concern resolved**: the repo's signal-safety rules deliberately bless one infallible read in multi-signal memos — it IS the backup subscription preventing permanent memo death, and every `CURRENT_ROOM` write site is `defer()`-wrapped (verified all 8), so the read can't be polled mid-write. The real hazard was the read *order*, fixed per the adversarial finding above. |
| DM count included hidden/archived threads the panel's DM rail filters out → phantom badge with no way to clear it | adversarial S1 | should-fix | **Fixed**: `count_unread_dms` now mirrors the rail's per-thread accumulation and skips threads per `is_thread_hidden_for` (strict-<= revival, both-direction revival clock), via a pure testable core. Also corrects the same pre-existing lie in the document-title count. |
| `mark_all_rooms_as_read` on tab-hide wipes unread on mobile app-switch, limiting the badge to one foreground session | adversarial S2 | should-fix | **Deferred with decision**: pre-existing policy shared by title + rail badges; changing it alters long-standing title semantics. Filed #446 for the product call (cc @sanity). |
| Button `aria-label` overrides descendant text → screen readers never hear the count; `aria-label` on a generic span is ignored | Codex P2 + adversarial N1 | P2/nit | **Fixed**: count folded into each button's accessible name ("Open room list, N unread"); badge span is `aria-hidden`. Copy no longer claims "other rooms" (DMs included; welcome screen counts all). |
| No 99+ cap on the count | adversarial N2 | nit | **Not done, deliberate**: room-list and DM-rail badges are also uncapped; 3 digits still fit the button. Capping all three consistently is a separate cosmetic change. |
| DM contribution untested | adversarial N3 | nit | **Fixed**: three new unit tests on the pure DM core (inbound/last-seen, hidden-skip + revival, third-party exclusion). |

Both reviewers explicitly cleared: rsx memo reads (codebase idiom), double-counting (rooms/DMs disjoint; open DM thread marks read on render), duplicate testid (mutually exclusive match arms), Playwright raciness (deferred writes handled by retrying assertions), perf (O(same as existing room_items memo)).

Verification pass on the fix commit is running (fresh Codex + adversarial re-read); will confirm before merge alongside CI.

[AI-assisted - Claude]

sanity:
Verification pass complete on the fix commit:

- **Codex re-review** (full branch incl. fixes): clean — "no blocking correctness issue was found."
- **Adversarial re-read**: all four dispositions VERIFIED (subscription reorder, hidden-thread parity with the DM rail incl. strict-<= boundary semantics, aria naming, #446 deferral coherence); no new blocking/should-fix issues. Its one residual test gap (outbound-message-revives-hidden-thread) is pinned by a new unit test in d7f1526b.

Merging once CI is green on d7f1526b.

[AI-assisted - Claude]

PR #448, opened 2026-07-22
feat(ui): expand randomized name suggestions

sanity:
## Problem

Users reported that the nickname River suggests when you join a room feels "samey" — different people keep getting handles built from the same handful of words.

The generator (`ui/src/nickname.rs`) derives a two-word handle deterministically from the member's verifying key, one word from `FIRST_NAMES` and one from `LAST_NAMES`. Both pools held 30 words, so the whole space was 30 x 30 = **900 combinations**. In a room with a few dozen members, repeated words (and occasionally repeated full handles) are common at that size.

## Approach

Grow both pools to 100 words, giving **10,000 combinations** (11x more). All 60 original words are kept and the existing 90s-hacker-culture tone is matched — the new words are the same kind of thing (`Halcyon`, `Obsidian`, `Bitwise`, `Peregrine`, `Crucible`, `Almanac`). Sample output:

```
Lumen Vector      Chrome Nightjar    Xenon Nightjar     Neural Corsair
Eclipse Glider    Titanium Gateway   Chrome Storm       Crash Peregrine
Scarlet Peregrine Phantom Gateway    Bitwise Override   Hollow Osprey
Nitro Lattice     Phantom Bellows    Digital Cobra
```

Every word is ASCII, capitalized, 3-9 characters and inoffensive, so the longest possible handle is 19 bytes — comfortably inside the 50-byte default `max_nickname_size` (a generated default that exceeded a room's limit would make the join / self-heal UPDATE fail outright, so this is checked exhaustively in a test against the real `Configuration::default()`).

Two smaller things came along with it:

- The two pools are now **disjoint**, so no handle can read "Echo Echo". A test pins this.
- `fold` accumulates in `u64` instead of `usize`. `usize` is 32-bit on the wasm32 build that actually ships and 64-bit natively, so `accumulator % pool_len` gave *different* words in the browser than in the native unit tests. `u64` makes both agree, so the tests measure what users actually get.

**Note on determinism:** the derivation stays a pure function of the key (the member-info self-heal path depends on that and is unaffected). But because the pool lengths change, a given key now maps to a different handle than it did before. That is cosmetic only — a stored `member_info` nickname always wins; the generated handle is a fallback used when no entry exists. The visible effect is that a member who never set a nickname and has no published `member_info` may show a different default handle after upgrading.

Scope is client-only: this touches `ui/src/nickname.rs` and nothing under `common/`, `contracts/` or `delegates/`, so no WASM bytes and no contract/delegate key change.

## Testing

`cargo test -p river-ui --bins` — 496 passed, 0 failed (this is the same command CI runs for the UI crate). New / updated tests in `ui/src/nickname.rs`:

- `pools_are_100` — both pools are exactly 100 entries, 10,000 combinations (replaces `pools_are_30`)
- `pools_have_no_duplicates` — no repeated word within either pool
- `pools_are_disjoint` — no word in both pools, so no doubled handle
- `words_are_short_ascii_and_capitalized` — ASCII-alphabetic, capitalized, 3-9 chars
- `handles_are_within_max_handle_len` — exhaustive over all 10,000 handles vs `MAX_HANDLE_LEN`
- `handles_fit_the_default_nickname_limit` — exhaustive over all 10,000 handles vs the real `Configuration::default().max_nickname_size`, rather than a hardcoded copy of it
- `fold_reaches_most_of_both_pools` — 2,000 keys must reach at least 80 of the 100 words in each pool (measured: 100/100 for both, so the floor has plenty of slack)
- `distinct_keys_mostly_distinct_handles` — tightened from >30/50 unique to >180/200 unique now that the space is 10,000

Also ran `cargo fmt --all` and `cargo clippy -p river-ui --bins` (clean). Deliberately did not run `cargo make build` — it rebuilds the WASMs.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/nickname.rs
+++ b/ui/src/nickname.rs
@@ -116,7 +116,7 @@ pub const LAST_NAMES: [&str; 100] = [
     "Mantis",
     "Panther",
     "Jaguar",
-    "Cougar",
+    "Ocelot",
     "Wolfhound",
     "Serpent",
     "Foundry",
@@ -179,7 +179,7 @@ pub fn generate_default_nickname(vk: &VerifyingKey) -> String {
 mod tests {
     use super::*;
     use ed25519_dalek::SigningKey;
-    use std::collections::HashSet;
+    use std::collections::{HashMap, HashSet};
 
     /// Every handle the generator can emit, for the exhaustive checks.
     fn all_handles() -> impl Iterator<Item = String> {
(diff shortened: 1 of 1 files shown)


sanity:
## Review summary

Adversarial review run (Light tier: UI-only, no WASM impact — confirmed `git diff origin/main --name-only` returns exactly `ui/src/nickname.rs`).

**No blocking findings.** Addressed in `c5f33b44`:

- **(important)** `fold_reaches_most_of_both_pools` claimed to catch "a pool size sharing factors with the hash multiplier", but a reachability floor is a total-collapse detector only. Verified: degenerate multipliers (2, 32, and 100 — one sharing factors with the pool size) all still reach 100/100 buckets over 2,000 keys, as does a synthetic distribution putting 90% of its mass on 10 words. Renamed to `fold_spreads_evenly_across_both_pools` and added a max-share bound, which is what actually catches skew.
- **(minor)** Dropped `Cougar` for `Ocelot` — the big-cat reading is clear from its cluster, but `Velvet Cougar` / `Scarlet Cougar` are reachable and read as the slang sense.

**Correction to the PR body:** the longest possible handle is **18** bytes (`Obsidian Mainframe`), not 19 — no FIRST word is 9 chars. Still far under the 50-byte default.

**Verified clean:**
- The `usize`->`u64` fix is real and the claim checks out: simulating the old `usize` accumulator at both widths over 20,000 keys, the full handle differed between wasm32 and x86_64 for **99.5%** of keys (`2^32 mod 30 = 16 != 0`). The native tests genuinely were not testing what browsers produced. No width-dependent arithmetic remains before the final modulo.
- Distribution is statistically uniform: 200,000 keys over 100 buckets gives chi-square 96.9 (first) / 103.3 (last) against 99 df, well under the p=0.001 critical value of 148.2.
- `generate_default_nickname` is only ever called **for self** — other members lacking `member_info` render "Unknown", never a generated handle — so a pool change cannot desync two clients' view of a third party. Renames are limited to members with no published `member_info`, no cached `self_member_info` and no `self_nickname`, and are cosmetic.
- All 140 new words read: no duplicates within or across pools, pools disjoint, all ASCII/capitalized/3-9 chars, and no impersonation-adjacent words (no Admin/Moderator/System/Support/Freenet/Owner/Staff/Bot).
- Sealed (private-room) nicknames are measured on **plaintext** length (`SealedBytes::declared_len()`), so the AES-GCM tag does not count toward `max_nickname_size` — the raw-length test does cover the private path.

**Filed as follow-up:** #450 — the UI heal path has no `max_nickname_size` clamp (the CLI does), so an owner setting an unusually small limit fails the whole delta. Largely pre-existing; this PR widens the exposed set by one value.

[AI-assisted - Claude]

PR #452, opened 2026-07-22
fix(ui): show the deputy 🛡 shield in the member-info modal legend (#451)

### LINKED ISSUE #451: Member Info modal legend is missing the deputy 🛡 shield
## Problem

The Member Info modal shows a row of chips explaining the icons that appear next to that member in the members list — 👑 Room Owner, ⭐ You, 🔑 Invited by You, 🎪 Invited You.

The deputy shield **🛡**, added with the member-deputies feature (#410), was never added to that legend. So a member who shows the shield next to their name in the members list has no explanation for it when you open their Member Info page.

## Expected

The modal should show a 🛡 Deputy chip under the same viewer-relevant condition the member-list row uses (a global moderator appointed by the owner, a deputy in the viewer's ancestor chain, or a deputy the viewer appointed), with the appointer name(s) in the tooltip — mirroring the row's badge.

## Notes

The row computes this in `ui/src/components/members.rs` (`deputized_by` via `relevant_deputizers` over the canonical `MemberInfo.deputies`); the modal (`ui/src/components/members/member_info_modal.rs`) hand-rolls a narrower subset of tags and omits the shield. The fix should share the row's exact computation so the two can't drift.

[AI-assisted - Claude]

sanity:
## Problem

The Member Info modal shows a row of chips explaining the icons that appear next to that member in the members list — 👑 Room Owner, ⭐ You, 🔑 Invited by You, 🎪 Invited You. The deputy shield **🛡**, added with the member-deputies feature (#410), was never added to that legend, so a member who shows the shield next to their name in the list has no explanation for it in their Member Info page.

## Approach

The row's shield is **viewer-scoped** — it shows only for deputies relevant to the viewer (a global moderator appointed by the owner, a deputy in the viewer's own ancestor chain, or a deputy the viewer appointed) — and its tooltip names the appointer(s). To make the modal match that exactly, and to keep the two from drifting again, the row's computation is extracted into three shared `pub(crate)` helpers in `members.rs`:

- `build_deputizers_of` — the reverse map from canonical `MemberInfo.deputies`.
- `viewer_relevant_deputizer_set` — the viewer's strict ancestors ∪ self.
- `relevant_deputizer_names` — the resolved appointer display names for a target, i.e. exactly what the 🛡 badge shows.

`MemberList` now calls these instead of its inline copies (behavior unchanged — the existing deputy-ordering and viewer-scoping tests still pass), and the modal renders a `🛡 Deputy` chip driven by the **same** `relevant_deputizer_names`, with an "appointed by …" tooltip.

Scope note: the modal legend is also missing `🌐 In Your Network` and `🔭 In Your Invite Chain` (relationship hints that exist in the row). Those were left alone — this PR is scoped to the reported shield gap.

## Not a WASM change

UI-only. `git diff origin/main --name-only` touches exactly `ui/src/components/members.rs` and `ui/src/components/members/member_info_modal.rs` — nothing under `common/`, `contracts/`, `delegates/`, no `.wasm`, no Cargo files. No migration ritual.

## Testing

- `member_display_parts_shows_deputy_shield_with_appointer_tooltip` — the row renders 🛡 with the appointer tooltip iff `deputized_by` is non-empty.
- `shared_deputy_helpers_name_a_global_moderator` — end-to-end pin of the shared helpers: an owner-appointed moderator shows the shield named "room owner" even to an unrelated viewer; a non-deputy shows none.
- `modal_renders_deputy_shield_via_shared_helper` — source-grep pin that the modal renders the chip via the shared helper, so it can't silently drop the shield or fork a drifting copy of the logic again.

`cargo test -p river-ui --bins`: **499 passed**. `cargo fmt` applied; clippy clean (no new warnings).

Closes #451

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -489,6 +489,121 @@ fn relevant_deputizers(
         .collect()
 }
 
+/// Reverse map: for each deputy member, the members who have deputized them,
+/// built from every member's CANONICAL signed `MemberInfo.deputies` (#410).
+///
+/// Routed through each member_id's canonical record (highest
+/// `member_info_rank`), not a raw scan of `member_info.member_info` — `verify`
+/// accepts duplicate member_info records per member_id (migration safety), and
+/// unioning deputies across ALL of a member's duplicate records (rather than
+/// reading only the converged/canonical one) can keep a revoked deputy grant
+/// showing even after the revoke has won (freenet/river#411 round 8).
+pub(crate) fn build_deputizers_of(
+    member_info: &river_core::room_state::member_info::MemberInfoV1,
+) -> HashMap<MemberId, Vec<MemberId>> {
+    let mut deputizers_of: HashMap<MemberId, Vec<MemberId>> = HashMap::new();
+    let member_ids_with_info: HashSet<MemberId> = member_info
+        .member_info
+        .iter()
+        .map(|mi| mi.member_info.member_id)
+        .collect();
+    for appointer in member_ids_with_info {
+        let Some(canonical) = member_info.canonical(appointer) else {
+            continue;
+        };
+        for deputy in &canonical.member_info.deputies {
+            deputizers_of.entry(*deputy).or_default().push(appointer);
+        }
+    }
+    deputizers_of
+}
+
+/// The viewer-relevant set for the 🛡 deputy badge (#410): the viewer's STRICT
+/// ancestors (members whose deputy could ban the viewer) unioned with the
+/// viewer themselves (deputies they appointed). Strict ancestors are EMPTY for
+/// the owner, so the owner sees the shield only for their own appointees; the
+/// owner is always included so a global moderator (a deputy of the owner) is
+/// relevant to every viewer.
+pub(crate) fn viewer_relevant_deputizer_set(
+    members: &MembersV1,
+    owner_id: MemberId,
+    self_member_id: MemberId,
+) -> HashSet<MemberId> {
+    // Strict ancestors of the viewer: the owner (of every non-owner) plus the
+    // walk up the invite chain. `self` is deliberately NOT a strict ancestor.
+    let mut set = HashSet::new();
+    if self_member_id != owner_id {
+        set.insert(owner_id);
+    }
+    let invited_by: HashMap<MemberId, MemberId> = members
+        .members
+        .iter()
+        .map(|m| (m.member.id(), m.member.invited_by))
+        .collect();
+    let mut guard = HashSet::new();
+    guard.insert(self_member_id);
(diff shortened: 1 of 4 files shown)


sanity:
## Review — Codex (external) pass

**Addressed in 1d0ee540.** Codex found one P3: `build_deputizers_of` iterated a `HashSet` of appointers, so a member with multiple relevant deputizers could show `room owner, you` in the member row and `you, room owner` in the modal legend (or reorder between renders) — directly undermining the identical-tooltip contract this PR relies on. The nondeterminism was pre-existing in the row but became visible once the modal computes the same list independently. Fixed by iterating appointers sorted by MemberId, pinned by `build_deputizers_of_orders_appointers_deterministically`.

Codex confirmed the deputy badge logic is otherwise shared correctly between the two call sites.

An independent adversarial Claude review focused on refactor behavior-equivalence is also running; I'll fold in anything it finds.

[AI-assisted - Claude]

sanity:
## Review — adversarial Claude pass (refactor equivalence)

**Verdict: no blocking or important-correctness bugs.** The reviewer verified the three extracted helpers produce output identical to the old inline row code:

- `viewer_relevant_deputizer_set` reproduces `self_ancestors ∪ {self}` exactly (owner-insert guard, invite-chain walk, cycle guard); empty-for-owner holds.
- `build_deputizers_of` produces the same map; `self_ancestors` had no other use (grep-confirmed), so collapsing it is safe.
- `relevant_deputizer_names`'s `name_of` matches (owner→"room owner", self→"you", else decrypted nickname, else "Unknown").
- `deputy_display_order` (member ordering) is provably unaffected — it selects by `(invite_depth, unique base_rank)`, independent of appointer-list order, so the determinism sort doesn't reorder the list.
- Modal: `owner_id`/`self`/`target` come from the same sources as the row, so the shield shows under **exactly** the same (viewer, target) condition. `title:` is a Dioxus text attribute (DOM-escaped), **not** `dangerous_inner_html` — XSS-safe with attacker-controlled nicknames (#227 class). Render-time reads only; no signal-safety concern.

**Finding addressed (6e9e54f7):** the modal shield had only a source-grep pin, so an arg-swap or inverted condition would pass every test. Closed with an end-to-end Playwright render test (`member-info-deputy-tag.spec.ts`) — example data now has the owner deputize the "(Member)" member, and the spec asserts the deputy's modal shows the 🛡 chip (with the "appointed by" tooltip) while a non-deputy's does not. **25 passed across all 5 browser projects**, and the existing member-list spec still passes.

**Filed as follow-up (#453):** the modal legend still omits 🌐 / 🔭 and its 🔑 semantics differ from the row — pre-existing and out of this shield-scoped PR.

[AI-assisted - Claude]

PR #454, opened 2026-07-23
docs(cli): document riverctl dm send

sanity:
## Problem

Reported by Ivvor on Matrix:

> I see `riverctl dm accept` to process a DM invite, but no specific DM send, unless you just send the raw code as a message I guess.

`riverctl dm send` **is not missing** — it has existed since v0.1.55 (#243, May 2026) and works. I verified it end-to-end on a live node while triaging: Alice sent, Bob's `dm list` decrypted it.

The gap is documentation:

1. The **top-level `README.md`** riverctl command list omitted `dm` entirely — it listed `room`, `message`, `member`, `invite`, `identity`.
2. **`cli/README.md`** carried DMs only as a single row in the reference table (`dm | send, list, purge, accept`). Every other command family — messages, invites, identity, members — has a worked example section with copy-pasteable commands. DMs had none.

So a reader finds `dm accept` in that table and reasonably concludes there's no send.

## Approach

- Add `riverctl dm` to the top-level README command list.
- Add a "Direct messages" section to `cli/README.md`, in the same style as the neighbouring sections, covering `send` / `list` / `purge` / `accept`.

Beyond the bare commands it documents the three things that actually trip people up:

- the recipient is a **member ID** (the short 8-char form `member list` prints is accepted);
- the **purge token** comes from the `purge token: …` line printed under each inbound DM by `dm list`;
- `dm accept` takes the **carrier** room (the room whose DM thread contains the invitation), *not* the room you're joining.

It also notes the sender-side plaintext cache: only the recipient can decrypt a DM, so your own sent messages render from a local cache and show as ciphertext-only from a different machine.

## Verification

Every command and output format documented here was checked against riverctl 0.2.1 — `--help` signatures for `send`/`list`/`purge`/`accept`, plus a real round trip on a local node (create room → invite → accept → `dm send` → `dm list`), including confirming the `purge token:` line renders as described.

## Scope

Docs only. No code, no WASM, no contract change, no republish. Fixes the discoverability problem behind the report; no behavioural change.

[AI-assisted - Claude]


sanity:
Closing — opened from the wrong branch by mistake. The `gh pr create` ran in a different worktree, so this PR points at `fix/443-action-payload-bytes` (whose commits are already merged via #447) instead of the docs branch. Reopening the intended docs change as a separate PR. No review needed here.

[AI-assisted - Claude]

PR #457, opened 2026-07-23
feat(cli): riverctl dm invite — send a clickable invitation via DM

### LINKED ISSUE #456: riverctl can accept invite DMs but can't send them (no clickable-invite parity with the UI)
Reported by Ivvor on Matrix:

> OK, so how do I send an invite code via DM that the recipient can just click on? When I send the bare invite code it just displays the raw code at the other end.

## Gap

The River UI's "Share invite via DM" (#252) sends a structured `DirectMessageBody::Invite` DM that the recipient's UI renders as a clickable **Accept card**. riverctl can *accept* such invite DMs (`dm accept`) but can't *originate* one — `dm send` composes plaintext only, so a bare invite code sent from the CLI shows up as inert text.

A full invitation URL pasted as a DM is at least clickable (it autolinks), but a bare code is not, and there's no CLI way to produce the proper card.

## Ask

Add `riverctl dm invite <carrier-room> <recipient> --room <target-room> [-m <msg>]` — the send-side counterpart of `dm accept`, building the same `DirectMessageBody::Invite` the UI does so CLI-sent invites also render as a clickable card.

[AI-assisted - Claude]

sanity:
## Problem

Reported by Ivvor (#456): sending a bare invite code via `dm send` just shows raw text at the recipient's end. The River UI's "Share invite via DM" (#252) sends a structured `DirectMessageBody::Invite` DM that renders as a clickable **Accept card**, but riverctl could only *accept* such invite DMs (`dm accept`), not *originate* one — `dm send` composes plaintext only.

## Approach

Adds `riverctl dm invite <carrier-room> <recipient> --room <target-room> [-m <msg>]`, the send-side counterpart of `dm accept` and the CLI analog of the UI picker:

- `carrier-room` — the room the DM travels in (you and the recipient are both members).
- `--room target-room` — the room you're inviting them to (any room you're a member of).
- `-m` — optional note shown above the Accept button.

Key correctness point — **byte-identical to what the UI produces**, so a UI recipient gets the real card:

- `build_invitation` is extracted from `create_invitation` (which now calls it, then base58-encodes), so the CBOR `invitation_payload` a `dm invite` embeds is exactly what `invite create` and the UI produce.
- The command wraps it in the same `InvitePayload` / `DirectMessageBody::Invite` and `encode_body`s it — the wire body `dm accept` and the UI Accept card already decode.

`dm send` and `dm invite` now share one delivery core, `deliver_dm`, so the anti-silent-drop guards (#269) and the pruned-sender rejoin bundling (#256 / Ivvor Bug #1) stay byte-identical for both. The only per-kind differences are the wire body (raw UTF-8 for legacy `Text` vs magic-byte+CBOR for `Invite`), the outbound-cache label, and the success noun.

## Scope

**CLI-only.** `git diff origin/main --name-only`: `cli/src/api.rs`, `cli/src/commands/dm.rs`, `cli/README.md`, `README.md`, `cli/Cargo.toml`, `Cargo.lock`. No `common/`, contract, delegate, or `.wasm` change — it only *uses* existing `river-core` DM-body types, so no migration. `river-core` stays 0.1.17; `riverctl` bumped 0.2.1 → 0.2.2 for republish.

## Testing

- `invite_dm_body_round_trips_through_accept_path` — the body `dm invite` builds decodes back through the exact `decode_body` + `decode_invitation_from_payload` path `dm accept` and the UI card use, into the same invitation; `dm list` labels it `[Invitation to room …]`.
- `normalize_personal_message_blanks_to_none` — a blank `--message` becomes `None` so the UI hides the message box.
- **Live end-to-end** (local node, release binary): Alice `dm invite`d Bob to a second room via a carrier room → Bob's `dm list` showed `[Invitation to room 3ena64R7…] come join my other room!` → Bob `dm accept`ed → Bob is now a member of the target room. Alice's own `dm list` shows the same invite label from the outbound cache.

`cargo test -p riverctl` green (226+ lib + integration); `cargo clippy -p riverctl` clean; `cargo fmt` applied.

## Docs

`cli/README.md` gains an "Inviting someone via DM" subsection; the top-level README's `riverctl dm` line now lists `invite`.

Closes #456

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4724,7 +4724,7 @@ dependencies = [
 
 [[package]]
 name = "riverctl"
-version = "0.2.1"
+version = "0.2.2"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ cargo install riverctl
 Commands include:
 - `riverctl room` - Room management (create, list, info)
 - `riverctl message` - Send and receive messages
-- `riverctl dm` - End-to-end-encrypted direct messages to a co-member (send, list, purge, accept)
+- `riverctl dm` - End-to-end-encrypted direct messages to a co-member (send, invite, list, purge, accept)
 - `riverctl member` - Member management
 - `riverctl invite` - Create and accept invitations
 - `riverctl identity whoami` - Your own member ID in a room (matches the `author` on your messages)
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "riverctl"
-version = "0.2.1"
+version = "0.2.2"
 edition = "2021"
 authors = ["Freenet Project"]
 description = "Command-line interface for River decentralized chat on Freenet"
--- a/cli/README.md
+++ b/cli/README.md
@@ -64,8 +64,29 @@ Only the recipient can decrypt a DM, so `dm list` renders your own sent
 messages from a local plaintext cache. A DM you sent from a different machine
 shows as ciphertext-only there.
 
-The UI can also share an invitation *inside* a DM, which `dm list` shows as
-`[Invitation to room …]`. Accept it with:
+### Inviting someone via DM
+
+You can hand a room invitation to a co-member *as a DM*. The recipient's River
+UI renders it as a clickable **Invitation card** with an Accept button — unlike a
+bare invite code pasted into `dm send`, which just shows up as raw text.
+
+```bash
+# Invite <recipient> (a member of the carrier room) to <target-room>.
+riverctl dm invite <carrier-room-vk> <recipient> --room <target-room-vk> -m "come join!"
+```
+
+- `<carrier-room-vk>` is the room the DM travels in — you and the recipient must
+  both be members of it.
+- `--room <target-room-vk>` is the room you are inviting them to (any room you're
+  a member of, including the carrier room itself). A fresh single-use invite
+  credential is minted, exactly as `invite create` does.
+- `-m/--message` is an optional note shown above the Accept button.
+
+The recipient accepts with `dm accept` (below), or clicks the card in the UI.
+
+### Accepting an invitation that arrived as a DM
+
+`dm list` shows an invite DM as `[Invitation to room …]`. Accept it with:
 
 ```bash
 riverctl dm accept <carrier-room-vk>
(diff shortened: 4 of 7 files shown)


sanity:
## Review round 1 — addressed in 34d8468f

Codex (external) and an adversarial Claude pass both ran. They disagreed on two points; reconciling in Codex's favour after re-checking the code:

**Two Codex P2s — both real, both fixed:**

1. **Signing-key override leaked into the target invitation.** `Storage::get_room` applies a `--signing-key-file` override to *every* room, so `build_invitation(&target)` would have signed the target-room invitation with the carrier sender identity (a non-member of the target room) — minting an invitation the target contract rejects on accept, while the command reported success. (The Claude pass missed this — it assumed `build_invitation` fetches the target's key without tracing `resolve_signing_key`.) Fixed: `build_invitation` now takes the inviter key explicitly; `dm invite` signs with the target room's **own stored key** via a new `Storage::stored_signing_key` (override bypassed), while the DM envelope keeps the possibly-overridden carrier identity. `create_invitation` still honours the override for its one room (unchanged output).

2. **Self-invite was advertised but un-acceptable.** With `--room` == carrier, the recipient is already a member, so the re-accept guards refuse the card. `execute_invite` now rejects equal carrier/target IDs; help + README drop the "carrier room itself" claim.

**From the adversarial pass:** added `dm_send_uses_legacy_raw_text_wire_shape`, a source-scrape pin that `dm send` still emits raw UTF-8 (no `encode_body` magic byte) — the highest-value hardening it flagged.

Re-verified on a live node: self-invite rejected; cross-room invite accepted (recipient joins the target room); `invite create`/`accept` still works through the refactored `build_invitation`.

Re-running the external review on the final content now (the code changed materially since round 1).

[AI-assisted - Claude]

sanity:
## Review round 2 — clean

Codex re-reviewed the final content and found **no actionable correctness issues**.

Round 2 added one more Codex P2, now fixed (9b59bccc): `dm invite` built the invitation from the **stale local cache** of the target room, so a stored-but-since-pruned target identity would deliver — and report success for — an invitation whose `invited_by` isn't a current member, which the recipient can't accept. Now it fetches the target's **current network state** and refuses if the inviting identity isn't a member (and the fresh state also gives the invitee up-to-date `room_secrets`). Pinned by `room_has_member_recognizes_owner_and_members`.

**Full-tier review complete:** external Codex (3 rounds, final clean) + one adversarial Claude pass; every finding fixed. All three P2s were the same class — "reports success on an invitation the recipient can't accept" — each now converted to a clear upfront error.

Re-verified live after every change: self-invite rejected; cross-room invite accepted (recipient joins the target room); `invite create`/`accept` still works through the refactored `build_invitation`. 228 tests green, clippy clean.

Merging once CI is green on 9b59bccc, then republishing riverctl 0.2.2 via the `riverctl-v0.2.2` tag.

[AI-assisted - Claude]

PR #463, opened 2026-07-24
docs(ui): recommend DM invites in the invite-member modal

sanity:
## Problem

The invite-member modal's only guidance was the one-person warning:

> **Share privately with one person only.** Each invitation creates a unique identity. If multiple people use the same link, they will share one identity and cause conflicts. Generate a new invitation for each person.

It says nothing about sending an invitation **via DM** (#252, #457), which is the flow we actually want people to use: ask whether they'd like to join, and once they agree, click their name in the member list → **Share invite** → River drops an invitation card into the DM thread that they accept in one click. Copying a bearer credential into an outside channel is the fallback, not the default.

## Change

Copy-only. Two blocks now, in priority order:

1. **Recommended (accent callout, above the link/code):** "Best way to invite someone: send it in a DM. If you already share a room with them, ask whether they'd like to join — then click their name in the member list and choose **Share invite**. River builds the invitation and drops it into your DM as a card they accept in one click, with nothing to copy, paste, or leak."
2. **Fallback (existing yellow warning, reframed):** "No DM yet? Share the link or code below privately, with one person only. Each invitation creates a unique identity and is good for exactly one person. If two people use the same link or code, they share one identity and neither works correctly. Click **New Invitation** for every additional person."

The warning now names the **New Invitation** button (which is right there in the modal) instead of the vaguer "generate a new invitation".

No logic, no state, no WASM change → no delegate/contract migration needed.

## Test coverage

- New classes `bg-accent-soft` / `border-accent` / `border-l-4` all verified present in the generated Tailwind CSS (both already used elsewhere in the UI).
- Playwright layout suite run against the example-data build before publishing.

### DIFF
--- a/ui/src/components/members/invite_member_modal.rs
+++ b/ui/src/components/members/invite_member_modal.rs
@@ -186,11 +186,32 @@ pub fn InviteMemberModal(is_active: Signal<bool>) -> Element {
                             );
 
                             rsx! {
-                                // Warning
+                                // Recommended path first: ask, then send the
+                                // invitation as a DM via the built-in
+                                // "Share invite" flow (#252, #457). It hands
+                                // the recipient a one-click Accept card and
+                                // never puts a bearer credential through an
+                                // outside channel, so it belongs ABOVE the
+                                // link/code it is meant to displace.
+                                div { class: "mb-4 p-3 bg-accent-soft border-l-4 border-accent rounded-r-lg",
+                                    p { class: "text-sm text-text",
+                                        span { class: "font-medium", "Best way to invite someone: send it in a DM. " }
+                                        "If you already share a room with them, ask whether they'd like to join — then click their name in the member list and choose "
+                                        span { class: "font-medium", "Share invite" }
+                                        ". River builds the invitation and drops it into your DM as a card they accept in one click, with nothing to copy, paste, or leak."
+                                    }
+                                }
+
+                                // Fallback path (no shared room yet): the
+                                // link/code below are bearer credentials — a
+                                // single reusable identity — hence the
+                                // one-person-only warning.
                                 div { class: "mb-4 p-3 bg-warning-bg border-l-4 border-yellow-500 rounded-r-lg",
                                     p { class: "text-sm text-text",
-                                        span { class: "font-medium", "Share privately with one person only. " }
-                                        "Each invitation creates a unique identity. If multiple people use the same link, they will share one identity and cause conflicts. Generate a new invitation for each person."
+                                        span { class: "font-medium", "No DM yet? Share the link or code below privately, with one person only. " }
+                                        "Each invitation creates a unique identity and is good for exactly one person. If two people use the same link or code, they share one identity and neither works correctly. Click "
+                                        span { class: "font-medium", "New Invitation" }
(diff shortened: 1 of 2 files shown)


sanity:
## Review

**Scope:** copy-only change to `InviteMemberModal`, plus additive `data-testid` hooks and a new Playwright spec. No state, no logic, no WASM → no delegate/contract migration (`check-delegate-migration` and `check-room-contract-migration` both pass, confirming the WASMs are byte-identical).

### Correctness

- ✅ The claim the new copy makes is real and reachable: `member_info_modal.rs:410` renders the **Share invite** button (`member-info-share-invite-button`) → `open_invite_via_dm_picker` → `invite_via_dm_picker_modal`, which signs an invitation against the *candidate* room and sends it as a structured DM the recipient accepts in one click.
- ⚠️ **Found and fixed in d6873fc:** the first wording said "click their name in the member list" as if the button were in *this* room's member list. It isn't — the invitee isn't in this room yet. The button lives in the member-info modal of a room you **already share**, and the picker then asks *which* room to invite them to. Copy now reads "in any room you already share with them, click their name in the member list, choose **Share invite**, and pick this room."
- ✅ The "if you already share a room" conditional also covers the disabled-button case gracefully: `share_button_enabled` is false when the user is in no other rooms, and the copy doesn't promise the button in that situation.
- ✅ The one-person warning is preserved, not weakened. Both the link and the portable code are the same bearer credential (see the existing comment at `invite_member_modal.rs:301-310`), and the reframed text now names both explicitly ("the same link or code") rather than just "link", closing a small gap in the old copy.
- ✅ The pasteable `default_msg` still carries its own IMPORTANT one-person notice, so a recipient who only sees the pasted message is unaffected by this change.

### Style / conventions

- Block structure, spacing (`mb-4 p-3 … border-l-4 … rounded-r-lg`) and the `span.font-medium` lede match the existing warning and `not_member_notification.rs`.
- `bg-accent-soft` / `border-accent` are pre-existing tokens (`ui/assets/tailwind.css` `@theme`, used in `room_list.rs` and `message_input.rs`); verified both compile into the generated `styles.css`, so this isn't relying on a class Tailwind won't emit.
- `data-testid` naming follows the documented `<surface>-<role>` kebab-case convention.
- Comments explain *why* the recommendation sits above the fallback, matching this file's existing comment density.

### Risks

- **Modal height.** Two callouts + link + portable code + message preview makes the body taller. The container is `max-h-[90vh] overflow-y-auto`, so it scrolls rather than clipping; the Playwright responsive suite (320/480/767/768/1280) is the guard. Worth an eye on the 320px case after publish.
- No security impact: no change to what is generated, signed, or exposed. Arguably a net security improvement, since it steers users away from pasting a bearer credential into third-party channels.

### Test coverage

- New `ui/tests/invite-modal-guidance.spec.ts` pins: DM recommendation present, names **Share invite**, ordered above the warning (bounding-box comparison), and the one-person warning still names **New Invitation**.
- Existing `portable-invite-code.spec.ts` / `copy-clipboard-feedback.spec.ts` continue to exercise this modal unchanged.

### Verdict

Ship it once `build` and `ui-playwright-tests` go green.

PR #465, opened 2026-07-25
fix(ui): reveal DM archive ✕ on touch devices (#462)

### LINKED ISSUE #462: fix(ui): DM archive ✕ is hover-gated on touch — invisible on phones, unreachable on tablets
Reported by Ivvor in the official room (2026-07-24):

> Quick mobile touch issue I noticed on River. It's not obvious how to banish a DM convo to the archive as the little 'x' only appears on mouse over.

Confirmed, and it's worse on tablets than on phones. The DM rail's archive ✕ (`ui/src/components/room_list/dm_rail_section.rs:220-228`) gates its reveal on a **viewport breakpoint** used as a proxy for "has a hover pointer":

```
absolute right-1 top-1/2 -translate-y-1/2 p-1 rounded text-text-muted
opacity-40 md:opacity-0 group-hover:opacity-100 group-focus-within:opacity-100
hover:text-red-400 hover:bg-surface focus:opacity-100
```

Tailwind v4 emits `group-hover:*` inside `@media (hover: hover)`. Verified in the compiled stylesheet (`ui/assets/styles.css`, the sheet `app.rs:462` actually loads):

```css
@media (hover:hover){.group-hover\:opacity-100:is(:where(.group):hover *){opacity:1} ...}
...
@media (width >= 48rem){ ... .md\:opacity-0{opacity:0}}
```

So on any touch device (`hover: none`) `group-hover:opacity-100` can never match, and the only thing left is the `md:` breakpoint.

## Three distinct defects

**1. Phone (<768px) — undiscoverable.** `opacity-40` applies, `md:opacity-0` doesn't, `group-hover` never fires. The ✕ is permanently at 40% opacity, 12×12px, in `text-text-muted`. Technically rendered; in practice invisible. This is what Ivvor hit.

**2. Tablet (≥768px with `hover: none`) — impossible, and it eats taps.** iPad/Android tablets in most orientations exceed the `md` breakpoint (48rem = 768px) and report `hover: none`. There, `md:opacity-0` applies and nothing can ever reveal the button: `group-hover` is media-gated out, and `group-focus-within`/`focus` need a keyboard. **Archiving a DM thread is unreachable on tablets.**

Worse, the button is `opacity: 0` but still hit-testable — no `pointer-events: none`, unlike the `.hover-actions` fix from #402. It is absolutely positioned over the right edge of the row (the row button carries `pr-9` to clear it), so a tap aimed at the right of a DM row opens... the archive action, not the thread. The Undo toast catches it, but it reads as the app randomly archiving conversations.

Specificity note for anyone re-checking: `md:opacity-0` is `(0,1,0)` while `group-hover:opacity-100:is(...)` is `(0,2,0)`, so desktop hover still wins over the breakpoint rule. Desktop is fine; only touch is broken.

**3. Tap target is 20×20px.** `p-1` (4px) around a 12px icon. Below WCAG 2.5.8's 24×24 minimum, and far below Apple HIG (44) / Material (48).

## The fix already exists in this codebase

#402 solved exactly this class of bug for the message hover-actions bar, and the DM rail never adopted it. `ui/assets/main.css` defines:

- `.hover-actions` — gets `pointer-events: none` under `@media (hover: none)`, so an invisible bar can't eat taps
- `.touch-actions` — `display: none` by default, `display: flex` under `@media (hover: none)`, decorating an always-visible, properly-sized (`w-8 h-8`) kebab

Suggested changes:

- [ ] **Drop `md:opacity-0`.** Gate on hover capability, not viewport width — the breakpoint is the root cause of the tablet case. Reuse the `.hover-actions` / `.touch-actions` pattern rather than inventing a third mechanism.
- [ ] **On `hover: none`, make the ✕ fully visible** (not 40%) and give it a real tap target — at least 44×44 (`w-11 h-11`, or `p-3` with a larger icon), widening the row's `pr-9` to match.
- [ ] **On `hover: hover`, keep the current opacity-0 → group-hover reveal** — desktop behaviour is fine and shouldn't change.
- [ ] **Add `pointer-events: none` to any opacity-0 state that can't be revealed**, so it can never capture a tap (the #402 lesson).
- [ ] Consider swipe-to-archive or a long-press menu as the more idiomatic touch affordance, consistent with the kebab already used in the conversation view.

Repo-wide audit of `group-hover:opacity` / `*:opacity-0` (4 hits): `conversation.rs:3040` is the #402-fixed `.hover-actions` bar and is fine; `dm_rail_section.rs:222` is this bug; `message_actions.rs:113` (`QuickReactionButton`) has the same bare `opacity-0 group-hover:opacity-100` with no touch fallback, but the component has no callers, so it is dead code rather than a live defect — worth deleting or fixing alongside. So this is the only live instance.

Regression coverage: the Playwright suite can assert, at a touch-emulated ≥768px viewport, that the archive control is both visible and ≥44px — that is the check that would have caught this.

[AI-assisted - Claude]

sanity:
## Problem

Ivvor (official room, 2026-07-24): "the little 'x' only appears on mouse over" — the archive ✕ in the Direct Messages rail is unreachable on touch.

The button gated its reveal on the `md:` viewport breakpoint as a stand-in for "has a hover pointer":

```
opacity-40 md:opacity-0 group-hover:opacity-100 group-focus-within:opacity-100
```

Tailwind v4 emits `group-hover:*` inside `@media (hover: hover)` (confirmed in the compiled `styles.css`), so on any touch device that reveal can never fire and only the breakpoint decides visibility:

- **Phone (<768px):** `opacity-40` applies, `md:opacity-0` doesn't, `group-hover` never fires → the ✕ is permanently at 40% opacity, 12px, muted grey. Effectively invisible. This is what Ivvor hit.
- **Tablet (≥768px + touch):** `md:opacity-0` applies and nothing can reveal it — but the `opacity-0` button is still hit-testable and overlays the row's right edge, so a tap meant to open the thread archives it instead (the Undo toast then makes it look like the app randomly archives conversations).
- Tap target was ~20×20px, below WCAG 2.5.8's 24 and Apple HIG's 44.

Desktop was unaffected: `group-hover`'s `:is()` selector outranks `md:opacity-0` on specificity, so hover reveal still won there.

## Approach

Gate on hover **capability**, not viewport width — the same `.hover-actions` / `.touch-actions` split #402 introduced for the message action bar (and which the DM rail never adopted).

- The button is `opacity-0` at rest, so the clean hover reveal is preserved on desktop (`@media (hover: hover)`).
- `.dm-archive-btn` in `main.css` forces it fully visible with a **44px tap target** under `@media (hover: none)`.
- `.dm-rail-row-btn` widens the row's right padding on touch so the enlarged ✕ doesn't overlap the nickname or unread badge.

No behavioural change on desktop; the `md:`-as-hover-proxy is removed.

## Testing

Measured computed styles in a real browser (Playwright) against the **shipped** bundled stylesheets:

| Context | new `.dm-archive-btn` | old `md:opacity-0` |
|---|---|---|
| Touch, phone (390px) | **opacity 1, 44×44** | opacity 0.4, 19×28 (invisible) |
| Touch, tablet (820px) | **opacity 1, 44×44, hit-testable** | invisible-but-tappable |
| Desktop, at rest | opacity 0 (hover reveal preserved) | opacity 0.4 |

On a touch viewport the new ✕ renders as a crisp, full-opacity mark with clear separation from the unread badge; the old one is a faint 40% smudge crammed against the badge.

Also added two source-scrape pin tests in `dm_rail_section.rs` (`cargo test -p river-ui --bins`, both pass) that fail if a refactor reintroduces the viewport-gated reveal (`md:`-breakpoint) or drops the `@media (hover: none)` touch CSS. The existing `dm-archive-ux.spec.ts` can't drive this end-to-end because the example-data build ships no DMs, so no rail row renders.

Closes #462

[AI-assisted - Claude]

### DIFF
--- a/ui/assets/main.css
+++ b/ui/assets/main.css
@@ -144,3 +144,31 @@
         pointer-events: none;
     }
 }
+
+/* DM-rail Archive ✕ — touch reveal (freenet/river#462).
+ *
+ * Same failure mode as #402 above: the archive button in the Direct Messages
+ * rail (`dm_rail_section.rs`) is `opacity-0` and reveals via Tailwind
+ * `group-hover:opacity-100`, which Tailwind v4 wraps in `@media (hover: hover)`.
+ * On a touch device that reveal can never fire, so before this rule the button
+ * was invisible on phones and — because it is `opacity-0` yet still
+ * hit-testable and overlays the row's right edge — an invisible-but-tappable
+ * hazard on tablets (a tap meant to open the thread archived it instead).
+ *
+ * Gate on hover CAPABILITY, not viewport width (the previous `md:opacity-0`
+ * proxy is what broke tablets, which exceed the `md` breakpoint yet have no
+ * hover pointer). On `hover: none` force the button fully visible with a
+ * >=44px tap target (WCAG 2.5.8 / Apple HIG), and widen the row's right
+ * padding to match so the enlarged ✕ never overlaps the nickname or unread
+ * badge. Desktop (`hover: hover`) is untouched: opacity-0 -> group-hover. */
+@media (hover: none) {
+    .dm-archive-btn {
+        opacity: 1;
+        min-width: 2.75rem;
+        min-height: 2.75rem;
+    }
+
+    .dm-rail-row-btn {
+        padding-right: 3.5rem;
+    }
+}
--- a/ui/assets/main.css
+++ b/ui/assets/main.css
@@ -155,13 +155,28 @@
  * hit-testable and overlays the row's right edge — an invisible-but-tappable
  * hazard on tablets (a tap meant to open the thread archived it instead).
  *
- * Gate on hover CAPABILITY, not viewport width (the previous `md:opacity-0`
+ * Gate on pointer CAPABILITY, not viewport width (the previous `md:opacity-0`
  * proxy is what broke tablets, which exceed the `md` breakpoint yet have no
- * hover pointer). On `hover: none` force the button fully visible with a
- * >=44px tap target (WCAG 2.5.8 / Apple HIG), and widen the row's right
- * padding to match so the enlarged ✕ never overlaps the nickname or unread
- * badge. Desktop (`hover: hover`) is untouched: opacity-0 -> group-hover. */
-@media (hover: none) {
+ * hover pointer). Force the button fully visible with a >=44px tap target
+ * (WCAG 2.5.8 / Apple HIG), and widen the row's right padding to match so the
+ * enlarged ✕ never overlaps the nickname or unread badge.
+ *
+ * The query is `(hover: none), (any-pointer: coarse)`, NOT `(hover: none)`
+ * alone. `hover` reports the PRIMARY input, so a touchscreen laptop or a touch
+ * Chromebook answers `hover: hover` while the user's finger is on the glass —
+ * they would keep an `opacity-0` button that is still hit-testable over the
+ * row's right edge, which is the invisible-but-tappable hazard this fix exists
+ * to remove, just moved to a different device class. `any-pointer: coarse` is
+ * true whenever ANY available pointer is a finger, so it catches them.
+ *
+ * The cost is that a hybrid device shows the ✕ always-visible rather than on
+ * hover. That is the right way round: a mouse user on such a device sees a
+ * button that is merely always present, while a finger user on the alternative
+ * sees nothing and archives threads by accident.
+ *
+ * A mouse-only desktop matches neither term and is untouched:
+ * opacity-0 -> group-hover. */
+@media (hover: none), (any-pointer: coarse) {
     .dm-archive-btn {
         opacity: 1;
         min-width: 2.75rem;
--- a/ui/src/components/room_list/dm_rail_section.rs
+++ b/ui/src/components/room_list/dm_rail_section.rs
@@ -1799,72 +1799,149 @@ mod tests {
 
     // ---- #462: archive ✕ must be reachable on touch devices ----
     //
-    // The archive reveal can't be driven end-to-end in Playwright (the
-    // example-data build ships no DMs, so no rail row renders — see
-    // `ui/tests/dm-archive-ux.spec.ts`), so these pin the fix by
-    // source-scraping the component markup and the static stylesheet.
-    // They fail if a refactor reintroduces the viewport-breakpoint reveal
-    // that stranded touch users, or drops the touch-visibility CSS.
+    // The archive reveal can't be driven end-to-end in Playwright through the
+    // real rail (the example-data build ships no DMs, so no row renders — see
+    // `ui/tests/dm-archive-ux.spec.ts`), so the CASCADE is pinned in the
+    // browser by `dm-archive-touch.spec.ts`, which probes the shipped
+    // stylesheets directly. These two pin the SOURCE side: that the markup
+    // still carries the hooks that stylesheet targets, and that the rule is
+    // still inside a pointer-capability media query rather than at top level.
+    //
+    // Both cut production off at `mod tests` for the reason spelled out on
+    // `rail_builders_read_infallible_guard_before_first_fallible_read` above:
+    // `include_str!` reads this file back, assert MESSAGES included, so an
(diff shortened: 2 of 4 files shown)


sanity:
## Review round

Rebased onto current `main` (the base was from 2026-07-25; `main` has since taken #494, #517, #518, #520, #521). One conflict in `dm_rail_section.rs` — both sides added tests to the same module (main's #499 rail tests, this branch's #462 pins); both kept.

Reviewed with a skeptical adversarial lens. **The CSS fix is correct** — the reviewer traced the cascade and it wins for two independent reasons: `main.css` is entirely unlayered while every Tailwind utility sits inside `@layer utilities`, and unlayered declarations outrank every declared layer regardless of source order; independently, `app.rs` loads `styles.css` before `main.css`. So the fix is robust even if someone reorders those two `Stylesheet` lines.

Three things needed fixing before merge, all in `cdd3cb7e`.

### The gate was too narrow — the hazard just moved

`@media (hover: none)` keys off the **primary** input, so a touchscreen laptop or a touch Chromebook answers `hover: hover` while the user's finger is on the glass. Below 768px those devices went from `opacity-40` (dim but visible) to `opacity-0` — still hit-testable over the row's right edge. That is exactly the invisible-but-tappable hazard this PR exists to remove, relocated to a different device class rather than fixed.

Now `(hover: none), (any-pointer: coarse)`, which is true whenever *any* available pointer is a finger. The cost is that a hybrid device shows the ✕ always-visible instead of on hover, which is the right way round: a mouse user there sees a button that is merely always present, while a finger user under the alternative sees nothing and archives threads by accident.

### Both pins passed against the mutations they existed to catch

`DM_RAIL_SRC` was the whole file — test module and assert messages included — so `contains("dm-archive-btn")` was satisfied by its own failure message, and `contains("group-hover:opacity-100")` by a production comment. **Deleting `dm-archive-btn` from the class attribute, the single hook the entire fix hangs on, left the test green.** This file documents the correct convention 200 lines above (cut at `mod tests`) and its existing pins follow it; these two had silently opted out.

The CSS pin was worse. It asserted that `@media (hover: none)` appears somewhere *before* the rule — and the #402 block opens one far above it, so moving `.dm-archive-btn` out to top level (which would force the ✕ permanently visible on a mouse-only desktop and destroy the hover reveal) passed too.

Both rewritten: production cut at `mod tests`, positive checks anchored on the class *attribute* rather than a bare substring, and the CSS check brace-scans to find the innermost media query still open at the rule and asserts on that query's text. All three mutations — drop the class, move the rule to top level, narrow the query back — now go red.

Also fixed a latent panic in the old check: it byte-sliced a fixed 200-byte window out of `main.css`, which contains ✕, — and ⋮ in its comments, so an appended rule could land the index mid-codepoint.

### Nothing measured the actual cascade

The source pins cannot see the thing that really broke — whether `.dm-archive-btn { opacity: 1 }` beats Tailwind's `opacity-0` at run time. Two one-line edits would silently lose it (pulling `main.css` into an `@layer`; reordering the `Stylesheet` tags while the layering changed) and no source scrape would notice.

`dm-archive-touch.spec.ts` probes the shipped stylesheets in a real engine using a synthetic element with the same classes — the real rail can't be driven end-to-end because the example build ships no DMs. It branches on `matchMedia`, so it asserts opacity 1 + a ≥44px target + the widened row padding on coarse-pointer projects and opacity 0 on mouse-only ones, which makes it meaningful on all five Playwright projects rather than skipped on most.

Mutation-verified in the browser, not just asserted: deleting the touch rule from `main.css` and rebuilding turns every coarse-pointer project red while the mouse-only ones stay green.

### Left as-is, deliberately

- The tap zone is 48px wide while the drawn ✕ stays 12px. Enlarging the icon is a visual-design call, and the undo toast already covers a mis-tap; noted rather than changed.
- `QuickReactionButton` (`message_actions.rs`) still carries a bare `opacity-0 group-hover:opacity-100` with no touch fallback. It has no callers — dead code — so it is out of scope here, but it is the same latent bug if it is ever wired up.

721 Rust tests pass; 15/15 Playwright locally across both DM-archive specs on all five engines.

[AI-assisted - Claude]

PR #466, opened 2026-07-25
perf(ui): stop re-rendering the whole message history on every state change

sanity:
## Problem

A River user reported the chat interface has "overactive reactivity" with "a lot of reflow/repaint", and guessed it was a flexbox side effect.

Profiling says the reactivity half of that is right and the layout half is not. On a 90-message room, sending one message costs **1466 ms** of main-thread time at 6x CPU throttle (mid-range phone). Of that, layout + style recalc is **24 ms (1.6%)**. Scripting is **1187 ms (81%)** — and only 2-8 DOM nodes actually change. The browser was doing almost no layout work; the app was computing a render it then threw away.

Measured with Chromium CDP `Performance.getMetrics` + a `MutationObserver`, against an example-data build padded to 90 messages (production caps `max_recent_messages` at 100).

## Root cause

Dioxus decides whether a child component can be skipped by comparing **all** of its props with `PartialEq` — event handlers included. The derive excludes callbacks from the *move* logic, not from the *equality* check:

```rust
let equal = self == new;   // derived PartialEq over every field
#move_event_handlers
if !equal { /* move regular fields */ }
equal
```

A closure written inline in `rsx!` becomes a fresh `Callback` on every render, and `Callback`'s `PartialEq` is pointer identity (`ptr_eq(&other.callback) && origin == origin`). So the four inline handlers passed to `MessageGroupComponent` (`on_react`, `on_request_delete`, `on_edit`, `on_reply`) made it **permanently un-memoizable**: every re-render of `Conversation` re-rendered every message group.

Instrumented call counts on an 85-group room:

| interaction | group renders before | after |
|---|---|---|
| send one message | 340 (4 full passes) | **4** |
| scroll the history | 85 (1 full pass) | **0** |
| type a character | 0 | 0 |
| idle | 0 | 0 |

Scrolling re-rendered the entire history because crossing the bottom sentinel flips `is_at_bottom`, which `Conversation` reads in render.

## Approach

**1. Stable callback identities.** `use_callback` allocates the handle once and swaps the inner closure each render, so identity is stable (props compare equal → the group memoizes) while captured state stays current. Freshness does not depend on the memoize path: `use_callback` itself replaces the inner closure on every `Conversation` render, and Dioxus additionally re-points the old handle at the newest closure inside `memoize` (called unconditionally, before the equality result is used).

Considered and rejected: hand-writing the props struct with a manual `PartialEq` that skips the handler fields. That is the documented opt-out, but it means restating 12 fields and re-deriving `Props` by hand, and it silently breaks if a field is added later.

**2. Cache the `Intl.DateTimeFormat` instances.** `toLocaleTimeString` / `toLocaleString` with an options object construct a new formatter per call — **67-74 µs** each in Chromium, vs **4.4-4.6 µs** for a cached one (~15x). These ran once per message group per render, so a 90-message room paid ~12 ms of pure date formatting per re-render. Output is byte-identical (verified for both formats).

## Results

6x CPU throttle, 94 messages:

| scenario | scripting before → after | longest task before → after |
|---|---|---|
| send a message | 1176 ms → **228 ms** | 1114 ms → **109 ms** |
| send another | 617 ms → **130 ms** | 524 ms → **76 ms** |
| scroll history | 253 ms → **12 ms** | 253 ms → **0** |
| type 26 chars | 120 ms → 120 ms | 0 → 0 |

The ~1.1 s main-thread freeze per sent message becomes ~109 ms.

## Testing

- Two source-scrape pin tests (the idiom this repo already uses) fail CI if either regression returns: handlers must be built with `use_callback` and the call site must pass those handles, and the date formatters must stay cached.
- `cargo test -p river-ui --bins`: 506 passed.
- Full Playwright suite across chromium, firefox, webkit, mobile-chrome, mobile-safari: **425 passed, 0 failed**. This directly exercises the handlers changed here — "Reply from the kebab opens the composer reply preview", "React from the kebab opens the emoji picker and closes the menu", "kebab opens a menu with Reply / Edit / Delete on own messages".
- UI-only change: no delegate, contract, or `common/` code touched, so no WASM migration.

## Drive-by

The two sandboxed-iframe layout tests hardcoded `http://localhost:8082` as the iframe `src` instead of honouring the configured `baseURL`, so they failed against any other port (after 3x30 s retries). They now use `baseURL`, whose default is unchanged. Verified pre-existing by reproducing on the unmodified base commit.

## Not addressed here

Named rather than silently dropped:

- `Conversation` still re-renders **4 times** per sent message. Each pass is now cheap, but 1 would be better.
- The composer's textarea auto-resize (`height: auto` → read `scrollHeight` → set height) forces ~3 layouts per keystroke — 67 ms layout + 108 ms style for 26 characters at 6x throttle. This is the one genuine reflow cost in the app, and it is the part of the original report that was literally about layout.

Neither is a regression from this PR.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -4533,9 +4533,41 @@ mod tests {
     // -----------------------------------------------------------------
     // Message-group memoization (over-rendering regression guard)
     // -----------------------------------------------------------------
+    /// The production half of this file, i.e. everything above this test module.
+    ///
+    /// `include_str!("conversation.rs")` pulls in the test module too, so a
+    /// source scrape that searched the whole string would match the literals
+    /// written in the assertions below and pass no matter what the real code
+    /// does. Cutting the file at the test-module header is what makes these
+    /// scrapes able to fail.
+    fn production_src() -> &'static str {
+        let src = include_str!("conversation.rs");
+        let marker = "#[cfg(test)]\nmod tests {";
+        let cut = src
+            .rfind(marker)
+            .expect("conversation.rs must contain the `#[cfg(test)] mod tests {` header");
+        &src[..cut]
+    }
+
+    #[test]
+    fn production_src_excludes_this_test_module() {
+        // Guard for the guard: if the marker ever stops matching (reformat,
+        // renamed module) `production_src` would silently start returning the
+        // whole file and the scrapes below would go back to being self-fulfilling.
+        let prod = production_src();
+        assert!(
+            !prod.contains("fn production_src_excludes_this_test_module"),
+            "production_src() must cut the file before the test module"
+        );
+        assert!(
+            prod.contains("fn MessageGroupComponent"),
+            "production_src() must still contain the production code it scrapes"
+        );
+    }
+
     #[test]
     fn message_group_callbacks_are_stable_handles() {
-        let src = include_str!("conversation.rs");
+        let src = production_src();
 
         // Dioxus decides whether a child component can be skipped by comparing
         // ALL of its props with PartialEq — event handlers included. A closure
@@ -4594,6 +4626,15 @@ mod tests {
             src.contains("_timeFmt") && src.contains("_fullFmt"),
             "format_time_local / format_full_datetime_local must reuse cached \
              Intl.DateTimeFormat instances rather than building one per call."
+        );
+        // A cached formatter pins the timezone it was built with, so the cache
+        // must be invalidated when the device timezone changes mid-session —
+        // otherwise message times drift away from the date separators, which
+        // read the current zone directly.
+        assert!(
+            src.contains("getTimezoneOffset()") && src.contains("_fmtOffset"),
+            "The cached Intl formatters must be rebuilt when the device timezone \
+             changes; otherwise times render in the stale zone until reload."
         );
         assert!(
             !src.contains("date.toLocaleTimeString(") && !src.contains("date.toLocaleString("),
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -1991,6 +1991,54 @@ pub fn Conversation() -> Element {
         }
     };
 
+    // Stable handles for the per-message-group callbacks.
+    //
+    // A closure written inline in `rsx!` becomes a fresh `Callback` on every
+    // render, and `Callback`'s `PartialEq` is pointer identity. The props
+    // `PartialEq` that Dioxus uses to decide whether a child can be memoized
+    // compares EVERY prop, callbacks included — so inline closures made
+    // `MessageGroupComponent` un-memoizable: every re-render of `Conversation`
+    // re-rendered all N message groups even when nothing about them changed.
+    // Measured on a 90-message room: one sent message re-rendered every group
+    // 4 times, and merely scrolling past the bottom sentinel re-rendered all of
+    // them once.
+    //
+    // `use_callback` allocates the handle once and swaps the inner closure each
+    // render, so the identity is stable (props compare equal → group memoizes)
+    // while the captured state stays current. Dioxus additionally re-points the
+    // old handle at the newest closure inside `memoize`, so a memoized group
+    // still invokes the latest logic.
+    let on_react_cb = use_callback({
+        let handle_toggle_reaction = handle_toggle_reaction.clone();
+        move |(msg_id, emoji): (MessageId, String)| {
+            let handle_toggle_reaction = handle_toggle_reaction.clone();
+            handle_toggle_reaction(msg_id, emoji);
+        }
+    });
+    let on_request_delete_cb = use_callback(move |msg_id: MessageId| {
+        pending_delete.set(Some(msg_id));
+    });
+    let on_edit_cb = use_callback({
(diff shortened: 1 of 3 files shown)


sanity:
## External review pass (Codex) — both findings fixed

Ran `codex review --base main` as the independent non-Claude pass. Two findings, both valid and both now addressed in b8339050.

### [P3] The callback pin test was self-fulfilling — good catch, this was real

`include_str!("conversation.rs")` pulls in the test module itself, so every literal the assertions searched for (`"on_react: on_react_cb"`, `"let on_react_cb = use_callback("`, …) was present in the test's own source. `src.contains(..)` was true regardless of what the production code did. The test could never have failed.

Fixed by cutting the file at the test-module header before scraping, plus a guard test (`production_src_excludes_this_test_module`) so a reformat or module rename can't silently restore the self-fulfilling behaviour.

**Verified the fix by reinstating the regression**: put an inline closure back at the call site and removed the timezone check, and confirmed both tests fail with the intended messages:

```
test components::conversation::tests::message_group_callbacks_are_stable_handles ... FAILED
  The MessageGroupComponent call site must pass the stable use_callback handle.
  Passing an inline `move |..| { .. }` closure here silently un-memoizes every
  message group. Missing: on_react: on_react_cb

test components::conversation::tests::local_date_formatters_are_cached ... FAILED
  The cached Intl formatters must be rebuilt when the device timezone changes;
  otherwise times render in the stale zone until reload.
```

### [P2] Cached formatters went stale on a timezone change

Correct. `Intl.DateTimeFormat` resolves its timezone once at construction, so a device timezone change mid-session would leave message times in the old zone while `local_date_key` (which reads the `Date`'s local fields directly) already reflected the new one — times and date separators disagreeing. The per-call form being replaced picked such a change up immediately, so the cache has to as well.

`ensureFormatters()` now rebuilds when the current UTC offset changes. Measured cost of each candidate check:

| check | cost |
|---|---|
| `new Date().getTimezoneOffset()` | **0.16 µs** |
| `fmt.format(ms)` (the work itself) | 1.5 µs |
| `new Intl.DateTimeFormat().resolvedOptions().timeZone` | **84 µs** |
| original `toLocaleTimeString(undefined, opts)` | 71 µs |

Re-resolving the zone *name* — the obvious correct-looking check — is more expensive than not caching at all, and asking the cached formatter for its own `resolvedOptions()` only ever reports the zone it was built with, so neither can do this job. The offset check is the only cheap signal available; its limitation (two zones sharing a current offset are indistinguishable, and render identically anyway) is documented at the call site.

**Output identity re-verified end-to-end** against the original per-call form: **0 mismatches over 4000 timestamps spanning a full year** in America/Chicago, Europe/London, Asia/Kolkata and Pacific/Auckland — covering both DST halves, a half-hour offset, and southern-hemisphere DST — while still running **14-17x faster**. The headline figure in the code comment is corrected from the raw-format 40x to the measured end-to-end 15x.

### Re-verification

- `cargo test -p river-ui --bins`: 507 passed.
- Full Playwright suite, all 5 browser projects: 424 passed, 0 failed, 1 flaky (`title-unread-badge` on mobile-safari, passed on retry).

That flake is pre-existing and unrelated — it passes 5/5 in isolation on both this branch and the unmodified base commit, and only appeared under full 5-project parallel load. Filed separately rather than dismissed.

[AI-assisted - Claude]

sanity:
Closing this without merging. The analysis is good and I have moved all of it into **#523** so nothing is lost; the code stays on branch **`perf/message-group-memoization`** (head `b8339050`), not deleted.

### Why it isn't rebasing

Its `conversation.rs` half was written against the pre-#501/#505 render body. Since then #501 and #505 rewrote exactly that code around the history window — backfill sentinel, `HistoryWindow::resolve`, windowed `rows` — so the conflict is the whole rendered-history block, and the branch's `rows.into_iter().enumerate()` / `is_last_group` / `rows_len` shape no longer exists. That is a re-implementation, not a rebase, and I would rather it be done deliberately than smuggled through a conflict resolution.

### Why the numbers need redoing, not just carrying over

The headline measurements (340 group renders per send; 1176 ms → 228 ms of scripting at 6x throttle) were taken when **the whole history rendered**. With windowing, only the visible tail — about 60 items — renders at all. So the window has already absorbed an unknown share of the win, and the remaining benefit is unmeasured.

The *mechanism* is untouched by any of that and still true: `Callback`'s `PartialEq` is pointer identity, a closure written inline in `rsx!` is a fresh `Callback` every render, and the derived props `PartialEq` compares handler fields — so `MessageGroupComponent` is permanently un-memoizable today. That part needs no re-verification, only the payoff does.

### The cheap half is worth taking on its own

The `Intl.DateTimeFormat` cache in `ui/src/util.rs` is self-contained, conflict-free against current `main`, and independently measurable (~70 µs → ~4.5 µs per call, once per message group per render). Its one subtlety — a cached formatter resolves its timezone at construction, so it has to rebuild when the device offset changes or message times will disagree with the date separators above them — is already worked out on the branch, including why the two obvious alternatives (`resolvedOptions()` on the cached formatter; re-resolving the zone name) do not work.

That is the piece I would land first if picking this up.

[AI-assisted - Claude]

PR #470, opened 2026-07-25
feat(cli): add deputy-status read commands to riverctl

sanity:
## Problem

riverctl can WRITE deputy status (`member deputize`, `member revoke-deputy`) but
has no way to READ it. The data is in each member's own `AuthorizedMemberInfo.deputies`
(`common/src/room_state/member_info.rs`), but `member list` never showed it and
`debug room-state` never included it.

The only way to tell whether a `deputize` had landed was to re-run `deputize` and
look for `Member is already a deputy; nothing to do`. That is a bad API, and it
gives no answer at all to the more useful question: "does this member have
moderation authority, and from whom?"

## Approach

Two read commands, plus deputy annotations on the two existing read surfaces:

| Command | Question it answers |
|---|---|
| `member deputies <ROOM> [MEMBER_ID]` | forward: who has X deputized? (defaults to your own identity) |
| `member deputized-by <ROOM> <MEMBER_ID>` | reverse: who has deputized X? |
| `member list <ROOM>` | each member row now shows who deputized them |
| `debug room-state <ROOM>` | every grant in the room |

All of them support `-f json`. The two new commands emit one consistent shape,
`{room_id, direction, subject, grants[]}`, where each grant carries `deputizer`,
`deputy`, `scope` (`room-wide` / `invite-subtree`) and `active`.

### Why two subcommands rather than one command with a `--of` flag

The two directions have genuinely different subjects and different output, and a
flag that silently swaps which side of the relation the argument refers to is
exactly the ambiguity this feature exists to remove. Two names that read as
English sentences (`X's deputies` / `X is deputized by`) are harder to misread.

### Semantics

- **Per-deputizer, never a global flag.** Every rendering names the deputizer
  explicitly. This deliberately does not reproduce the UI's viewer-scoped green
  shield badge, which is a recurring source of confusion.
- **Canonical records only.** All reads route through `MemberInfoV1::deputies_of`,
  which selects the highest-rank record per member. `MemberInfoV1::verify`
  deliberately accepts a state carrying more than one record per member
  (migration-safety), so a bare first-match scan can read an already-revoked
  record and report authority the contract does not enforce.
- **Inert grants are shown, not hidden.** A grant naming a member who has left
  the room is listed but marked `inactive`, matching `MembersV1::is_ban_authorized`,
  which only honours deputy authority between current members. Hiding it would
  make a lingering grant invisible; claiming it is active would be wrong.
- **Ambiguous prefixes are an error.** Unlike the write paths, which take the
  first `.find()` match, an input matching more than one member id is rejected
  with the list of matches instead of silently picking one.

### Scope

CLI/read-only. No contract, WASM or serialized-state changes:
`git diff origin/main --stat` touches only `cli/`.

## Testing

19 new unit tests (`cargo test -p riverctl`: 244 lib tests pass).

`cli/src/deputies.rs` covers forward lookup, reverse lookup, both directions
agreeing, empty deputies, a member with no `member_info` record, short-id
resolution (case-insensitive, unknown id, ambiguous prefix), canonical-record
selection (a v1 grant superseded by a v2 revoke must not be reported), a grant
naming a pruned deputy, a grant from a pruned deputizer, deterministic ordering
of `all_grants` / `deputizers_by_deputy`, and the human status line naming the
deputizer.

`cli/src/commands/member.rs` covers the clap surface: the optional `MEMBER_ID`
on `deputies`, the required one on `deputized-by`, and the kebab-case spelling.

Also smoke-tested against the live Official room
(`4uNUKFzZQCnzo4K2ecZ16cMsYEEfoaRS35z6exEsbvm4`) on a local node, which
surfaced a real inactive grant to a pruned member:

```
$ riverctl member deputies <room> PMAQEUP5
  (unknown) (3YBS76GA)  inactive: (unknown) (3YBS76GA) is not currently in this room
  Ian Clarke (7XSOGJTK)  active: may ban within Invite Bot (PMAQEUP5)'s invite subtree
  HostFat (DLQVWAIV)  active: may ban within Invite Bot (PMAQEUP5)'s invite subtree
  ...

$ riverctl member deputized-by <room> 7XSOGJTK
Ian Clarke (7XSOGJTK) has been deputized by 1 member(s):
  Invite Bot (PMAQEUP5)  active: may ban within Invite Bot (PMAQEUP5)'s invite subtree
```

https://claude.ai/code/session_01HP4Xk5kf2FujRiFq7qye38

[AI-assisted - Claude]

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -200,11 +200,24 @@ often nobody.
 
 Both accept `--format json`, emitting `{room_id, direction, subject, grants[]}`.
 Each grant carries `deputizer`, `deputy`, `scope` (`room-wide` /
-`invite-subtree`), `reaches_members` (how many members the grant actually
-covers) and `active`. A grant whose deputizer or deputy has left the room is
-reported with `active: false`; the contract does not honour such a grant.
-`in_room` is `true` for the room owner even though the owner never appears in
-the member list.
+`invite-subtree`), `active`, and `reaches_members`.
+
+`reaches_members` is how many members the grant can currently be used against,
+computed from the contract's own rules: `0` for an inactive grant, every member
+for an owner grant, otherwise the deputizer's invite subtree minus anyone in it
+who has themselves deputized this deputy (a member cannot be banned by someone
+they deputized). It is often `0` even for a live grant, because most members
+have invited nobody. `reaches_members > 0` is the check for "this member
+actually has moderation authority here".
+
+A grant whose deputizer or deputy has left the room is reported with
+`active: false`; the contract does not honour such a grant. `in_room` is `true`
+for the room owner even though the owner never appears in the member list.
+
+Nicknames are quoted and escaped in terminal output. They are set by the member
+themselves and are printed next to an authority claim, so an unescaped one could
+forge a line that reads as a real deputy grant. The `-f json` output carries the
+faithful nickname.
 
 `member list --format json` gains `deputies` (who this member deputized) and
 `deputized_by` (who deputized them) alongside the existing `member_id` and
--- a/cli/README.md
+++ b/cli/README.md
@@ -173,13 +173,39 @@ riverctl member set-nickname <room-owner-vk> "New Nickname"
 riverctl member ban          <room-owner-vk> <member-vk>   # Owner only.
 ```
 
+### Deputies
+
+A deputy can ban within their deputizer's invite subtree. Deputies are
+**per-deputizer**, not a room-wide flag: "is X a deputy?" is only meaningful as
+"whose deputy is X?", so both queries name the deputizer explicitly.
+
+```bash
+riverctl member deputize      <room-owner-vk> <member-id>  # Grant.
+riverctl member revoke-deputy <room-owner-vk> <member-id>  # Revoke.
+
+riverctl member deputies      <room-owner-vk> [member-id]  # Who has X deputized?
+riverctl member deputized-by  <room-owner-vk> <member-id>  # Who has deputized X?
+```
+
+`member deputies` defaults to your own identity in the room.
+`member deputized-by` is the one that answers "does this member have moderation
+authority, and from whom?" — the room owner's grant is room-wide, anyone else's
+is limited to their own invite subtree.
+
+Both accept `--format json`, emitting `{room_id, direction, subject, grants[]}`
+where each grant carries `deputizer`, `deputy`, `scope`
+(`room-wide` / `invite-subtree`) and `active`. A grant is inactive when either
+party is no longer in the room, which the contract does not honour. `member
+list` also annotates each member with who deputized them, and `debug room-state`
+reports every grant in the room.
+
 ## Command reference
 
 | Group      | Commands                                                                |
 |------------|-------------------------------------------------------------------------|
 | `room`     | `create`, `list`, `join`, `leave`, `republish`, `config`                |
 | `message`  | `send`, `list`, `stream`, `edit`, `delete`, `react`, `unreact`, `reply` |
-| `member`   | `list`, `set-nickname`, `ban`                                           |
+| `member`   | `list`, `set-nickname`, `ban`, `deputize`, `revoke-deputy`, `deputies`, `deputized-by` |
 | `invite`   | `create`, `accept`                                                      |
 | `dm`       | `send`, `list`, `purge`, `accept`                                       |
 | `identity` | `whoami`, `export`, `import`                                            |
--- a/cli/README.md
+++ b/cli/README.md
@@ -170,8 +170,12 @@ riverctl identity import < my-identity.token     # On another machine.
 ```bash
 riverctl member list         <room-owner-vk>
 riverctl member set-nickname <room-owner-vk> "New Nickname"
-riverctl member ban          <room-owner-vk> <member-vk>   # Owner only.
+riverctl member ban          <room-owner-vk> <member-id>
 ```
+
+`member ban` is not owner-only: the room owner can ban anyone, and so can a
+member banning within their own invite subtree, or a deputy of such a member
+(see below).
 
 ### Deputies
 
@@ -187,17 +191,33 @@ riverctl member deputies      <room-owner-vk> [member-id]  # Who has X deputized
 riverctl member deputized-by  <room-owner-vk> <member-id>  # Who has deputized X?
 ```
 
-`member deputies` defaults to your own identity in the room.
+`member deputies` defaults to your own identity in the room (the identity that
+would sign, so `--signing-key-file` applies).
 `member deputized-by` is the one that answers "does this member have moderation
-authority, and from whom?" — the room owner's grant is room-wide, anyone else's
-is limited to their own invite subtree.
+authority, and from whom?" The room owner's grant covers every member; anyone
+else's covers only the members they invited, directly or indirectly, which is
+often nobody.
 
-Both accept `--format json`, emitting `{room_id, direction, subject, grants[]}`
-where each grant carries `deputizer`, `deputy`, `scope`
-(`room-wide` / `invite-subtree`) and `active`. A grant is inactive when either
-party is no longer in the room, which the contract does not honour. `member
-list` also annotates each member with who deputized them, and `debug room-state`
-reports every grant in the room.
+Both accept `--format json`, emitting `{room_id, direction, subject, grants[]}`.
+Each grant carries `deputizer`, `deputy`, `scope` (`room-wide` /
+`invite-subtree`), `reaches_members` (how many members the grant actually
+covers) and `active`. A grant whose deputizer or deputy has left the room is
+reported with `active: false`; the contract does not honour such a grant.
+`in_room` is `true` for the room owner even though the owner never appears in
+the member list.
+
+`member list --format json` gains `deputies` (who this member deputized) and
+`deputized_by` (who deputized them) alongside the existing `member_id` and
+`nickname`. `debug room-state` gains `deputy_grant_count` and `deputy_grants`.
+
+Two caveats worth knowing:
+
+- The read commands reject an ambiguous partial member ID and list the matches;
+  the write commands (`ban`, `deputize`, `revoke-deputy`) still resolve a
+  partial ID to the first match. Pass the full 8-character ID to be certain.
+- Like every riverctl command that reads a room, these fetch through the
+  standard path, which may PUT migrated state or publish a `member_info` heal
+  for your own identity. They never write deputy state.
 
 ## Command reference
 
--- a/cli/src/commands/debug.rs
+++ b/cli/src/commands/debug.rs
@@ -1,4 +1,5 @@
 use crate::api::ApiClient;
+use crate::deputies::{grant_status_line, party_label, DeputyGrant, RoomDeputies};
 use crate::output::OutputFormat;
 use anyhow::{anyhow, Result};
 use clap::Subcommand;
@@ -53,6 +54,12 @@ struct RoomStateSummary {
     max_members: usize,
     privacy_mode: String,
     configuration_version: u32,
+    /// Number of `deputizer -> deputy` grants in the room (#410). Deputies are
+    /// per-deputizer, so this counts grants, not "deputy members".
+    deputy_grant_count: usize,
+    /// Every grant, read from the CANONICAL `member_info` record per member so
+    /// a revoked grant lingering in a duplicate record is not reported.
+    deputy_grants: Vec<DeputyGrant>,
 }
 
 #[derive(Serialize)]
@@ -277,7 +284,11 @@ pub async fn execute(command: DebugCommands, api: ApiClient, format: OutputForma
         }
         DebugCommands::RoomState { room_owner_key } => {
             let owner_vk = parse_owner_key(&room_owner_key)?;
-            let room_state = api.get_room(&owner_vk, false).await?;
+            let mut room_state = api.get_room(&owner_vk, false).await?;
+            // Private-room nicknames are sealed; collect the local member's
+            // secrets so deputy grants render names rather than ciphertext.
+            let secrets = api.room_display_secrets(&owner_vk, &mut room_state);
+            let deputy_grants = RoomDeputies::new(&room_state, &owner_vk, &secrets).all_grants();
 
             let config = &room_state.configuration.configuration;
             let summary = RoomStateSummary {
@@ -289,6 +300,8 @@ pub async fn execute(command: DebugCommands, api: ApiClient, format: OutputForma
                 max_members: config.max_members,
                 privacy_mode: format!("{:?}", config.privacy_mode),
                 configuration_version: config.configuration_version,
+                deputy_grant_count: deputy_grants.len(),
+                deputy_grants,
             };
 
             match format {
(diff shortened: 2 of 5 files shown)


sanity:
## Multi-model review: consolidated result

Three independent reviewers on commit `7c03a5f0`: an external **Codex** pass (the non-Claude model), a **skeptical/adversarial** Claude pass, and a **code-first** Claude pass (code read before the description). Findings deduplicated below; every one is either fixed in `0c97e77c` or dismissed with a reason.

Nothing blocking was found. Three reviewers independently converged on the same real defect (`member list` reading non-canonical records), which is the one that would have shipped a wrong answer.

### Fixed

| # | Finding | Raised by | Fix |
|---|---------|-----------|-----|
| 1 | **`member list` iterated the raw `member_info` vector.** `verify` accepts duplicate records per member, so a duplicate-carrying state yielded two rows for one member, pairing a *losing* record's nickname with the *canonical* record's deputy annotation. In JSON, two objects with the same `member_id`. | all three | One row per member via `members_with_info()`; nickname read through `canonical`. Pinned by `duplicate_records_collapse_to_one_row_with_the_canonical_nickname`. |
| 2 | **`active: true` over-claimed.** A non-owner deputizer's grant only fires where they are a strict invite ancestor of the target, so a deputizer who invited nobody confers nothing, while the line still read "may ban within X's invite subtree". Most members have invited nobody, so this was the common case. | skeptical | Added `reaches_members` (the deputizer's invite-subtree size; the whole member list for an owner grant). The line now states the count, and says plainly when it reaches no one. Pinned by `reach_is_the_deputizers_invite_subtree_not_merely_that_a_grant_exists`. |
| 3 | **`active`'s doc had the mechanism wrong.** It claimed a non-member cannot be a strict invite ancestor; `is_ban_authorized` inserts the inviter into `strict_ancestors` *before* the membership check, so that was false. And `MemberInfoV1::verify` rejects a record for a non-member, so the state the pruned-deputizer test builds cannot occur. | code-first | Doc rewritten: the load-bearing gate is the deputy-side `banner_is_member` check; the deputizer-side check is defence-in-depth. Test renamed and re-scoped to say so. |
| 4 | **`ResolveError::Ambiguous` was ordered by `MemberId` hash**, which has no relation to the base32 the user sees, so the error listed matches in apparently random order. | code-first | Sorted by printed id, asserted in the ambiguity test. |
| 5 | **`own_member_id` re-inlined the identity derivation** that `api::author_member_id` explicitly says not to re-inline, and did not honour `--signing-key-file`. | code-first | Routes through `Storage::self_identity`, the same derivation `identity whoami` and every send path use. |
| 6 | **Terminal injection via nickname.** Nicknames are attacker-controlled and `unseal_nickname_display` is an unfiltered lossy UTF-8 decode. Cosmetic before; now a forged row would claim moderation authority. | skeptical | Control characters replaced before printing. Pinned by `a_hostile_nickname_cannot_forge_an_extra_output_row`. |
| 7 | **No serde-shape test for the `-f json` contract**, though `debug.rs` already establishes that convention. | code-first | `json_shape_is_pinned_so_scripts_do_not_break_silently` pins the party fields and both kebab-case `scope` values. |
| 8 | **Complexity comment asserted a property the code did not have** (`deputizers_by_deputy` was itself O(N²) via a `canonical` scan per candidate, and `member list` called `deputies_of` again per row). | both Claude passes | Canonical deputy lists resolve once in `RoomDeputies::new`; lookups are now map hits. The comment is accurate. |
| 9 | **Ambiguity test asserted a pigeonhole that does not hold** (10 ids over a 32-symbol alphabet share a first character only ~79% of the time). Deterministic today, a latent flake after any seed change. | skeptical | The test now searches 199 seeds for a genuine collision, which pigeonhole *does* guarantee. |
| 10 | **README `member ban ... # Owner only.`** has been wrong since #410, four lines above the new section that explains deputies can ban. `<member-vk>` was also wrong (it takes a member ID). | skeptical | Both corrected. |
| 11 | Network fetch happened before the local-storage check, so the no-`MEMBER_ID` form paid a full GET before erroring. | code-first | Own identity resolved first. |
| 12 | Explanatory footer went to stdout, unlike every other `member` subcommand's chatter; and neither new command printed the usual progress preamble. | code-first | Footer moved to stderr; preamble added to both. |
| 13 | README/PR did not name the new JSON fields (`deputies`, `deputized_by`, `deputy_grant_count`, `deputy_grants`), and a dangling clause made the inactive-grant sentence misread. | code-first | Both fixed; README now also documents the read/write short-id asymmetry and the fetch-path side effects. |
| 14 | `"may ban room-wide"` overstates: `is_ban_authorized` refuses `target == owner` outright. | code-first | Reworded to "may ban any of the N member(s) in the room". |
| 15 | `suffix.yellow()` colorized an empty string for every member without a deputizer. | skeptical | Only emitted when non-empty. |

### Dismissed, with reasons

- **"Read-only" oversells it: `get_room` runs `ensure_room_migrated` (may PUT) and `heal_member_info` (may publish a `member_info` delta).** (Codex P1, both Claude passes as a nit.) Correct, and now documented in the README. Not changed: this is the behaviour of *every* riverctl command that reads a room, including `member list` and `debug room-state` as they already existed. Giving these two commands a bespoke non-healing fetch path would make them inconsistent with the rest of the CLI for no gain, and rewiring `get_room` is a change to every command, well outside a read feature. The constraint that actually mattered here — no contract, WASM or serialized-state changes — holds: `git diff origin/main --stat` touches only `cli/`.
- **Add the full verifying key to the JSON** so automation has more than 40 bits to key on. Real, but every other riverctl surface (`member list`, `ban`, `deputize`, `revoke-deputy`) both prints and accepts the 8-character short ID. Introducing a second identifier in one command's JSON only would create exactly the inconsistency this feature exists to remove. Worth doing uniformly across the CLI, not here.
- **Converge the write paths' short-id resolution** (`ban` / `deputize` / `revoke-deputy` still take the first match; the new read commands reject ambiguity). Changing a write command from "resolves" to "errors" could break existing scripts and is a behaviour change to commands this PR does not touch. Documented in the README instead.
- **`debug bans` should report whether each ban still enforces** (`BansV1::ban_is_enforcing` is public). Genuinely useful and genuinely adjacent, but it is ban status, not deputy status. Suggested as a follow-up.
- **Repeating the deputizer's name on every row of the forward view.** Kept deliberately. The ambiguity this feature exists to remove is "a deputy of whom?", so every row states it and stands alone when copied or grepped. The repetition is cosmetic.
- **`in_room: false` for the room owner could confuse a script author.** The owner is never in `members.members`, so this is the honest literal reading; `is_owner` disambiguates, `in_room` is `true` for the owner (it ORs in the owner check), and the README now says so.

### Verified clean by more than one reviewer

No panic in either string slice (`MemberId`'s `Display` is always exactly 8 ASCII base32 characters). Reverse-lookup completeness (deputies live only in `MemberInfo.deputies`, so a member with no record provably has none). Output ordering fully deterministic, with no `HashMap` iteration reaching output. No delta construction or state mutation in the deputy module. Private-room nicknames neither leak plaintext nor render ciphertext. `member list` and `debug room-state` JSON changes are purely additive.

Also smoke-tested against the live Official room after the fixes: the invite bot's 5 grants resolve correctly, including one to a member who has since left (reported inactive), and `reaches_members` = 156 against `member_count` = 157, i.e. everyone but the bot itself.

[AI-assisted - Claude]

sanity:
## Review round 2 (re-run on `0c97e77c`, since the code had changed)

Reviews are per code content, so the round-1 result went stale the moment the fixes landed. Re-ran the external **Codex** pass and a fresh **adversarial** Claude pass against the fix commit. Both independently found the same top two, and both were right. Fixed in `90799312`.

### Fixed

**1. The nickname sanitizer did not stop the attack it was added for.** `char::is_control()` is Unicode category **Cc only**. It misses **Cf**, which is where the bidi overrides live: a nickname of `"Bob\u{202e}"` reverses the rendering of the entire rest of the row, including the member id and the `deputy of:` annotation. Zero-width characters (U+200B, U+FEFF) and the Zl/Zp separators passed through too.

Nicknames now render through `str`'s `Debug`, which escapes Cc, Cf and Zl/Zp using the standard library's own printability table rather than a hand-maintained range list that would silently miss a category, and leaves apostrophes alone so ordinary names read normally.

**2. The quoting matters independently of the escaping.** Also raised in round 2: a nickname of *purely printable* characters —

```
Bob (AAAAAAAA)  deputy of: Room Owner
```

— renders as a plausible second row, and `colored` drops the distinguishing colour the moment output is piped, which is exactly how a moderator would read this. Nicknames are now quoted, so the attacker-controlled span has a visible boundary. Output reads `"Ian Clarke" (7XSOGJTK)`.

**3. Escaping in the data model corrupted `-f json`.** Baking the sanitized nickname into `DeputyParty.nickname` meant a bridge relaying nicknames got U+FFFD instead of the real string, a fidelity regression versus the pre-#470 `member list -f json`. Escaping moved to the print sites; JSON carries the faithful value, where JSON's own escaping already makes it safe.

**4. `reaches_members` overstated, in two ways.** Both reviewers flagged it.
- It was populated for **inactive** grants, so a grant naming a departed deputy serialized as `{"active": false, "reaches_members": 7}`. A script filtering `reaches_members > 0` for real moderation authority got a false positive. Now `0` when inactive.
- It ignored **`is_ban_authorized` step 4**: a member cannot be banned by someone they themselves deputized. Those subtree members are now excluded. The owner's grant is untouched by this, correctly: step 3 is absolute and is checked *before* the guardrail.

**5. The `invite_subtrees` doc claimed to mirror a cycle guard the contract does not have.** `get_downstream_members` has no visited set; it relies on `verify` having rejected circular invite chains. The guard here is *additional*, because this reads a fetched state directly and must not hang on a malformed one. Comment corrected so the next reader is not misled about the contract.

**6. `sort_by_key(to_string)` + `dedup()` could miss a duplicate.** The printed id is 40 bits of a 64-bit hash, so two distinct ids can print identically and need not land adjacent under a printed-id-only sort. Sorts by `(printed, id)` now.

**7. Smaller items:** replaced the `unreachable!()` in the `Deputies` arm with an error (it was provably unreachable, but it is a panic path in a user-facing CLI resting on an invariant two statements away); the inactive line names both parties when both have left; `N member(s)` replaced with proper singular/plural.

### Test gap the round-2 review found

`assert!(!label.contains('\u{1b}'), "ANSI escapes must be stripped")` **passed vacuously** — the fixture nickname contained no escape character. The test now exercises four real vectors (newline, ANSI escape, bidi override, zero-width), plus a fully-printable forgery attempt, and asserts the JSON side keeps the faithful value.

25 tests for this feature now, up from 19. New coverage: reach excluding guardrail-blocked targets, inactive grants reaching zero, deep-chain subtree counts, the both-absent inactive line, singular reach.

### Dismissed

**`member list`'s row set is records-driven, so a member in `members.members` with no `member_info` record never appears** — and such a member can hold real deputy authority. Correct, and worth fixing, but it is pre-existing behaviour with no regression here, and changing which members `member list` shows is a user-visible change to an existing command that has nothing to do with deputies. `member deputized-by <ID>` already answers correctly for such a member, since short-id resolution covers the full member set. Better as its own change.

**Sanitization is local to this module, not to riverctl** (`message.rs`, `dm.rs` and several `api.rs` sites still print nicknames raw). True and out of scope; worth an issue.

### Verified clean across both round-2 reviewers

`invite_subtrees` was traced exhaustively against the contract's traversal: deep chains, 2-cycles, 3-cycles, a cycle with a tail, self-invite, and a member listed twice all produce exactly the set `is_ban_authorized`'s ancestor walk implies, and termination is guaranteed. Owner reach correctly equals the member count and correctly does *not* exclude the deputy themselves (step 3 fires before the step-4 guardrail). The precomputation loses nothing relative to the previous on-the-fly version. `members_with_info()` shows every member the old raw iteration showed. `Storage::self_identity` resolves the same identity the old derivation did, including under `--signing-key-file`. No borrow or lifetime hazard from `deputies_of` returning a slice. No panic reachable in production code.

Re-smoke-tested against the live Official room: `member_count` 139, the invite bot's active grants report a reach of 138 (everyone but the bot), and its grant to a departed member reports `active: false`, `reaches_members: 0`.

[AI-assisted - Claude]

sanity:
## Scope note: no version bump in this PR

Dropped the `chore(cli): bump riverctl to 0.2.3` commit that was briefly on this branch. Publishing is being centralised so concurrent agents don't race each other on publishes from main, so this PR is code-only and the version bump plus the `riverctl-v*` tag will be handled separately.

`cli/Cargo.toml` is back at `0.2.2`. Force-pushed `1b6f3001` -> `78d88434`; the only difference between those two trees is the version line in `cli/Cargo.toml` and its `Cargo.lock` counterpart, so every reviewed source file is byte-identical to what the round-2 review ran against. `git diff origin/main --stat` still touches only `cli/`, with no `.wasm` and nothing under `contracts/`.

Recommendation for whoever cuts the release: this is a new feature with no `river-core` change, so `0.2.2` -> `0.2.3` and `river-core` stays at `0.1.17` (the release workflow skips republishing an already-published version).

[AI-assisted - Claude]

sanity:
## Review round 3 (two independent Codex runs on `78d88434`) — one real bug

Both runs independently found the same medium-severity correctness bug, and they were right. Fixed in `76bc4ca3`.

### `reaches_members` undercounted contract-authorized reach

I had approximated the "you cannot ban the member who deputized you" guardrail by hand, excluding every subtree target that listed the deputy. That guardrail is `is_ban_authorized` **step 4**, checked *after* the absolute grants, so the blanket exclusion is wrong in two real cases:

- **The deputy is the target's own invite ancestor** (step 2). Codex's counterexample: `owner -> Alice -> Deputy -> Target`, where Alice deputizes Deputy and Target also deputizes Deputy. The contract authorizes the ban; my code said it did not.
- **The owner has separately appointed the deputy as a global moderator** (step 3), which no target's own grant can block.

It also made the zero-reach line *factually false*: it could claim there was nobody in the subtree the deputy may ban while the contract permitted the ban.

**Fix:** call `MembersV1::is_ban_authorized` per target instead of re-deriving its rules. It is `pub`, so the CLI can use the contract's own predicate rather than a mirror that drifts from it — which is the better shape regardless, since this is the second round in which my summary of those rules turned out to be wrong. The scope it is evaluated over stays the grant's: every member for an owner grant, the deputizer's invite subtree otherwise.

Both counterexamples are now pinned by `the_guardrail_does_not_apply_where_an_absolute_grant_already_does`.

### Also fixed this round

- The room-wide status line needed its own singular form. "may ban any of the 1 member in the room" is not English.
- The `invite_subtrees` doc claimed the owner is excluded from the map. The owner is a `children` key and does get an entry; `reach_of` simply does not consult it for an owner grant. Doc corrected, since the deep-chain test asserts that entry exists.

### Test gaps the round-3 audit found

- The `\r` and `\t` assertions in the hostile-nickname test were passing **vacuously**: no fixture contained either character. It now carries carriage-return and tab fixtures, plus U+2028/U+2029 (Zl/Zp, which are neither Cc nor Cf).
- No regression test existed for the printed-id-collision dedup fix. Added, constructing two `MemberId`s that are genuinely distinct but print identically — the printed form is 8 base32 characters, i.e. only the low 5 of the 8 underlying little-endian bytes — since grinding a real 40-bit collision is not viable in a unit test.
- Added the room-wide singular test.

256 lib tests pass; fmt and clippy clean on the changed files.

### Verified clean by both runs

`display_nickname` via `str`'s `Debug` does escape Cc, Cf (including U+202E) and Zl/Zp while leaving apostrophes alone, and the quoting closes the fully-printable forgery in both `party_label` and the member-list row. Inactive grants consistently report `reaches_members: 0` with no human/JSON contradiction. `invite_subtrees` returns correct strict-descendant sets, bounded at ~20k entries worst case at `max_members: 200`. The `(printed, id)` sort makes `dedup` correct. The `unreachable!()` replacement is safe. Every `grant_status_line` branch is correct.

### Follow-ups filed

- freenet/river#472 — `debug bans` does not show whether a ban still enforces. A ban by a revoked or departed deputy is inert and currently invisible, which is misleading to a moderator, and this PR makes the gap more visible by letting them discover that nobody deputized the banner.
- freenet/river#474 — nicknames are still printed unescaped outside the deputy commands (`message.rs`, `dm.rs`, five sites in `api.rs`).

[AI-assisted - Claude]

Issue #284, opened 2026-05-18
fix(private-rooms): replace alarming "secret v0 not available" placeholder during invite-acceptance sync window


sanity:
## Symptom

User accepts invitation to a private room. For ~5-30 seconds afterwards, every message in the room renders as \`[Encrypted message - secret v0 not available (have: [])]\`. Then the back-fill arrives and messages decrypt.

The placeholder text is alarming — it suggests the room is permanently broken. It's actually a transient sync state.

## Root cause

When the invitee accepts:
1. GET response for the new room contract arrives. State contains messages + member list. **No \`encrypted_secrets\` blob for the invitee's \`member_id\` yet** — the owner's chat-delegate hasn't issued one yet.
2. Invitee's \`repopulate_secrets_from_state\` runs, decrypts 0 secrets. Local \`secrets\` HashMap empty.
3. Renderer falls back to the \"secret vN not available (have: [])\" placeholder.
4. Invitee PUTs join_event. Owner's delegate sees the new member via \`ContractNotification\`, runs back-fill, publishes blob.
5. Update notification reaches the invitee → \`repopulate_secrets_from_state\` runs again → blob decrypted → messages decrypt.

The window between (3) and (5) is what users see. It's a real network round-trip — order of seconds in practice.

## Fix shape

In the renderer (\`ui/src/components/conversation.rs\`, message bubble rendering), when:
- room is private AND
- local \`room_data.secrets\` is empty AND
- the message is private-encrypted

show a friendly placeholder instead:
- First ~30 seconds: \"Decrypting your invitation — this should only take a moment\" with a spinner.
- After 30 seconds with still no secrets: \"Still waiting for the room owner to share keys. Try refreshing if this persists.\"

Optionally also hide the message list entirely behind a centred \"Syncing your invitation…\" placeholder until first decryption succeeds. Less jarring when decryption flips on.

## Logs (user reproduction, build 94fcd8c2)

\`\`\`
get_response.rs:237 GET response: decrypted 0 room secret(s) for member MemberId(QC2UMYUC)
get_response.rs:257 Private room but no secret available for encrypting nickname, using public
\`\`\`

Then several seconds of state updates pass. Eventually the user sends a message at \`secret_version: 2\` — confirming the back-fill landed by then.

## Related

[...]

PR #471, opened 2026-07-25
fix(ui): don't render a reply quote that can't be verified against live state

sanity:
## Problem

When a message quotes (replies to) a message whose author is later **banned**, the banned user's text stays on screen in every reply that quoted it.

`ReplyContentV1` embeds a **snapshot** of the quoted message — `target_author_name` and `target_content_preview` (`common/src/room_state/content.rs:279`). It is written and signed by the *replier* and validated by nothing. The renderer refreshed that snapshot from the live message when it could find it, but fell back to the embedded copy when it could not. An enforced ban removes the member, which makes `ChatRoomStateV1::post_apply_cleanup` purge their messages (step 4b), so after a ban the lookup always fails and the snapshot is always what gets rendered. Ban someone for abuse and the abusive text survives, verbatim and attributed, in every quote of it.

Three neighbouring holes have the same shape and are closed by the same change:

- **Deleted target.** Deleting a message left its text visible in quotes of it.
- **Forged snapshot.** Any member could post a reply naming an arbitrary author and arbitrary "quoted" text, with a `target_message_id` pointing at nothing — rendered as `↩ @Alice: <whatever>`.
- **Action/event target.** A `target_message_id` pointing at a reaction or a join event resolved fine and rendered machine copy (`[Reaction 👍 to …]`, `joined the room`) as though it were the author's words, with a scroll-to-original handler aimed at a DOM id that never exists.

## Approach

Render the quote **only when it can be verified against live room state**, and take *both* halves from that message:

- Target present, quotable and readable → show it, reading the current text (edits propagate) and the author's current nickname via `member_info.canonical()` (renames propagate, a forged `target_author_name` is ignored, and the quote can't disagree with the message header when duplicate `member_info` records exist).
- Otherwise → a neutral, inert placeholder: `↩ Original message unavailable`.

The snapshot is never displayed. It is still read for `target_message_id`, which is what identifies the quoted message.

`ReplyStrip` is an enum, so "unavailable" and "quote" are structurally exclusive and the renderer is an exhaustive match — the `.claude/rules/bug-prevention-patterns.md` "paired `Option` fields that must co-occur" row is exactly this shape.

### Also fixed: riverctl

`riverctl` had the identical hole with no live-state check at all (`cli/src/api.rs`), and its `--format json` / monitor-stream `reply_to` feeds bridges, which would republish a banned member's text. It now resolves the same way (`ReplyContextDisplay`). The human-readable output distinguishes the cases (`[reply to an unavailable message]`); the JSON emits `reply_to: null` for an unverifiable quote rather than a new shape, so no existing consumer can break on it — at the cost of a bridge not being able to tell "not a reply" from "reply we couldn't verify".

### Why this isn't keyed on the ban list

The obvious design — "if the quoted author is in `bans`, show a placeholder" — **cannot be implemented**, and a check written that way would be dead code:

- The snapshot carries the quoted author's **name**, not a `MemberId`.
- `MessageId` is `fast_hash(signature)` (`common/src/room_state/message.rs:906`), so the target id can't yield an author.
- Banning purges the author's `member_info` record too (`common/src/room_state.rs:330`), with no exemption — a banned member has **no nickname anywhere in state** to match the snapshot name against.
- There is no UI-side cache (localStorage, delegate storage, `MESSAGE_HTML_CACHE`, `RoomData`) that maps `MemberId`→name or `MessageId`→author past the purge.

So the only case where "resolve target → author → check `bans`" works is when the quoted message is still present — and a banned author's message never is. Absence is the only observable, so that is what this keys on.

### The trade-off, stated plainly

A target that merely **aged out** of the bounded `recent_messages` window also loses its preview. `max_recent_messages` defaults to **100** and is owner-configurable, so how often this bites is a function of room config. Two consequences:

- The wording is deliberately neutral. Ian asked for something like "Replied message banned"; the UI cannot honestly assert "banned", because the common cause of absence is an ordinary old message. `Original message unavailable` covers every cause without mislabelling any of them. Pinned by `quote_of_aged_out_message_is_also_hidden`.
- A timestamp discriminator (if the oldest retained message is newer than the reply, the target provably aged out) is **rejected on purpose**: `MessageV1::time` is self-signed and unvalidated, so an ally of a banned member could backdate a reply to force that branch and resurrect the text. Recorded in the doc comment so it isn't re-proposed.

## Limitation

This **hides** the text, it does not erase it. The snapshot remains in the contract state, so anyone reading raw state still sees it. Genuinely removing it would mean rewriting other members' signed messages.

## Testing

11 unit tests on `resolve_reply_context` (`ui/src/components/conversation.rs`) plus a CLI regression test:

| Test | Pins |
|---|---|
| `banned_authors_quoted_text_is_not_rendered` | The motivating case. Asserts the quote renders *before* the purge and is gone after, so it can't pass on a dead fixture. |
| `quote_of_aged_out_message_is_also_hidden` | The deliberate trade-off, so it isn't "fixed" by accident. |
| `quote_shows_current_text_and_current_nickname_of_present_target` | Normal reply still works; edits and renames propagate; target id preserved for scroll-to-original. |
| `quote_author_uses_the_canonical_member_info_record` | Duplicate `member_info` records resolve the same way the message header resolves them. |
| `quote_of_deleted_message_is_hidden` | Deleted target. |
| `forged_snapshot_is_ignored_in_favour_of_the_real_target` | Snapshot naming the wrong author / quoting text the target never contained, plus a target id pointing at nothing. |
| `quote_of_an_action_or_event_message_is_hidden` | Reaction and join-event targets. |
| `private_target_is_quoted_only_when_its_secret_is_available` | Private room: decryptable target quotes; rotated-past and cold-start secrets fall through to the placeholder instead of quoting `[Encrypted message - secret vN unavailable]` as if it were the author's words. |
| `quote_of_target_with_unsynced_author_is_attributed_unknown` | `Unknown` attribution rather than the snapshot name. |
| `quote_preview_is_truncated` | 100-char clamp. |
| `plain_message_has_no_reply_strip` | Non-replies unaffected. |
| `quoted_text_is_dropped_once_the_quoted_message_is_gone` (cli) | Same before/after purge assertion for riverctl. |

The helper was extracted from `group_messages` because that function reads the `RECEIVE_TIMES` `GlobalSignal`, which panics outside a Dioxus runtime.

**Render coverage:** the placeholder branch is exercised end-to-end. `ui/src/example_data.rs` now carries a reply whose `target_message_id` is fabricated (reproducing the post-ban end state deterministically), and a new Playwright test asserts the placeholder renders, is inert, is mutually exclusive with the quote strip, and that **neither half of the snapshot reaches the DOM**. The existing a11y test was scoped to `[data-testid="reply-strip"]` so it keeps asserting the interactive strip's ARIA contract; the placeholder carries its own class rather than `reply-strip`, so it doesn't inherit hover/focus-expand rules written for a focusable, ellipsized element.

`cargo test -p river-ui`: 515 passed. `cargo test -p riverctl`: 271 passed. `cargo fmt` and `cargo clippy` clean (remaining warnings are pre-existing on main).

## Review

Full multi-model review: Codex (external) plus three Claude reviewers with distinct lenses (skeptical bug-hunt, code-first, testing). Findings addressed in `review round 1` / `1b`:

- **riverctl had the same hole** (all four reviewers) — fixed here rather than deferred.
- **Undecryptable private target rendered a decryption diagnostic as the quote** while reporting it verified (Codex P2, skeptical I2, testing I1) — text resolution is now fallible.
- **Action/event targets were quotable** (Codex P3, skeptical I5, testing I2) — filtered.
- **Paired `Option` fields let both strips render** (skeptical I3, code-first, testing I3) — replaced with an enum.
- **Non-canonical nickname lookup** (Codex P3, skeptical N8, code-first) — routed through `canonical()`, shared with the message header.
- **Placeholder had zero render coverage** (all four) — example data + Playwright test.
- **Placeholder contrast below WCAG 1.4.3 on self bubbles** (Codex P2) — now matches the quote strip's colours instead of a dimmer `text-white/70`.
- **`.reply-strip` selector collision + inherited focus styling** (skeptical N7, testing B2) — separate class, selectors scoped to the testid.
- **Doc/description inaccuracies** (code-first, skeptical) — `max_recent_messages` is 100 not 200; only *enforced* bans purge; the purge applies to any removed member; the rejected timestamp discriminator is recorded.

## Scope

No `.wasm`, nothing under `contracts/` or `common/`, so the live Official room's contract key is untouched. Rebased onto `main` after #470 merged; the rebase was conflict-free.

[AI-assisted - Claude]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,7 @@ test-contract/
 .playwright-mcp/
 
 # Node modules (Playwright tests, Tailwind build)
-node_modules/
+node_modules
 
 # Playwright test results
 ui/tests/test-results/
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -697,26 +697,54 @@ fn reply_context(
 pub(crate) fn reply_context_display(
     room_state: &ChatRoomStateV1,
     msg: &river_core::room_state::message::AuthorizedMessageV1,
-) -> Option<(String, String)> {
+) -> ReplyContextDisplay {
     reply_context_display_with_secrets(room_state, msg, &HashMap::new())
+}
+
+/// What a message's reply context should render as, resolved against LIVE room
+/// state. Mirrors the UI's `ReplyStrip` (`ui/src/components/conversation.rs`);
+/// the two must stay in step, since they render the same contract data.
+#[derive(Debug, PartialEq)]
+pub(crate) enum ReplyContextDisplay {
+    /// Not a reply.
+    NotAReply,
+    /// A reply whose quoted message could not be read back from room state.
+    Unavailable,
+    /// A quote verified against the message it quotes. Both fields are re-read
+    /// from that message; neither comes from the replier's snapshot.
+    Quote { author: String, preview: String },
 }
 
 /// Like [`reply_context_display`], but able to decrypt the reply context of a
 /// **private** reply when the caller supplies the room's `secrets` map. A
 /// private reply seals its `ReplyContentV1` (target author name + quoted
 /// preview) alongside the reply text, so without the secret the whole reply
-/// context is opaque and no `[reply to …]` prefix could be shown. Public
-/// replies decode directly, exactly as before. Returns `None` for a non-reply,
-/// or when a private reply's secret is unavailable / undecodable.
+/// context is opaque.
+///
+/// The quoted author name and preview stored in `ReplyContentV1` are a SNAPSHOT
+/// written and signed by the REPLIER, validated by nothing, so they are only
+/// trustworthy while the quoted message can be re-read from room state. This
+/// resolves against `recent_messages` and returns [`ReplyContextDisplay::Quote`]
+/// ONLY for a target that is actually there, taking both the text and the
+/// attribution from it.
+///
+/// The case that matters: an ENFORCED ban removes the member, which makes
+/// `post_apply_cleanup` purge their messages, so the snapshot embedded in every
+/// reply that quoted them becomes the last surviving copy of their text.
+/// Rendering it would keep a banned member's abusive message in circulation —
+/// and `riverctl`'s JSON output feeds bridges, which would republish it. See
+/// `resolve_reply_context` in `ui/src/components/conversation.rs` for why this
+/// cannot instead be keyed on the room's ban list, and for the deliberate
+/// trade-off that a merely aged-out target is treated the same way.
 pub(crate) fn reply_context_display_with_secrets(
     room_state: &ChatRoomStateV1,
     msg: &river_core::room_state::message::AuthorizedMessageV1,
     secrets: &HashMap<u32, [u8; 32]>,
-) -> Option<(String, String)> {
+) -> ReplyContextDisplay {
     use river_core::room_state::content::{DecodedContent, ReplyContentV1, CONTENT_TYPE_REPLY};
     use river_core::room_state::message::RoomMessageBody;
     if msg.message.content.content_type() != CONTENT_TYPE_REPLY {
-        return None;
+        return ReplyContextDisplay::NotAReply;
     }
     let reply = match msg.message.content.decode_content() {
         // Public reply — decoded directly.
@@ -730,17 +758,57 @@ pub(crate) fn reply_context_display_with_secrets(
                 ..
             } = &msg.message.content
             else {
-                return None;
+                return ReplyContextDisplay::NotAReply;
             };
-            let secret = secrets.get(secret_version)?;
-            let plaintext =
-                river_core::ecies::decrypt_with_symmetric_key(secret, ciphertext, nonce).ok()?;
-            ReplyContentV1::decode(&plaintext).ok()?
+            let decoded = secrets
+                .get(secret_version)
+                .and_then(|secret| {
+                    river_core::ecies::decrypt_with_symmetric_key(secret, ciphertext, nonce).ok()
+                })
+                .and_then(|plaintext| ReplyContentV1::decode(&plaintext).ok());
+            match decoded {
+                Some(reply) => reply,
+                // We cannot even tell what this reply quotes, so we certainly
+                // cannot verify it.
+                None => return ReplyContextDisplay::Unavailable,
+            }
         }
     };
-    let rendered = render_mentions_for_terminal(room_state, &reply.target_content_preview);
-    let preview = truncate_reply_preview(&rendered);
-    Some((reply.target_author_name, preview))
+
+    let messages = &room_state.recent_messages;
+    let resolved = messages
+        .messages
+        .iter()
+        .find(|m| m.id() == reply.target_message_id)
+        // Soft-deleted messages stay in `messages`, and action / event records
+        // are never rendered as message rows. None of them is quotable.
+        .filter(|target| {
+            !messages.is_deleted(&reply.target_message_id)
+                && !target.message.content.is_action()
+                && !target.message.content.is_event()
+        })
+        .and_then(|target| {
+            // Never fall back to a `<encrypted>` / decode-failure placeholder
+            // here: an undecryptable target has NOT been re-read, so it must be
+            // reported unavailable rather than quoted.
+            let text = messages
+                .effective_text(target)
+                .or_else(|| decrypt_private_body_text(&target.message.content, secrets))?;
+            let author = room_state
+                .member_info
+                .canonical(target.message.author)
+                .map(|info| unseal_nickname_display(&info.member_info.preferred_nickname, secrets))
+                .unwrap_or_else(|| "Unknown".to_string());
+            Some((author, text))
+        });
+
+    match resolved {
+        Some((author, text)) => ReplyContextDisplay::Quote {
+            author,
+            preview: truncate_reply_preview(&render_mentions_for_terminal(room_state, &text)),
+        },
+        None => ReplyContextDisplay::Unavailable,
+    }
 }
 
 /// Rebuild a room's message `actions_state` (edits / deletes / reactions) using
@@ -4259,10 +4327,19 @@ impl ApiClient {
                 // Re-emission of an edited message — distinguish it from a fresh
                 // one for a downstream relay reading the human stream.
                 let edit_prefix = if is_edit { "[edit] " } else { "" };
-                let reply_prefix = reply
-                    .as_ref()
-                    .map(|(author, preview)| format!("[reply to {}: {}] ", author, preview))
-                    .unwrap_or_default();
+                let reply_prefix = match &reply {
+                    ReplyContextDisplay::Quote { author, preview } => {
+                        format!("[reply to {}: {}] ", author, preview)
+                    }
+                    // The quoted message can no longer be read back (its author
+                    // was banned and their messages purged, it was deleted, or
+                    // it aged out), so the replier's snapshot of it is
+                    // unverifiable and is not printed.
+                    ReplyContextDisplay::Unavailable => {
+                        "[reply to an unavailable message] ".to_string()
+                    }
+                    ReplyContextDisplay::NotAReply => String::new(),
+                };
                 let reactions_str = reactions
                     .map(|r| {
                         if r.is_empty() {
@@ -4308,9 +4385,16 @@ impl ApiClient {
 
                 // Reply context (null for non-replies) so a relay can thread the
                 // message; previously absent from the monitor's JSON output.
-                let reply_to = reply
-                    .as_ref()
-                    .map(|(author, preview)| json!({ "author": author, "preview": preview }));
+                // An UNVERIFIABLE quote is emitted as null rather than as a
+                // new shape — see the matching comment on the backfill path in
+                // `commands/message.rs`. A bridge must never receive the
+                // replier-authored snapshot of a banned member's message.
+                let reply_to = match &reply {
+                    ReplyContextDisplay::Quote { author, preview } => {
+                        Some(json!({ "author": author, "preview": preview }))
+                    }
+                    ReplyContextDisplay::Unavailable | ReplyContextDisplay::NotAReply => None,
+                };
 
                 // Output as JSONL (one JSON object per line). `type` is "edit"
                 // for a re-emitted message whose content changed, else "message".
@@ -7036,15 +7120,84 @@ mod display_text_tests {
             nonce,
             0,
         );
-        let (state, msg) = state_with_message(body);
+        let (mut state, msg) = state_with_message(body);
 
-        // Opaque without the secret.
-        assert_eq!(reply_context_display(&state, &msg), None);
+        // Opaque without the secret — we cannot even tell what it quotes.
+        assert_eq!(
+            reply_context_display(&state, &msg),
+            ReplyContextDisplay::Unavailable
+        );
 
-        // Author + quoted preview decrypt with the secret.
+        // The quoted message itself must be present and readable for the quote
+        // to render, so put it in the room, encrypted under the same secret,
+        // and name its author. (`reply` above targets `FastHash(0)`; build the
+        // target so its id matches by re-signing a message and re-pointing the
+        // reply at it.)
+        let alice_sk = SigningKey::from_bytes(&[42u8; 32]);
+        let (t_ct, t_nonce) = river_core::ecies::encrypt_with_symmetric_key(
+            &secret,
+            &river_core::room_state::content::TextContentV1::new("quoted original".to_string())
+                .encode(),
+        );
+        let target = AuthorizedMessageV1::new(
+            MessageV1 {
+                room_owner: msg.message.room_owner,
+                author: MemberId::from(&alice_sk.verifying_key()),
+                content: RoomMessageBody::private(
+                    river_core::room_state::content::CONTENT_TYPE_TEXT,
+                    river_core::room_state::content::TEXT_CONTENT_VERSION,
+                    t_ct,
+                    t_nonce,
+                    0,
+                ),
+                time: SystemTime::UNIX_EPOCH,
+            },
+            &alice_sk,
+        );
+        let target_id = target.id();
+        state.recent_messages.messages.push(target);
+        state
+            .member_info
+            .member_info
+            .push(AuthorizedMemberInfo::new_with_member_key(
+                MemberInfo {
+                    member_id: MemberId::from(&alice_sk.verifying_key()),
+                    version: 0,
+                    preferred_nickname: SealedBytes::public(b"Alice".to_vec()),
+                    deputies: Vec::new(),
+                },
+                &alice_sk,
+            ));
+        // Re-issue the reply now pointing at the real target.
+        let reply = ReplyContentV1::new(
+            "my reply".to_string(),
+            target_id,
+            "Alice".to_string(),
+            "quoted original".to_string(),
+        );
+        let (ciphertext, nonce) =
+            river_core::ecies::encrypt_with_symmetric_key(&secret, &reply.encode());
+        let msg = AuthorizedMessageV1::new(
+            MessageV1 {
+                content: RoomMessageBody::private(
+                    CONTENT_TYPE_REPLY,
+                    REPLY_CONTENT_VERSION,
+                    ciphertext,
+                    nonce,
+                    0,
+                ),
+                ..msg.message.clone()
+            },
+            &SigningKey::from_bytes(&[11u8; 32]),
+        );
+
+        // Author + quoted text decrypt with the secret.
         let secrets = HashMap::from([(0u32, secret)]);
-        let (author, preview) = reply_context_display_with_secrets(&state, &msg, &secrets)
-            .expect("private reply context decrypts");
+        let ReplyContextDisplay::Quote { author, preview } =
+            reply_context_display_with_secrets(&state, &msg, &secrets)
+        else {
+            panic!("private reply context should resolve to a quote");
+        };
         assert_eq!(author, "Alice");
         assert_eq!(preview, "quoted original");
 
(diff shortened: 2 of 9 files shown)


sanity:
## Independent review — verdict: **no blocking findings; merge once CI is green**

Reviewed the **actual checked-out code** at `470b84d3` (matches `refs/pull/471/head`), not the diff text. Three findings below are worth fixing; none of them blocks this landing, and given the vulnerability is being actively exploited I'd ship it and fix #1–#3 in a follow-up rather than hold it.

**Note on review freshness:** the branch moved twice during this review (`21a6ba27` → `470b84d3`). Reviewers initially read uncommitted working-tree changes; those are now committed as the round-2 commit. Every load-bearing claim below was **re-verified against committed `470b84d3`**, not against what the reviewers first saw.

---

### The core security property holds

I traced this myself rather than taking it on trust. `rg 'target_author_name|target_content_preview'` over the whole repo — every production read of the snapshot is accounted for:

- `ui/src/components/conversation.rs:687-688,707-708` — inside `extract_reply_context`. Both production callers discard both halves: `conversation.rs:246` binds `(_snapshot_author, _snapshot_preview, target_id)`, `app/notifications.rs:496` binds `(_, _, target_id)`.
- `cli/src/api.rs:677-678` — reads the snapshot, but `#[cfg(test)]` at `api.rs:667` sits directly above the enclosing `fn reply_context`. Test-only.
- `cli/src/api.rs:3820-3852`, `cli/src/private_room.rs:471-487` — **write** sites (unchanged, necessarily).
- `ui/src/example_data.rs` — feature-gated demo data.

No wholesale-serialization leak either: `ReplyContentV1` only ever exists as opaque CBOR inside `RoomMessageBody`, and `as_text` / `to_display_string` return `reply.text` only. Neither strip half uses `dangerous_inner_html` — both render as Dioxus text nodes (`conversation.rs:3234-3235`), and the preview is reduced to plain text by `strip_markdown`, which drops `Node::Html` and link URLs.

**Output-path coverage is complete.** UI conversation (single render site, exhaustive `match` on `ReplyStrip`), composer strip (built from the rendered message, never a snapshot), notifications (`ReplyContentV1.text` only), document title (no message text), room list (no preview exists), DMs (no reply content type at all), riverctl `message list`, `--format json`, and both monitor paths (one choke point at `output_message`). Truncation is `char`-based on both sides — no byte-boundary panic.

**Verification correctness.** `MessageId` is a 64-bit non-cryptographic `fast_hash` over the signature, so `find(|m| m.id() == target_id)` is hash-matched, not proven-identical. Not exploitable: the preimage is an ed25519 signature the attacker cannot choose, so a targeted collision is ~2^64 grinds, and the failure mode is benign — author and text both come from the *same* resolved message, so a collision yields "real message + its real author", never a mix.

The **regression matrix checks out**: no case that used to render a legitimate quote stops doing so. Scroll-to-original is strictly better than before — the resolvable set is now a subset of rendered `msg-{id}` rows, where previously an action/event target produced a live-looking strip pointing at an element that never existed.

---

### Important (fix, but not blocking)

**1. `quote_of_an_action_or_event_message_is_hidden` passes with the filter deleted — and that filter is load-bearing.**
`ui/src/components/conversation.rs:5091`

The test asserts `Unavailable` for an action and an event target, but never populates `edited_content`. Without it, resolution reaches those targets via `effective_text` → `as_public_string` → `as_text`, which already returns `None` for both. **Delete the `!is_action() && !is_event()` filter at `conversation.rs:271-272` and the test stays green.**

That matters because the `is_event()` half really is doing work. I verified the mechanism:
- `MessagesV1::rebuild_actions_state` builds `message_authors` filtering only actions (`common/src/room_state/message.rs:273-277`), so **event messages are editable**.
- `effective_text` consults `actions_state.edited_content` *first*, before any content-type decode (`message.rs:396-404`) — bypassing the whitelist that otherwise makes this safe.

So a member can post a join event, then an EDIT action targeting it (`actor == original_author` passes — you author your own join event), and populate `edited_content[event_id]` with arbitrary text. With the filter removed that becomes a "verified" quote. Impact of the escape is bounded — the text would be attributed to the attacker themselves, not to a third party — but it's a weakened form of exactly what this PR closes, currently held shut by a filter no test pins.

*Fix:* insert `messages.actions_state.edited_content.insert(target.id(), "attacker-chosen text".into());` before resolving, so the assertion distinguishes filter-present from filter-absent.

*Related:* the doc comment at `conversation.rs:250-254` justifies the filter with `"[Reaction 👍 to …]"` machine copy, but `target_plaintext` never calls `to_display_string()`. The stated rationale is wrong, which is likely why the test exercises the wrong mechanism.

**2. UI/CLI still diverge on a malformed *public* reply — and a comment claims otherwise.**
`ui/src/components/conversation.rs:249-251` vs `cli/src/api.rs:749-762`

Round 2 achieved parity for the *private* undecodable case, but not the public one. For a body with `content_type == CONTENT_TYPE_REPLY` and garbage `data` (nothing validates the payload, so this is postable):
- UI → `ReplyStrip::Unavailable` ("Original message unavailable")
- CLI → `decode_content()` returns `None` → `_` arm → the `let RoomMessageBody::Private {..} … else` returns **`NotAReply`** → no prefix, `reply_to: null`

Neither leaks attacker content, so impact is cosmetic. But the new comment at `conversation.rs:249-251` — *"so the two clients agree (riverctl reports the same)"* — is false for this case, in a security-critical code path whose whole thesis is UI/CLI parity. Either align the CLI branch or scope the comment.

**3. The new emit helpers have no tests.**
`cli/src/api.rs:861` (`reply_prefix_display`), `cli/src/api.rs:882` (`reply_to_json`)

Extracting these into shared helpers was the right call and fixes the duplication risk. But nothing asserts `Unavailable → "[reply to an unavailable message] "` or `Unavailable → null`. This is precisely the bridge-facing surface the PR names as the reason the CLI fix matters. They're now pure functions — cheap to test directly.

---

### Minor

**4. Scope sentence in the description is now inaccurate.** The PR says *"nothing under `contracts/` or `common/`"*, but `common/tests/ban_purges_authored_content_test.rs` (195 lines) is in the diff. It is **test-only** — no contract source, no `.wasm` — so the scope *conclusion* (the live room's contract key is untouched) still holds. I verified: `git diff origin/main --name-only` shows no `.wasm`, nothing under `contracts/`, nothing under `common/src/`. Just correct the sentence. The test itself is a good addition — it's the tripwire for the purge premise the whole design rests on.

**5. Preview rendering parity is overclaimed.** `cli/src/api.rs:704-706` says the two "must stay in step". The *resolution* logic now genuinely mirrors; the *rendering* does not and never did: UI truncates at 100 chars with **no marker** (`conversation.rs:293-296`), CLI at 50 chars **with `...`** (`truncate_reply_preview`); UI strips markdown, CLI doesn't. Pre-existing, but scope the claim to resolution. Silent truncation with no cut indicator is a mild misrepresentation vector even on genuine text.

**6. Description's test inventory is stale** — it lists 11 UI + 1 CLI; the tree has 12 UI + 3 CLI, and the quoted `cargo test` counts were measured on a tree that no longer exists.

**7. Aged-out trade-off is understated.** The description calls it "a function of room config". It's structural: since the reply must be inside the window while its target is older, in **any** room sitting at `max_recent_messages` the oldest replies structurally lose their quotes. The design call is still right — absence is the only observable, and the timestamp discriminator is correctly rejected as attacker-controlled (`MessageV1::time` is self-signed) — but say "guaranteed for any reply whose target predates the window."

---

### Notes — pre-existing, explicitly *not* this PR's job

**8. Nickname impersonation is untouched, and it is the actual reported incident.** Nothing constrains nicknames to be unique, so an attacker can still set `preferred_nickname = "Ian Clarke"` and have every surface say so. **What this PR fixes is the durability half**: once the impersonator is banned and purged, the forged quote now disappears instead of surviving forever as the last copy. Worth stating in the description, because "fixes the forged-quote-attributed-to-Ian problem" will otherwise read as fixing impersonation, which it does not.

**9. Message signatures are not verified on the delta path.** `MessagesV1::apply_delta` (`common/src/room_state/message.rs:103+`) validates privacy mode, size, dedup and author *membership*, but never calls `message.validate(vk)`. The only signature check is `MessagesV1::verify`, reachable via `validate_state`. This is **pre-existing and affects `main` identically — and worse**: without this PR a forged message renders as a normal message attributed to the victim. So it is not a regression here and does not defeat the fix, which strictly reduces surface. But it does mean "verified against live state" is conditional on an invariant River doesn't enforce in its own code. Worth a separate issue; a per-rendered-reply `validate()` would make the guarantee self-contained.

---

### Test quality: real, not vacuous

I specifically hunted the vacuous-fixture pattern flagged in a recent sibling PR. **It is not present in the headline tests.**

- `banned_authors_quoted_text_is_not_rendered` (`conversation.rs:4824`): the fixture genuinely carries `ABUSE` in `target_content_preview`. The before-half uses `expect_quote`, which **panics** on a non-`Quote` variant — a hard positive, so it cannot pass on a dead fixture. The after-half is a whole-value `assert_eq!(after, ReplyStrip::Unavailable)`, not a weak `!contains`.
- The CLI counterpart (`api.rs:8403`) is better still — its fixture fills the snapshot with `"SNAPSHOT AUTHOR MUST NOT RENDER"` / `"SNAPSHOT PREVIEW MUST NOT RENDER"`, so a regression fails loudly rather than silently matching real text.
- The Playwright fixture is non-vacuous: `example_data.rs` carries a reply with a fabricated `target_message_id` and `"snapshot text that must never be rendered"`, which renders verbatim on `origin/main`. Positive assertions (`toBeVisible`, `toContainText`) run **before** the negatives, so the negatives can't pass on an unrendered page — no flake surface.
- Round 2 fixed the `toHaveCount(1)`-on-`.first()` tautology by counting on the unfiltered locator (`message-layout.spec.ts:598-602`). Good.

No tests were removed (+11 UI, +1 CLI, 0 deletions). Three CLI mention tests were rewritten to give their fixtures a real target — necessary fixture repair, not weakening.

Remaining gaps beyond #1 and #3: no multi-byte/exact-boundary truncation case on the UI side (CLI already has one), and the non-self placeholder contrast branch lost Playwright coverage when the example reply became owner-authored.

---

### Process disclosure (per the multi-model review rule)

The external non-Claude pass **could not run**: `codex exec --sandbox read-only` failed with `You've hit your usage limit … try again at Jul 28th, 2026 12:14 PM`. Gemini CLI is unavailable org-wide (deprecated free tier). Per the documented fallback I substituted **three independent Claude reviewers with distinct adversarial lenses**, each blind to the others:

1. **Bypass** — can any attacker-controlled string still reach screen/stdout/JSON (snapshot trace, attribution laundering, target-id resolution, markdown/bidi injection, untouched render paths)
2. **Completeness + verification correctness** — code-first, every output path enumerated, UI/CLI equivalence walked
3. **Test quality** — vacuous-fixture hunt, would-it-fail-on-main for every test

I then verified every load-bearing claim myself against committed `470b84d3` rather than relying on the subagent reports — including the two findings above that the reports got partly wrong in severity.

---

### Merge readiness

- **Scope constraint verified clean.** No `.wasm`, nothing under `contracts/`, nothing under `common/src/`. The live Official room's contract key is untouched.
- **CI is PENDING** on `470b84d3` (`build`, `ui-playwright-tests`). Must be green on this exact SHA before merge.
- **PR is still a DRAFT** — mark ready.
- **Branch is 1 commit behind `main`** (#476, riverctl 0.2.3 bump). `git diff origin/main -- cli/Cargo.toml` misleadingly shows a 0.2.3→0.2.2 downgrade; this is a staleness artifact, **not** a change this PR makes — the branch never touches that file, so the merge preserves 0.2.3. No conflict, no action needed.

**Verdict: merge as-is once CI is green.** Findings #1–#3 are real and worth a follow-up, but none of them leaves attacker-controlled content on screen, and the fix closes the exploited hole correctly. My confidence in the core property (the replier's snapshot never renders on any production path) is high — I traced it exhaustively myself.

[AI-assisted - Claude]

sanity:
## Review update — **verdict changed: do NOT merge `470b84d3`**

Follow-up to my review above. Two things came out of a deeper pass, and one of them reverses my merge recommendation.

I no longer stand behind "merge as-is once CI is green" **for `470b84d3`**. The reason is a UX regression on a common private-room path, described below. The round-3 work currently uncommitted in the worktree already fixes it.

---

### 1. Regression at `470b84d3`: cold start flashes the placeholder over every private reply

`ui/src/components/conversation.rs:247-257` (at `470b84d3`):

```rust
return if content.content_type() == CONTENT_TYPE_REPLY {
    ReplyStrip::Unavailable
} else {
    ReplyStrip::None
};
```

`content_type` is cleartext on the `Private` variant, so this fires for **every** private reply whenever `secrets` lacks the reply's version — including when the map is simply empty. `RoomData.secrets` is `#[serde(skip)]` (`ui/src/room_data.rs:108-124`), so an established private room begins every cold start holding nothing, and a fresh joiner holds nothing until the delegate `encrypted_secrets` back-fill lands.

Old behaviour: `extract_reply_context` returned `(None, None, None)` and **no strip rendered**. New behaviour at this SHA: "Original message unavailable" renders directly above a body that simultaneously says "Decrypting messages — this should only take a moment...". Two contradictory messages, and the alarming one asserts the quoted message is gone when it is merely still sealed. That is the alarm class #284/#286 were fixed to remove.

To be precise about the window: this is **not** a same-tick race. `merge` + `repopulate_secrets_from_state` are atomic inside one `ROOMS.with_mut` (`response_handler.rs:1449-1466`), and the delta path likewise. It is the genuinely-empty case — a joiner awaiting back-fill, or the documented `merge`-returns-`Err` path that skips repopulation "until a clean reload" (`ui/src/room_data.rs:1743-1757`).

The test at this SHA pinned the wrong behaviour: `private_target_is_quoted_only_when_its_secret_is_available` looped `HashMap::new()` and asserted `Unavailable`.

**The uncommitted round-3 work fixes this correctly** via a `pending_decryption(content, secrets)` helper (`is_private && secrets.is_empty()`) routing cold start to `ReplyStrip::None`, with the test updated to match. I traced the predicate against the full matrix: it agrees exactly with `decrypt_message_content`'s own `secrets.is_empty()` discriminator, so body copy and strip stay consistent, and a rotated-past version (non-empty secrets, missing version) still correctly yields `Unavailable`. Round 3 also closes the malformed-public-reply divergence (my finding #2) and adds `pending_decryption` to the CLI.

---

### 2. Mutation testing: I stopped reading the tests and started reverting the code

Prompted by the nine bad tests found on the sibling PR #470, I did not trust a reading-based assessment. I cloned the tree to scratch at `470b84d3`, reverted each production guard one at a time, and ran the suite. **This did not touch the branch.**

| Mutation | Caught by |
|---|---|
| UI: core fix reverted — restore the snapshot fallback | **6 tests** ✅ |
| CLI: core fix reverted — restore the snapshot fallback | **2 tests** ✅ |
| UI: `is_deleted()` check disabled | **1 test** ✅ |
| UI: delete the `is_action()` + `is_event()` filter | **0 tests** ❌ |
| UI: `canonical()` → first-match `.find()` lookup | **0 tests** ❌ |

**The headline security property is genuinely well pinned.** Reverting the actual vulnerability — falling back to the replier's snapshot — fails 6 UI tests and 2 CLI tests. That is the single most important thing to know about this PR's tests, and it is now verified by execution rather than by reading. The #470 failure mode does **not** affect the core fix.

Two secondary guards are unpinned:

**`quote_of_an_action_or_event_message_is_hidden` (`conversation.rs:5091`)** — delete the filter at `conversation.rs:271-272` and all 12 tests still pass. The test never populates `edited_content`, so resolution fails for an unrelated reason (`as_text` already returns `None` for those variants). The filter is load-bearing: `message_authors` filters only actions (`common/src/room_state/message.rs:273-277`), so events **are** editable, and `effective_text:396-404` returns edited content before any content-type decode. Fix: insert `edited_content` for the event target before resolving.

**`quote_author_uses_the_canonical_member_info_record` (`conversation.rs:4951`)** — replace `canonical()` with a first-match `.find()` and all 12 tests still pass. This one is worth reading closely, because it is the exact shape #470 produced. The fixture is `info(vec![canonical, stale])`, and its own comment says:

> Listed with the STALE one last so a naive `.collect()` into a map would pick it.

That ordering does defeat a last-wins `HashMap` collect. But `canonical()` is `max_by_key` over `(version, signature)` (`member_info.rs:39-44`), and a first-wins `.find()` on this fixture returns the canonical record by luck of ordering. So the test pins one regression shape and sails past its sibling. Fix: assert with both orderings, or add a third record so no single position is correct.

Neither gap is a live vulnerability — both guards are present and correct in the code. They are regression risk: a future refactor can remove either one with CI green.

---

### Revised verdict

- **`470b84d3` (current PR head): do not merge.** The cold-start placeholder regression is user-visible on every private-room load, and it is already fixed in the working tree.
- **Round 3 looks right** on the two things I traced (`pending_decryption` predicate, CLI parity). Its restructuring of `extract_reply_context` into `extract_reply_target_id` is a real improvement: the snapshot stops being *returned* at all, so it becomes unreachable by construction rather than by the discipline of every caller binding `_`.
- **Commit and push round 3, then this needs a re-review on the new SHA** — per the per-code-content rule, and because round 3 touches `notifications.rs` and `build.yml`, which no review has covered.
- The two mutation-surviving tests should be strengthened in the same push. They are cheap fixes and both are one-liners.

Standing corrections to my review above: finding #2 (UI/CLI malformed-public divergence) is fixed by round 3, and my `cli/Cargo.toml` note remains a staleness artifact rather than a real change.

**Process note:** the external non-Claude pass remains unavailable (Codex usage limit, resets Jul 28; Gemini deprecated org-wide). Lenses run for this update: private-path regression, plus empirical mutation testing. This is the documented external-unavailable fallback, and the mutation results are execution-verified rather than read-and-inferred.

[AI-assisted - Claude]

sanity:
Heads-up for anyone reading this after the fact: a later review round found a regression in what merged here, fixed in #480.

`RoomData.secrets` is `#[serde(skip)]`, so every cold start of an established **private** room transiently holds no secrets until `repopulate_secrets_from_state` rehydrates them. This PR made an undecodable reply report "unavailable" rather than rendering no strip — so in that window every reply in the room renders an italic `↩ Original message unavailable` on every page load before flipping to the real quote. That is the alarming-placeholder class #284 was filed to remove, and the wording is wrong there: the original message is fine, we just have not decrypted it yet.

#480 distinguishes "still decrypting" (no secrets at all → render nothing) from "genuinely unreadable" (rotated past this version, purged, deleted, forged → placeholder), and also fixes:

- riverctl reported `NotAReply` where the UI reported `Unavailable` for a malformed public reply, contradicting the parity these comments assert.
- The action/event target filter was not actually pinned — both clients' tests passed with the filter deleted. They now give the target an edit, which is the real attack (an event is editable by its author, and `effective_text` reads `edited_content` before any content-type decode).
- `common/tests/ban_purges_authored_content_test.rs` was added here as a tripwire but CI runs only specific `--test` targets, so it was never executed or even compiled.

**If the River UI is republished from main before #480 lands, private-room users will see the flash.** Not data loss, but visible.

[AI-assisted - Claude]

PR #473, opened 2026-07-25
feat(ui): deputy shield on message authors, and no emoji in nicknames

sanity:
## Problem

Two related gaps in the conversation view.

**Moderators are invisible.** The 🛡 deputy shield exists in the member list and
the member-info modal, but not where moderation actually happens: next to the
name on a message. You cannot tell whether the person telling you to calm down
has the authority to ban you.

**A nickname can fake the badge.** A nickname is attacker-controlled bytes from
the member's own signed `MemberInfoV1.preferred_nickname`. Set it to `Alice 🛡`
and you render a shield River never granted; the same works for `👑` (room
owner) and `⭐` (you). Adding the badge to the conversation without closing this
would ship an impersonation surface.

## One correction to the spec, up front

The brief for this work was, verbatim:

> a deputy shield indicates "this user has been deputized WHICH ALLOWS THEM TO
> BAN ME", unless I'm also a deputy in which case they can't ban me but I can
> still see that they're a deputy.

**Being a deputy does not make you immune.** `MembersV1::is_ban_authorized`
grants owner-appointed moderators *absolute* authority at step 3, which is
checked *before* the step-4 "cannot ban the member who deputized you" guardrail
(`common/src/room_state/member.rs:299-313`). So two owner-appointed moderators
can ban each other, and a moderator looking at a peer moderator's message will
see **"Deputy (appointed by the room owner). Can ban you."** — the tooltip the
premise says should not appear.

This PR implements what the contract actually does and pins it
(`deputy_viewer_still_sees_the_shield_on_a_fellow_deputy`). The visibility half
of the ask is implemented exactly as stated: **the badge never disappears
because the viewer is immune.** Only the wording reflects reality.

If deputies *should* be immune to each other, that is a change to
`is_ban_authorized` — a contract change, which re-keys every live room, and
deliberately out of scope here. Say the word and it can be a separate PR.

### The badge predicate

The shield means the author holds deputy authority that **bears on the viewer**.

That predicate already existed. `relevant_deputizer_names` +
`viewer_relevant_deputizer_set` (added for #410/#451) are viewer-relative: a
member shows the shield when one of the people who deputized them is a **strict
invite-ancestor of the viewer**, **the room owner** (a global moderator,
relevant to everyone), or **the viewer themselves** (a deputy you appointed).
It never consults the viewer's own deputy status, so "still shows when I'm
immune" holds by construction.

Note this is *not* `deputies_of(self_member_id)` — that is `target_is_my_deputy`,
which drives the Deputize/Undeputize button and is the opposite ("who I
deputized") direction.

This PR wraps it in a shared `DeputyBadge` / `deputy_badges_for_viewer` used by
**all three** surfaces, so visibility *and* wording cannot drift the way #451
drifted:

| | |
|---|---|
| **Visibility** | non-empty `relevant_deputizer_names` (unchanged rule) |
| **Tooltip** | `Deputy (appointed by the room owner). Can ban you.` |

The ban clause is the real contract answer from `is_ban_authorized`, and it
accounts for **cascading bans**: a ban sweeps the target's whole invite subtree,
so someone who can ban your inviter removes you too. Asking only about a direct
ban would have told a user they were safe from someone who can delete them at
will.

The clause is dropped entirely on your own row ("can they ban you" is
meaningless about yourself, and `is_ban_authorized` answers `true` for a global
moderator asked about themselves).

Decisions the brief asked to be explicit about:

- **Deputy of a subtree outside the viewer's ancestry → no badge.** Their
  authority cannot reach the viewer. They still show it inside their own
  subtree's views.
- **The room owner never gets a shield.** Their authority is inherent, not
  delegated. This is enforced, not incidental: nothing in `MemberInfoV1::verify`
  stops a member listing the owner in their own `deputies`.
- **Self-deputisation grants nothing.** Also contract-legal, also now ignored —
  otherwise anyone in the viewer's invite chain could render a *real* shield on
  their own messages, which is the impersonation this PR exists to prevent, just
  achieved with a real badge instead of an emoji.
- **Self messages have no author line**, so your own badge appears only in the
  sidebar and the modal.

### Emoji in nicknames

Render-time stripping is the security boundary: `riverctl` writes `member_info`
straight to the contract and never runs UI validation. New
`ui/src/util/display_name.rs` owns one `display_nickname(sealed, secrets)` choke
point and **every** nickname render path goes through it — twelve inlined call
sites collapsed into one, covering message author lines, the member list, the
modal, the nickname field, reply-strip authors, mention chips (both renderers),
both @-autocomplete lists, reaction tooltips, DM thread and rail, notifications,
and the deputy tooltip's appointer names.

Three of those are not looked-up nicknames at all — they are name-shaped
strings the SENDER writes, spoofable with no `member_info` involved:

- `ReplyContentV1.target_author_name` (the `↩ @name:` in a reply strip), on both
  the public and the encrypted-room decode paths
- an unresolved mention's `[name]` snapshot in `@[Admin 🛡](rv:…)`
- the tooltip's appointer names

The reply strip additionally now prefers the **real** author of the quoted
message when it is still in the buffer, instead of the sender's claim about who
they were quoting.

**What is stripped:** emoji and pictographic blocks; invisible and blank
characters (zero-width, bidi overrides, soft hyphen, Hangul fillers, Braille
blank, the Default_Ignorable tail) — a nickname made only of those renders blank
while being non-empty, which gives a message a nameless header that reads as a
continuation of the group above it; private-use-area codepoints (Nerd Fonts map
PUA to icon glyphs including shields, and `U+E000`/`U+E001` are River's own
mention-substitution sentinels); and spaces that render identically to `U+0020`.

**What is not:** anything that could be part of a real name. CJK, Arabic,
Hebrew-with-niqqud, Cyrillic, Greek, Devanagari, Thai, Armenian, Georgian,
Ethiopic, Cherokee, accented and Vietnamese Latin, CJK punctuation, `U+3000`
(the Japanese name separator) and ordinary punctuation all pass through
byte-identical.

`U+200C` ZWNJ and `U+200D` ZWJ are the interesting case. They look like emoji
machinery, but they are **orthography**: Persian compounds (`علی‌رضا`), Sinhala
touching letters (`සූර්‍ය`), Malayalam chillu. Blanket-stripping them mangled
those names *and*, because the same predicate gates three inputs, locked those
users out of creating a room or accepting an invitation while telling them their
name contained emoji. They are now kept where they can be doing orthographic
work — between two non-ASCII letters, or trailing one — and dropped elsewhere,
which covers both the orphans the emoji strip leaves behind and the
`"Bo\u{200D}b"` clone of `"Bob"`.

**Input-time rejection is UX, not security.** The three nickname inputs show
"Nicknames can't contain emoji", mark the field `aria-invalid`, and refuse to
submit.

### Stated honestly

- This **hides** emoji, it does not prevent storage. A CLI user can still write
  `Alice 🛡` into the contract; the **River UI** never renders it — but
  `riverctl member list` and `riverctl message stream -f json` still print the
  raw value, so a bot consuming riverctl output sees the unsanitised name.
  Contract-level enforcement was rejected because it re-keys every live room.
- **Homoglyphs are not addressed and cannot be** without breaking the non-Latin
  names above. A Cyrillic `а` is a legitimate letter.
- Sanitising can still **create** a rendered-name collision (a joiner inside a
  CJK name, or a homoglyph). The cheap vectors are closed; the rest is the
  homoglyph problem. `duplicate_candidate_names` in the mention picker compares
  name strings, so it will not flag such a pair.
- Combining marks are preserved, so a "Zalgo" nickname can still be ugly. It
  cannot forge a badge.

### Changes to existing surfaces, called out

- The member-list and modal shield tooltips now carry the same
  `. Can/Cannot ban you.` clause (single-definition tooltip *is* the anti-drift
  mechanism).
- `BanButton` was passing a raw `SealedBytes` into a `String` prop, which Dioxus
  converts via `Display`: the ban dialog showed an unsanitised nickname, and
  `[Encrypted: N bytes, vN]` instead of a name in private rooms. Now fixed.
- Focusing and leaving your own nickname field used to republish `member_info`
  at version + 1 with no edit. Now guarded.
- Notification bodies flatten newlines out of the message text (a leading
  newline let an author paint a second line shaped like a notification from
  someone else).

## Testing

`cargo test -p river-ui --bins --all-features` — **537 passed**.
Playwright chromium, full suite — **86 passed, 6 skipped, 0 failed**. (One
pre-existing flake in `responsive-layout.spec.ts`'s sandboxed-iframe tests,
reproduced on another branch's dev server too — filed as #479.)

Highlights of the new coverage:

- `deputy_badge_is_viewer_relative_and_survives_viewer_immunity` — one room
  covering every case: deputy-of-owner → badge + "can ban you"; deputy of the
  viewer's inviter → badge; deputy of an unrelated subtree → no badge; **a
  deputy the viewer appointed → badge STILL shows, tooltip says "cannot ban
  you"**; no authority → no badge; owner → no badge.
- `can_ban_viewer_accounts_for_cascading_bans` — asserts the precondition (the
  direct ban really is denied) before asserting the tooltip, so it cannot pass
  for the wrong reason.
- `self_deputisation_grants_no_badge`, `owner_never_carries_a_deputy_badge`,
  `single_target_badge_matches_the_sweep`.
- `joiners_between_letters_are_preserved` and
  `joiners_are_dropped_where_they_only_clone` — the two halves of the ZWNJ/ZWJ
  rule.
- `real_names_in_other_scripts_are_untouched` — 24 names across 15 scripts, the
  guard against a rule that mangles real people's names.
- `nicknames_that_render_blank_become_unnamed`,
  `space_lookalikes_are_normalised_but_ideographic_space_is_not`,
  `reply_context_author_name_is_sanitised_on_both_paths`,
  `unresolved_mention_chip_snapshot_is_sanitised`.

**Regression gates:**

- `nickname_render_paths_go_through_display_nickname` walks the whole `ui/src`
  tree and fails on three shapes: the inlined unseal, `to_string_lossy` on the
  field, and a `SealedBytes` reaching a name-shaped component prop. That third
  check exists because the first two missed the live `BanButton` bug. Verified
  to fire by planting each violation.
- Source-grep pins in `conversation.rs` and `member_info_modal.rs` that both
  surfaces take visibility *and* tooltip from the shared helper.
- Example data gives the room owner a deliberately spoofed `… (Owner) 🛡👑`
  nickname, so `member-info-deputy-tag.spec.ts`'s existing `toHaveCount(1)`
  becomes a live regression gate for the strip, and gives the deputy one
  guaranteed message so the badge is deterministically present.

## Scope

UI-only — no `contracts/`, no `common/`, no serialized types, no WASM. The live
Official room's contract key is untouched.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/notifications.rs
+++ b/ui/src/components/app/notifications.rs
@@ -664,10 +664,10 @@ pub fn notify_new_messages(
     let sender_name = member_info
         .canonical(msg.message.author)
         .map(|ami| {
-            match unseal_bytes_with_secrets(&ami.member_info.preferred_nickname, room_secrets) {
-                Ok(bytes) => String::from_utf8_lossy(&bytes).to_string(),
-                Err(_) => ami.member_info.preferred_nickname.to_string_lossy(),
-            }
+            crate::util::display_name::display_nickname(
+                &ami.member_info.preferred_nickname,
+                room_secrets,
+            )
         })
         .unwrap_or_else(|| "Someone".to_string());
 
--- a/ui/src/components/app/notifications.rs
+++ b/ui/src/components/app/notifications.rs
@@ -373,7 +373,16 @@ pub fn show_notification(
     // Gateway iframe: hand the notification to the shell (real origin) over the
     // postMessage bridge — the sandboxed iframe can't use the Notifications API.
     if is_in_shell_iframe() {
-        let body = format!("{}: {}", sender_name, message_preview);
+        // Flatten line breaks out of the message text. The body renders as
+        // `sender: message`, so a message starting with a newline lets its author
+        // paint a second line that reads exactly like a notification from someone
+        // else — including a moderator with a 🛡 in it. The sender name itself is
+        // already sanitised.
+        let body = format!(
+            "{}: {}",
+            sender_name,
+            message_preview.replace(['\n', '\r'], " ")
+        );
         post_notification_to_shell(room_key, room_name, &body);
         return;
     }
@@ -415,7 +424,16 @@ fn create_notification_internal(
     sender_name: &str,
     message_preview: &str,
 ) {
-    let body = format!("{}: {}", sender_name, message_preview);
+    // Flatten line breaks out of the message text. The body renders as
+    // `sender: message`, so a message starting with a newline lets its author
+    // paint a second line that reads exactly like a notification from someone
+    // else — including a moderator with a 🛡 in it. The sender name itself is
+    // already sanitised.
+    let body = format!(
+        "{}: {}",
+        sender_name,
+        message_preview.replace(['\n', '\r'], " ")
+    );
     info!(
         "Creating notification - title: '{}', body: '{}'",
         room_name, body
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -2923,18 +2923,18 @@ fn MessageGroupComponent(
                             },
                             "{group.author_name}"
                         }
-                        // Deputy shield. Same glyph and `member-icon` styling
-                        // as the member-list row and the member-info modal, so
-                        // the three surfaces read as one badge. The nickname
-                        // above cannot forge it: nicknames are stripped of
-                        // emoji by `crate::util::display_name`.
+                        // Deputy shield. Same glyph, same visibility rule and
+                        // the same tooltip as the member-list row and the
+                        // member-info modal chip, so the three read as one
+                        // badge. The nickname above cannot forge it: nicknames
+                        // are stripped of emoji by `crate::util::display_name`.
                         if let Some(badge) = group.author_badge.as_ref() {
                             {
                                 let tooltip = badge.tooltip();
                                 rsx! {
                                     span {
                                         "data-testid": "message-author-deputy-badge",
-                                        class: "member-icon text-sm cursor-default",
+                                        class: "text-sm cursor-default",
                                         title: "{tooltip}",
                                         "aria-label": "{tooltip}",
                                         "🛡"
@@ -3853,6 +3853,90 @@ mod tests {
         for spoof in ["Mallory 🛡", "Mallory 👑", "Mallory \u{1F6E1}\u{FE0F}"] {
             assert_eq!(sanitize_display_name(spoof), "Mallory");
         }
+    }
+
+    /// A mention token carries a `[name]` SNAPSHOT written by the sender. When
+    /// the member reference doesn't resolve, that snapshot is what renders —
+    /// so `@[Admin 🛡](rv:<nobody>)` would paint a shield inside a message with
+    /// no nickname involved at all. Any member can type this into any message,
+    /// which makes it a broader vector than the nickname one.
+    #[test]
+    fn unresolved_mention_chip_snapshot_is_sanitised() {
+        let unknown = MemberId(freenet_scaffold::util::FastHash(0xDEAD_BEEF));
+        let me = MemberId(freenet_scaffold::util::FastHash(1));
+        let token = river_core::mention::encode_mention(unknown, "Admin 🛡");
+
+        let html = message_to_html_with_mentions(&token, &HashMap::new(), me);
+        assert!(
+            !html.contains('\u{1F6E1}'),
+            "an unresolved mention rendered a badge glyph from its snapshot: {html}"
+        );
+        assert!(
+            html.contains("Admin"),
+            "the name itself must survive: {html}"
+        );
+    }
+
+    /// Same snapshot, the other renderer: the quoted reply preview resolves
+    /// mention tokens to plain `@name` and falls back to the sender-written
+    /// snapshot for an unknown member.
+    #[test]
+    fn reply_preview_sanitises_an_unresolved_mention_snapshot() {
+        let unknown = MemberId(freenet_scaffold::util::FastHash(0xDEAD_BEEF));
+        let token = river_core::mention::encode_mention(unknown, "Admin 🛡");
+
+        let preview = clean_reply_preview(&token, &HashMap::new());
+        assert!(
+            !preview.contains('\u{1F6E1}'),
+            "reply preview rendered a badge glyph from a mention snapshot: {preview}"
+        );
+        assert!(preview.contains("@Admin"), "got: {preview}");
+    }
+
+    /// `ReplyContentV1.target_author_name` is a name the REPLYING client wrote
+    /// into the message body, never checked against the target message's real
+    /// author. It renders as `↩ @name:` directly under the real author line and
+    /// its badge, so it must be sanitised on BOTH the public and the private
+    /// (encrypted-room) decode paths. Sanitising only one of them is exactly
+    /// the bug this test exists to prevent.
+    #[test]
+    fn reply_context_author_name_is_sanitised_on_both_paths() {
+        use river_core::room_state::content::ReplyContentV1;
+
+        let target_id = MessageId(freenet_scaffold::util::FastHash(7));
+        let spoof = "Admin \u{1F6E1}";
+
+        // Public room.
+        let public = RoomMessageBody::reply(
+            "sure".to_string(),
+            target_id.clone(),
+            spoof.to_string(),
+            "the original".to_string(),
+        );
+        let (author, _, _) = extract_reply_context(&public, &HashMap::new());
+        assert_eq!(author.as_deref(), Some("Admin"), "public reply path");
+
+        // Private room: same payload, encrypted with a known secret.
+        let secret = [42u8; 32];
+        let plaintext = ReplyContentV1::new(
+            "sure".to_string(),
+            target_id,
+            spoof.to_string(),
+            "the original".to_string(),
+        )
+        .encode();
+        let (ciphertext, nonce) =
+            crate::util::ecies::encrypt_with_symmetric_key(&secret, &plaintext);
+        let private = RoomMessageBody::private(
+            river_core::room_state::content::CONTENT_TYPE_REPLY,
+            river_core::room_state::content::REPLY_CONTENT_VERSION,
+            ciphertext,
+            nonce,
+            3,
+        );
+        let secrets: HashMap<u32, [u8; 32]> = HashMap::from([(3u32, secret)]);
+        let (author, _, _) = extract_reply_context(&private, &secrets);
+        assert_eq!(author.as_deref(), Some("Admin"), "private reply path");
     }
 
     /// Issue #315 — pin that the markdown renderer never passes raw HTML
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -3911,51 +3911,13 @@ mod tests {
         assert!(preview.contains("@Admin"), "got: {preview}");
     }
 
-    /// `ReplyContentV1.target_author_name` is a name the REPLYING client wrote
-    /// into the message body, never checked against the target message's real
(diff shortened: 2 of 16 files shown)


sanity:
## Multi-model review

**External model pass could not run.** `codex exec` fails with `You've hit your usage limit … try again at Jul 28th, 2026 12:14 PM`, and the Gemini CLI is unavailable on this machine (deprecated individual tier, every call returns `IneligibleTierError`). Waiting three days for the Codex reset was not practical, so this used the **Claude-subagent fallback**: six independent reviewers across two rounds, each with a distinct adversarial lens and each blind to the others' findings. Recording the substitution so the weaker signal is auditable.

**Round 1** (on `0c6b061a`): badge-spoofing, code-first/predicate-correctness, test-coverage.
**Round 2** (on `40673731`, after the round-1 fixes): fix-verification, a fresh impersonation lens, big-picture/goal-alignment.

Every finding is fixed or answered below. Nothing was waved through.

### Blocking, fixed

| Finding | Fix |
|---|---|
| `extract_reply_context` sanitised only the **public** decode branch, so every reply in a **private** room could still forge a shield one line under the real author name. | Restructured to decode-then-return-once so a one-branch fix is impossible, and pinned with a test that encrypts a real payload and checks both paths. |
| `U+200C` ZWNJ / `U+200D` ZWJ are orthography in Persian, Sinhala and Malayalam, not just emoji machinery. Stripping them mangled real names **and** — since the same predicate gates three inputs — locked those users out of creating a room or accepting an invitation, while telling them their name contained emoji. | Kept between non-ASCII letters (and trailing one, for Malayalam chillu), dropped elsewhere. `joiners_between_letters_are_preserved` fails on the old rule. |
| A nickname of `room owner` forged an owner-level appointment in the badge tooltip: role labels and attacker-chosen nicknames shared one flat comma-joined list, and `room owner` is plain ASCII so the sanitiser cannot touch it. | Real nicknames are quoted; labels read `the room owner` / `you`. Pinned. |

### Should-fix, fixed

- **`BanButton` received a raw `SealedBytes`** into a `String` prop, which the Dioxus props derive silently converts via `Display`: the ban dialog showed an unsanitised nickname, and `[Encrypted: N bytes, vN]` instead of a name in private rooms. This is the one bug in the batch that **no test would have caught**, so the render-path scan gained a third check for exactly that shape, verified by planting a violation.
- **"Cannot ban you" could lie.** A ban cascades to the target's invite subtree, so someone who can ban your inviter removes you too. The clause now asks about the whole ancestor set. The test asserts the precondition (the direct ban really is denied) before asserting the tooltip.
- **Focusing and leaving your own nickname field republished `member_info` network-wide** at version + 1 with no edit, silently rewriting a `riverctl`-set nickname. Guarded.
- **Self-deputisation and deputising the owner** are contract-legal (`MemberInfoV1::verify` bounds only list length and signature) and would have rendered a *real* shield — the same impersonation, achieved without an emoji. Both now render nothing.
- **A nickname of only Default_Ignorable characters** rendered blank while being non-empty, skipping the `UNNAMED` placeholder and giving a message a nameless header that reads as a continuation of the group above it. Stripped, along with the legacy-computing blocks and text-presentation badge glyphs (`U+06DE` and friends).
- **Pixel-identical name clones** via a kept joiner (`Bo<ZWJ>b`) or NBSP. Closed for the ASCII cases; the residual (a joiner inside a CJK name) is documented in the module header alongside homoglyphs.
- **The reply strip showed the sender's claim** of who they replied to even when the real message was in the buffer, so a genuine quote could be reattributed to a moderator. It now prefers the real author.
- **Notification bodies** flatten newlines out of the message text.
- The pin test **sliced by bytes** and would have panicked on a multi-byte character near a match; it takes chars now. Its doc also told a future developer to move code below `#[cfg(test)]` — the exact thing that silently disarmed its first draft, which hid 2,800 lines of `conversation.rs` including a real offender. Removed.
- Modal uses a **single-target** badge lookup (it is not memoised, and the sweep decrypts every deputy's appointer nicknames per render).
- Dead is-empty filters on the @-autocomplete, create-room submit now disabled rather than silently no-op, `aria-label` on all three badge surfaces, em-dash removed from user-facing copy.

### Raised and answered, not fixed here

- **`target_content_preview` is not sanitised.** It is message content, where emoji are legitimate and the message body itself renders them unsanitised. The strongest form of the attack — attaching a real quote to a moderator's name — is closed by taking the author from the real target message. For a target that has scrolled out of the buffer the snapshot is all there is; that is inherent to the design and is now stated in the code.
- **Room names and descriptions are unsanitised.** Same `SealedBytes` surface, but they are owner-signed and no badge sits next to them. Out of scope.
- **`riverctl member deputized-by` does not filter self-deputisation** (`cli/src/deputies.rs`, from #470), so the CLI and the UI now disagree about what a deputy is. Not touched here because this PR is UI-only by constraint — flagging for a follow-up.
- **`cli/README.md`** should note that the UI strips emoji at render time, so a `riverctl member set-nickname "Alice 🛡"` user is not surprised. Same reason: not touched here.
- An **8-char base32 `MemberRef` collision** in mention resolution (~2⁴⁰ work) is pre-existing and outside this diff.

### Pre-existing flake found while running the suite

`responsive-layout.spec.ts`'s two sandboxed-iframe tests flake under parallel workers — reproduced on a *different* branch's dev server too, so not from this change. Filed as #479 rather than fixed here.

[AI-assisted - Claude]

sanity:
## Rebased twice; conflict resolutions reviewed

Main moved under this branch while the review was running. Rebased onto `6fabe8a0`, which pulled in #471 and #480 — both of which rewrite the reply-preview code this branch had touched. Conflict resolution is an unreviewed judgment call, so a fourth reviewer went over the merged result specifically. Two things it found, both fixed:

**The rebase silently disarmed this PR's own badge pin.** `author_deputy_badge_uses_the_shared_helper` cut the file at `rfind("#[cfg(test)]")`. #471 appended a *second* test module, so the cut moved to that module and swallowed the pin's own assertion literals — every needle then matched itself, and the test would have passed with the badge deleted outright. It now cuts at `mod tests` via a needle containing an escaped newline, which cannot self-match. This is the same failure mode documented on the render-path scan, arriving by a different route.

**The live quote-author path had no sanitisation coverage.** Taking #471's `resolve_reply_strip` meant this branch's test of `extract_reply_context`'s author field was testing a value both callers discard. Added `quote_author_label_is_sanitised`, which asserts the rendered quote attributes to `Alice` when the member's nickname is `Alice 🛡`.

**Where main's version won outright:**

- #471's `resolve_reply_strip` is a strict superset of this branch's reply-author fix. This branch preferred the real author when the target was in the buffer; #471 discards the sender-written snapshot entirely and renders a quote only when it can be re-read from live state. It also distinguishes "undecodable reply" from "not a reply", filters soft-deleted/action/event targets, and avoids quoting a decryption diagnostic as the author's words — three cases this branch's version got wrong.
- #480 then removed `target_author_name` from the decoder's return type altogether, which is stronger than sanitising it: there is no longer a value that could reach the screen. This branch's snapshot-sanitisation test was deleted as superseded, with a comment recording why.
- #471's `resolve_member_nickname` (canonical-record routing, which fixes a real duplicate-record hazard) was kept, with its body switched to `display_nickname` so it also sanitises. That is now the single helper both the author line and the quote label use.

Verified on the merged tree: 549 unit tests pass, `cargo fmt` and `clippy` clean, the render-path scan finds 0 offenders across all 60 files in `ui/src`, no test from either side was dropped, and the diff is still UI-only. Playwright chromium: 88 passed, 6 skipped, 0 failed; the three new specs also pass on firefox, webkit and mobile-chrome.

Waiting on CI for `0690c789` before this is mergeable.

[AI-assisted - Claude]

Issue #474, opened 2026-07-25
riverctl: member nicknames are printed to the terminal unescaped outside the deputy commands


sanity:
## Problem

`unseal_nickname_display` (`cli/src/api.rs`) is `String::from_utf8_lossy` with no filtering, and its output reaches the terminal verbatim from several riverctl surfaces. A nickname is set by the member themselves, so it is attacker-controlled.

Unescaped, a nickname can:

- **start a new output line** with an embedded `\n` or `\r`, forging a row that looks like real CLI output;
- **rewrite the rest of the line** with an ANSI escape (`\u{1b}[…`), erasing or repainting what follows;
- **reverse the rendering of the rest of the line** with a bidi override, U+202E. This is the Trojan Source trick, and it is not caught by `char::is_control()`, which only covers Unicode category Cc while U+202E is Cf;
- **hide content** with zero-width characters (U+200B, U+FEFF) or break lines on the Zl/Zp separators (U+2028, U+2029).

`colored` disables colour whenever stdout is not a TTY, so any styling that might have distinguished a forged row disappears exactly when output is piped, redirected or logged, which is how these commands are usually consumed.

## Where

freenet/river#470 fixed this for the deputy surfaces by routing nicknames through `deputies::display_nickname` (quote + escape via `str`'s `Debug`, which uses the standard library's own printability table and so covers Cc, Cf and Zl/Zp). Everywhere else still prints raw:

- `cli/src/commands/message.rs:236` and `:317` (message author names)
- `cli/src/commands/dm.rs:600` (DM peer name)
- `cli/src/api.rs:3678`, `:4115`, `:4179`, `:4251`, `:4298`

## Why it is worth fixing beyond the deputy commands

On its own a forged message-author line is a spoofing nuisance. It becomes more than that wherever CLI output is consumed by something that acts on it: a bridge, a log scraper, or a moderator eyeballing `message list` to decide who to ban. Escaping is cheap and the surface is small.

## Suggested fix

Promote `display_nickname` (or an equivalent) to a shared helper and use it at each print site. Two things to preserve from #470's version:

- **Escape at the print site, not in the data model.** `-f json` must keep the faithful nickname; JSON's own escaping already makes it safe, and a bridge relaying nicknames needs the real string.
- **Quote as well as escape.** Escaping alone does not stop a fully printable forgery such as a nickname of `Bob (AAAAAAAA)  deputy of: Room Owner`. The quotes are what give the attacker-controlled span a visible boundary.

[...]

Issue #472, opened 2026-07-25
riverctl: `debug bans` does not show whether a ban still enforces


sanity:
## Problem

`riverctl debug bans <room>` prints every ban with `banned_user_id`, `banned_by_id` and `banned_at_secs`, but gives no indication of whether the contract still **enforces** that ban.

Since deputy ban authority (#410), a ban can be present in state and completely inert. `BansV1` computes enforcement separately from storage: `MembersV1::check_banned_members` only collects a ban's target (and their downstream subtree) when `MembersV1::is_ban_authorized(ban.banned_by, ban.ban.banned_user, …)` returns true at the time state is applied. That predicate is re-evaluated against current state, so a ban stops enforcing when any of these later become true:

- the banner's deputy authority was **revoked** (their deputizer removed them from `deputies`);
- the banner was **pruned or left**, so `banner_is_member` is false and the deputy-derived branches (steps 3 and 5) no longer grant;
- the banner's deputizer left, so the banner is no longer a deputy of a genuine invite ancestor of the target;
- the target has since deputized the banner, tripping the step-4 guardrail (and no absolute grant applies).

A moderator reading `debug bans` sees the ban listed and reasonably concludes the member is banned. They may well be back in the room.

## Why it matters more now

freenet/river#470 adds `member deputized-by`, so a moderator can now discover that nobody has deputized member X. What they still cannot see is the consequence: every ban X issued is dead. The two halves of the picture are split across commands, and only one of them exists.

## Suggested fix

`BansV1::ban_is_enforcing` is already `pub` (`common/src/room_state/ban.rs`). Add an `enforcing: bool` to `BanInfo` in `cli/src/commands/debug.rs`, populate it from that predicate, and surface it in both the human and `-f json` branches. Mark the non-enforcing ones clearly in the human output, since that is the case a moderator needs to notice.

Worth considering alongside: a short reason for a non-enforcing ban (banner absent / authority revoked / guardrail), which is the information that makes it actionable.

## Notes

Found while reviewing #470 (deputy-status read commands). Deliberately left out of that PR: it is ban status rather than deputy status, and it changes an existing command's output shape.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-07-29T16:13:11.509964452+00:00
Issue hash: `ad920727`

## Value Assessment

[...]

PR #477, opened 2026-07-25
fix(cli): a deputy no longer counts themselves in their own reach

sanity:
## Problem

Follow-up to #470, found by a post-merge review pass. Two independent reviewers flagged it, one rating it high.

`MembersV1::is_ban_authorized` has no `banner == target` check, and self-ban is genuinely contract-permitted: a deputy whose deputizer is their own invite ancestor satisfies step 5 against themselves. So `reach_of` faithfully counted the deputy in their own reach.

Faithful to the contract, wrong as a user-facing number:

- It makes `reaches_members` answer the wrong question. A script filtering `reaches_members > 0` for real moderation authority gets a false positive from a deputy who can only "ban" themselves — which is precisely what the field's own doc says it exists to prevent.
- It produces a **false sentence**. In a room of `{owner, deputy}` the only counted target was the deputy, and the output read *"may ban the only other member in the room"* when there is no other member.

On the live Official room the effect was off-by-one: 106 reported where 105 is correct (107 members, minus the Invite Bot itself, minus the deputy).

**The test suite had locked the wrong number in.** Five expectations encoded it, which is why #470's three review rounds did not catch it — the tests agreed with the bug.

## Approach

One-line behavioural fix: exclude the deputy from their own reach. Everything else in the diff is correcting expectations, correcting a doc, and repairing tests that were weaker than they looked.

The `reaches_members` doc still described the hand-approximated guardrail that #470's round 3 **deleted**, so it contradicted both `reach_of`'s own doc and the tests, and would have walked the next maintainer straight back into that bug. Rewritten to state what the field actually measures — capability within the grant's scope, not attribution of that capability to this grant specifically — and to warn against reintroducing a local re-derivation of the contract's rules.

The room-wide status line now covers zero, one and many other members. Zero became reachable once the self-count went away.

## Test strengthening

The same review audited every test in the file and found four that were weaker than they looked. Each fix below is **verified** by reverting the production code and confirming the test fails:

| Test | Was | Now |
|---|---|---|
| canonical-record | pushed the revoke **before** the grant, so a bare first-match scan — the exact regression the module doc names — passed it | order swapped |
| dedup collision | inlined its own copy of the sort, so reverting the fix left it green | drives a signed deputies list through `RoomDeputies` (3 vs 2 with the fix reverted) |
| `an_inactive_grant_reaches_nobody` | reported zero with or without the `if active` gate | uses the case where the gate is load-bearing: a departed deputizer who still has a subtree via a child's `invited_by` (1 vs 0) |
| cycle guard | only covered a cycle the root was part of, which `*child != root` alone handles | replaced with the case the visited set actually covers — duplicate member entries giving one member two parents |

On that last one: the review proposed a cycle *disjoint* from the root as the hang case. That case cannot arise. `invited_by` is a function, so every cycle member's parent is inside the cycle and nothing outside can reach it. The visited set's real job is deduplication when the members vector carries two entries for one id with different `invited_by`, so that is what the test now builds.

New coverage: `member list --format json`'s shape (extracted into a testable helper, matching the pattern `debug.rs` already uses for `contract_get_summary_json`), the third `inactive:` arm, the room-wide plural line, and a sealed private-room nickname decrypting through `party`.

258 lib tests pass; `cargo fmt --check` and clippy clean on the changed files.

## Scope

CLI-only, and read-only within that. `git diff origin/main --stat` touches `cli/src/deputies.rs` and `cli/src/commands/member.rs`. No `.wasm`, nothing under `contracts/`, no version bump.

## Verified live

Against the Official room (`4uNUKFzZQCnzo4K2ecZ16cMsYEEfoaRS35z6exEsbvm4`), 107 members:

```
Deputies appointed by "Invite Bot" (PMAQEUP5) (5 grants):

  (unknown) (3YBS76GA)  inactive: (unknown) (3YBS76GA) is not currently in this room
  "Ian Clarke" (7XSOGJTK)  active: may ban 105 members in "Invite Bot" (PMAQEUP5)'s invite subtree
  ...
```

105 = 107 members, minus the bot (not in its own subtree), minus the deputy.

## Review note

The external (non-Claude) pass could not run: Codex is out of quota until 2026-07-28. Per the review policy's external-unavailable fallback I used diverse Claude subagents with distinct lenses instead — contract-parity and test-validity — which is what found this. I will run the same fallback on this PR.

Follows #470. Related: #472, #474.

[AI-assisted - Claude]

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -200,16 +200,16 @@ often nobody.
 
 Both accept `--format json`, emitting `{room_id, direction, subject, grants[]}`.
 Each grant carries `deputizer`, `deputy`, `scope` (`room-wide` /
-`invite-subtree`), `active`, and `reaches_members`.
+`invite-subtree`), `active`, and `members_deputy_can_ban`.
 
-`reaches_members` is how many **other** members the grant can currently be used
-against. Each candidate is decided by the contract's own authorization rule
+`members_deputy_can_ban` is how many **other** members this deputy can ban
+within the grant's scope. Each candidate is decided by the contract's own authorization rule
 rather than by a summary of it: `0` for an inactive grant; every other member
 for an owner grant; otherwise the members of the deputizer's invite subtree the
 contract authorizes this deputy to ban. The deputy is never counted against
 themselves — banning yourself is permitted but is not moderation authority. It
 is often `0` even for a live grant, because most members have invited nobody.
-`reaches_members > 0` is the check for "this member actually has moderation
+`members_deputy_can_ban > 0` is the check for "this member actually has moderation
 authority here".
 
 The number counts capability, not attribution: a member this deputy could also
--- a/cli/README.md
+++ b/cli/README.md
@@ -202,13 +202,20 @@ Both accept `--format json`, emitting `{room_id, direction, subject, grants[]}`.
 Each grant carries `deputizer`, `deputy`, `scope` (`room-wide` /
 `invite-subtree`), `active`, and `reaches_members`.
 
-`reaches_members` is how many members the grant can currently be used against,
-computed from the contract's own rules: `0` for an inactive grant, every member
-for an owner grant, otherwise the deputizer's invite subtree minus anyone in it
-who has themselves deputized this deputy (a member cannot be banned by someone
-they deputized). It is often `0` even for a live grant, because most members
-have invited nobody. `reaches_members > 0` is the check for "this member
-actually has moderation authority here".
+`reaches_members` is how many **other** members the grant can currently be used
+against. Each candidate is decided by the contract's own authorization rule
+rather than by a summary of it: `0` for an inactive grant; every other member
+for an owner grant; otherwise the members of the deputizer's invite subtree the
+contract authorizes this deputy to ban. The deputy is never counted against
+themselves — banning yourself is permitted but is not moderation authority. It
+is often `0` even for a live grant, because most members have invited nobody.
+`reaches_members > 0` is the check for "this member actually has moderation
+authority here".
+
+The number counts capability, not attribution: a member this deputy could also
+ban by some other route still counts, so revoking one grant may not change it.
+It also counts direct targets, while a ban cascades to the target's own invite
+subtree, so it understates how many members a ban would actually remove.
 
 A grant whose deputizer or deputy has left the room is reported with
 `active: false`; the contract does not honour such a grant. `in_room` is `true`
--- a/cli/src/commands/member.rs
+++ b/cli/src/commands/member.rs
@@ -1,6 +1,6 @@
 use crate::api::ApiClient;
 use crate::deputies::{
-    display_nickname, grant_status_line, party_label, ResolveError, RoomDeputies,
+    display_nickname, grant_status_line, party_label, DeputyParty, ResolveError, RoomDeputies,
 };
 use crate::output::OutputFormat;
 use anyhow::{anyhow, Result};
@@ -155,20 +155,7 @@ pub async fn execute(command: MemberCommands, api: ApiClient, format: OutputForm
                     let json_members: Vec<_> = members
                         .into_iter()
                         .map(|(party, own_deputies, granted_by)| {
-                            serde_json::json!({
-                                "member_id": party.member_id,
-                                // Unchanged shape: the pre-#470 output rendered
-                                // a member with no readable nickname as the
-                                // lossy placeholder string, never null, and
-                                // `party.nickname` is only None for a member
-                                // with no member_info record, which cannot
-                                // appear in this listing.
-                                "nickname": party.nickname.unwrap_or_default(),
-                                // Members this member has deputized.
-                                "deputies": ids_to_strings(&own_deputies),
-                                // Members who have deputized this member.
-                                "deputized_by": ids_to_strings(&granted_by),
-                            })
+                            member_list_json(&party, &own_deputies, &granted_by)
                         })
                         .collect();
                     println!("{}", serde_json::to_string_pretty(&json_members)?);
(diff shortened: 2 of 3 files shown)


sanity:
## Review of this PR — and I was wrong about one thing

External pass unavailable (Codex out of quota until 2026-07-28), so this is the policy's Claude-subagent fallback: two lenses, adversarial bug-hunt and semantics/user-facing correctness, each blind to the other. Both converged on the same three findings, and one of them **disproved a claim in my own PR description**. Fixed in the latest push, then rebased onto current `main` (clean, no conflicts).

### I was wrong: a root-disjoint invite cycle CAN arise

My PR body argued the reviewer's proposed hang case was impossible:

> `invited_by` is a function, so every cycle member's parent is inside the cycle and nothing outside can reach it.

`invited_by` is only a function while each member id appears once in `members.members`, and nothing in `MembersV1::verify` enforces that — as this PR's own fixture demonstrated by deliberately giving one member two parents. Two entries for `alice` produce:

```
owner -> alice,  alice -> bob,  bob -> alice   (second alice entry)
```

The walk from `owner` enters the alice/bob cycle, `*child != root` does not help (neither is the root), and without `seen.insert` gating the push it spins forever.

Worse, the replacement test I wrote **could not fail**: its fixture was acyclic, and `seen` being a `HashSet` makes double-counting impossible regardless of the guard. So I shipped exactly the vacuous-test failure mode this PR exists to fix, while claiming in the description that I had verified each repaired test against reverted code. That verification was real for three of the four and not for this one.

Now fixed: the test builds the cyclic state and runs `invite_subtrees` on a worker thread with a deadline, so a regression **fails** rather than hanging the test binary. Verified both ways — 10s timeout failure with the guard removed, 0.01s pass with it.

### Two user-facing strings the reach change left saying something false

Both flagged by the semantics lens.

- **The room-wide plural line** read `"may ban any of the 2 members in the room"` in a three-member room. `debug room-state` prints the real member count two lines above it, so the contradiction was visible in a single screen with nothing explaining the missing member. Now `"2 other members in the room"`.
- **The invite-subtree zero-reach line** said there is `"nobody in X's invite subtree that this deputy may ban"`. False when the deputy is themselves inside that subtree, since the contract does authorize the self-ban — the zero is our deliberate exclusion, not an absence of authority. Now `"nobody else"`. The room-wide zero arm was already safe because it said "other".

### The README carried the stale text this PR was written to kill

`cli/README.md` still described `reaches_members` as "every member for an owner grant" (no longer true) and "minus anyone in it who has themselves deputized this deputy" — the hand-approximated step-4 guardrail that #470 round 3 deleted and that `the_guardrail_does_not_apply_where_an_absolute_grant_already_does` exists to disprove. Fixing the internal rustdoc while leaving the identical wording in the file users actually read was the same mistake one level out. Rewritten.

### Doc honesty

The `reaches_members` doc claimed the answer "cannot drift from what the contract enforces". That overstates it: there are two deliberate *narrowing* deviations — the self-exclusion, and the inactive-grant gate that skips the predicate entirely. Both are now named, with the argument that the self-exclusion hides nothing (a self-ban cascade only reaches the deputy's descendants, who are already counted individually via step 2). Also documented that the number is capability rather than attribution, so revoking *this* grant may not change it, and that it counts direct targets while a ban cascades to the target's own subtree.

### Filed from this review: freenet/river#478

The semantics lens found the same underlying fact on a far more dangerous surface. `ui/src/components/members/member_info_modal.rs` renders `BanButton` under `if !is_owner` with **no** `member_id != self_member_id` guard, while `DeputyButton` four lines below *does* guard self. `viewer_can_ban` is a thin wrapper over `is_ban_authorized` with no self-guard either. So a deputy opening their own Member Info sees an enabled "Ban User" button, and the resulting ban is contract-valid and cascades through `get_downstream_members` to their entire invite subtree.

I verified both call sites directly. The CLI bug cost one wrong integer; that one costs a moderator their membership and everyone they invited.

### Verified clean by both lenses

The one-line production fix is correct and complete: the deputy is the only party needing exclusion (the deputizer is kept out of their own subtree by `*child != root`; the owner is absent from `members.members` and refused outright by `is_ban_authorized`). Every changed count was hand-traced against the contract predicate, including the two untouched tests that could have shifted. The new `RoomWide` reach-0 arm is reachable, correctly worded, and shadows nothing (guard-arm order 0, 1, default). `member_list_json` is output-identical to the inline version it replaced. The private-room test's `seal_bytes` + secrets-map shape matches what `room_display_secrets` actually produces.

258 lib tests pass; fmt and clippy clean.

[AI-assisted - Claude]

sanity:
## Review provenance (external pass unavailable)

Recording this so the weaker-signal decision is auditable rather than invisible.

**The external, non-Claude pass could not run.** Codex is out of usage quota until **2026-07-28** — a real invocation, not an assumption: `codex exec` returned `You've hit your usage limit ... try again at Jul 28th, 2026 12:14 PM`. Gemini has been unavailable since 2026-06-24 (`IneligibleTierError` on the deprecated individual free tier), so there was no second external model to fall back to. Waiting three days for quota was not proportionate for a fix that gates an unpublished release.

**Substituted:** the policy's Claude-subagent fallback, using **Opus** for every reviewer, with three distinct adversarial lenses, each blind to the others' findings:

| Lens | What it was pointed at | What it found |
|---|---|---|
| **Adversarial bug hunt** | panics, off-by-one, match-arm ordering, fixture validity | the disjoint-cycle counterexample that disproved my PR-body claim; the README staleness |
| **Semantics / user-facing correctness** | is every printed sentence true for the state that produces it | the room-wide plural line asserting a wrong room size; the zero-reach line being false when the deputy is in the subtree; the UI self-ban button, now filed as #478 |
| **Big picture / goal alignment** | does the change serve the goal, CI-chasing, stale docs, is the field worth keeping | the `reaches_members` naming problem and the window closing at the tag; `reach_of`'s doc still claiming pure delegation; `api.rs`'s existing self-ban refusal as the better justification |

Each finding is fixed in `744ad167` or answered above. Three lenses on the earlier rounds converged independently on the same defect three separate times, which is the argument for the multi-reviewer shape over a single pass — but it is a weaker signal than a genuine external model, and this note exists so nobody later mistakes it for one.

## Two corrections to my own PR description

The description is what survives in the history, so:

**1. "That case cannot arise" was wrong.** I wrote that a root-disjoint invite cycle is impossible because `invited_by` is a function. It is only a function while each member id appears once in `members.members`, and nothing in `MembersV1::verify` enforces that — as this PR's own fixture demonstrates by giving one member two parents. The reviewer's original counterexample was correct. I verified it by simulation before rewriting the test: with the visited guard the walk terminates at 2; without it, it never terminates.

**2. "Replaced with the case the visited set actually covers" understates it.** The original root-in-cycle case was **kept**; the disjoint-cycle case was **added** alongside it. Nothing was removed.

## The distinction worth stating plainly

The old number was **faithful to `MembersV1::is_ban_authorized`** and still **wrong as a user-facing number**. `is_ban_authorized(P, P)` genuinely returns true for most deputies — step 3 (owner-appointed) or step 5 (a strict non-owner ancestor deputized them) both fire against self, and the ancestor walk seeds `visited` with the target so step 2 never does.

So the bug was never in the predicate. It was in what the number *means*: a deputy who can only remove themselves holds no moderation authority, and a moderator reading "can ban 106 members" to decide whether to revoke a grant is misled by the one that is the deputy. The fix deliberately reports something the contract predicate does not, and the field doc now names both places it does so (this exclusion, and the inactive-grant gate) rather than claiming parity it does not have.

That is also why the rename matters: `reaches_members` reads as a claim about the grant, and the value is a property of the deputy within a scope.

[AI-assisted - Claude]

PR #480, opened 2026-07-25
fix(ui,cli): don't flash "Original message unavailable" while a private room decrypts

sanity:
## Problem

Follow-up to #471, which merged before this round of review landed. Three findings, one of them a regression #471 introduced.

### 1. Private-room cold start flashes the placeholder over every reply (regression)

#471 made an undecodable reply report "unavailable" rather than rendering no strip, so the UI and riverctl would agree on the same input. That is wrong for one important case.

`RoomData.secrets` is `#[serde(skip)]`, so **every** cold start of an established private room transiently holds no secrets until `repopulate_secrets_from_state` rehydrates them from the encrypted blobs — the comment at `ui/src/components/conversation.rs` says so explicitly, and it is why #284 replaced the alarming body placeholder with a calm "Decrypting messages…". In that window every reply is undecodable, so after #471 every reply in the room renders an italic `↩ Original message unavailable` on every page load before flipping to the real quote.

That is the exact alarm class #284 was filed to remove, and the wording is actively wrong there: the original message is fine, we just have not decrypted it yet. Before #471 no strip rendered at all in that window.

**Fix:** distinguish "still decrypting" from "genuinely unreadable". `pending_decryption` is true only when the body is `Private` **and** we hold *no secrets at all* — deliberately not "this version is missing", because a rotated-past version really is unavailable and must keep showing the placeholder. The pending case renders no strip, matching both the pre-#471 behaviour and what the message body itself does.

The parity argument that motivated #471's change does not apply here: riverctl collects secrets synchronously before any output, so it has no transient window to be in parity with. riverctl gets the same rule anyway, for the case where it holds no local storage for a room.

### 2. riverctl and the UI diverged on a malformed public reply

A `Public` body with `content_type == 3` and undecodable CBOR — postable by any member, since the contract treats content as opaque bytes — got `Unavailable` from the UI and `NotAReply` from riverctl, contradicting the parity the code comments assert. No text leaked either way. Fixed by restructuring the decode so both the private-undecryptable and malformed-public cases flow through one path.

### 3. The action/event target filter was not actually pinned

Both clients' tests passed **with the filter deleted** — resolution already failed inside the text lookup because an action/event body has no text, so the filter was never the thing under test.

That matters because the filter is load-bearing for a real reason, and not the one the comment gave. An **event is editable by its author** (`rebuild_actions_state_with_decrypted` excludes only `is_action()` from `message_authors`), and `effective_text` consults `edited_content` *before* any content-type decode. So a member could post a join event, edit it to arbitrary text, and quote it — surfacing text that is rendered nowhere else in the app, since events render through `format_event_summary`.

Both tests now give the target an edit, which is the actual attack. **Verified by deleting the filter in each client and confirming the corresponding test fails.**

## Also in this round

- **CI gap.** `common/tests/ban_purges_authored_content_test.rs` was added by #471 as the tripwire for the contract behaviour the whole fix rests on — but `build.yml` runs only specific `--test` targets, so it was never executed *or even compiled*. Added as its own step.
- **Structural hardening.** `extract_reply_context` narrowed to `extract_reply_target_id`. It no longer returns the replier-authored snapshot at all, so "the snapshot is never rendered" is enforced by the signature rather than by every caller remembering to discard it. Both callers only ever wanted the id.
- **`cli/README.md`** documented `reply_to` as null only for non-replies. It is now also null for a reply whose quote could not be verified — precisely what a bridge author needs to know, and the PR that changed it did not update the doc.
- **Naming.** `resolve_reply_context` → `resolve_reply_strip` (it returns a `ReplyStrip`); `ReplyStrip::None` → `ReplyStrip::NotAReply`, matching riverctl's clearer variant name.
- **Comment corrections** where a comment overstated what the code does: the strict/lenient decrypt difference is two cases not one; the CLI/UI mirror claim now scopes to *resolution* (rendering deliberately differs — 50 chars with "..." vs 100 chars markdown-stripped); the "add a content type" note pointed at the wrong CLI function and omitted the public decode path; the enum's rationale described a failure mode the old code did not have.

## Testing

`cargo test`: 21 targets green (`river-ui` 516, `riverctl` 259+, `river-core`).

New/changed pins:
- Cold start (no secrets) → no strip; rotated-past version → placeholder. Asserted separately in both clients so the two cases cannot be conflated again.
- Malformed public reply → `Unavailable` in riverctl.
- Action/event filter: tests now fail without the filter (verified by deleting it).
- The `common/tests` premise test now actually runs in CI.

`cargo fmt` and `cargo clippy` clean. Room-contract WASM hash unchanged (`scripts/check-room-contract-migration.sh --ci origin/main HEAD`), so the live Official room's key is untouched.

## Review

Full multi-model review across three rounds on #471 and this follow-up. Round 1 had an external Codex pass plus three Claude lenses. **Codex is unavailable for rounds 2 and 3** — `codex exec` fails with "You've hit your usage limit… try again at Jul 28th 2026 12:14 PM", and the Gemini CLI is deprecated on the individual tier. Per the review policy I used the Claude-subagent fallback with widened perspective diversity instead of dropping independent review: round 2 ran skeptical + testing lenses, round 3 ran skeptical + big-picture + code-first. Every finding in this PR came out of that fallback — including the cold-start regression, which no single lens caught twice.

Worth re-running a real Codex pass on this after Jul 28 if it is still open.

## Known follow-ups (not in this PR)

- The **send path still mints the snapshot**: every new reply persists `target_author_name` / `target_content_preview` into signed contract state, data no current River client reads. Older builds and the currently-published web container still render it, so users only see the fix once the UI is republished.
- `common/src/room_state/content.rs`'s `ReplyContentV1` doc comment still says the snapshot exists "so that the reply remains meaningful even if the target message is later deleted or scrolled out of the recent window" — the design intent this work deliberately reverses. Left alone to keep `common/src/` untouched; a comment-only change cannot move the WASM hash, so it is safe to fix whenever someone is in there.
- The non-self placeholder styling branch has no Playwright coverage: the example fixture is owner-authored so the `is_self` branch is deterministic.

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,6 +126,17 @@ jobs:
       # --test target from migration_test, so it needs its own invocation.
       run: cargo test -p river-core --test room_contract_migration_test
 
+    - name: Test ban purges authored content (reply-quote premise)
+      env:
+        RUST_MIN_STACK: 8388608
+        CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      # Pins that an enforced ban purges the banned member's messages AND their
+      # member_info. Both clients' reply-quote suppression keys on that absence
+      # (there is no other link from a reply's quoted-author NAME back to a
+      # MemberId), so if this sweep regressed, a banned member's text would
+      # start rendering again with every client-side test still green.
+      run: cargo test -p river-core --test ban_purges_authored_content_test
+
     - name: Test riverctl (CLI crate unit tests, incl. #292 recovery)
       env:
         RUST_MIN_STACK: 8388608
--- a/cli/README.md
+++ b/cli/README.md
@@ -256,7 +256,9 @@ Run `riverctl <group> --help` or `riverctl <group> <cmd> --help` for full flags.
 | `edit` | A previously-streamed message's content changed | same as `message` (with the new `content` and `edited: true`) |
 | `delete` | A previously-streamed message was deleted | `message_id`, `room`, `author`, `nickname`, `timestamp` (no `content`) |
 
-`reply_to` is `null` for non-replies, otherwise `{ "author", "preview" }`. `edit`/`delete` events are emitted only for messages the stream actually surfaced. Bridges should key off `type` and tolerate unknown future types.
+`reply_to` is `{ "author", "preview" }` only when the quoted message could be read back from live room state. It is `null` both for non-replies **and** for a reply whose quoted message could not be verified — its author was banned (which purges their messages), it was deleted, or it aged out of the room's `max_recent_messages` window. The quoted author and preview stored in a reply are a snapshot written by the *replier* and validated by nothing, so riverctl never emits them unverified; a bridge would otherwise republish a banned member's text after the ban. The human-readable output distinguishes the two cases with a `[reply to an unavailable message]` prefix; the JSON deliberately does not, so no existing consumer breaks on a new shape.
+
+`edit`/`delete` events are emitted only for messages the stream actually surfaced. Bridges should key off `type` and tolerate unknown future types.
 
 The top-level `author` is the sending member's ID; get your own with `riverctl
 identity whoami <room-owner-vk>` and compare against it to recognise your own
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -702,8 +702,14 @@ pub(crate) fn reply_context_display(
 }
 
 /// What a message's reply context should render as, resolved against LIVE room
-/// state. Mirrors the UI's `ReplyStrip` (`ui/src/components/conversation.rs`);
-/// the two must stay in step, since they render the same contract data.
+/// state. Mirrors the UI's `ReplyStrip` (`ui/src/components/conversation.rs`).
+///
+/// The RESOLUTION (which targets are quotable, and when a quote is refused)
+/// must stay in step between the two clients — they read the same contract
+/// data, and a divergence means one client shows text the other suppressed.
+/// The RENDERING deliberately differs: riverctl truncates to 50 chars with a
+/// "..." marker and leaves markdown intact; the UI truncates to 100 and strips
+/// markdown.
 #[derive(Debug, PartialEq)]
 pub(crate) enum ReplyContextDisplay {
     /// Not a reply.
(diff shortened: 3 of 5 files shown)


sanity:
## Independent review of #480 — **approve, one non-blocking gap**

Reviewed the committed code at `6f9915a1`, with mutation testing rather than reading. This is the follow-up to my review of #471 (which I reversed on the cold-start regression this PR fixes). Per that reversal I re-verified only what changed — the scope check, snapshot trace, and output-path coverage from the #471 pass still hold and aren't re-litigated here.

**Verdict: no blocking findings.** One test gap survives from my #471 table and should be closed, but it is not a live vulnerability and I would not hold the PR for it.

---

### 1. The cold-start fix is correct and survived the commit unchanged

`ui/src/components/conversation.rs:337`:

```rust
fn pending_decryption(content: &RoomMessageBody, secrets: &HashMap<u32, [u8; 32]>) -> bool {
    matches!(content, RoomMessageBody::Private { .. }) && secrets.is_empty()
}
```

This is exactly the predicate I traced while it was uncommitted. The load-bearing detail is the one the doc comment calls out: the discriminator is **"no secrets at all"**, not "this version is missing". That is the right split —

- empty map → cold start, nothing is verifiable and nothing is being hidden (the snapshot is unreadable too) → render as if not a reply
- non-empty map, missing version → a genuinely rotated-past secret → keep showing the placeholder

and it matches `decrypt_message_content`'s own `secrets.is_empty()` discriminator, so the body copy ("Decrypting messages…") and the strip stay consistent instead of contradicting each other. That contradiction was the regression on `470b84d3`.

**Mutation-verified**, not just read: forcing `pending_decryption` to return `false` (reintroducing the flash) fails **2 UI tests** and **1 CLI test**. The fix is genuinely pinned on both sides.

### 2. The core vulnerability is now structurally unreachable in the UI

This is the strongest change in the PR and it deserves calling out. `extract_reply_context` became `extract_reply_target_id` (`conversation.rs:717`), returning `Option<MessageId>`. The snapshot no longer leaves the function.

I verified the consequence: in `ui/src/components/conversation.rs`, `target_author_name` and `target_content_preview` now appear **only in comments** (lines 199, 300, 713). My "revert the core fix" mutation could not even be applied to the UI — the snapshot isn't in scope to fall back to.

That converts the guarantee from "every caller must remember to bind `_`" into "the compiler will not let you". Reintroducing this bug now requires restructuring the extraction, not a one-line slip during a refactor. That is a real improvement over #471's shape.

On the CLI side the snapshot is still in scope (the resolver decodes the full `ReplyContentV1` for `target_message_id`), so the mutation is still meaningful there — and it is caught by **2 tests**.

### 3. The CI gap is closed, and the wiring actually works

`.github/workflows/build.yml` adds `cargo test -p river-core --test ban_purges_authored_content_test`. I ran that exact command against this SHA:

```
test owner_ban_purges_the_banned_members_messages_and_member_info ... ok
test result: ok. 1 passed; 0 failed
```

So the premise test added by #471 — which was on disk but never compiled by CI — now runs. That matters more than it looks: both clients' quote suppression keys on the *absence* created by that sweep, so if the sweep ever regressed (tombstones, a "soft ban", scoping it to inactivity-prune), a banned member's text would start rendering again with every client-side test still green. This is the tripwire for the whole design and it was previously unwatched.

---

### Mutation table — #480 vs #471

Same method as before: scratch clone at the PR SHA, revert one guard at a time, run the suite. The branch is never touched.

| Mutation | #471 (`470b84d3`) | #480 (`6f9915a1`) |
|---|---|---|
| UI: core fix reverted (snapshot fallback) | 6 tests ✅ | **structurally unreachable** ✅✅ |
| CLI: core fix reverted | 2 tests ✅ | 2 tests ✅ |
| UI: `pending_decryption()` → `false` | n/a | **2 tests ✅** |
| CLI: `pending_decryption()` → `false` | n/a | **1 test ✅** |
| UI: delete `is_action()` + `is_event()` filter | **0 tests** ❌ | **1 test ✅ FIXED** |
| UI: `is_deleted()` disabled | 1 test ✅ | 1 test ✅ |
| UI: `canonical()` → first-match `.find()` | **0 tests** ❌ | **0 tests** ❌ still open |

Baseline both sides: 12 UI tests, 259 riverctl tests, all passing.

**The action/event gap is properly fixed.** `quote_of_an_action_or_event_message_is_hidden` now populates `edited_content` for the event target, so it exercises the real mechanism — events are editable (`common/src/room_state/message.rs:273-277` filters only actions when building `message_authors`) and `effective_text` reads `edited_content` before any content-type decode. Deleting the filter now fails the test, where before it did not.

---

### The one remaining gap (non-blocking)

**`quote_author_uses_the_canonical_member_info_record` (`conversation.rs:4980`) still does not pin `canonical()`.** Replacing `member_info.canonical(member_id)` with a first-match `.find()` leaves all 12 tests green.

The fixture is unchanged from #471 — `info(vec![canonical, stale])` — and its comment still reads:

> Listed with the STALE one last so a naive `.collect()` into a map would pick it.

That ordering does defeat a last-wins `HashMap` collect, which is the shape the comment names. But `canonical()` is `max_by_key` over `(version, signature)` (`member_info.rs:39-44`), and on this fixture a first-wins `.find()` returns the canonical record by luck of position. The test pins one regression shape and sails past its sibling.

Not a live vulnerability — `canonical()` is present and correct, and the impact if it regressed is a quote attributing to a stale nickname rather than anything attacker-controlled. It is regression risk on a security-adjacent path, and the fix is one line: add a third record, or assert with both orderings so no single position is accidentally correct.

I'd take this as a follow-up rather than a blocker.

---

### Merge readiness

- **Scope clean**: no `.wasm`, nothing under `contracts/`, nothing under `common/src/`. Only `common/tests/` is touched indirectly via the CI wiring. Contract key untouched.
- **CI**: `build` **pass**, both migration checks pass, CLA pass. `ui-playwright-tests` still **pending** at time of writing — must be green on `6f9915a1` before merge.
- Rebased on current `main` (`64a77f58`), no conflict.

**Process note:** the external non-Claude pass remains unavailable (Codex usage limit, resets Jul 28; Gemini deprecated org-wide). This is the documented external-unavailable fallback. For this pass the primary evidence is execution-based — mutation testing — which is stronger than a reading-based review and is what caught the two unpinned guards in the first place.

[AI-assisted - Claude]

PR #481, opened 2026-07-25
fix(cli): mirror the UI's pending-decryption rule on the reply-quote target path

sanity:
## Problem

Third and final follow-up to #471 / #480, from the review round that ran after #480 merged. Two reviewers independently found the same gaps.

### 1. riverctl and the UI diverge on a public reply quoting a private target

#480 added `pending_decryption` (hold no secrets at all → render nothing, rather than asserting "unavailable") and applied it in **two** places in the UI: when the reply itself won't decode, and when the target's plaintext won't resolve. riverctl only got the first.

That second path is reachable, and not exotically: the composer **deliberately sends a reply publicly when the room secret is missing** (`ui/src/components/conversation.rs`, `warn!("Room is private but no secret available, sending reply as public")`). So a public reply can quote a private target, and with no secrets held the UI renders no strip while riverctl prints `[reply to an unavailable message]`.

Cosmetic — neither client renders any snapshot text — but #480's own doc comment says "the RESOLUTION must stay in step between the two clients", and it wasn't. riverctl now mirrors the check.

### 2. The behaviour change #480 made to malformed public replies was never pinned

#480 changed riverctl's verdict on a `Public` body with `content_type == 3` and undecodable payload from `NotAReply` to `Unavailable`, aligning it with the UI. That is the correct direction — any member can post one, since the contract treats message content as opaque bytes — but no test covered it in **either** client. Both now do.

These tests also pin something subtle: that `pending_decryption`'s `matches!(…, Private { .. })` guard is load-bearing. Simplifying it to a bare `secrets.is_empty()` would silently swallow the malformed-public case.

## Also in this round

- **CI: a catch-all instead of a fifth allowlist entry.** #480 added `cargo test -p river-core --test ban_purges_authored_content_test`. But the named-target list is exactly why that test shipped in #471 and was never run *or even compiled* — and eight other `common/tests/*.rs` files are in the same dead zone (~136 tests, including `summary_determinism_test` and `deputy_ban_test`). Replaced with `cargo test -p river-core --tests`, which cannot silently omit a newly-added file. Verified all 13 targets pass locally.
- **`pending_decryption`'s doc overstated its premise.** It claimed empty-secrets is always the transient cold-start window. It isn't: a public room always has empty secrets, and any member of one can post a `Private` body (nothing rejects it) to flip the flag for everyone. The failure mode is benign — both branches render zero attacker-supplied text, so all it can suppress is the "unavailable" marker — but the comment shouldn't assert something false. The riverctl copy also called it "waiting", which doesn't fit a one-shot CLI that either resolves secrets at startup or never will; it now says why the behaviour is mirrored anyway.
- **`cli/README.md`** claimed the human-readable output always distinguishes an unverifiable quote. It doesn't for the no-secrets case, and the list of `null` reasons omitted "could not be decrypted".
- **Four stale `resolve_reply_context` pointers** left by #480's rename to `resolve_reply_strip` — one of them written by #480 itself, and one in the very test file it wired into CI. These comments are the only mechanism keeping the two clients' resolution in step, so a pointer that doesn't resolve is worse than usual: finding 1 above is exactly the drift they exist to prevent.
- A stale test doc comment that described the pre-#480 assertion.

## Testing

`cargo test`: 21 targets green (`river-ui` 518, `riverctl` 261, `river-core` 13 targets / 343).

New pins, both verified **fail-first** by reverting the fix:
- `public_reply_to_undecryptable_private_target` (both clients) — quote resolves with the secret, `Unavailable` when rotated past it, no strip when holding nothing. Removing riverctl's new branch makes the CLI copy fail.
- `undecodable_public_reply_reports_unavailable` (both clients).

`cargo fmt` and `cargo clippy` clean. Room-contract WASM hash unchanged, so the live Official room's key is untouched.

## Review

Two independent Claude lenses (skeptical bug-hunt, testing) on #480's merged content; both surfaced findings 1 and 2 independently.

**Codex remains unavailable** — `codex exec` fails with "You've hit your usage limit… try again at Jul 28th, 2026 12:14 PM", and the Gemini CLI is deprecated on the individual tier. Per the review policy this is the Claude-subagent fallback with widened perspective diversity, not a dropped review. Across #471/#480/this PR the sequence was: round 1 Codex + 3 Claude lenses; round 2 two lenses; round 3 three lenses; round 4 two lenses. A real Codex pass after Jul 28 would be worth it if any of this is still open.

## Known follow-ups (not in this PR)

- The **send path still mints the snapshot**: every new reply persists `target_author_name` / `target_content_preview` into signed contract state, data no current River client reads. Older builds and the currently-published web container still render it, so users only see the fix once the UI is republished.
- `common/src/room_state/content.rs`'s `ReplyContentV1` doc still describes the snapshot as existing so a reply "remains meaningful even if the target message is later deleted or scrolled out of the recent window" — the intent this work reverses. Left alone to keep `common/src/` untouched; a comment-only edit cannot move the WASM hash, so it is safe whenever someone is in there.
- The action-message half of the action/event filter tests is synthetic (`message_authors` excludes actions, so the contract can never populate `edited_content` for one). The event half is the reachable attack; worth a note if anyone extends those tests.
- `resolve_reply_strip` does a linear scan per message inside the conversation memo, i.e. O(n²) per recompute. Fine at `max_recent_messages` = 100, but it compounds the known conversation-memo jank (#466 territory).

[AI-assisted - Claude]


sanity:
Pushed `92cb4eaf`: closes the `canonical()` mutation gap carried over from #471/#480.

The fixture was `vec![canonical, stale]` — ordered so a naive last-wins `.collect()` into a map picks the stale record, which it does catch. But position 0 happened to be the *correct* record, so a first-match `.find()` sailed straight through while the test looked deliberate. It pinned one regression shape and missed its sibling.

Now asserts **both** orderings. Verified fail-first by reverting `resolve_member_nickname` to a first-match `.find()`:

```
quote_author_uses_the_canonical_member_info_record ... FAILED
  left: "Alice v1"
 right: "Alice v2"
```

Also filed #482 for the underlying vulnerability — reply quotes are forgeable, because `ReplyContentV1` stores a replier-authored, contract-unvalidated snapshot of the quoted author and text. It includes the live impersonation incident as evidence, and frames removing `target_author_name` / `target_content_preview` as an actionable follow-up now that contract changes are permitted.

[AI-assisted - Claude]

PR #483, opened 2026-07-25
fix(ui): never offer the Ban action on your own Member Info

### LINKED ISSUE #478: UI: a deputy sees an enabled "Ban User" button on their own Member Info, and self-banning cascades to their whole invite subtree
## Problem

A member with deputy authority who opens **their own** Member Info modal sees an enabled "Ban User" button. Clicking it produces a contract-valid ban that removes them **and their entire invite subtree**.

## Mechanism

`MembersV1::is_ban_authorized` (`common/src/room_state/member.rs:318`) has no `banner == target` check, and self-ban genuinely passes it:

- `target == owner_id`? No (the viewer is not the owner).
- `banner == owner_id`? No.
- Step 2, strict ancestors: never fires for self, because the ancestor walk seeds `visited` with `target` (`member.rs:346`), so a member is never their own strict ancestor.
- **Step 3** fires if the owner listed this member in `deputies` (owner-appointed global moderator).
- **Step 5** fires if any strict non-owner ancestor listed them — the ordinary case, since you usually deputize someone you invited.

So `is_ban_authorized(P, P) == true` for most deputies.

`viewer_can_ban` (`ui/src/components/members/member_info_modal.rs:23-38`) is a thin wrapper over that predicate with no self-guard, and its result is passed straight to `BanButton` (`:451-455`). The button is rendered under `if !is_owner` only — there is no `member_id != self_member_id` condition. Note that `DeputyButton`, four lines below at `:458`, *does* guard self:

```rust
if member_id != self_member_id {
    DeputyButton { … }
}
```

`BanButton` (`ui/src/components/members/ban_button.rs:168`) has no guard either.

Nothing downstream rejects it: `validate_single_ban` does not refuse `banned_by == banned_user`, and enforcement re-runs the same predicate, so the ban applies. `MembersV1::check_banned_members` then extends removal to `get_downstream_members(target)` (`member.rs:286-287`), so the member's whole invite subtree goes with them.

## Impact

One misclick costs a moderator their membership and removes every member they invited, transitively. On the live Official room the Invite Bot's deputies each sit above ~105 members, so the blast radius is most of the room.

It is also reachable by accident in a way an ordinary ban is not: the natural mental model is "the Ban button appears on people I can ban", and the modal shows it on yourself.

## Suggested fix

Guard the render site the same way `DeputyButton` already is:

```rust
if member_id != self_member_id {
    BanButton { … }
}
```

Belt and braces: also return `false` from `viewer_can_ban` when `viewer == target`, so any future call site inherits the guard.

Worth considering separately: whether `validate_single_ban` should reject `banned_by == banned_user` outright at the contract level. That is a wire-format-adjacent change and needs its own discussion, but the UI fix is safe and independent.

## Notes

Found by a reviewer during freenet/river#477, which fixes the CLI-side manifestation of the same underlying fact: `riverctl` was counting a deputy as one of their own bannable targets, so `reaches_members` reported authority that is really just the ability to remove yourself. The CLI bug cost one wrong integer; this one costs a user their subtree.

Related: #470, #477.

[AI-assisted - Claude]

sanity:
## Problem

A member with deputy authority who opened **their own** Member Info modal saw an enabled "Ban User" button. `member_info_modal.rs` rendered `BanButton` under `if !is_owner` with no `member_id != self_member_id` condition — while `DeputyButton`, a few lines below in the same `div`, already guarded self.

The resulting ban is **contract-valid**, and enforcement cascades removal to `get_downstream_members`, so one misclick removed the moderator **and their entire invite subtree**. (Blast radius figure per #478; not independently re-derived here.)

Only deputies were affected. The owner never was — `is_ban_authorized` denies `target == owner_id` outright before any grant — and neither was an ordinary member with no deputy authority.

## This is contract-PERMITTED behaviour gated at the interaction layer, not a contract fix

Please do not later "simplify" any of these guards away believing the contract prevents self-bans. It does not:

- `is_ban_authorized(P, P)` genuinely returns `true` for most deputies. The strict-ancestor walk seeds its `visited` set with `target`, so a member is never their own strict ancestor and step 2 never fires — but **step 3** (the owner listed them in `deputies`) or **step 5** (a strict non-owner ancestor listed them, the ordinary case since you usually deputize someone you invited) does.
- `validate_single_ban` does not refuse `banned_by == banned_user`, and enforcement re-runs the same predicate, so the ban applies. The cascade is in `MembersV1::banned_member_ids`, retained out by `ChatRoomStateV1::post_apply_cleanup`.

Whether the *contract* should reject `banned_by == banned_user` is a separate, wire-format-adjacent question and is deliberately **not** decided here. This PR is `ui/` only — `git diff origin/main --name-only` is three files, all under `ui/`, so no WASM is re-keyed.

> Note for anyone reading #478: the issue (and this PR's first revision) attributed the cascade to `MembersV1::check_banned_members`. That is wrong. `check_banned_members` walks invite chains for `has_banned_members`, which itself has no production caller. The cascade is `banned_member_ids`. The issue's line citation was right; only the function name was wrong.

## Approach — three layers, each with a distinct job

**1. `viewer_can_ban` refuses `viewer == target`** (`ui/src/components/members.rs`). The primary guard. It is the UI's single definition of "may this viewer ban this target", so every current and future caller inherits it.

**2. The `BanButton` render site is gated on `member_id != self_member_id`**, matching `DeputyButton` beside it. Redundant today, and kept deliberately: the asymmetry between a guarded Deputize and an unguarded Ban is what hid this bug, so the two actions in the row now read the same way.

**3. `BanButton::execute_ban` refuses `member_to_ban == banned_by` before constructing the `UserBan`** (`ui/src/components/members/member_info_modal/ban_button.rs`). Added in review. The first two decide whether the button *appears*; this one decides whether a ban is ever *built*, so it is the layer that does not depend on a caller having asked the right question first. Nothing downstream would catch it: the chat delegate signs whatever bytes it is handed, and `validate_single_ban` does not refuse a self-ban.

That third layer is the one that makes the riverctl argument actually land. `ApiClient::ban_member` (`cli/src/api.rs`) returns `Err("Cannot ban yourself")` — but it refuses at ban *construction*, and the first two guards are on the read/render path. Before this, the UI had no write-path refusal at all.

## Other callers audited

| Site | Verdict |
|---|---|
| `member_info_modal.rs` (`can_ban` → `BanButton`) | **The bug.** Fixed. |
| `members.rs` `badge_for_target` (calls `is_ban_authorized` directly) | **No gap.** `target == self_member_id` is already excluded. The remaining self-evaluation is `is_ban_authorized(A, A)` for `A` a strict ancestor of the viewer — but the same `.any()` also evaluates `is_ban_authorized(A, viewer)`, unconditionally `true` via step 2 for any strict ancestor, so the self path can never decide the result. Confirmed independently by two reviewers, who verified the two ancestor walks use identical semantics. |
| `ui/src/room_data.rs`, `room_synchronizer.rs` `UserBan` constructions | All inside `#[cfg(test)]`. |

`ban_button.rs` is the only production `UserBan` construction in `ui/`, and `BanButton` has one render site.

## Testing

Three tests, one per guard, because **any one guard alone hides or blocks the ban** — so no single behavioural test can detect the loss of just one.

1. `viewer_can_ban_refuses_self_even_though_the_contract_permits_it` — behavioural. Covers both self-ban routes (step 3 and step 5) and asserts the **raw contract predicate first**, so it cannot pass for the wrong reason: if a future contract change starts refusing self-bans, the precondition fails loudly rather than letting the UI guard be deleted under a still-green test. It then asserts one case per authority route (steps 2, 3, 5) reaching a **third party**, so the guard can't be widened. The step-5-over-a-third-party case was added in review — that route was asserted nowhere in `ui/`, despite being the route the second self-ban case depends on.
2. `ban_button_render_site_is_self_guarded` — source pin, line-based: the guard must be on the immediately preceding line of code, at **every** `BanButton {` occurrence.
3. `execute_ban_refuses_self_before_building_the_ban` — source pin asserting the comparison exists, runs before **every** `UserBan` construction, and that its body actually `return`s.

Both pins run over comment-stripped source (`//` and `/* */`), so a comment that merely quotes a guard cannot satisfy them. The `#[cfg(test)]` cut is a shared `production_source()` helper used by every pin in the module, so it cannot drift between them.

### Mutation verification

Every guard was reverted in turn and the suite re-run. Each mutation fails **only** its own test, confirming the three tests are independent:

| Mutation | Result |
|---|---|
| Remove `viewer == target` from `viewer_can_ban` | behavioural test FAILS |
| Remove the render-site `if` | render pin FAILS |
| Add a second, unguarded `BanButton` render site | render pin FAILS |
| Disarm attempt: `/* */` comment quoting the guard, real guard deleted | render pin FAILS |
| Remove the `execute_ban` guard | write-path pin FAILS |
| Drop just the `return;`, keep the `warn!` | write-path pin FAILS |
| Move the guard after the `UserBan` construction | write-path pin FAILS |
| Over-block: require owner-appointment (drops step-2/step-5 authority) | behavioural test FAILS |

`cargo test -p river-ui --bins` 553 passed; `cargo test -p river-core --lib` 205 passed. `cargo fmt --all -- --check` clean; `cargo clippy -p river-ui --bins --all-targets` reports nothing in any changed file.

## Follow-ups deliberately not done here

- **A `common/`-side test that a self-ban cascades.** It would pin the harm at its source, and nothing in `common/` currently tests `banned_by == banned_user`. Out of scope: this change is `ui/`-only because a `common/` change re-keys the live room contract.
- **A Playwright test.** It would be vacuous today: in `ui/src/example_data.rs` the local user is never deputized, so `is_ban_authorized(self, self)` is already `false` and the Ban button is absent with or without all three guards. Making it real needs a fixture change (deputize `(You)` in one example room), which would touch other e2e tests.
- **The empty `mt-4 flex` row** on your own Member Info when both buttons are hidden. Pre-existing — `BanButton` already rendered nothing whenever `can_ban` was false — not a regression from this PR.

Closes #478

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -639,6 +639,27 @@ pub(crate) fn relevant_deputizer_names(
 /// NOT bare invite-chain ancestry, so a DEPUTY sees the Ban action for members in
 /// their deputizer's subtree. Bare downstream ancestry (`is_downstream`, still
 /// used for the "🔑 Invited by You" relationship tag) would have hidden it.
+///
+/// SELF-BAN IS REFUSED HERE, and that refusal is the UI's own — do NOT
+/// "simplify" it away believing the contract already prevents it, because it
+/// does not (freenet/river#478). `is_ban_authorized(P, P)` genuinely returns
+/// `true` for most deputies: the strict-ancestor walk seeds its visited set with
+/// `target`, so a member is never their own strict ancestor and step 2 never
+/// fires, but step 3 (owner-appointed global moderator) or step 5 (a strict
+/// non-owner ancestor deputized them — the ordinary case, since you usually
+/// deputize someone you invited) does. `validate_single_ban` does not refuse
+/// `banned_by == banned_user` either, and enforcement re-runs the same
+/// predicate, so a self-ban is accepted and then CASCADES:
+/// `MembersV1::check_banned_members` extends removal to
+/// `get_downstream_members(target)`, taking the member's whole invite subtree
+/// with them.
+///
+/// Refusing at the predicate rather than only at the render site means every
+/// caller inherits the guard, and it makes the UI's read path agree with
+/// riverctl's write path, which already refuses self-ban outright
+/// (`ApiClient::ban_member`: "Cannot ban yourself"). Whether the CONTRACT should
+/// reject `banned_by == banned_user` is a separate, wire-format-adjacent
+/// question and is deliberately not decided here.
 pub(crate) fn viewer_can_ban(
     members: &MembersV1,
     member_info: &river_core::room_state::member_info::MemberInfoV1,
@@ -646,6 +667,9 @@ pub(crate) fn viewer_can_ban(
     target: MemberId,
     owner_id: MemberId,
 ) -> bool {
+    if viewer == target {
+        return false;
+    }
     let members_by_id = members.members_by_member_id();
     MembersV1::is_ban_authorized(viewer, target, &members_by_id, member_info, owner_id)
 }
@@ -3947,5 +3971,96 @@ mod tests {
         let decoded =
             Invitation::from_encoded_string(&encoded).expect("legacy invitation should decode");
         assert!(decoded.room_secrets.is_empty());
+    }
+    /// freenet/river#478: a deputy must never be offered the Ban action on
+    /// THEMSELVES. The contract permits that ban — `is_ban_authorized(P, P)` is
+    /// `true` for most deputies — and `MembersV1::check_banned_members` then
+    /// cascades the removal to `get_downstream_members(P)`, so one misclick
+    /// costs the moderator their membership AND their whole invite subtree.
+    ///
+    /// Both self-ban routes are covered, because they reach `true` through
+    /// different branches of `is_ban_authorized`: step 3 (the owner listed you
+    /// in their `deputies`) and step 5 (a strict non-owner ancestor listed you
+    /// — the ordinary case, since you usually deputize someone you invited).
+    ///
+    /// Each case asserts the raw CONTRACT predicate first. Without that
+    /// precondition the test would pass for the wrong reason: if a future
+    /// contract change started refusing self-bans, the UI guard could be
+    /// deleted and this test would stay green while pinning nothing.
+    #[test]
+    fn viewer_can_ban_refuses_self_even_though_the_contract_permits_it() {
+        use river_core::room_state::member::MembersV1;
+        use river_core::room_state::member_info::MemberInfoV1;
+
+        let mut rng = rand::thread_rng();
+        let owner_sk = SigningKey::generate(&mut rng);
+        let global_mod_sk = SigningKey::generate(&mut rng);
+        let parent_sk = SigningKey::generate(&mut rng);
+        let subtree_mod_sk = SigningKey::generate(&mut rng);
+        let bystander_sk = SigningKey::generate(&mut rng);
+
+        let id = |sk: &SigningKey| MemberId::from(&sk.verifying_key());
+        let owner_id = id(&owner_sk);
+
+        // owner -> {global_mod, parent, bystander}; parent -> subtree_mod.
+        let members = MembersV1 {
+            members: vec![
+                authorized_member(&owner_sk, &global_mod_sk.verifying_key()),
+                authorized_member(&owner_sk, &parent_sk.verifying_key()),
+                authorized_member(&owner_sk, &bystander_sk.verifying_key()),
+                member_invited_by(&parent_sk, owner_id, &subtree_mod_sk.verifying_key()),
+            ],
+        };
+        let member_info = MemberInfoV1 {
+            member_info: vec![
+                // Owner appoints a global moderator (step-3 authority).
+                signed_member_info(&owner_sk, "Owner", vec![id(&global_mod_sk)]),
+                signed_member_info(&global_mod_sk, "GlobalMod", vec![]),
+                // Parent deputizes someone they invited (step-5 authority).
+                signed_member_info(&parent_sk, "Parent", vec![id(&subtree_mod_sk)]),
+                signed_member_info(&subtree_mod_sk, "SubtreeMod", vec![]),
+                signed_member_info(&bystander_sk, "Bystander", vec![]),
+            ],
+        };
+        let members_by_id = members.members_by_member_id();
+
+        for (label, sk) in [
+            ("owner-appointed global moderator (step 3)", &global_mod_sk),
+            (
+                "deputy of a strict non-owner ancestor (step 5)",
+                &subtree_mod_sk,
+            ),
+        ] {
+            let who = id(sk);
+            assert!(
+                MembersV1::is_ban_authorized(who, who, &members_by_id, &member_info, owner_id),
+                "precondition: the CONTRACT must permit this self-ban ({label}). \
+                 This test pins a UI-layer guard and is worthless if the \
+                 contract already refuses"
+            );
+            assert!(
+                !viewer_can_ban(&members, &member_info, who, who, owner_id),
+                "the UI must refuse a self-ban ({label}): it is contract-valid \
+                 and cascades to the member's whole invite subtree \
+                 (freenet/river#478)"
+            );
+        }
+
+        // The guard must be exactly `viewer == target` and nothing wider — both
+        // deputies keep the authority they genuinely have over other members.
+        assert!(viewer_can_ban(
+            &members,
+            &member_info,
+            id(&global_mod_sk),
+            id(&bystander_sk),
+            owner_id
+        ));
+        assert!(viewer_can_ban(
+            &members,
+            &member_info,
+            id(&parent_sk),
+            id(&subtree_mod_sk),
+            owner_id
+        ));
     }
 }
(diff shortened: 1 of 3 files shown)


sanity:
Closing as duplicate of #484, which is the canonical fix for #478.

Both PRs guard all three layers (`viewer_can_ban`, the render gate, and `execute_ban` at the action boundary). #484 is being kept because it carries a posted three-lens review and documented mutation testing — five mutants, all verified red, including two pins that were initially wrong (a render-gate test using final brace depth instead of minimum, and a guard-deleted-but-still-quoted-in-a-comment shape). This PR has no review posted.

No reflection on the work here — two agents were dispatched onto #478 independently, which was a coordination failure on my side, not a problem with this branch. If anything in this PR pins something #484 does not, say so and it will be folded in rather than lost.

[AI-assisted - Claude]

sanity:
## Review record

**External model pass attempted and unavailable.** `codex exec` failed with `You've hit your usage limit ... try again at Jul 28th, 2026 12:14 PM` — a real invocation, not an assumption. Gemini CLI is unavailable project-wide (individual free tier deprecated, every call returns `IneligibleTierError`). With both external reviewers down and the change wanted sooner than Jul 28, this used the **Claude-subagent fallback** with widened perspective diversity, which Ian confirmed is acceptable for this change.

**Two full rounds**, re-run rather than reused, because the code changed after round 1 and a review of superseded code is not a review.

| Round | SHA | Lenses |
|---|---|---|
| 1 | `86f58cb0` | test-validity, skeptical bug-hunt, code-first |
| 2 | `c46ae1cb` | test-validity, skeptical bug-hunt (scoped to the newly added code) |

Round 2 exists because round 1's fixes added a guard and rewrote the tests — the exact "reviewers approved an hour ago" trap.

### Findings and disposition

Everything below was found by a reviewer, not by me. All are fixed; nothing was dismissed.

**Round 1**

| Finding | Severity | Disposition |
|---|---|---|
| Cascade attributed to `MembersV1::check_banned_members`, which never calls `get_downstream_members`, is private, and is reachable only from a function with no production caller. Found independently by two reviewers | MINOR ×2 | **Fixed.** Corrected to `banned_member_ids` in both comments, the test doc, and the PR body. This mattered more than a normal comment nit: those comments exist to be a verifiable proof that stops a future maintainer deleting the guard, and a citation leading to dead code invites them to conclude the rationale is stale |
| No write-path guard. Both guards were read/render-path, while the PR's headline argument cites riverctl's *write*-path refusal. `execute_ban` constructs and signs the `UserBan` with no self-check. Converged on independently by two reviewers | MAJOR | **Fixed.** Added the refusal in `execute_ban` |
| The "not hypothetical" justification for the render-site guard cited the deputy-badge code's direct `is_ban_authorized` call — but that code produces a tooltip and could never feed `can_ban` | MINOR | **Fixed.** Dropped; the comment now leads with the `DeputyButton`-symmetry argument, which is the one that holds, and says plainly the guard is redundant today |
| Source pin disarmable by a comment quoting the guard verbatim | MINOR ×2 | **Fixed**, then found insufficient in round 2 (see below) |
| Pin only checked the FIRST `BanButton {`; a second unguarded site would pass | MINOR | **Fixed.** Checks every occurrence |
| Step-5 authority over a third party asserted nowhere in `ui/`, though it is the route the second self-ban case depends on | MAJOR | **Fixed.** Added a `sibling` member; the over-blocking check is now one case per authority route (steps 2, 3, 5) |

**Round 2**

| Finding | Severity | Disposition |
|---|---|---|
| The write-path pin never pinned the abort. Dropping `return;` while keeping the `warn!` compiles clean, does nothing, and left the test green. Found by both reviewers | MAJOR | **Fixed.** Asserted, and checked against every `UserBan` construction rather than the first |
| `strip_comments` did not strip `/* */` — the same disarm via the other comment syntax, i.e. the hole the helper was added to close. Its doc asserted "never a silent pass", which was false. Found by both reviewers | MAJOR / MEDIUM | **Fixed.** Block comments stripped; the doc now states honestly what remains unhandled and which direction it fails in |
| The 200-char window had ~131 chars of slack; four extra comment lines would fail the pin with a message claiming the guard is missing when it is right there | MINOR | **Fixed** by removing the window: the pin is now line-based (guard on the immediately preceding line of code), and comment-only lines collapse to nothing |
| The write-path refusal stranded the confirmation dialog — `execute_ban` never resets `show_confirmation` on the success path, so returning early left a "Yes, Ban User" button that silently did nothing | MEDIUM | **Fixed.** Closes the dialog before returning. Unreachable today, but stranding the UI in precisely the case the guard exists for is the wrong failure mode |
| `check_banned_members` *is* called (by `has_banned_members`); it is `has_banned_members` that has no production caller | LOW | **Fixed** |
| Hardcoded `member.rs:286-287` citation will rot | LOW | **Fixed.** Line numbers dropped |
| TOCTOU justification overstated — a ROOMS write invalidates the memo and re-renders, unmounting the button | LOW | **Fixed.** Softened to defense in depth |
| Guard sat after a deep clone of the whole room state | LOW | **Fixed.** Moved ahead of the clones |

Two round-2 suggestions are **not** done, and are recorded as follow-ups in the PR description rather than silently dropped: a `common/`-side test that a self-ban cascades (out of scope — `common/` re-keys the live contract), and a Playwright test (vacuous today, since `example_data.rs` has no deputized local user, so the button is absent with or without the guards).

### Mutation verification

Every guard was reverted in turn, at the current SHA, after the tests were rewritten. Eight mutations, each failing **only** its own test — including the four disarm routes reviewers identified. The table is in the PR description.

[AI-assisted - Claude]

PR #484, opened 2026-07-25
fix(ui,cli): refuse bans whose cascade would remove the banner (#478)

### LINKED ISSUE #478: UI: a deputy sees an enabled "Ban User" button on their own Member Info, and self-banning cascades to their whole invite subtree

sanity:
## Problem

Bans **cascade**. `MembersV1::check_banned_members` inserts the banned member and then extends with `get_downstream_members`, the full transitive closure of their invite subtree. That is by design.

The consequence: a deputy who bans anyone **above them in the invite chain** is inside that ancestor's closure, so the cascade removes **the deputy themselves and their entire subtree**. Same blast radius as a direct self-ban, reached by a different route.

It is enabled today, not theoretical. `MembersV1::is_ban_authorized` grants owner-appointed **global moderators** authority at step 3 (`common/src/room_state/member.rs:377`), ahead of the step-4 "cannot ban the member who deputized you" guardrail — so a global mod is authorized over their own ancestors, and over themselves. Nothing downstream rejects it: `verify` accepts the ban, and the cascade runs. On the Official room the Invite Bot's deputies carry roughly 105 members each.

The original PR (#478) closed only the **direct** case (`banned_by == banned_user`). This closes the transitive one, which the review of the first round flagged as still open.

## Approach

Replaces the direct-only self-check with **one rule, stated once**:

> Compute the set the ban would remove — the target **plus** their transitive downstream closure — and refuse if the banning member is in it.

That subsumes the direct case (`banner == target`) rather than sitting beside it as a second special case. A future reader sees a single "you may not ban yourself out of the room" invariant.

Three pieces, in `ui/src/components/members.rs`:

- **`ban_removal_set`** — mirrors what `check_banned_members` actually builds. The contract's `get_downstream_members` is private to the contract crate, so the walk is recomputed rather than approximated (`cli/src/deputies.rs::invite_subtrees` is riverctl's existing mirror of the same thing). It carries a **visited-set guard** the contract omits: the contract can rely on `verify` having rejected circular invite chains, while the UI walks whatever state it currently holds.
- **`self_removing_ban_reason`** — the rule. Returns the user-visible reason, `None` when it does not apply. The *decision* is one computation; only the *wording* branches, because "you can't ban yourself" and "you can't ban someone you joined through" read very differently to a user.
- **`ban_gate`** — applies the rule to the **output** of `is_ban_authorized`, never inside its priority ladder. That placement is the load-bearing part: a check wired into the wrong branch would not fire for exactly the global moderators who can reach the most members. Mutant **M6** pins this.

### Why `viewer_can_ban` became `ban_gate`

The bool predicate could not express *why* the action was withheld, and "silently disable the button" is a bug report waiting to happen. `BanGate` has three answers:

| Answer | Meaning | UI |
|---|---|---|
| `Allowed` | authority held, ban is safe | Ban button renders |
| `WouldRemoveViewer(reason)` | authority held, but the ban removes **you** | button replaced by the reason text |
| `NoAuthority` | never had the action here | nothing (unchanged) |

Distinguishing the last two matters: without it, every ordinary member opening their own inviter's profile would be told they can't do something they never could. `a_non_global_deputy_is_refused_an_ancestor_for_lack_of_authority` pins the distinction.

This is a **replacement, not a removed layer** — `viewer_can_ban` was a thin wrapper with no other callers, and every assertion that used it now asserts on `ban_gate` (strictly stronger: the tests distinguish `NoAuthority` from `WouldRemoveViewer`, so an over-broad guard can no longer hide as "no authority").

### Enforcement points

1. **`ban_gate`** — the predicate. Refuses.
2. **rsx render gate** — the modal renders the explanation in place of `BanButton`, so the UI never offers an action that will be refused.
3. **`execute_ban`** (action boundary) — re-checks the **same** predicate, deliberately not a re-derivation. This is the one that makes the capability *unreachable* rather than merely un-rendered: the handler trusts a `member_to_ban` prop and a `can_ban` prop, so any future call site passing `can_ban: true` would otherwise run the cascading ban. The pin asserts the check does not consult `can_ban` at all (mutant **M10**).

**Cascade semantics are unchanged.** Banning a non-ancestor still cascades exactly as before. This adds a guard; it does not alter who gets removed.

### riverctl

It did **not** refuse the transitive case — only `banned_member_id == my_member_id`. Now fixed:

- `cli/src/deputies.rs` gains `ban_removal_set` + `self_removing_ban_reason` (built on the existing cycle-guarded `invite_subtrees`), and `ban_member` routes through it.
- `RoomDeputies::reach_of` stops counting targets whose ban would remove the deputy. The existing comment there already required this — it said riverctl's read path must not advertise authority the write path refuses, and *"keep the two in step: if either is relaxed, revisit the other."* Before this, a global moderator's reported reach included their own ancestors, which `ban_member` now rejects.

The rule cannot live in `river-core` without changing the room-contract WASM (and therefore the contract key and a migration), so the two clients carry mirrored copies with cross-references in both directions.

## Testing

`cargo test -p river-ui --bins` — 558 passed. `cargo test -p riverctl --lib` — 267 passed. `cargo fmt` clean; `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` clean.

Every case below is proved against **one shared fixture** (`owner → alpha → beta → deputy → child`, plus an unrelated `stranger`, with `deputy` an owner-appointed global moderator), so the ALLOWED cases and the REFUSED cases come from the same room — an over-broad guard cannot hide behind a friendlier fixture. Every refusal asserts, as a **precondition**, that `is_ban_authorized` returns `true` and that the removal set really contains the banner; without that a test could pass for want of authority rather than because of the guard. (The review of round 1 found exactly this trap in `example_data.rs`, where "(You)" has an empty `deputies` list.)

### Mutation table

Every guard was reverted **individually**, the tests re-run, and the guard restored. 16 mutants, 16 caught.

| # | Mutation | Result |
|---|---|---|
| M1 | rule deleted entirely (UI) | FAILED — self + ancestor |
| M2 | **rule narrowed to the DIRECT case only (UI)** — the exact pre-PR behaviour | FAILED — ancestor (self still passes, as it must) |
| M3 | rule inverted, refuse everything (UI) | FAILED — over-broadness test |
| M4 | removal set walks only ONE level | FAILED — grandparent + closure test |
| M5 | cycle guard removed from the removal set | HANG (caught by timeout) |
| M6 | **rule skipped for owner-appointed global mods** — the ordering trap | FAILED — self + ancestor |
| M7 | render gate deleted from the modal | FAILED — render pin |
| M8 | explanation element deleted from the rsx (button hidden, no reason) | FAILED — render pin |
| M9 | action-boundary check deleted | FAILED — boundary pin |
| M10 | **action boundary gated on the caller's `can_ban` prop** | FAILED — boundary pin |
| M11 | source-scrape cut point disarmed so the prod slice includes test code | FAILED — anti-disarm assertion |
| M12 | rule narrowed to the DIRECT case only (riverctl) | FAILED |
| M13 | rule inverted, refuse everything (riverctl) | FAILED — over-broadness test |
| M14 | `ban_member` reverted to the old direct-only check | FAILED — wiring pin |
| M15 | `reach_of` reverted to excluding only the deputy | FAILED |

M8 was caught only **after** it exposed a real weakness: the first draft of the render pin asserted the modal *destructured* `BanGate::WouldRemoveViewer`, which stays green if someone deletes the element that displays the text. Fixed in `9682cee5` and re-run.

M11 exists because of the failure mode where a rebase moves a source-scrape's cut point and every needle starts matching itself. `assert_prod_only` now fails loudly if a production slice still contains `#[test]`.

### Coverage

- deputy bans direct inviter → refused (M2)
- deputy bans grandparent → refused (M2, M4)
- deputy bans themselves → still refused, #478 not regressed
- deputy bans an unrelated member → **still allowed** (M3/M13)
- deputy bans their own downstream → **still allowed** (M3/M13)
- deputy bans a deputy scoped to an ancestor's subtree → **still allowed**
- owner bans anyone, including mid-chain → **still allowed** (the owner is nobody's invitee, so never in a removal set)
- owner-appointed global moderator path guarded (M6) — and shown non-vacuous by a subtree-scoped deputy getting `NoAuthority` for the same target
- action boundary holds against a caller claiming the action is permitted (M10)
- removal set is the full transitive subtree, and terminates on a forged cycle (M4, M5)

No existing test was weakened, deleted or `#[ignore]`d.

## Scope

UI + CLI only — 5 files, no `common/`, no `contracts/`, no `delegates/`, no serialized types, **no WASM change, so no migration entry needed**.

Pre-existing and **not** touched here: `cargo clippy -D warnings` fails across the workspace on files this branch does not modify (`common/src/room_state/ban.rs` doc-indent lints under rust 1.94, plus dead-code and boolean-simplification lints in `cli/src/version_check.rs`, `cli/src/commands/debug.rs`, `ui/src/components/app/notifications.rs`, and others). CI's clippy workflow is disabled (`clippy.yml.disabled`), which is how the backlog accumulated. Fixing it means editing `common/`, which trips the room-contract migration checks — out of scope for this PR, worth its own.

## Review status

**The previous review is superseded for the changed code.** It reviewed a direct-only guard built on `viewer_can_ban` with a `member_to_ban == banned_by` check at the action boundary; all three of those layers have been rewritten, and the predicate they were built on no longer exists. Its own "Reported, not fixed here" section named this transitive gap as the follow-up — this is it.

Two items from that review remain open and are **not** addressed here:

- **The ban dialog still does not disclose the cascade.** It says only "This action cannot be undone", nothing about the invite subtree going with the target. `get_downstream_members` is computable client-side, so "this will also remove N other members" is cheap.
- **The contract still accepts a self-removing ban.** All four enforcement points (three UI, one CLI) are client-side. A stale web container, an old build, or a custom client can still trigger the cascade.

Closes #478

[AI-assisted - Claude]

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -4727,9 +4727,15 @@ impl ApiClient {
 
         let banned_member_id = target_member.member_info.member_id;
 
-        // Prevent self-banning
-        if banned_member_id == my_member_id {
-            return Err(anyhow!("Cannot ban yourself"));
+        // Prevent banning yourself out of the room, directly or transitively
+        // (freenet/river#478). ONE rule: refuse when the ban's cascade would
+        // remove US — the target being self, or the target being one of our own
+        // invite ancestors, whose `get_downstream_members` closure contains us.
+        // Both are contract-VALID, so nothing downstream stops them.
+        if let Some(reason) =
+            crate::deputies::self_removing_ban_reason(&room_state, my_member_id, banned_member_id)
+        {
+            return Err(anyhow!(reason));
         }
 
         // Prevent banning the room owner
--- a/cli/src/deputies.rs
+++ b/cli/src/deputies.rs
@@ -307,10 +307,10 @@ impl<'a> RoomDeputies<'a> {
     /// or is an owner-appointed global moderator. An earlier version of this
     /// function approximated the guardrail by hand and undercounted both cases.
     ///
-    /// ONE local narrowing is applied on top: the deputy is not counted against
-    /// themselves (see the filter below). Scope is the grant's: every member for
-    /// an owner grant (whose subtree is the room), otherwise the deputizer's
-    /// invite subtree.
+    /// ONE local narrowing is applied on top: targets whose ban would remove the
+    /// deputy themselves are not counted (see the filter below). Scope is the
+    /// grant's: every member for an owner grant (whose subtree is the room),
+    /// otherwise the deputizer's invite subtree.
     fn reach_of(&self, deputizer: MemberId, deputy: MemberId) -> usize {
         let empty = HashSet::new();
         let targets = if deputizer == self.owner_id {
@@ -321,19 +321,21 @@ impl<'a> RoomDeputies<'a> {
         targets
             .iter()
             .filter(|target| {
-                // Exclude the deputy themselves. `is_ban_authorized` has no
-                // `banner == target` check and self-ban really is contract-
-                // permitted (a deputy whose deputizer is their own invite
-                // ancestor satisfies step 5 against themselves), but counting
-                // it makes the number answer the wrong question: a deputy who
-                // can only "ban" themselves holds no moderation authority.
+                // Exclude every target whose ban would remove the DEPUTY: the
+                // deputy themselves, and the deputy's own invite ancestors
+                // (whose cascade sweeps the deputy up). `is_ban_authorized` has
+                // no such check and both really are contract-permitted (step 3
+                // grants an owner-appointed global moderator authority over
+                // their own ancestors), but counting them makes the number
+                // answer the wrong question: authority you can only spend by
+                // removing yourself is not moderation authority.
                 //
-                // riverctl's own WRITE path already refuses it -- `ban_member`
-                // returns "Cannot ban yourself" (`api.rs`) -- so before this
-                // the read path advertised a capability the CLI will not
-                // exercise. Keep the two in step: if either is relaxed, revisit
-                // the other.
-                **target != deputy
+                // riverctl's own WRITE path refuses exactly this set --
+                // `ban_member` goes through `self_removing_ban_reason` -- so
+                // the read path must not advertise a capability the CLI will
+                // not exercise. Keep the two in step: if either is relaxed,
+                // revisit the other.
+                !self.ban_would_remove(deputy, **target)
                     && MembersV1::is_ban_authorized(
                         deputy,
                         **target,
@@ -343,6 +345,17 @@ impl<'a> RoomDeputies<'a> {
                     )
             })
             .count()
+    }
+
+    /// Whether a ban of `target` would remove `banner` from the room — the same
+    /// question [`self_removing_ban_reason`] answers, off the subtrees this
+    /// struct already built.
+    fn ban_would_remove(&self, banner: MemberId, target: MemberId) -> bool {
+        banner == target
+            || self
+                .subtrees
+                .get(&target)
+                .is_some_and(|downstream| downstream.contains(&banner))
     }
 
     /// Build the resolved view of a single `deputizer -> deputy` grant.
@@ -464,6 +477,58 @@ fn invite_subtrees(state: &ChatRoomStateV1) -> HashMap<MemberId, HashSet<MemberI
         }
     }
     subtrees
+}
+
+/// Everyone a ban of `target` would remove from the room: `target` themselves
+/// PLUS their entire transitive invite subtree — exactly what the contract's
+/// `check_banned_members` inserts (the banned user) and then extends with
+/// (`get_downstream_members`).
+pub fn ban_removal_set(state: &ChatRoomStateV1, target: MemberId) -> HashSet<MemberId> {
+    let mut removed = invite_subtrees(state).remove(&target).unwrap_or_default();
+    removed.insert(target);
+    removed
+}
+
+/// NOBODY MAY BAN THEMSELVES OUT OF THE ROOM (freenet/river#478).
+///
+/// ONE rule, stated once: compute the set the ban would remove
+/// ([`ban_removal_set`]) and refuse if `banner` is in it. That covers the direct
+/// self-ban (`banner == target`) and the transitive one (`target` is a strict
+/// invite ancestor of `banner`, so the cascade sweeps the banner up with them)
+/// as the single case they are — do NOT split this back into two special cases.
+///
+/// The CONTRACT permits both: `MembersV1::is_ban_authorized` has no self-check,
+/// and step 3 (owner-appointed global moderator) fires before anything that
+/// would stop it, so an owner-appointed deputy is authorized to ban themselves
+/// AND their own ancestors. The resulting ban is fully valid and cascades to the
+/// banner's whole invite subtree. So this is a client-side gate, not a contract
+/// fix — and it must be applied to the OUTPUT of `is_ban_authorized`, never
+/// wired into one branch of its priority ladder, or it would miss exactly the
+/// global moderators who can reach the most members.
+///
+/// The UI enforces the identical rule in
+/// `ui/src/components/members.rs::self_removing_ban_reason`; the two clients are
+/// deliberately kept in step (the rule cannot live in `river-core` without
+/// changing the room-contract WASM, and therefore the contract key).
+///
+/// Returns the user-visible reason when the ban is refused, `None` when the rule
+/// does not apply.
+pub fn self_removing_ban_reason(
+    state: &ChatRoomStateV1,
+    banner: MemberId,
+    target: MemberId,
+) -> Option<&'static str> {
+    if !ban_removal_set(state, target).contains(&banner) {
+        return None;
+    }
+    Some(if banner == target {
+        "Cannot ban yourself. The ban would remove you, and everyone you \
+         invited, from the room."
+    } else {
+        "Cannot ban a member you joined the room through. A ban also removes \
+         everyone the banned member invited, so this one would remove you, and \
+         everyone you invited, along with them."
+    })
 }
 
 /// Render a nickname for TERMINAL output: quoted, with every non-printable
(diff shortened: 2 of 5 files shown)


sanity:
## Review: three independent Claude lenses

Codex is unavailable (usage limit until Jul 28) and the Gemini CLI is deprecated on this machine, so this used three independent Claude reviewers with distinct adversarial lenses, each blind to the others: **completeness/over-broadness of the guard**, **test validity** (specifically: does each test fail when the code it pins is reverted), and **badge spoofing** against the shipped #473 code.

All three agree the fix is correct and not over-broad. Everything they found is fixed below or reported.

### Fixed in `c680be11`

**`execute_ban` re-checked nothing.** It trusts a `member_to_ban` prop, so any future call site passing `can_ban: true` would have run the cascading ban regardless of the render gates. Both the skeptical and testing reviewers independently asked for this. It is now checked at the action boundary, so the capability is unreachable rather than merely un-rendered — which is also what lets the source-scrape be a nicety instead of the guard.

**The source-scrape pin had two constructible false-pass shapes**, both demonstrated rather than hypothesised:

- A guard deleted but still *quoted in a comment* satisfied `rfind`. Now strips line comments first, reusing the shape `util::ecies`'s pin already uses for exactly this reason.
- A **second, unguarded `BanButton`** after the guarded one passed, because only the first occurrence was checked. Now every occurrence must be guarded.

I re-ran both mutants after the fix; both now fail.

**Re-indented the guarded rsx block.** rustfmt does not touch macro bodies and there is no fmt gate in CI, so the body sat at the same column as its guard — hard to read in exactly the place where a future reader must not mis-read the nesting.

### Mutation verification

Every guard in this PR has a pin that was individually verified to fail when *that guard alone* is reverted:

| Guard | Pin | Reverted alone → |
|---|---|---|
| `viewer_can_ban` self-check | `viewer_can_ban_refuses_self` | FAILED |
| rsx render gate | `ban_action_is_not_rendered_for_self` | FAILED |
| `execute_ban` action check | `execute_ban_refuses_self_at_the_action_boundary` | FAILED |
| guard quoted in a comment | `ban_action_is_not_rendered_for_self` | FAILED |
| second unguarded `BanButton` | `ban_action_is_not_rendered_for_self` | FAILED |

The testing reviewer also confirmed no existing test was weakened or made vacuous (the diff is +/−0 on existing assertions; the only deletions are re-indentation), and that both tests run under CI's `cargo test -p river-ui --bins` without `--all-features`.

It also talked me **out** of a Playwright test here: a spec that opens your own member-info modal and asserts no "Ban User" would pass with the fix reverted, because in `example_data.rs` the owner deputizes the *other* member and self gets `deputies: Vec::new()` — so `is_ban_authorized(self, self)` is already false in every example room. Making it honest would require deputizing "(You)", which breaks `member-info-deputy-tag.spec.ts`'s `toHaveCount(1)`. Real cost, no signal.

### Reported, not fixed here

**Banning your own ancestor has the identical blast radius.** `get_downstream_members` is the full transitive closure, so a deputy who bans any strict ancestor removes themselves and their whole subtree too — `is_ban_authorized` step 3 (owner-appointed global mod) grants this before the step-4 guardrail, so it is enabled today. Same harm as #478, different route. It needs a product decision (block it? warn?) rather than a reflexive guard, since banning an ancestor can be legitimate.

**The ban dialog does not disclose the cascade.** It says only "This action cannot be undone" — nothing about the invite subtree going with the target. `get_downstream_members` is computable client-side, so "this will also remove N other members" is cheap and is arguably the fix that closes both the above and the residual.

**The contract still accepts `banned_by == banned_user`.** These three layers are all UI. A stale web container, an old build or a custom client can still trigger the cascade. `riverctl` already refuses self-bans (`cli/src/api.rs:4732`), so the UI was the last open door among our own clients — but the contract-side hole is untracked.

I have not opened issues for these three; flagging to @sanity to decide whether they want them filed or folded into the impersonation follow-up.

### Separately: the spoofing lens found live gaps in the already-merged #473

Reported separately since that code is already deployed — see my message to the team. Headline: `is_display_hidden` misses `U+FFF0..FFF8` and most of plane 14, so `Ian\u{FFF0} Clarke` renders pixel-identical to `Ian Clarke`, is accepted by the nickname input (not riverctl-only), and defeats the @-mention picker's duplicate-name disambiguation — directly relevant to the campaign running right now.

[AI-assisted - Claude]

PR #488, opened 2026-07-25
fix(ui): close invisible-character and tooltip-injection gaps in the badge work

sanity:
## Problem

An adversarial review of the badge/nickname code **#473 already shipped** found four name-impersonation vectors that survived it. They are live (published as 30000357), and the Official room is under an active impersonation campaign, so they are worth closing on their own rather than waiting for the impersonation-marker work.

### 1. ~3,700 invisible characters survived the strip (the important one)

`is_display_hidden` enumerated `U+FFF9..FFFB` and two hand-picked plane-14 sub-ranges, which left `U+FFF0..U+FFF8` and most of plane 14 (`E0080..E00FF`, `E01F0..E0FFF`) through. Every one is Default_Ignorable — HarfBuzz, which backs Chrome, Firefox and Android, treats the whole of plane 14 that way — so it renders as nothing:

```
"Ian\u{FFF0} Clarke"   renders pixel-identical to  "Ian Clarke"
"Ian\u{E0FFF} Clarke"  same
"\u{FFF0}"             a nickname that renders completely blank
```

Two consequences beyond the obvious clone:

- **Not riverctl-only.** `contains_hidden_chars` gates the nickname `<input>`, so it returned `false` for these and the UI accepted them from an ordinary paste-and-save.
- **It defeats the @-mention picker's own anti-collision mechanism.** `duplicate_candidate_names` compares name *strings*, so two members whose names differ only by an invisible character show as two identical `@Ian Clarke` rows with **no** disambiguating member id.

Plane 14 is now covered end to end rather than by enumerated sub-ranges, which is also what stops this recurring the next time someone adds a range.

### 2. Enclosing combining marks rebuilt the glyphs the table strips

`U+2460..U+24FF` (Ⓐ, ①) is stripped, but the enclosing marks that compose the same thing were not:

```
"Mod A\u{20DD}"  → Ⓐ  (enclosing circle)
"Mod !\u{20E4}"  → ! in a triangle, reads as ⚠
"Mod A\u{20E0}"  → 🚫-style prohibition overlay
```

Also `U+10102` AEGEAN CHECK MARK, which renders as ✓ wherever Noto Sans Symbols is installed (stock Ubuntu/Fedora), and the halfwidth clones of the geometric shapes.

### 3. The shield tooltip's appointer quoting was decorative, not protective

The tooltip joins appointers into one flat string and leaves the role labels (`the room owner`, `you`) unquoted precisely so a nickname cannot masquerade as them. But the quote characters were never stripped *from the nickname*, so:

- nickname `Bob”, the room owner, “Carol`
- put a sockpuppet in your `deputies`
- every viewer in your subtree sees `Deputy (appointed by “Bob”, the room owner, “Carol”). Can ban you.`

That unquoted `the room owner` is the forged global-moderator appointment the quoting exists to prevent. Sanitising cannot help — the payload *is* the quote characters. Requires being the victim's invite-ancestor, which an impersonator who invites people satisfies.

### 4. Two smaller ones

- **`deputies` was not deduplicated.** `MAX_DEPUTIES` bounds the length but nothing rejects repeats, so `[sockpuppet; 64]` named the same appointer 64 times and buried the tooltip's real content.
- **Notification previews flattened only `\n`/`\r`,** not `U+2028`, `U+2029` or `U+0085` — leaving exactly the second-line forgery the existing comment describes ("a moderator with a 🛡 in it").

## Testing

`cargo test -p river-ui --bins --all-features` — 555 passed. `cargo fmt` + `clippy` clean.

Five new tests. **Each was verified to fail when its own fix is reverted** — I reverted all four production changes together and confirmed all four tests go red, then restored:

| Test | Reverted → |
|---|---|
| `default_ignorable_characters_cannot_clone_a_name` | FAILED |
| `composed_and_stray_badge_glyphs_are_stripped` | FAILED |
| `appointer_names_cannot_inject_a_role_label` | FAILED |
| `duplicate_deputy_entries_are_collapsed` | FAILED |

The first also asserts `contains_hidden_chars` rejects each character, so the input-side gap is pinned as well as the render-side one.

## What the review could NOT break

Recorded because a clean result is only useful if you know what was tried: every BMP `Emoji=Yes` codepoint is covered; all Nerd Fonts ranges (Pomicons, Powerline, Font Awesome, Codicons, Octicons, Weather, Material Design) are covered via the PUA ranges; bidi cannot detach the badge from its owner because the name and the shield are separate flex items; a member cannot forge the owner's `deputies` because `MemberInfoV1::verify` checks the signature against the key named in `member_id`; and the mention chip escapes both text and attributes.

Known and unchanged: homoglyphs are deliberately not stripped (a Cyrillic `а` is a legitimate letter) — that is what the impersonation *markers* are for, and they follow separately. Mathematical Alphanumeric Symbols (`𝗜𝗮𝗻 𝗖𝗹𝗮𝗿𝗸𝗲`) also survive; they have no legitimate name use and are worth a decision, but stripping them is a behaviour change rather than a gap-closure so I left it out of this PR.

## Scope

UI-only — no `contracts/`, no `common/`, no serialized types, no WASM.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/notifications.rs
+++ b/ui/src/components/app/notifications.rs
@@ -357,6 +357,32 @@ pub fn request_permission_on_first_message() {
     });
 }
 
+/// Every character a renderer may start a new line on, flattened to a space by
+/// [`notification_body`]: LF, CR, U+0085 NEXT LINE, U+2028 LINE SEPARATOR and
+/// U+2029 PARAGRAPH SEPARATOR.
+const NOTIFICATION_LINE_BREAKS: [char; 5] = ['\n', '\r', '\u{0085}', '\u{2028}', '\u{2029}'];
+
+/// Build the `sender: message` body a notification displays.
+///
+/// The body is ONE flat string, so a message that can start a new line lets its
+/// author paint a second line reading exactly like a notification from someone
+/// else, including a moderator with a 🛡 in it. Every line-break character is
+/// therefore flattened to a space. The sender name needs no such treatment: it
+/// arrives already sanitised from
+/// [`crate::util::display_name::display_nickname`].
+///
+/// BOTH notification paths build their body here (the gateway shell bridge in
+/// [`show_notification`], and the direct Notifications API in
+/// [`create_notification_internal`]), so the guard cannot be half-applied and a
+/// third path cannot ship without it. Pinned by
+/// `both_notification_paths_go_through_notification_body`.
+fn notification_body(sender_name: &str, message_preview: &str) -> String {
+    format!(
+        "{sender_name}: {}",
+        message_preview.replace(NOTIFICATION_LINE_BREAKS, " ")
+    )
+}
+
 /// Show a notification for new messages in a room
 ///
 /// # Arguments
@@ -373,16 +399,7 @@ pub fn show_notification(
     // Gateway iframe: hand the notification to the shell (real origin) over the
     // postMessage bridge — the sandboxed iframe can't use the Notifications API.
     if is_in_shell_iframe() {
-        // Flatten line breaks out of the message text. The body renders as
-        // `sender: message`, so a message starting with a newline lets its author
-        // paint a second line that reads exactly like a notification from someone
-        // else — including a moderator with a 🛡 in it. The sender name itself is
-        // already sanitised.
-        let body = format!(
-            "{}: {}",
-            sender_name,
-            message_preview.replace(['\n', '\r', '\u{2028}', '\u{2029}', '\u{0085}'], " ")
-        );
+        let body = notification_body(sender_name, message_preview);
         post_notification_to_shell(room_key, room_name, &body);
         return;
     }
@@ -424,16 +441,7 @@ fn create_notification_internal(
     sender_name: &str,
     message_preview: &str,
 ) {
-    // Flatten line breaks out of the message text. The body renders as
-    // `sender: message`, so a message starting with a newline lets its author
-    // paint a second line that reads exactly like a notification from someone
-    // else — including a moderator with a 🛡 in it. The sender name itself is
-    // already sanitised.
-    let body = format!(
-        "{}: {}",
-        sender_name,
-        message_preview.replace(['\n', '\r', '\u{2028}', '\u{2029}', '\u{0085}'], " ")
-    );
+    let body = notification_body(sender_name, message_preview);
     info!(
         "Creating notification - title: '{}', body: '{}'",
         room_name, body
@@ -815,6 +823,70 @@ mod notify_gate_tests {
             },
             author_sk,
         )
+    }
+
+    /// A notification body renders as one flat `sender: message` line. Any
+    /// character the renderer breaks on lets a message author paint a second
+    /// line that reads as a notification from someone else, which is the exact
+    /// impersonation this guard exists to stop.
+    #[test]
+    fn notification_body_flattens_every_line_break() {
+        for (label, c) in [
+            ("LINE FEED", '\n'),
+            ("CARRIAGE RETURN", '\r'),
+            ("NEXT LINE", '\u{0085}'),
+            ("LINE SEPARATOR", '\u{2028}'),
+            ("PARAGRAPH SEPARATOR", '\u{2029}'),
+        ] {
+            let forged = format!("hi{c}Ian Clarke \u{1F6E1}: everyone is banned");
+            let body = notification_body("Mallory", &forged);
+            assert!(
+                !body.contains(c),
+                "U+{:04X} {label} survived into a notification body: {body:?}",
+                u32::from(c)
+            );
+            assert!(
+                body.starts_with("Mallory: hi Ian"),
+                "U+{:04X} {label} was not flattened to a space: {body:?}",
+                u32::from(c)
+            );
+        }
+    }
+
+    #[test]
+    fn notification_body_leaves_ordinary_text_alone() {
+        assert_eq!(
+            notification_body("Alice", "hello there"),
+            "Alice: hello there"
+        );
+        // The multi-message summary path passes an empty sender name.
+        assert_eq!(notification_body("", "3 new messages"), ": 3 new messages");
+    }
+
+    /// Both notification paths must build their body through
+    /// [`notification_body`]. They are `wasm`-only, so no behavioural test in
+    /// this crate can reach them: re-inlining the `format!` on either one
+    /// re-opens the hole on that path alone and every other test stays green.
+    #[test]
+    fn both_notification_paths_go_through_notification_body() {
+        let source = include_str!("notifications.rs");
+        let (production, _) = source
+            .split_once("mod notify_gate_tests")
+            .expect("this test module's own declaration marks the end of production code");
+
+        assert_eq!(
+            production
+                .matches("notification_body(sender_name, message_preview)")
+                .count(),
+            2,
+            "expected exactly two call sites (`show_notification` and \
+             `create_notification_internal`); a path stopped delegating"
+        );
+        assert!(
+            !production.contains("\"{}: {}\""),
+            "a notification path re-inlined the `sender: message` join, which \
+             skips the line-break strip that only `notification_body` applies"
+        );
     }
 
     #[test]
--- a/ui/src/components/app/notifications.rs
+++ b/ui/src/components/app/notifications.rs
@@ -381,7 +381,7 @@ pub fn show_notification(
         let body = format!(
             "{}: {}",
             sender_name,
-            message_preview.replace(['\n', '\r'], " ")
+            message_preview.replace(['\n', '\r', '\u{2028}', '\u{2029}', '\u{0085}'], " ")
         );
         post_notification_to_shell(room_key, room_name, &body);
         return;
@@ -432,7 +432,7 @@ fn create_notification_internal(
     let body = format!(
         "{}: {}",
         sender_name,
-        message_preview.replace(['\n', '\r'], " ")
+        message_preview.replace(['\n', '\r', '\u{2028}', '\u{2029}', '\u{0085}'], " ")
     );
     info!(
         "Creating notification - title: '{}', body: '{}'",
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -585,16 +585,35 @@ pub(crate) fn viewer_relevant_deputizer_set(
     set
 }
 
-/// Display names of `target`'s deputizers that are RELEVANT to the viewer,
-/// resolved to `"room owner"` / `"you"` / the appointer's decrypted nickname —
-/// exactly what the member-list 🛡 badge shows. An empty result means the
-/// shield does not show for this viewer.
+/// One appointer behind a 🛡 shield, kept as a TYPE rather than a string so a
+/// surface cannot accidentally treat a nickname as a role label.
+///
+/// `Owner` and `You` are trusted: River decides them, and their wording is a
+/// literal in this file. `Member` carries an attacker-chosen nickname and is
+/// the reason this enum exists — see [`DeputyBadge::tooltip`] for what may and
+/// may not be done with it.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub(crate) enum Appointer {
+    /// The room owner. Rendered as the trusted label `"the room owner"`.
+    Owner,
+    /// The viewer themselves. Rendered as the trusted label `"you"`.
+    You,
+    /// Any other member, named by their decrypted `preferred_nickname`.
+    ///
+    /// **Attacker-controlled.** Never interpolate this into a flat sentence
+    /// alongside a role label.
+    Member(String),
+}
+
+/// `target`'s deputizers that are RELEVANT to the viewer, in the order the
+/// member-list 🛡 badge shows them. An empty result means the shield does not
+/// show for this viewer.
 ///
 /// This is the single definition shared by the member-list row and the
 /// member-info modal legend, so the shield's visibility and its
 /// `Deputy (appointed by …)` tooltip cannot drift between the two
 /// (freenet/river#451).
-pub(crate) fn relevant_deputizer_names(
+pub(crate) fn relevant_appointers(
     member_info: &river_core::room_state::member_info::MemberInfoV1,
     room_secrets: &HashMap<u32, [u8; 32]>,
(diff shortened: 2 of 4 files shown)


sanity:
## Test-gap closure (commit `d072c039`)

Mutation testing against `52787223` found three classes of gap where the PR's
production code could be reverted with all 554 tests still green. This commit
closes them. No sanitisation behaviour changes.

### What changed

**`notifications.rs` had no coverage at all.** The `sender: message` body was
built by an inline `format!` in two `wasm`-only functions, so no native test
could reach either one, and reverting the separator strip on either path (or
both) left the suite green. Extracted `notification_body(sender_name, preview)`,
called from both sites, unit-tested directly. That also removes the duplicated
guard: nothing previously stopped a third notification path shipping without it.

**Range endpoints were unpinned.** The existing tests sample interior points, so
they catch a deleted range but not one narrowed by one. Added
`widened_range_endpoints_are_hidden`, covering both ends of all five widened
ranges plus `U+20E0` and `U+20E3` (named in the widening comment, previously
punchable-out).

**Nothing constrained the ranges from above.** Widening `0xFFF0..=0xFFFB` to
swallow `U+FFFD` REPLACEMENT CHARACTER, or plane 14 to `0xEFFFF`, passed every
test. Added `characters_just_outside_the_widened_ranges_stay_visible`.

Also corrected the plane-14 comment: `E0000..E0FFF` is Unicode's
Default_Ignorable set for that plane, not the whole plane.

### Mutation table

Every new assertion was verified live: the production code was reverted and the
test confirmed to FAIL, then restored. 30 mutations, all killed by the intended
test; 2 negative controls confirm the harness can report SURVIVED.

| # | Mutation | Verdict | Killed by |
|---|---|---|---|
| N1 | `show_notification` back to an inline `format!` | KILLED | `both_notification_paths_go_through_notification_body` |
| N2 | `create_notification_internal` back to an inline `format!` | KILLED | `both_notification_paths_go_through_notification_body` |
| N3 | drop `\n` from the strip set | KILLED | `notification_body_flattens_every_line_break` |
| N4 | drop `\r` | KILLED | `notification_body_flattens_every_line_break` |
| N5 | drop `U+0085` NEXT LINE | KILLED | `notification_body_flattens_every_line_break` |
| N6 | drop `U+2028` LINE SEPARATOR | KILLED | `notification_body_flattens_every_line_break` |
| N7 | drop `U+2029` PARAGRAPH SEPARATOR | KILLED | `notification_body_flattens_every_line_break` |
| E1 | `0x20D0..=0x20F0` low end → `0x20D1` | KILLED | `widened_range_endpoints_are_hidden` |
| E2 | `0x20D0..=0x20F0` high end → `0x20EF` | KILLED | `widened_range_endpoints_are_hidden` |
| E3 | punch `U+20E0` out of the range | KILLED | `widened_range_endpoints_are_hidden` |
| E4 | punch `U+20E3` out of the range | KILLED | `widened_range_endpoints_are_hidden` |
| E5 | `0xFFED..=0xFFEE` low end → `0xFFEE` | KILLED | `widened_range_endpoints_are_hidden` |
| E6 | `0xFFED..=0xFFEE` high end → `0xFFED` | KILLED | `widened_range_endpoints_are_hidden` |
| E7 | `0xFFF0..=0xFFFB` low end → `0xFFF1` | KILLED | `widened_range_endpoints_are_hidden` |
| E8 | `0xFFF0..=0xFFFB` high end → `0xFFFA` | KILLED | `widened_range_endpoints_are_hidden` |
| E9 | `0x10100..=0x101FC` low end → `0x10101` | KILLED | `widened_range_endpoints_are_hidden` |
| E10 | `0x10100..=0x101FC` high end → `0x101FB` | KILLED | `widened_range_endpoints_are_hidden` |
| E11 | `0xE0000..=0xE0FFF` low end → `0xE0001` | KILLED | `widened_range_endpoints_are_hidden` |
| E12 | `0xE0000..=0xE0FFF` high end → `0xE0FFE` | KILLED | `widened_range_endpoints_are_hidden` |
| W1 | widen to `0x20CF..=0x20F1` | KILLED | `characters_just_outside_the_widened_ranges_stay_visible` |
| W2 | widen to `0xFFEC..=0xFFEF` | KILLED | `characters_just_outside_the_widened_ranges_stay_visible` |
| W3 | widen to `0xFFEF..=0xFFFD` (swallows `U+FFFD`) | KILLED | `characters_just_outside_the_widened_ranges_stay_visible` |
| W4 | widen to `0x100FF..=0x101FD` | KILLED | `characters_just_outside_the_widened_ranges_stay_visible` |
| W5 | widen plane 14 to `0xDFFFF..=0xEFFFF` | KILLED | `characters_just_outside_the_widened_ranges_stay_visible` |

Plus the exact pre-PR state of each hunk, which is the mutation that actually
matters (a realistic revert, not a synthetic off-by-one):

| # | Pre-PR revert | Verdict | Killed by |
|---|---|---|---|
| R1 | `0x20D0..=0x20F0` → `0x20E3` | KILLED | `widened_range_endpoints_are_hidden` |
| R2 | `0xFFF0..=0xFFFB` → `0xFFF9..=0xFFFB` | KILLED | `widened_range_endpoints_are_hidden` |
| R3 | delete `0xFFED..=0xFFEE` | KILLED | `widened_range_endpoints_are_hidden` |
| R4 | `0x10100..=0x101FC` → `0x101DB` | KILLED | `widened_range_endpoints_are_hidden` |
| R5 | `0xE0000..=0xE0FFF` → `0xE0000..=0xE007F` | KILLED | `widened_range_endpoints_are_hidden` |
| R6 | strip set → `['\n', '\r']` | KILLED | `notification_body_flattens_every_line_break` |

Negative controls, to show the harness is capable of reporting a survivor rather
than reporting KILLED for everything:

| # | Control | Verdict | Expected |
|---|---|---|---|
| C1 | narrow an unpinned pre-existing range (`0x1F000` → `0x1F001`), run the endpoint test | SURVIVED | SURVIVED |
| C2 | revert BOTH call sites, run only `notification_body_flattens_every_line_break` | SURVIVED | SURVIVED |

C2 is the reason the wiring pin exists: the helper's own behavioural test stays
green when both call sites stop using it, so behaviour coverage alone would not
have caught the original gap.

### Notes for review

- `both_notification_paths_go_through_notification_body` scrapes the source via
  `include_str!` and cuts at `mod notify_gate_tests`, not at `#[cfg(test)]` — a
  `#[cfg(test)]` cut point has silently disarmed a source pin in this repo
  before.
- The gap-4 neighbours are deliberate: `U+FFFD` is what `String::from_utf8_lossy`
  emits, and `display_nickname`'s undecryptable-nickname fallback goes through
  it, so a range that swallowed it would blank a whole nickname. `U+FFEF` sits in
  the one-codepoint gap between `0xFFED..=0xFFEE` and `0xFFF0..=0xFFFB`, so that
  single assertion pins the top of one range and the bottom of the other.
- No existing test was weakened, deleted or ignored. 554 → 559 tests.

[AI-assisted - Claude]

sanity:
## Blocking review findings addressed (`d7850d20`, `c6c5b6c2`)

All four items fixed. 566 tests (554 at branch point). 51 mutations run across
three batches, every one behaving as expected; tables below.

### 1. Enclosing marks — the widening missed half its own category

`0x20D0..=0x20F0` caught 7 of Unicode's 13 general-category-**Me** characters.
Added `0x0488..=0x0489`, `0x1ABE`, `0xA670..=0xA672`. `"Mod A\u{A670}"` now
strips to `"Mod A"`.

`every_enclosing_mark_is_stripped` enumerates the **complete** Me category
rather than the six that were missing, so a fourteenth member is a visible
one-line edit rather than a silent gap. The module header now states the Me
exception to "combining marks are preserved" — Mn decorates the preceding
character, Me encloses it, and only the latter draws a badge.

### 2. Tooltip role-label forgery — the quote strip was not a fix

Confirmed: the forging primitive is the **comma**. `Bob, the room owner, Carol`
is plain ASCII with no Unicode trickery and defeats every quote-based defence,
and `title=` is inherently flat so markup cannot help there.

Fixed by removing names from the tooltip entirely. Appointers are now a **type**
rather than a pre-formatted string:

```rust
pub(crate) enum Appointer { Owner, You, Member(String) }
```

`Owner`/`You` are trusted (their wording is a literal in `members.rs`);
`Member` is attacker-controlled. `tooltip()` is built from trusted literals plus
a count and names nobody:

```
Deputy (appointed by the room owner and 2 other members). Can ban you.
```

The names moved to the member-info modal, one DOM element each
(`member-info-deputy-appointers`), where a comma inside a name cannot span two
of them. `appointer_names()` carries a do-not-join contract, pinned by
`modal_lists_appointer_names_as_separate_elements`.

The misleading test is gone. Its replacement,
`no_nickname_content_reaches_the_shield_tooltip`, asserts the tooltip equals the
trusted-literal form and contains no fragment of the payload, across seven
payloads — the plain-ASCII one **first**, since that is the case a
character-denylist fix silently passes.

### 3. Two-word Malayalam names

`.is_none_or(|n| joins_letters(Some(n)) || n.is_whitespace())`. Verified the
Latin clone surface stays shut: the preceding character must still be non-ASCII,
so `"Bo\u{200D}b"` → `"Bob"`. `word_final_joiners_survive_in_multi_word_names`
covers `"മോഹന്‍ കുമാർ"`, the two-chillu form, word-final Persian ZWNJ, and the
single-token form, and re-asserts the three Latin drop cases.

One honest note: this does widen the **non-ASCII** residual already documented in
the module header (an interior joiner can clone a CJK name). Word-final
positions are now also available. No new capability class, more positions for the
existing one; recorded in the code comment rather than left implicit.

### 4. Japanese IVS names — and a gap the first attempt introduced

`U+E0100..U+E01EF` are judged in context like ZWJ/ZWNJ: kept directly after an
ideograph, dropped elsewhere. `contains_hidden_chars` is position-sensitive to
match, so it can no longer reject a name the sanitiser keeps. `辻󠄀`, `邊`, `﨑`,
`髙` round-trip and are accepted at the input.

**My first attempt was wrong and mutation testing caught it.** I split
`is_display_hidden` into `E0000..E00FF | E01F0..E0FFF` with the carve-out
holding the complementary `E0100..E01EF`. Two complementary hand-written ranges
drift: narrowing the carve-out by **one** codepoint left `U+E01EF` in neither —
not stripped, not judged, kept verbatim in every name while rendering as nothing.
A free clone vector, and all 565 tests passed (mutation **B4e**, SURVIVED).

Rather than patch the test, `c6c5b6c2` changes the shape: the strip covers the
whole plane-14 range and the carve-out is **layered on top** in
`sanitize_display_name` step 1 and `contains_hidden_chars`, so the gap cannot
exist. Judging both on the raw string also removes the divergence where the input
check and the render strip could disagree. Added
`no_plane_14_codepoint_escapes_both_the_strip_and_the_carve_out`, a sweep of all
4096 codepoints. B4e is now KILLED.

### Mutation tables

Blocking fixes — production reverted, test confirmed to FAIL, restored:

| # | Mutation | Verdict | Killed by |
|---|---|---|---|
| B1a | drop U+0488/U+0489 from the Me set | KILLED | `every_enclosing_mark_is_stripped` |
| B1b | drop U+1ABE | KILLED | `every_enclosing_mark_is_stripped` |
| B1c | drop U+A670..U+A672 (the reported vector) | KILLED | `every_enclosing_mark_is_stripped` |
| B1d | remove the whole added Me group | KILLED | `every_enclosing_mark_is_stripped` |
| B1e | narrow A670..A672 by one at the low end | KILLED | `every_enclosing_mark_is_stripped` |
| B1f | narrow 0x0488..=0x0489 to 0x0488 | KILLED | `every_enclosing_mark_is_stripped` |
| B2a | tooltip back to the flat comma-joined list | KILLED | `no_nickname_content_reaches_the_shield_tooltip` + 1 |
| B2b | quote-wrap each name and join (the fix that does NOT work) | KILLED | `no_nickname_content_reaches_the_shield_tooltip` + 4 |
| B2c | classify a nickname of `room owner` as trusted `Owner` | KILLED | `deputizer_names_are_sanitised` |
| B2d | phrase says `the room owner` when only members appointed | KILLED | `appointer_phrase_is_built_from_trusted_literals_only` + 2 |
| B2e | modal joins the appointer names into one element | KILLED | `modal_lists_appointer_names_as_separate_elements` |
| B3a | trailing joiner allowed only at end-of-STRING (pre-fix) | KILLED | `word_final_joiners_survive_in_multi_word_names` |
| B3b | allow a joiner after ASCII (reopens the `Bob` clone) | KILLED | `joiners_are_dropped_where_they_only_clone` + 1 |
| B4a | remove the IVS carve-out from sanitize step 1 | KILLED | `ideographic_variation_sequences_are_preserved` |
| B4b | keep EVERY variation selector regardless of context | KILLED | `variation_selectors_not_after_an_ideograph_are_still_dropped` |
| B4c | `contains_hidden_chars` accepts any variation selector | KILLED | `variation_selectors_not_after_an_ideograph_are_still_dropped` |
| B4d | drop CJK Compatibility Ideographs from the ideograph set | KILLED | `ideographic_variation_sequences_are_preserved` |
| B4e | narrow the IVS range by one at VS256 | KILLED | `ideographic_variation_sequences_are_preserved` |
| B4f | re-split the range into complementary halves (behaviour-neutral) | SURVIVED (expected) | — |

B4f is expected to survive: re-splitting alone changes no behaviour. It is B4e
that must be killed, and now is. Before `c6c5b6c2` it was not.

The original brief's 24 + 6 + 2 mutations were re-run against the final tree and
all still behave as expected (tables in the [earlier comment](https://github.com/freenet/river/pull/488#issuecomment-5079482516); `E11`/`E12`/`W5`/`R5` target the
plane-14 range that `c6c5b6c2` restored, and remain KILLED).

Negative controls, to show the harness can report a survivor:

| # | Control | Verdict | Expected |
|---|---|---|---|
| C3 | narrow an unpinned pre-existing range, run the Me test | SURVIVED | SURVIVED |
| C4 | revert the tooltip, run only the Me test | SURVIVED | SURVIVED |

### Note for reviewers

The one existing test I removed is the misleading tooltip assertion, as
directed. Every other test was kept; the badge-shape tests were adapted to
`Vec<Appointer>` mechanically, and the two that asserted quoted nicknames now
assert the stronger property (a nickname classifies as `Member`, never as a
trusted role label). No `#[ignore]`, no weakened assertions.

[AI-assisted - Claude]

PR #489, opened 2026-07-25
feat(ui): warn on names visually confusable with a moderator's

sanity:
## Problem

An impersonation campaign is running against the Freenet Official room: attackers join under nicknames that render identically to Ian Clarke's and to the room's moderators'. Two defenses exist and neither closes it.

- Invisible characters are stripped from nicknames — but a homoglyph is not invisible.
- A cron bans exact visual matches of a protected name — but `Ιan Clarke` with a Greek capital iota (U+0399) is not an exact match. That is how the tenth impersonator got in after the first nine were banned.

**No character-blocking table can fix this.** U+0399 is a letter. So is Cyrillic `а`. Rejecting them breaks real Greek and Russian names — the users least able to argue with a rule that mangles their name. The answer has to be detection and warning, not rejection.

Branch `feat/impersonation-warning-badge` already carried a confusable-detection engine (`ui/src/util/confusable.rs`), but **nothing called it**: `git grep "confusable::"` outside the module returned zero hits, and its doc comments pointed at a function that did not exist. This PR is that missing wiring, plus the work of making the engine actually hold up.

### The protected set comes from live room state

`impersonation_checker_for_viewer` builds it from the room **owner** plus the **same deputy-badge map the 🛡 shield already uses** (`deputy_badges_for_viewer`), taken as a parameter rather than recomputed. There is one answer to "who is a deputy", so a deputize or revoke moves the protected set on the next render with nothing to maintain by hand. Reusing the viewer-relative badge map also scopes it correctly: a protected name is one whose authority *this viewer* would be fooled by, which is exactly the set whose shield they can see.

**Every badged deputy contributes a name, whoever appointed them.** An earlier revision of this PR admitted only owner- or viewer-appointed grants, to keep the set unwritable by an attacker. That gate had to go: in the live Freenet Official room *all five* deputy grants are issued by `Invite Bot` (`PMAQEUP5`), not by the owner (`NRKA4WVX`) — verified with `riverctl debug room-state`. The gate protected exactly one name, `Room Owner`, and left `Ian Clarke`, `HostFat`, `Ivvor` and `ofansifkapital-xmpp` unprotected, i.e. it switched the feature off in the only room that has needed it.

The cost is accepted deliberately and is pinned by a test rather than left to be rediscovered: `MemberInfoV1::verify` validates only the *length* of a `deputies` list, never who is in it, so any strict ancestor of the viewer can deputise a sockpuppet, name it after an innocent member, and have that member render a warning across the attacker's invite subtree. `a_sockpuppet_can_inject_a_protected_name` asserts the current behaviour and says in its doc that re-narrowing the set is a policy change, not a bug fix.

### Identity beats name, always

The exemption is keyed on `ProtectedName::source`, a `MemberId` derived from a keypair, so **a member is never accused of impersonating an identity that is genuinely theirs**. It is deliberately *not* the broader "anyone privileged is immune", which an attacker exploited: a deputised sockpuppet earned a genuine 🛡 and suppressed its own ⚠, and on the author line that is worse than shipping nothing, because the real Ian is the *owner*, whose 👑 renders only in the member list.

### Only tier 1 renders

The engine reports `Identical` (same skeleton) and `NearMiss` (small edit distance). The UI renders `Identical` only, on measured evidence: River assigns 10,000 generated handles and `Amber Worm` / `Ember Worm` are one edit apart, so the near-miss tier accuses a pair of members River itself created, before any attacker acts. Both facts are pinned. `Ivvor2` — a real member — is one edit from the moderator `Ivvor`, and has its own named test row so nobody flips this decision without seeing who it lands on.

### The guards are as wide as the matcher

The filters deciding which names *enter* the protected set compared exact strings while the matcher they gate compares folds — a guard weaker than the thing it guards. `amber worm`, `AmberWorm` and `Arnber Worm` all passed `is_generated_handle` and all tier-1 match the member River assigned `Amber Worm`; no attacker is needed, a moderator styling their own name lower-case is enough. `"Unknown"` was not filtered at all, despite being the no-`member_info` fallback both surfaces render, so one deputy named `Unknown` would flag every member waiting on a sync gap. All three placeholders now compare skeletons.

### The tooltip does not assert things that are false

It switched on the *impersonated* role alone, so it said "this member is NOT a moderator" about members who provably hold a grant: two deputies with colliding names each render 🛡 **and** ⚠, and the room owner can render 👑 alongside the same sentence. It now switches on the flagged member's own privilege and says "Name conflict … check the member ID". The badge is **never** suppressed — a deputised sockpuppet carries a genuine shield, and suppressing on it is the immunity this PR exists to deny.

### Fold coverage

Five classes of total bypass were closed:

| Class | What evaded |
|---|---|
| Math Greek blocks | `% 52` ran past U+1D6A3 where the alignment stops, so `𝝞` (sans-bold capital IOTA) folded to `e`. Wrong both ways: bold Alpha folded to `E`. |
| Latin small capitals | `ɪᴀɴ ᴄʟᴀʀᴋᴇ` — all `Ll`, so both folds left it verbatim at edit distance 8. |
| Precomposed Latin past U+017F | 327 letters, all of Vietnamese. `Iạn Clarke` is a speck at member-list size. |
| Unicode `Cf` invisibles | 13 codepoints survived the sanitiser *and* the fold, so they reached the rendered nickname as a pixel-identical clone. |
| Partial block coverage | Lisu 2 of 26 folded, Cherokee 1 of 35, four Coptic lowercase missing while their capitals folded. |

Cherokee went from 1-of-35 folded to the whole block and Lisu from 2-of-26, which is the largest new false-positive surface here — so the sweep in `foldable_scripts_do_not_collide_with_latin_names` gained entries that *contain* the newly folded letters: `ᏣᎳᎩ` (Tsalagi) carries Ꮃ and Ꭹ, `ꓡꓲꓢꓴ` (Li-Su) carries all four of ꓡ ꓲ ꓢ ꓴ. Cherokee is a living script, but its letters are syllables, so what a Cherokee name folds to is not a word in any language and cannot land on a moderator by accident.

Tables are **generated, not curated**: the Latin ones from NFD, the `Cf` set from the Unicode category, and the shape-homoglyph blocks from UTR39 `confusables.txt` 17.0 — which matters because several Lisu letters are rotated or mirrored Latin capitals, and reading them off a chart gets it wrong in both directions.

The costs are enumerated in `documented_accepted_collisions`, which fails if any row silently stops holding. New rows record what the fold genuinely accepts, including two that hit the real moderator list: `Host Fat` collides with `HostFat` (spaces are removed, which is what catches `IanClarke`), and `Iwor` collides with `Ivvor` via `vv -> w`.

## Testing

650 tests, all green. Every fix has a test that fails without it, and the ones most likely to be pinned by a vacuous assertion were verified **mechanically** — revert the fix, confirm the test goes red, restore:

| Fix | Mutation | Result |
|---|---|---|
| Math Greek bound | restore `0x1D7CB` upper bound | RED |
| Small capitals | disable the fold step | RED |
| Wiring argument (member list) | `member_id` → `self_member_id` | RED |
| Wiring argument (author line) | `author_id` → `self_member_id` | RED |
| Clipping pin | reintroduce `truncate` in a different class position | RED |
| Precomposed Latin | remove both table ranges | RED (2 tests) |
| `Cf` invisibles | remove the `is_format_control` call | RED |
| Appointer gate removal | reinstate the gate | RED (5 tests) |
| `ProtectedRole::Owner` | change it to `Deputy` | RED |
| Citation pin | introduce a citation to a test that does not exist | RED |

The highest-value new test is `the_live_freenet_official_topology_protects_every_moderator`. Every existing fixture deputised its moderators from the *owner*, which passes under both the old policy and the new one — that is precisely why the appointer gate survived review. This one builds the real room's shape (owner deputizes nobody, `Invite Bot` deputizes all four moderators, the bot invited the viewer) and requires each moderator's name to be protected against a homoglyph while each genuine moderator stays unbadged.

`every_cited_pin_test_exists` now makes the most repeated defect in this feature's history fail CI: three separate comments claimed a `shield_and_warning_are_mutually_exclusive` pin, the test was never written, and the invariant it named was provably false — reviewers read the citation and believed the property was covered. I then did it three more times in this PR (two renamed tests, one never written); all three are fixed and the class is now caught mechanically.

Pins that were cited in comments but **never existed** are now written (`a_shield_and_a_warning_can_both_render`, `the_protected_set_is_not_gated_on_the_appointer`), and the comments that cited them — three of which asserted a mutual exclusivity that is provably false — are corrected. `check_name`, which bypasses the identity exemption and would flag every real moderator, is `#[cfg(test)]` rather than `pub` with a comment asking callers not to use it.

The wiring pins previously asserted only that `impersonation_warning_for_display(` was *called*, so the worst available mutation — passing `self_member_id` — compiled, passed everything, and branded the genuine owner and every genuine moderator. They now pin the argument list.

## Known gaps, stated rather than implied

- **Rendered on two surfaces**: the member-list row and the message author line. Not the member-info modal, DM thread, DM rail, mention chips, notifications or reaction tooltips. The doc comment used to claim "every render surface must go through this"; it now enumerates. Since `title=` tooltips do not fire on touch, the badge is currently a bare glyph on mobile — the modal wiring that fixes this is a follow-up.
- **No example-data fixture or Playwright spec yet**, so the badge has never been looked at in a browser. Both surfaces are now addressable — `message-author-impersonation-warning` on the author line and `member-list-impersonation-warning` in the member row — so the spec is unblocked; it is a follow-up, in progress separately.

Follow-up to #488 (invisible-character stripping), which closed the other half of this attack.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -7,8 +7,12 @@ use crate::components::app::{
     MobileView, CURRENT_ROOM, EDIT_ROOM_MODAL, MEMBER_INFO_MODAL, MOBILE_VIEW, NOTIFICATION_MODAL,
     ROOMS,
 };
-use crate::components::members::{deputy_badges_for_viewer, DeputyBadge};
+use crate::components::members::{
+    deputy_badges_for_viewer, impersonation_checker_for_viewer, impersonation_warning_for_display,
+    DeputyBadge,
+};
 use crate::room_data::{NotificationMode, SendMessageError};
+use crate::util::confusable::{ImpersonationChecker, ImpersonationWarning};
 use crate::util::display_name::{display_nickname, sanitize_display_name};
 use crate::util::ecies::{encrypt_with_symmetric_key, unseal_bytes_with_secrets};
 use crate::util::{
@@ -84,6 +88,14 @@ struct MessageGroup {
     /// See [`DeputyBadge`] for the predicate and why visibility and the
     /// "can ban you" wording are separate questions.
     author_badge: Option<DeputyBadge>,
+    /// The ⚠ impersonation warning for this author, or `None`. Same value, and
+    /// therefore the same tier rule and tooltip, the member-list row shows —
+    /// both go through
+    /// [`crate::components::members::impersonation_warning_for_display`].
+    ///
+    /// Mutually exclusive with `author_badge`: a shielded author is a deputy,
+    /// and a deputy is never warned about.
+    author_impersonation: Option<ImpersonationWarning>,
     is_self: bool,
     first_time: DateTime<Utc>,
     /// True if any message in this group had a future timestamp that was clamped
@@ -433,6 +445,10 @@ fn group_messages(
     // Which members show a 🛡 shield in THIS viewer's conversation, built once
     // per render by `deputy_badges_for_viewer`. Absent ⇒ no shield.
     deputy_badges: &HashMap<MemberId, DeputyBadge>,
+    // The ⚠ impersonation checker for this viewer, built once per render by
+    // `impersonation_checker_for_viewer` — passed in rather than built here so
+    // the protected set is folded once, not once per call.
+    impersonation: &ImpersonationChecker,
 ) -> Vec<DisplayItem> {
     let mut items: Vec<DisplayItem> = Vec::new();
     let group_threshold = Duration::from_secs(5 * 60); // 5 minutes
@@ -544,10 +560,16 @@ fn group_messages(
                 group.messages.push(grouped_message);
             }
         } else {
+            // Once per GROUP, not once per message: consecutive messages from
+            // one author share a header, so the warning is a property of the
+            // group.
+            let author_impersonation =
+                impersonation_warning_for_display(impersonation, author_id, &author_name);
             items.push(DisplayItem::Messages(MessageGroup {
                 author_id,
                 author_name,
                 author_badge: deputy_badges.get(&author_id).cloned(),
+                author_impersonation,
                 is_self,
                 first_time: message_time,
                 time_clamped,
@@ -1410,6 +1432,15 @@ pub fn Conversation() -> Element {
                         MemberId::from(&key),
                         self_member_id,
                     );
+                    // The ⚠ impersonation checker, from the SAME badge map, so
+                    // "who is a deputy" has one answer across the conversation
+                    // and the member list. Built once here, like the badges.
+                    let impersonation = impersonation_checker_for_viewer(
+                        &room_state.member_info,
+                        &room_data.secrets,
+                        MemberId::from(&key),
+                        &deputy_badges,
+                    );
                     let groups = group_messages(
                         &room_state.recent_messages,
                         &room_state.member_info,
@@ -1417,6 +1448,7 @@ pub fn Conversation() -> Element {
                         &room_data.secrets,
                         &member_names,
                         &deputy_badges,
+                        &impersonation,
                     );
                     return Some((groups, self_member_id, member_names));
                 }
@@ -2937,6 +2969,31 @@ fn MessageGroupComponent(
                                 });
                             },
                             "{group.author_name}"
+                        }
+                        // Impersonation warning. Sits immediately after the
+                        // name, before the shield slot — the two are mutually
+                        // exclusive (a shielded author is a deputy, and a
+                        // deputy is never warned about), so this is a single
+                        // badge position in practice.
+                        //
+                        // The nickname above cannot forge this glyph: U+26A0 is
+                        // inside the range `display_name::is_display_hidden`
+                        // strips, so it can never survive into a rendered
+                        // nickname. `the_warning_glyph_cannot_appear_in_a_nickname`
+                        // pins that.
+                        if let Some(warning) = group.author_impersonation.as_ref() {
+                            {
+                                let tooltip = warning.tooltip();
+                                rsx! {
+                                    span {
+                                        "data-testid": "message-author-impersonation-warning",
+                                        class: "text-sm cursor-default",
+                                        title: "{tooltip}",
+                                        "aria-label": "{tooltip}",
+                                        {crate::util::confusable::WARNING_GLYPH}
+                                    }
+                                }
+                            }
                         }
                         // Deputy shield. Same glyph, same visibility rule and
                         // the same tooltip as the member-list row and the
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -9,7 +9,7 @@ use crate::components::app::{
 };
 use crate::components::members::{
     deputy_badges_for_viewer, impersonation_checker_for_viewer, impersonation_warning_for_display,
-    DeputyBadge,
+    privilege_in_view, DeputyBadge,
 };
 use crate::room_data::{NotificationMode, SendMessageError};
 use crate::util::confusable::{ImpersonationChecker, ImpersonationWarning};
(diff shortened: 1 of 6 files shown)


sanity:
## Review record — Full tier

**Risk tier: Full.** This is a trust/authority surface (it labels members as impersonating a moderator), the diff is ~3,900 lines, and a false positive publicly brands an innocent member. Two earlier rounds on this branch had already found seven blocking issues, including a badge that could be *aimed* at a chosen victim.

### External model: unavailable

Codex was invoked and is **out of credit** (confirmed by the repo owner, not assumed). Gemini CLI remains deprecated (`IneligibleTierError`). Per `multi-model-review.md`, when every external reviewer is down the substitution is diverse Claude subagents with distinct adversarial lenses — recorded here so the weaker signal is visible rather than silent.

### Substitution: five independent lenses, each blind to the others

| Lens | Focus |
|---|---|
| false-positive | innocent members wrongly branded; non-Latin name corpora |
| bypass/evasion | can an attacker impersonate without tripping the warning |
| weaponization | can the badge be aimed; privileged-set writability |
| test-validity | vacuous fixtures, pin-test integrity, mutation coverage |
| code-first | read code before description; wiring, mobile, perf, stale docs |

### What they found (all addressed)

**Blocking evasions**
- **Mathematical Greek blocks mis-folded.** `(cp - 0x1D400) % 52` is only valid to `0x1D6A3`; past that it runs into five Greek blocks of 58 and emits garbage. `U+1D75E` (MATH SANS BOLD CAPITAL IOTA) folded to `e`, so `𝝞𝗮𝗻 𝗖𝗹𝗮𝗿𝗸𝗲` evaded while the pixel-identical `𝗜𝗮𝗻 𝗖𝗹𝗮𝗿𝗸𝗲` was caught — one character, same generator, same font, defeating the PR's own passing test case.
- **Latin small capitals entirely unfolded** — `ɪᴀɴ ᴄʟᴀʀᴋᴇ` evaded at edit distance 8. These render in every default font stack, and small-caps generators are the most common source of this attack on chat platforms.

**False positives reachable with no attacker at all**
- The guards were **exact-string** while the matcher compares **folds**. A moderator styling their own name lowercase (`cipher daylight`) branded every member auto-assigned that generated handle — people who never typed anything.
- `"Unknown"` — the no-`member_info` fallback rendered for every member mid-sync — was unfiltered, so one protected member named `Unknown` could flag a dozen innocents at once.
- Space **removal** (not collapse) collides genuinely different names: `Mac Donald`/`Macdonald`, `Le Roy`/`Leroy`, `Jo Anna`/`Joanna`, and live in this room, `Host Fat`/`HostFat`.

**Correctness-of-claims**
- Two pin tests cited in doc comments *and* in the PR's own mutation table **did not exist**: `shield_and_warning_are_mutually_exclusive` and `impersonation_checker_uses_the_identity_aware_entry_points`. The first also asserted an invariant the PR's own tests disprove. The second was the only stated guard on a `pub fn check_name` that bypasses the identity exemption — a caller would have put ⚠ on the *real* owner and every real deputy.
- The wiring pins asserted only that `impersonation_warning_for_display(` was called, not **which member id** was passed. Swapping `member_id` for `self_member_id` compiled and passed the entire suite while flagging every genuine moderator.

**Coverage**
- 328 precomposed Latin letters outside `U+00C0..U+017F` (including all Vietnamese) unfolded; 31 `Cf` codepoints survived both the sanitiser and the fold, reaching the *rendered* nickname as a pixel-identical clone; Cyrillic combining marks and the Lisu/Cherokee/Coptic blocks only partially covered.

### The finding that mattered most

The protected set was gated on `Appointer::Owner`, justified in-code by *"That matches the real topology — Ian appoints the moderators."* **It does not.** Every deputy grant in Freenet Official is issued by Invite Bot, so as written this shipped protecting exactly one name — `Room Owner` — and left every actual moderator freely impersonable. The gate is removed; the owner accepted the tradeoff explicitly, and `a_sockpuppet_can_inject_a_protected_name` pins that cost honestly rather than hiding it.

Because the widening also means a sockpuppet's name becomes protected, the *real* moderator gets flagged too. Suppression was deliberately rejected — it would hand a deputised sockpuppet exactly the immunity `a_deputised_sockpuppet_cannot_suppress_its_own_warning` exists to deny. Instead the tooltip's wording is driven by the flagged member's own privilege.

### Verification standard

Every fix was checked by **reverting it, observing the test go red, and restoring** — not by assertion. Confirmed red-then-green for: the math-Greek bound, small capitals, both wiring-argument swaps, `ProtectedRole::Owner`, the clipping pin, both precomposed-Latin ranges, `is_format_control`, reinstating the appointer gate (5 tests), and a deliberately bad test citation.

Two guards were added beyond the findings:
- `the_live_freenet_official_topology_protects_every_moderator` — builds the real room's shape (owner deputises nobody; Invite Bot deputises all four). Every pre-existing fixture deputised from the owner and therefore passed under both policies, which is precisely why the gate survived earlier review.
- `every_cited_pin_test_exists` — fails CI when a comment cites a pin test that does not exist, after three further instances were found in this PR alone. A mechanical check beats discipline for this class.

CI green on `023d982f`. 650 tests.

[AI-assisted - Claude]

PR #494, opened 2026-07-25
feat(ui): explain the impersonation warning where a phone user can read it

sanity:
## Problem

#489 shipped the ⚠ impersonation warning on two surfaces, and on both of them the explanation lived only in a `title=` attribute. **`title=` never fires on touch.** So on a phone the feature conveyed "there is a ⚠ here" and stopped, and the reader's natural next move — tapping the name — opened a member-info modal that said nothing about it at all.

Separately, nobody had ever seen the badge in a browser. `example_data.rs` seeds a spoof nickname so the 🛡 shield is visible in dev builds, but nothing in it produced two members whose names fold together, so `cargo make dev-example` could not render a ⚠ under any circumstances.

## Approach

**The modal now explains the warning in visible text.** It goes through `impersonation_warning_for_display` with `privilege_in_view` — the same entry point the member list and the conversation author line use — so this surface cannot show a badge the others do not, and it inherits the privileged/unprivileged tooltip branch rather than restating it. The chip's label is *derived* from the tooltip's own leading clause ("Impersonation warning" / "Name conflict") instead of being written by hand, so this surface cannot invent wording that drifts from the single definition in `ImpersonationWarning::tooltip`. The imitated name is shown here and nowhere else, as its own DOM node — safe for the same reason `appointer_names` is, since a comma inside an attacker-chosen nickname forges structure in a flat string but cannot span two elements.

This does pay for the whole badge map, which `deputy_badge_for_viewer` deliberately avoids. That cost is documented at the call site: the modal opens on a deliberate click and renders one member, and the alternative is not a cheaper warning but no explanation at all on a phone. The shield still uses the single-target helper so its Playwright-pinned behaviour stays on its own path.

**Example data gains an impostor** so the badge is visible in dev builds and reachable from a browser test. It is a plain member, not a second deputy — a second deputy would break `member-info-deputy-tag.spec.ts`, which asserts exactly one shield, and it would be the wrong shape anyway since the point is a member with *no* authority wearing a moderator's name. Its nickname is derived from the deputy's at fixture-build time rather than hardcoded, because those names are re-rolled every run and a literal `lan Clarke` would collide only on the runs where the generator happened to pick `Ian Clarke`.

## Testing

Five Playwright tests (member-list row, conversation author line, the modal's visible explanation, the deputy's modal, and a 320px layout case) plus three Rust tests. Every new behaviour was checked by mutation rather than by trusting a green.

**The 320px test was vacuous as first written, and passed against the clipping bug.** This is the most useful thing in this PR for whoever reads it next. Reintroducing the regression that `7c2511e8` fixed (`truncate` back on the row button) and measuring the real DOM:

| build | condition | badge right | row right | verdict |
|---|---|---|---|---|
| buggy | plain 320px | 165 | 312 | **passed against the bug** |
| buggy | row squeezed | 230 | 143 | RED |
| clean | row squeezed | 130 | 142 | GREEN |

The fixture's generated nicknames are short, so at 320px the row never overflows, nothing can be clipped, and the assertion was measuring a condition that never occurs. The test now forces the overflow itself and carries a premise assertion that the name is actually being clamped, so it fails loudly rather than silently going vacuous again.

Two further corrections came out of the same exercise, both worth knowing:

- My first premise check measured overflow on the *button*, which is wrong — on a correct layout the button never overflows, because the name clamps instead. That version failed on the clean build.
- The premise originally ran *before* the property assertion. Putting `truncate` back on the button stops it being a flex container, which makes both child spans inline, and `scrollWidth`/`clientWidth` are 0 for inline elements — so the mutation check failed with "the name is not being clamped", which is true but describes a symptom rather than the clipping. The property is now asserted first and the mutation fails with "the ⚠ is clipped past the right edge of its row".

**Adding a flagged member to example data broke an unrelated spec**, and the mechanism is worth recording. `getByRole` matches the accessible name by case-insensitive *substring*, and a member row's accessible name includes the `aria-label` of every badge inside it. The warning's label reads "…but their name **close**ly resembles one", so every row carrying a ⚠ now also answers to `getByRole("button", { name: "Close" })`. The Export Identity spec failed on a strict-mode violation at its Close button, pointing at a member row, with nothing to suggest this feature was involved. Fixed with `exact: true`, which is also simply correct for a button whose entire label is "Close". Found only by running the whole suite — the new spec passed throughout.

Also verified: breaking `confusable_variant` so the names stop folding turns both example-data tests red, and deleting the modal's visible sentence turns the modal pin red. `every_example_room_shows_the_impersonation_warning` asserts the fixture reaches its own premise in all three rooms — deputy badged, checker non-empty, exactly one member flagged, and that member not a badged deputy — which was re-run after the guard changes in #489 landed rather than assumed to still hold.

`AGENTS.md` gains a short note on telling whether a UI test measured the build you think it did: `dx serve` served a stale build during this work and reported a mutation check as passing, which inverts the one technique that proves a browser test is not vacuous.

Full suite green: 649 Rust tests, 5/5 Playwright, `cargo fmt` and prettier clean.

[AI-assisted - Claude]

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,7 +94,9 @@ the wrong thing, all hit in one session (2026-07-25):
 - **A stale `dx` earlier on PATH.** `/opt/rust/bin/dx` was 0.7.3 against the
   crate's dioxus 0.7.9. It refuses to build and serves a 404 "dx is not serving
   a web app" page that reads exactly like a broken change; `~/.cargo/bin/dx`
-  was correct. Check `dx --version` against `ui/Cargo.toml` before debugging.
+  was correct. Check `dx --version` against the RESOLVED dioxus version —
+  `cargo tree -p dioxus` or `Cargo.lock`, NOT `ui/Cargo.toml`, which declares
+  `0.7.3` and would have confirmed the stale binary as correct.
 - **Port 8082 is shared.** It is the suite's default `baseURL`, so if another
   worktree already has a server on it, `npx playwright test` silently tests
   THAT build. Use a free port and set `PLAYWRIGHT_BASE_URL`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,6 +81,24 @@ npx playwright test --project=webkit --grep "iframe"
 npx playwright test --project=mobile-safari --grep "Mobile"
 ```
 
+**Verify what you are actually testing.** Three ways a run silently measures
+the wrong thing, all hit in one session (2026-07-25):
+
+- **`dx serve` can serve a stale build.** The warning above is not theoretical.
+  A mutation check — reintroduce a bug and expect the test to go red — "passed"
+  only because the watcher never rebuilt and the browser was still running the
+  previous binary. That inverts the result of the one technique that proves a
+  test is not vacuous. Prefer the static-server flow above; if you must use
+  `dx serve`, read a marker out of the DOM (a row's `class` attribute is
+  enough) to confirm which build is loaded before believing a pass OR a fail.
+- **A stale `dx` earlier on PATH.** `/opt/rust/bin/dx` was 0.7.3 against the
+  crate's dioxus 0.7.9. It refuses to build and serves a 404 "dx is not serving
+  a web app" page that reads exactly like a broken change; `~/.cargo/bin/dx`
+  was correct. Check `dx --version` against `ui/Cargo.toml` before debugging.
+- **Port 8082 is shared.** It is the suite's default `baseURL`, so if another
+  worktree already has a server on it, `npx playwright test` silently tests
+  THAT build. Use a free port and set `PLAYWRIGHT_BASE_URL`.
+
 **Test coverage:**
 - Desktop 1280px: 3-column layout, no overflow
 - Tablet 768px: narrower sidebars via CSS clamp
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -675,7 +675,7 @@ fn group_messages(
             // Passing `self_member_id` compiles and passes every behavioural
             // test while putting ⚠ on the genuine owner and every genuine
             // moderator; the argument is pinned by
-            // `impersonation_warning_is_wired_into_both_render_surfaces`.
+            // `impersonation_warning_is_wired_into_every_render_surface`.
             let author_impersonation = impersonation_warning_for_display(
                 impersonation,
                 author_id,
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -1177,16 +1177,21 @@ pub(crate) fn impersonation_checker_for_viewer(
 ///
 /// This used to claim "every render surface must go through this", which was
 /// false and is the class of stale comment this file's credibility depends on
-/// not having. The warning is rendered on **two** surfaces today:
+/// not having. The warning is rendered on **three** surfaces today:
 ///
-/// * the member-list row ([`MemberList`]), and
-/// * the conversation's message author line.
+/// * the member-list row ([`MemberList`]),
+/// * the conversation's message author line, and
+/// * the member-info modal, which is the only one that renders the warning's
+///   sentence as VISIBLE TEXT rather than a `title=` tooltip. That matters
+///   because `title=` never fires on touch: on a phone the other two can show
+///   the glyph and nothing else, so the modal is where a reader who taps the
+///   name finds out what it means (freenet/river#494).
 ///
-/// It is NOT rendered on the member-info modal, the DM thread and DM rail,
-/// @mention chips and autocomplete, notifications, or reaction tooltips. Those
-/// show a nickname with no warning beside it. The DM thread is the most
-/// consequential of the gaps: it is a 1:1 surface, which is where a victim is
-/// most likely to act on perceived authority.
+/// It is NOT rendered on the DM thread and DM rail, @mention chips and
+/// autocomplete, notifications, or reaction tooltips. Those show a nickname
+/// with no warning beside it. The DM thread is the most consequential of the
+/// gaps: it is a 1:1 surface, which is where a victim is most likely to act on
+/// perceived authority.
 ///
 /// Adding a surface is a one-line call to this function plus the glyph; the
 /// reason not to do it blindly is that each surface needs its own
@@ -1423,7 +1428,7 @@ pub fn MemberList() -> Element {
             // `self_member_id` here compiles, passes every behavioural test, and
             // puts ⚠ on the genuine owner and every genuine moderator — the
             // argument is pinned by
-            // `impersonation_warning_is_wired_into_both_render_surfaces`.
+            // `impersonation_warning_is_wired_into_every_render_surface`.
             let impersonation_warning = impersonation_warning_for_display(
                 &impersonation,
                 member_id,
@@ -6894,7 +6899,7 @@ mod tests {
     /// `display_name::nickname_render_paths_go_through_display_nickname`
     /// documents).
     #[test]
-    fn impersonation_warning_is_wired_into_both_render_surfaces() {
+    fn impersonation_warning_is_wired_into_every_render_surface() {
         let members_src = include_str!("members.rs");
         let conversation = include_str!("conversation.rs");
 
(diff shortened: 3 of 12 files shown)


sanity:
## Review — Light tier + adversarial read

**Tier: Light.** Additive UI + fixture + tests, no protocol/contract/wire surface, no WASM change (both migration checks pass untouched). But the modal wiring is non-trivial logic on a trust surface, so it got a full adversarial read rather than a rubber stamp.

**External model unavailable** — Codex is out of credit; Gemini CLI remains deprecated. Claude-subagent substitution, same as #489. Recorded so the weaker signal is visible.

**Verdict: no BLOCKING findings.** Three IMPORTANT, addressed before publish.

### Clean, and verified rather than assumed

- **The modal passes the right member.** `member_id` is the modal target; `target_nickname` uses the *identical* `canonical → display_nickname` derivation `MemberList` uses, so the two surfaces cannot disagree on the string they compare. `self_member_id` appears nowhere in the warning path. This was checked specifically because #489's reviews found that swapping it at a render site compiles, passes everything, and puts ⚠ on the genuine owner and every real moderator.
- **Attacker-controlled text is contained.** No `dangerous_inner_html` anywhere in the file (#227 stands). The imitated name is its own DOM node, never interpolated into a sentence, and arrives already glyph-stripped so it cannot forge a 🛡 or 👑. The chip's label is *derived* from the tooltip (`split(':').next()`) rather than hand-written, so it follows the privilege branch automatically and this surface cannot invent wording that drifts.
- **The `exact: true` fix is complete.** The sweep was re-done independently and wider than the original: **121** name-based locators across the suite — `getByRole`, `getByText`, `getByLabel`, `getByTitle`, `getByPlaceholder`, `filter({hasText})` — matched against *all three* tooltip wordings. Exactly one collision, `"Close"`, matching both "Impersonation warning" wordings via "**close**ly resembles". Anchored patterns are structurally immune.
- **The derived-impostor rule is provably tier-1.** Branches 1 and 2 hold for any input, because `I` and `l` both map to the same sentinel in `fold_bar_class` and every later step then operates on byte-identical strings. Branch 3 holds for ASCII, and both name pools are verified pure ASCII.
- **No existing spec's counts break.** Each was checked individually, including the one asserting exactly one 🛡.
- **Four of the five new specs cannot pass with the feature broken.** The fifth is a pure negative, sound because it is paired with a positive in the same file — and it is what catches the id swap end-to-end.
- **Perf at ~135 members is fine.** The modal builds the deputizer maps twice per render, all O(members); well under a millisecond, and smaller than a sweep the modal already ran.

### Fixed before publish

1. **The modal's 4th argument was pinned by nothing.** The other two surfaces each carry an argument pin whose own comment calls the id swap "the worst mutation this feature has". In this fixture both the impostor's and the viewer's `privilege_in_view` are `None`, so swapping `member_id` → `self_member_id` there changed nothing observable in any test. The failure it allowed is the sharp one: a viewer who is themselves badged — **every moderator in Freenet Official** — opens an impostor's modal and reads *"Name conflict: … Both are real; check the member ID"*, the wording for a legitimate clash, shown for a genuine impersonation, to the people most likely to act on it.
2. **The authoritative surface list said "two surfaces" and named the modal among those NOT rendering it** — in the same paragraph that congratulates itself for not being the stale comment this file's credibility depends on. This PR makes the modal the third.
3. **The uppercase "fallback" was the majority case.** Exhaustively: branch 1 6.5%, branch 2 42.1%, uppercase **51.4%**. More than half of dev-example runs showed `BOB SMITH (MEMBER)` flagged against `Bob Smith (Member)` — correct by the fold, but it reads as a false positive and trains developers to distrust the badge, the inverse of the fixture's purpose.

### Worth carrying forward

**A badge's `aria-label` becomes part of its whole row's accessible name**, and Playwright matches names by case-insensitive substring. That is how a member row came to answer to `getByRole("button", {name: "Close"})` and broke an unrelated Export Identity spec — a failure pointing at a member row with nothing to suggest this feature existed. Now that the sentence is also *visible text*, `getByText`/`hasText` are collision surfaces too. The mitigation note lives on `ImpersonationWarning::tooltip`, where someone rewording it will actually see it.

Also noted, not a finding: `playwright.config.ts` sets `retries: 2`, so a genuinely flaky spec reports green as "flaky" — worth remembering when reading green CI here as evidence.

### On the 320px test

It was **vacuous as first written** and passed against the reintroduced clipping bug: the fixture's generated names are too short to overflow at 320px, so the assertion measured a condition that could not occur. Found only by mutating. It now forces the overflow and asserts the name is actually clamping as a premise, and the collision was re-demonstrated on the static release harness — the one CI uses — not just through a dev server.

[AI-assisted - Claude]

sanity:
## Review round (post-rebase)

Rebased onto current `main` (which had moved past this branch's base twice — #517 and #518). Codex is quota-exhausted, so per the project's review rule this ran as two independent blind Claude lenses on the rebased head: a **skeptical** adversarial pass and a **code-first** pass (code read before the description). Neither found a blocker. Everything below is fixed in `9d5c0db8`.

### The rebase itself

Three conflicts in `ui/src/example_data.rs` against main's #501/#505 windowing fixtures: two "both sides added a parameter to `add_example_messages`" and one "both sides added tests". Both kept.

The consequential one: the guaranteed impostor message takes the standard message block from 12 to 13, which pushed the at-cap fixture room one message over its cap and turned `at_cap_room_is_exactly_at_its_message_cap` red. `AT_CAP_MAX_RECENT_MESSAGES` is raised 160 → 161 rather than shrinking the fillers, which must stay an even count to keep the same-author pairing #505 depends on. Both reviewers independently checked the arithmetic (148 fillers + 13 = 161) and that main's windowing invariants survive: 148 paired fillers still give 74 display items past the 60-item window, and the deep room keeps its 80-message prune headroom.

### Findings fixed

**`confusable_variant` could mangle the role suffix instead of the name.** The digit-swap loop scanned the whole string, and `" (Member)"` contains an `e` — so a name with none of `o/e/s/a/t` would have produced `Quinn Zhu (M3mber)`, which reads as a broken label rather than a spoofed name, while satisfying every existing assertion. `FIRST_NAMES` already contains `Finn` and `Quinn`; this was one last-name addition away. The search is now constrained to the name, and the test asserts the role suffix comes through untouched.

**The fixture test sampled where it could enumerate.** Two 500-draw loops covered ~40% of the 1,058-name product per run, so a pool regression would have surfaced as an intermittent CI failure — a broken test by this repo's rules — while the doc claimed the property was measured exhaustively. Replaced with one exhaustive pass over a new test-only `all_full_names()`.

**A cross-surface pin assertion was unfailable.** `members.rs` scanned the whole of `member_info_modal.rs` for `impersonation_warning_for_display(`, which also appears inside that file's own argument pin — so deleting the production call could not fail it. Now cut at `#[cfg(test)]`.

**The chip label's colon was load-bearing and unpinned.** The modal derives its label from the tooltip's leading clause; a reword that dropped the colon would silently turn the chip into a paragraph. Documented on `tooltip()` and pinned by `tooltip_leads_with_a_short_label_before_a_colon`.

**`deputy_badge_for_viewer`'s doc became false.** It claimed the modal avoids the full badge sweep; since this PR the modal builds the map too. The doc now says so, records that the duplicate sweep is accepted on a click-opened single-member surface, and names the equivalent one-line cleanup. *Not fixed here:* re-plumbing the 🛡 shield to read from that map. It is pinned by a source test and two Playwright specs, and changing its derivation path is a separate change to a separate feature.

**`AGENTS.md` sent you to the wrong file.** It said to check `dx --version` against `ui/Cargo.toml`, which declares dioxus `0.7.3` — so following it with the stale `dx 0.7.3` on PATH confirms the exact broken binary the note exists to catch. Now points at `cargo tree -p dioxus` / `Cargo.lock`.

Plus: three sibling comments still saying "12 standard messages"; a Playwright deputy-name read that stripped a fixed glyph list out of the whole row (fixture-dependent badges would poison it); an over-promising assertion message about the owner, who is deliberately absent from `members.members`; and `invite-via-dm-picker.spec.ts`'s "no two members share a nickname prefix" comment, which this PR invalidated and which now gives the reason that actually holds.

### New coverage, and a vacuous test caught by mutation

The PR's premise is that `title=` never fires on touch — yet every modal test pinned itself to 1280px, so nothing would have noticed if the modal were unreadable at the width the feature is for. Added a 320px modal test.

Its first version **passed against the bug.** Deleting `max-w-full break-words` from the imitated-name chip and rebuilding left it green: an ancestor clips the page, so `document.documentElement.scrollWidth` never grows past the viewport. That is the same vacuity the 320px row test was written to avoid, made again one test over. What does move is the chip itself — with the rule removed its right edge measures ~2375px against a 320px panel, so the imitated name is simply not on screen. The assertion now compares those two edges, and is red on the mutated build and green on the clean one.

### Verified and ruled out

- **The `aria-label` collision sweep was re-run mechanically** across all 24 specs — every `getByRole`, `getByText`, `filter({hasText})`, `toContainText` and `getByTitle` — against all three tooltip wordings and the modal's new visible strings. `"Close"` remains the only collision, and it is fixed with `exact: true`.
- No `dangerous_inner_html` in the new markup; the imitated name is an escaped Dioxus interpolation in its own DOM node, so the comma-forgery class from #488 stays unreachable.
- No existing fixture assumption broken by the fourth member: no spec asserts a fixed member count, the exactly-one-shield assertion holds (the impostor is a plain member), and `room_state.verify()` passes at 161 == cap.

Lenses: skeptical (adversarial), code-first (blind to the description). 705 tests pass with `--features example-data,no-sync`, 700 without; 6/6 Playwright locally against a static example-data build.

[AI-assisted - Claude]

Issue #487, opened 2026-07-25
fix(dm): inbound DMs should systematically fail to auto-scroll (is_near_bottom read after the bubble is in the DOM)


sanity:
Found while fixing #486 (the conversation-view scroll latch). Filing rather than folding it into that PR, since it is a different component with a different mechanism.

## The problem

`ui/src/components/direct_messages/dm_thread_modal.rs` decides whether to auto-scroll with:

```rust
let near_bottom = is_near_bottom("dm-scroll-container", 50.0);
let should_scroll = is_first || outbound_changed || near_bottom;
```

`is_near_bottom` is called from **inside the effect** — that is, *after* the new bubble is already in the DOM. So the distance it measures already includes the height of the message we are deciding whether to scroll to.

Median rendered message height in River is ~62px, and the tolerance is 50px. A typical inbound DM therefore pushes the measured distance past the tolerance before the check runs, so `near_bottom` reads false and the scroll is skipped — for the ordinary case, not an edge case.

## Why it is not obvious in use

The `outbound_changed` bypass means your own messages always scroll, which masks it: the thread looks like it follows the conversation, because most of the time you are the one talking. Inbound-only bursts are where it should show.

Same class of mistake as #486 cause 1: measuring "is the end of the history on screen right now" *after* the update, and treating the answer as "was the reader following the conversation before it".

## Not verified against a running build

This is a code reading, not a reproduction — I have not driven a DM thread and counted missed scrolls. Worth confirming before fixing, since the exact behaviour depends on the rendered height of the specific message and on where the effect lands relative to layout.

## Fix direction

Same shape as #486: decide from the position *before* the update, or track a reader-intent pin that only reader-initiated scrolling can change, rather than re-measuring the DOM after the content has already grown.

[AI-assisted - Claude]

sanity:
Notes from fixing #486 (PR #495, merged as `bfbe9793`), while it is fresh. The
two bugs are the same species, but less of the fix transfers than it looks like,
and the parts that do not are the ones that would bite.

## The diagnosis here is right, and there is a second defect alongside it

`is_near_bottom("dm-scroll-container", 50.0)` is read inside the effect, so it
measures a DOM that already contains the bubble it is deciding whether to scroll
to. That is the same mistake as #486's cause 1.

But `dm_thread_modal.rs:349` still triggers off `last_dm_bubble`, an `onmounted`
on the last bubble — which is #486's cause **2**, unfixed. Any content change
[...]

Issue #468, opened 2026-07-25
perf(ui): composer textarea auto-resize forces ~3 layouts per keystroke


sanity:
Split out of #466. This is the one part of the original user report that really is about layout/reflow rather than reactivity.

## Observation

`auto_resize_message_input` (`ui/src/components/conversation/message_input.rs`) runs on every `oninput`:

```rust
el.style().set_property("height", "auto").ok();
let new_height = el.scroll_height().min(168);
el.style().set_property("height", &format!("{}px", new_height)).ok();
```

Write style, read `scrollHeight`, write style — a forced synchronous layout on every keystroke.

Measured over 26 keystrokes on a 90-message room (Chromium, CDP `Performance.getMetrics`):

| | unthrottled | 6x CPU throttle |
|---|---|---|
| LayoutCount | 79 (~3 per keystroke) | 79 |
| LayoutDuration | 9.1 ms | **67 ms** |
| RecalcStyleDuration | 12.8 ms | **108 ms** |
| TaskDuration | 111 ms | **789 ms** |

That is ~6.6 ms of layout + style per keystroke at 6x throttle, ~30 ms of total main-thread work per keystroke. Unchanged by #466, which addressed the scripting cost elsewhere — the typing path was already isolated from the message list by `MessageInput` owning its own text signal.

## Context

This is a real cost but a modest one, and it is the *only* place in the app where layout is a meaningful share of the work. For comparison, sending a message spends ~1.6% of its time on layout+style and ~81% on scripting.

The user report that prompted #466 guessed the problem was flexbox reflow. It mostly was not — but this is the piece where that instinct points at something real.

## Possible directions

- Skip the write when the computed height is unchanged (avoids the style invalidation; the read still forces layout).
- Only reset to `auto` when the content could have shrunk, rather than on every input.
- Measure against a hidden mirror element so the live textarea is not written twice per keystroke.

Any change here needs care: textarea sizing has Playwright coverage (`"textarea height resets after sending a multi-line message"`, `"edit container fits within viewport at narrow widths"`).

[AI-assisted - Claude]

sanity:
This is a correctness bug as well as a perf one, which I don't think the issue currently captures.

[...]

PR #495, opened 2026-07-26
fix(ui): stop the conversation view latching away from the newest message

### LINKED ISSUE #486: fix(ui): conversation stops following new messages (permanent latch, not intermittent)
## Symptom

New messages arrive in a room and the view does not scroll to reveal them. It reads as intermittent, but it is a **permanent latch**: once it triggers, the view never follows again until the reader manually scrolls back to the bottom or switches rooms.

Reproduced live on the Freenet room with CDP + rAF sampling + a MutationObserver.

## Recorded evidence

After the gate latched at t=60s:

- **54 further arrivals**, 35 of which fired the scroll trigger, and **zero scrolled**.
- Distance from the bottom ratcheted 147 → 285 → 387 → … → **2725px** (about 3.2 viewport heights).
- The gate flipped exactly as designed: open at 95px, closed at 105px. The design is the bug.

## Cause 1 — the gate latches false forever (dominant)

`should_scroll = forced || *is_at_bottom.peek()`.

`is_at_bottom` is written by an IntersectionObserver on a 1px sentinel with `rootMargin: "0px 0px 100px 0px"`, deferred through `setTimeout(0)`. That signal means *"is the sentinel in view right now"*, **not** *"was the reader following the conversation before this update"*. Once the gap exceeds 100px it reads false, and nothing re-arms it except a manual scroll back down or a room switch.

## Cause 2 — the trigger can silently miss

`last_chat_element` is written only by `onmounted` on the last bubble, and the effect's only reactive dependency is `last_chat_element()`. An update that changes content height **without remounting the last bubble** is invisible to the effect.

Note the direction: a remount is the *working* path. The broken path is when the last row is **not** remounted — mid-list insertion, a merged join summary (joins within 1h merge into a summary keyed on the FIRST event's id, so no remount), reactions, edits, late-resolving heights.

## Cause 3 — the composer latches it with no network involved

Growing the composer to its 168px maximum shrinks the chat container 846 → 726px, moving the distance 0 → 120px, past the 100px margin. Typing a long message latches the gate on its own.

This makes #468 a **cause** of this bug, not merely a performance cost.

## Fix direction

Both halves are wrong in the same direction; fixing either alone leaves the bug standing.

1. **Replace the gate.** Keep the IntersectionObserver for the scroll-to-latest *button*, where its semantics are correct. Gate auto-scroll on a separate "reader is pinned to the bottom" flag that changes only on reader-initiated scrolling, and that ignores settles caused by our own programmatic scrolls.
2. **Replace the trigger.** Fire on any content change rather than on an `onmounted` of the last bubble.

## Related

- #468 — composer auto-resize; cause 3 above makes it a functional bug, not only a perf cost.
- #466 — message-group memoization. Does not conflict and does not fix this; memoization does not suppress mounts. It narrows the race window only, so the failure should look identical after it lands.

[AI-assisted - Claude]

sanity:
## Problem

New messages arrive in a room and the view does not follow them. It reads as
intermittent, but it is a **permanent latch**: once it triggers, the view never
follows again until the reader manually scrolls back to the bottom or switches
rooms. Recorded live on the Freenet room with CDP + rAF sampling: after the gate
latched at t=60s there were **54 further arrivals, 35 of which fired the scroll
trigger, and zero scrolled**, while the distance from the bottom ratcheted
147px → 285 → 387 → … → **2725px**, about 3.2 viewport heights.

Two independent defects, both in the same direction, so fixing either alone
leaves the bug standing.

**The gate.** `should_scroll = forced || *is_at_bottom.peek()`. `is_at_bottom` is
written by an IntersectionObserver on a 1px sentinel with a 100px `rootMargin`.
That signal answers *"is the end of the history on screen right now?"*, not *"was
the reader following the conversation before this update"*. Once the gap exceeds
100px it reads false and nothing re-arms it. Growing the composer to its 168px
maximum shrinks the history by more than the margin on its own, so typing a long
message latched it with no network activity at all — which is why #468 is a
**cause** of this bug rather than only a performance cost.

**The trigger.** `last_chat_element` was written only by an `onmounted` on the
last bubble, and that was the effect's only reactive dependency. Every content
change that leaves the last row in place was invisible to it: a mid-list insert,
a reaction, an edit, or another join folding into an existing "N people joined"
summary (keyed on its FIRST event, so absorbing a new one grows it without
remounting anything).

## Approach

**Gate → a pin that tracks intent.** A `scrollend` listener (a debounced `scroll`
where `scrollend` is missing) re-measures the distance once per settle. Whose
settle it is, is decided by **position**: `last_scroll_top` holds the offset our
last scroll asked for, a settle that lands there is ours and leaves the pin
alone, a settle anywhere else is the reader's.

Deciding it with a "this settle is ours" flag was tried first and does not
survive two settles being in flight at once. Clearing a long draft grows the
container, the browser clamps `scrollTop` on its own, and that clamp's settle was
consumed by the flag the previous scroll had raised — the recorded offset then
pointed somewhere the view had already left, and both follow paths stood down.
Caught by the new suite on mobile Safari (62px short of the newest message, still
short 12s later), not by reading the code. Position also needs no
`wheel`/`pointerdown` listeners to guess at intent, so reader scrolls that
produce no gesture event (native scrollbar drag on Firefox, find-in-page,
focus-driven scrolling) stop being a hole.

The IntersectionObserver stays exactly where its semantics are correct — the
scroll-to-latest button — and its `rootMargin` is now built from the same
constant, so the doc claim that the two agree about where the bottom is became
true.

**Trigger → any content change.** The effect subscribes to the grouped-message
memo, and a ResizeObserver covers the size changes no re-render can see. It
watches **both** the content wrapper and the scroll container, because they see
different halves: the wrapper sees the history reflowing (late-loading image,
font swap, rewrap on a narrower screen), and only the container sees the *window*
shrinking, which is what the composer does. De-latching the gate alone would
have left the newest message below the fold after a long draft until something
else happened.

**Resolving the pin takes a `scrollend`**, which arrives milliseconds after the
reader actually moves, so both scroll paths also refuse to act over a reader
scroll the pin has not seen yet, and the deferred one re-reads the pin when it
runs rather than trusting what its gate saw. Both reproduced on Firefox, WebKit
and mobile Safari while Chromium stayed green.

**Also here, because the new trigger depends on it:** `Utc::now()` was read per
message, per render, inside the grouping loop as the clamp target for
future-dated timestamps, so a clock-skewed message re-timed itself on every
render. The memo is now the auto-scroll trigger, so it has to be a pure function
of the content rather than of when the render happened. It clamps against when
this client first saw the message instead. Two consequences are handled rather
than left to surprise someone: the old comparison carried the whole propagation
delay as implicit slack and the new one carries none, so a minute of tolerance
keeps a peer with a slightly fast clock from having every message marked "clock
may be wrong"; and the clamp target is now persisted data, so `receive_times`
rejects values that are not plausible arrival times.

## Testing

Six Playwright tests across all five browser projects (30 cases), plus Rust unit
tests for the clock change and source-grep pins for the wiring.

**Verified two ways.** Against the pre-fix code, all 30 fail — but only the three
#486 tests fail at their own assertions:

| assertion | cases |
|---|---|
| the composer grew over the newest message and the view did not follow it | 5 |
| content grew above the last row and the view did not follow it | 5 |
| the history reflowed taller and the view did not follow it | 5 |
| filler messages should have been followed (**setup** of the 3 guard tests) | 15 |

The guard tests constrain the *fix*, not the bug — the pre-fix code also refuses
to scroll a parked reader, for the wrong reason — so a revert says nothing about
their teeth. Those were established by **mutating the fix** instead: making the
pin always true fails all three; making the settle handler able to clear the pin
but never re-arm it fails the reader-returns leg specifically.

Every test asserts its own premise rather than assuming it. That is not
decoration: at 1280 → 700 this fixture reflows by exactly **zero** pixels, so the
reflow test as first written passed without the view ever having been pushed off
the bottom. It narrows to 380 now and fails loudly if the history does not get
taller. The composer test measures the *window* rather than the gap for the same
reason in reverse — under the fix the gap closes again immediately, so "the view
is off the bottom" is a precondition a complete fix erases.

Delivery goes through `window.__riverTest`, installed only in builds with example
data **and** sync switched off (it writes messages signed by a non-member key
straight into `ROOMS`). The composer is not a substitute: that path raises
`force_scroll` and deliberately bypasses the pin.

Full local runs on the branch: `cargo test -p river-ui --bins` 660 passed; full
Playwright suite 485 passed / 0 failed / 0 flaky; the autoscroll spec 4x30 clean,
and the previously flaky composer case 8x clean (it was failing ~1 in 3 before
the position-based redesign). No new clippy warnings against `main`.

## Review

Codex is out of credit, so this had three independent Claude adversarial reviews
with distinct lenses (skeptical bug-hunt, code-first, test-quality) instead. All
findings are addressed in the diff; the ones that changed behaviour were the
composer-resize gap, the `time_clamped` widening and its now-inaccurate tooltip,
the persisted-clamp-target validation, and the `rootMargin` constant. One review
claim was checked and found false — the composer-clear leg does not pin
`reader_moved_up_since`'s `.min(max_scroll_top(..))` (deleting it, the test still
passes) — and the comment now says so rather than claiming coverage it lacks.

Not folded in: #487 (the DM thread modal has the same class of bug, filed
separately and deliberately kept out of this PR).

Closes #486

[AI-assisted - Claude]

### DIFF
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -35,10 +35,12 @@ dioxus-free-icons = { version = "0.10.0", features = ["font-awesome-solid"] }
 
 # Web-related
 web-sys = { workspace = true, features = [
+    "AddEventListenerOptions",
     "Clipboard",
     "Document",
     "DomTokenList",
     "Element",
+    "Event",
     "EventTarget",
     "CssStyleDeclaration",
     "HtmlAnchorElement",
@@ -53,6 +55,9 @@ web-sys = { workspace = true, features = [
     "Notification",
     "NotificationOptions",
     "NotificationPermission",
+    "ResizeObserver",
+    "ScrollBehavior",
+    "ScrollToOptions",
     "Storage",
     "VisibilityState",
     "Window",
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -99,6 +99,13 @@ pub fn App() -> Element {
     // Must be called from within a Dioxus component where the runtime is active.
     crate::util::capture_runtime();
 
+    // Playwright's inbound-message hooks. `example-data` builds that also have
+    // sync switched off, and nothing else — the hooks mint messages signed by a
+    // non-member key, so they must never reach a real node. See
+    // `example_data::install_test_hooks`.
+    #[cfg(all(feature = "example-data", feature = "no-sync"))]
+    crate::example_data::install_test_hooks();
+
     // In the gateway iframe, notifications are shown by the shell page; listen
     // for its `notification_click` replies to route to the clicked room
     // (freenet/river#408). No-op when served as a top-level page. Installs once.
--- a/ui/src/components/app/receive_times.rs
+++ b/ui/src/components/app/receive_times.rs
@@ -38,11 +38,32 @@ fn parse_map(s: &str) -> HashMap<i64, f64> {
         let mut parts = pair.splitn(2, ':');
         if let (Some(k), Some(v)) = (parts.next(), parts.next()) {
             if let (Ok(key), Ok(val)) = (k.parse::<i64>(), v.parse::<f64>()) {
-                map.insert(key, val);
+                if is_plausible_arrival_ms(val) {
+                    map.insert(key, val);
+                }
             }
         }
     }
     map
+}
+
+/// Is `ms` a time this client could plausibly have received a message at?
+///
+/// These values come back out of `localStorage`, which anything running in the
+/// origin can write, and they are no longer only a diagnostic: `group_messages`
+/// clamps a future-dated sender timestamp to the arrival time, so a garbage
+/// entry would move that message's rendered time, its 5-minute grouping and its
+/// day separator. A NaN, a negative, or a value from before this project
+/// existed is not an arrival time; neither is one in the future, since we
+/// record arrivals as they happen. Rejecting them here means every reader gets
+/// a sane map rather than each one re-checking.
+#[cfg(target_arch = "wasm32")]
+fn is_plausible_arrival_ms(ms: f64) -> bool {
+    /// 2020-01-01, comfortably before any River build existed.
+    const EARLIEST_PLAUSIBLE_MS: f64 = 1_577_836_800_000.0;
+    // A little slack for a clock that ticks between the write and the read.
+    const FUTURE_SLACK_MS: f64 = 60.0 * 1000.0;
+    ms.is_finite() && ms >= EARLIEST_PLAUSIBLE_MS && ms <= now_ms() + FUTURE_SLACK_MS
 }
 
 /// Serialize to "key:value,key:value,..." format
@@ -121,10 +142,30 @@ pub fn record_receive_times(message_ids: &[MessageId]) {
     });
 }
 
+/// A snapshot of the first-seen times, taken once per grouping pass.
+///
+/// `group_messages` needs the same map for every message it walks, and reading
+/// the `RECEIVE_TIMES` `GlobalSignal` per message both costs a borrow each time
+/// and makes the function unusable outside a Dioxus runtime (so untestable).
+/// Taking the reference once and threading it through fixes both.
+pub type ReceiveTimes = HashMap<i64, f64>;
+
+/// When this client FIRST saw a message, in ms since the epoch.
+///
+/// This is local wall-clock at arrival, so unlike the sender-supplied
+/// timestamp it cannot be in the future and cannot be moved by a remote clock.
+pub fn first_seen_ms(times: &ReceiveTimes, message_id: &MessageId) -> Option<f64> {
+    times.get(&message_id.0 .0).copied()
+}
+
 /// Get the propagation delay for a message, if known.
 /// Returns delay in seconds, or None if unknown or negative (clock skew).
-pub fn get_delay_secs(message_id: &MessageId, send_time_ms: i64) -> Option<i64> {
-    let recv_ms = *RECEIVE_TIMES.read().get(&message_id.0 .0)?;
+pub fn get_delay_secs_from(
+    times: &ReceiveTimes,
+    message_id: &MessageId,
+    send_time_ms: i64,
+) -> Option<i64> {
+    let recv_ms = first_seen_ms(times, message_id)?;
     let delay_ms = recv_ms as i64 - send_time_ms;
     let delay_secs = delay_ms / 1000;
     if delay_secs >= 0 {
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -1,7 +1,9 @@
 use crate::components::app::document_title::count_unread_behind_rooms_panel;
 #[cfg(target_arch = "wasm32")]
 use crate::components::app::notifications::request_permission_on_first_message;
-use crate::components::app::receive_times::{format_delay, get_delay_secs};
+use crate::components::app::receive_times::{
+    first_seen_ms, format_delay, get_delay_secs_from, ReceiveTimes,
+};
 use crate::components::app::sync_info::{RoomSyncStatus, SYNC_INFO};
 use crate::components::app::{
     MobileView, CURRENT_ROOM, EDIT_ROOM_MODAL, MEMBER_INFO_MODAL, MOBILE_VIEW, NOTIFICATION_MODAL,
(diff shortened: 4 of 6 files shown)


sanity:
## Review record

**External model unavailable.** Codex is out of credit (its quota is exhausted, not
a transient error), and the Gemini CLI has been unusable since 2026-06-24
(`IneligibleTierError` on every call). With no external reviewer available, this
used the Claude-subagent fallback: **three independent reviewers, each with a
distinct adversarial lens, each blind to the others' findings** —

1. **skeptical bug-hunt** — state-machine holes in the scroll flags, event
   orderings, leaked closures, re-entrancy, feature gating;
2. **code-first** — read the code before the description, then flag every place
   the implementation and the stated intent diverge;
3. **test-quality** — this repo has a history of vacuous tests, so: would each
   test fail if the fix were reverted, is its premise reachable and asserted,
   could it pass against a wrong implementation.

This is a weaker signal than an external pass. Recorded here so that is visible
rather than implied. Tier: **Full** (the change is cross-cutting and touches a
surface with a prior production incident).

### Findings and resolutions

**Behaviour changes made in response:**

| Finding | Resolution |
|---|---|
| ResizeObserver watched only the content wrapper, so growing the composer — the issue's own cause 3 — left the newest message below the fold. All three reviewers, independently. | The observer now watches the scroll container too. This is what forced the test restructure below: a complete fix erases "the view is off the bottom" as a precondition. |
| `settle_is_ours` could latch true: `scroll_top()` truncates while `scroll_height()`/`client_height()` round, so on fractional layout the guard armed a suppression for a scroll that could not move. | Superseded entirely — the boolean is gone. Whose settle it is, is now decided by position. See below. |
| `time_clamped` silently widened: the old comparison carried the whole propagation delay as slack, the new one carried none, so every message from a peer with a slightly fast clock would render with the "clock may be wrong" marker — under a tooltip that then said something false ("original timestamp was in the future"). | Added a one-minute tolerance, reworded the tooltip, and fixed the two struct doc comments that still named the old clamp target. Two new unit tests: one for the newly-reachable case, one asserting skew inside the tolerance is left alone. |
| The clamp target is now unvalidated persisted `localStorage`, so a corrupt entry could move a message's rendered time, grouping and day separator. | `receive_times` rejects values that are not plausible arrival times, at the parse boundary. |
| `BOTTOM_THRESHOLD_PX` claimed to be shared with the IntersectionObserver's `rootMargin`, which was a hardcoded literal. | The margin is built from the constant, so the claim is now true. |
| The test hooks were gated on `example-data` but not `no-sync`, and `cargo make dev-example` enables the first without the second — a developer running it against a live node would inject contract-invalid messages. | Gated on both, at the call site and the definition. |
| `last_scroll_top` was not reset on room switch, so an offset measured in the previous room could suppress the follow in the next one. | Reset with the pin on room change. |
| No coverage of a reader scroll producing no gesture event (Firefox scrollbar drag, find-in-page, focus-driven scroll). | New test. It also turned out to be moot in the end, since the redesign stopped listening for gestures at all. |
| The settle handler's re-arm branch was never reached by any test — the button re-arms the pin directly — so narrowing it to `distance <= 0` would have passed. | The reader now scrolls back to the bottom themselves in the test; verified by mutation. |
| `waitForTimeout(1_000)` before an assertion that the view did NOT move: fail-lenient under CI load, and `retries: 2` would hide it. | Waits on the message being rendered, then asserts `scrollTop` unchanged rather than distance-from-bottom (which also moves when content grows). |
| Assorted doc-comment overstatements: "only the reader can change it" (the forced path re-arms), "would otherwise sort" (nothing sorts), "adds no new reactivity" (false for a room with no messages), a stale comment claiming `group_messages` panics outside a Dioxus runtime. | All corrected. |

**One review claim was checked and found false.** The composer-clear leg was
said to pin `reader_moved_up_since`'s `.min(max_scroll_top(..))`. Deleting that
clamp and re-running, the test still passed — the browser fires a settle for its
own clamping and the reference is repaired before the next arrival. The comment
now records that it narrows a race window rather than deciding an outcome, instead
of claiming coverage it does not have.

**The most valuable finding came from the tests, not the reviewers.** Fixing the
above left the composer test failing about 1 run in 3 on mobile Safari. Traced
rather than retried: the view sat 62px short of the newest message and stayed
there for 12+ seconds, because clearing a long draft grows the container, the
browser clamps `scrollTop` on its own, and that clamp's settle was consumed by the
`settle_is_ours` flag the previous scroll had raised. One boolean cannot represent
two settles in flight. Replaced with position-based discrimination, which cannot
go stale because every settle rewrites the reference, and which made the four
gesture listeners unnecessary. 8/8 clean on the previously flaky case afterwards,
4x30 clean on the full spec, and the whole Playwright suite 485 passed / 0 failed
/ 0 flaky.

**Not folded in:** #487 — the DM thread modal has the same class of bug, filed
separately and deliberately kept out of this PR.

[AI-assisted - Claude]

PR #497, opened 2026-07-26
Expose verified reply target IDs in CLI JSON

sanity:
## Summary

Add two backward-compatible fields to a resolved `reply_to` object in `riverctl` JSON/JSONL output:

- `message_id`: the cryptographically resolved target message ID
- `author_id`: the resolved target's canonical River member ID

The existing `author` and `preview` fields remain unchanged. All four values are read from the live target message after River's existing verification; none comes from the untrusted reply snapshot.

This enables moderation/bridge consumers to correlate an authenticated moderator reply to the exact message without guessing from nickname and preview text.

## Verification

- `cargo test -p riverctl monitor_tests --lib` (18 passed)
- `cargo test -p riverctl unverifiable_quotes_are_omitted_from_json_and_neutral_in_text --lib`
- `cargo test -p riverctl private_reply --lib`
- `cargo test -p riverctl reply_json_includes_verified_target_identity --lib`
- `cargo clippy -p riverctl --lib --no-deps -- -D warnings`
- `cargo fmt --all -- --check`

Workspace clippy with dependencies is currently blocked by pre-existing Rust 1.94 `doc_lazy_continuation` errors in `common/src/room_state/ban.rs`; this PR does not modify that file.

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -46,6 +46,15 @@ const UPGRADE_HOP_TIMEOUT: Duration = Duration::from_secs(15);
 /// cyclic or runaway chain.
 const MAX_UPGRADE_HOPS: usize = 32;
 
+/// Optional fail-closed checks for non-interactive moderation callers. All
+/// predicates are evaluated against the fresh state fetched by the ban path.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct BanSafety {
+    pub require_exact_member_id: bool,
+    pub require_no_descendants: bool,
+    pub require_not_deputy: bool,
+}
+
 /// Decide the next contract to follow from `state`'s upgrade pointer.
 ///
 /// Returns `Some(next)` when `state` carries an `OptionalUpgradeV1` pointer to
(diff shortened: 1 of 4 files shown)


sanity (commented):
## Review: PR #497

**Tier: Full** — the guarded-ban commit is a destructive-operation safety surface, and production `riverctl` is already built from this branch.

### Verified

**The guards are fail-closed and opt-in.** All three (`require_exact_member_id`, `require_no_descendants`, `require_not_deputy`) `return Err` and refuse the ban rather than proceeding on doubt, which is the correct direction for an irreversible operation. Being opt-in flags means existing call sites are unaffected.

The deputy check reads current canonical `member_info` and scans every record's `deputies_of`, so it cannot be defeated by checking only the banning member's own grants.

**It composes with #516.** This is the load-bearing check, since a riverctl built after both must carry both. Merged this branch with current `main` locally:

- `cargo test` — **275 passed, 0 failed**
- `cargo clippy --all-targets` — clean
- `cargo fmt --check` — clean
- both features present in the merged tree (`require_not_deputy`, `classify_subscribe_response`)

CI on the PR is green across build, ui-playwright, and both migration checks. GitHub reports `MERGEABLE` / `CLEAN`.

3 tests accompany the guards.

### Why this is being merged now rather than left in draft

Production `/usr/local/bin/riverctl` is built from **this branch, not main**. That has two consequences worth recording:

1. The SUBSCRIBE interleaving fix (#516, merged) **cannot reach the moderator** until this lands — the merged fix is stranded in main while production runs a branch build.
2. Anyone rebuilding riverctl from `main` today would silently **strip these ban guards**. The moderator passes `--require-not-deputy` unconditionally, so a main-built binary rejects the argument and every ban fails. That is a latent moderation outage sitting behind an innocuous-looking rebuild.

Merging removes both traps: after this, `main` is the thing production is built from, which is where it should have been.

### Follow-up (not blocking)

Once merged, riverctl needs a release and the moderator a rebuild, otherwise the SUBSCRIBE fix still is not live — the room is currently flapping ~30 reader reconnects/hour with 55 `Unexpected response to SUBSCRIBE request` per two hours.

[AI-assisted - Claude]

PR #498, opened 2026-07-26
perf(ui): stop the conversation from cloning room state per message, and window the history

sanity:
## Problem

Users report River tabs sitting at multiple GB of RAM. Profiling a live client
(Chrome, Linux, `freenet` 0.2.109) with Playwright reproduced it:

| | small room open | after opening "Off Topic" (1133 msgs) |
|---|---|---|
| WASM linear memory | 68.8 MB | **1,647 MB** |
| DOM nodes | 1,710 | **24,305** |
| renderer RSS | 503 MB | 2,143 MB |

Two independent causes, both proportional to how much history is on screen:

**1. The room state was cloned twice per rendered message group.** The
conversation's react / delete / edit handlers each captured `current_room_data`
by value. `RoomData` owns `ChatRoomStateV1` directly — no `Rc` — so each capture
is a deep copy of every retained message, member and signature. Dioxus clones an
event-handler closure once per rendered row, and two of these are cloned in the
row loop, so the cost was O(groups x state_size).

`cargo run -p river-core --release --example room_state_clone_cost` (added here)
prices that multiplicand. Messages dominate it, not members:

| room shape | per-clone |
|---|---|
| 136 members, 1133 messages | 343 KB |
| 136 members, 0 messages | 59 KB |
| 1 member, 1133 messages | 275 KB |

**2. The whole room rendered at once.** No windowing, so 1133 messages meant
24,305 DOM nodes and a Dioxus tree to match.

WASM linear memory never shrinks, so the peak is permanent for the tab's life —
switching to a small room afterwards never gives it back. That is why the number
users see keeps climbing rather than tracking what is on screen.

## Approach

**Handlers look the room up when they run** (`current_room_data_snapshot()`)
instead of capturing it. The closures capture nothing, so cloning one per row is
free. Costs one clone per interaction instead of two per row, and reads fresher
state than a render-time snapshot did.

**The history renders a windowed tail.** `INITIAL_WINDOW_ITEMS` (60) display
items, growing by `WINDOW_GROWTH_ITEMS` when the reader reaches the top. Rooms
that fit inside the window — most of them, `max_recent_messages` defaults to
100 — render exactly as before: no sentinel, no backfill, no behaviour change.

Three details are load-bearing and are commented as such:

* The slice happens **before** the clone. `GroupedMessage` carries each
  message's rendered HTML, so cloning all groups per re-render paid most of the
  cost the window exists to avoid.
* A backfill captures scroll geometry first and restores the offset once the new
  rows land. Without it the view jumps by the height of everything revealed —
  and the sentinel would still be on screen, cascading until the whole room
  re-rendered.
* The window resets on room switch, so a freshly opened room is never rendered
  to the depth of the last one.

Deliberately **not** a both-ends virtual list. Message heights vary, and
estimated spacers fight the scroll anchoring that #486/#495 just fixed. A tail
window bounds what a reader lands on without touching that machinery.

## Testing

`HistoryWindow::resolve` is split out of the render so the arithmetic is
unit-testable: the newest item renders for any window size (including zero), a
room inside the window renders whole, and a room past it renders only its tail.

Source-grep pins guard both fixes — re-capturing the room data, or rendering
`groups` directly, are each a one-token edit that costs nothing measurable until
someone opens a room with a thousand messages in it. The capture pin was
mutation-tested (re-added the capture, confirmed red, restored, confirmed green).

End-to-end A/B against a seeded local room is in progress; numbers to follow in
a comment before this leaves draft.

[AI-assisted - Claude]

### DIFF
--- a/common/examples/room_state_clone_cost.rs
+++ b/common/examples/room_state_clone_cost.rs
@@ -121,11 +121,16 @@ fn build_room(members: usize, messages: usize, msg_len: usize) -> ChatRoomStateV
 }
 
 fn main() {
-    // Shape taken from the live "Off Topic" room profiled on 2026-07-26:
-    // ~1133 rendered messages, ~136 members.
-    let members = 136;
-    let messages = 1133;
-    let msg_len = 80;
+    // Defaults are the live "Off Topic" room profiled on 2026-07-26: ~1133
+    // rendered messages, ~136 members. Override to price another room's shape:
+    //   cargo run --release --example room_state_clone_cost -- <members> <messages> <body_len>
+    let a: Vec<String> = std::env::args().skip(1).collect();
+    let arg = |i: usize, default: usize| a.get(i).and_then(|s| s.parse().ok()).unwrap_or(default);
+    // At least one member: messages are authored round-robin across them, and
+    // a room with no members has nobody to sign one.
+    let members = arg(0, 136).max(1);
+    let messages = arg(1, 1133);
+    let msg_len = arg(2, 80);
 
     let state = build_room(members, messages, msg_len);
     println!(
--- a/common/examples/room_state_clone_cost.rs
+++ b/common/examples/room_state_clone_cost.rs
@@ -1,0 +1,162 @@
+//! Measures what ONE `ChatRoomStateV1` clone costs in resident memory.
+//!
+//! Motivation: the conversation view clones the whole room state per rendered
+//! message row (two event-handler closures, each capturing `RoomData` by
+//! value). That makes the UI's memory O(messages x state_size). This binary
+//! prices the multiplicand so the blow-up can be predicted from a room's shape.
+//!
+//! Run: `cargo run -p river-core --release --example room_state_clone_cost`
+
+use ed25519_dalek::SigningKey;
+use freenet_scaffold::ComposableState;
+use rand::rngs::OsRng;
+use river_core::room_state::configuration::{AuthorizedConfigurationV1, Configuration};
+use river_core::room_state::member::{AuthorizedMember, Member, MemberId, MembersV1};
+use river_core::room_state::member_info::{AuthorizedMemberInfo, MemberInfo, MemberInfoV1};
+use river_core::room_state::message::{
+    AuthorizedMessageV1, MessageV1, MessagesV1, RoomMessageBody,
+};
+use river_core::room_state::privacy::SealedBytes;
+use river_core::room_state::{ChatRoomParametersV1, ChatRoomStateV1};
+use std::time::{Duration, SystemTime};
+
+const BASE_SECS: u64 = 1_700_000_000;
+
+/// Resident set size in bytes, straight from the kernel. `statm` field 2 is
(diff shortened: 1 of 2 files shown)


sanity:
## A/B measurement (Playwright, seeded local room)

Measured against an isolated local node with a room seeded to **1,250 messages
from 60 authors** — the author mix matters, see the note below. Same room, same
node, same browser; only the published UI build differs.

| | baseline (`origin/main`) | this branch |
|---|---|---|
| WASM linear memory on room open | **1,503 MB** | **22 MB** |
| DOM nodes | 26,560 | 1,563 |
| messages rendered | 1,250 | 60 |
| renderer+browser RSS | 2,444 MB | 950 MB |
| after 12 scroll-to-top backfills | n/a (all rendered up front) | 33 MB at 780 messages |

For scale, the same shape extrapolated: baseline at the 780 messages the fix
reaches after twelve backfills would be ~937 MB, against 33 MB here.

The reproduction closely matches the live "Off Topic" room this started from
(1,647 MB WASM / 24,305 DOM nodes), so the local room is a fair stand-in.

**A methodology note worth recording.** The first attempt seeded 1,200 messages
from a *single* author and measured only 27.9 MB on baseline — no blow-up at
all. That is not because the bug is absent; it is because River groups
consecutive messages from one author, so 1,200 messages collapsed into a couple
of message groups. The cost is per GROUP (that is the unit whose props carried
the room-state clones), not per message. Re-seeding across 60 authors made
groups ≈ messages and reproduced it. A single-author fixture would have
"proved" there was nothing to fix.

## Behavioural checks

| check | result |
|---|---|
| backfill triggers on scroll-up | yes, 60 → 120 |
| reader's anchor message after backfill | unchanged, **0px drift** |
| reaction round-trip (exercises the rewritten handler) | applied |
| send message | delivered and visible |
| page errors | none |

## Review

`codex review --base origin/main` could not run — usage limit hit, resets
**Aug 1**. Per the multi-model-review rule I fell back to diverse Claude review,
run as three sequential lenses rather than blind-parallel subagents (subagents
are restricted in this session). Findings, all fixed in cbbae6d0:

* **Skeptical bug-hunt** — window reset fired on any `CURRENT_ROOM` write, so
  re-selecting the open room collapsed a reader's backfilled window; scroll
  restore violated the file's documented "our own scrolls write the offset they
  asked for" contract.
* **Code-first** — sentinel sat inside the `space-y-4` list, shifting the
  history 16px whenever it appeared or disappeared.
* **Failure mode for this change (scroll/render)** — handlers could silently
  drop an action if `ROOMS.try_read()` missed (now warns); and the sentinel as a
  point marker dead-ended the history at `scrollTop == 0` on any viewport
  shorter than the lead. That last one was found by Playwright, not by reading.

One deliberate non-change: the render-side room lookup still duplicates
`current_room_data_snapshot`'s body. Unifying them would shorten a guard's
lifetime in the render path that has had Firefox-specific bugs (#402, #486), and
the gain is cosmetic.

[AI-assisted - Claude]

Issue #490, opened 2026-07-25
Receiver-side rejection (ban sweep / membership prune) is invisible to the sender, so the exchange never terminates


sanity:
## Summary

A **receiver-side rejection leaves no trace in the receiver's summary**, so the sender can never learn its payload was refused and recomputes the identical payload on every fan-out, indefinitely.

This is structurally distinct from the cap-driven prune/resend loop fixed in #485. That one is closed by publishing a retention horizon in the summary, so the sender withholds what the receiver would prune. **No horizon can express this one**: the horizon governs only what the *sender* withholds, and the rejection here happens in `post_apply_cleanup`'s ban/membership sweep, which changes nothing the summary reports.

Independently identified by the protocol reviewer on #485 as finding **S2**.

## Reproduction

`common/tests/retention_loop_test.rs::ban_sweep_reopening_the_horizon_still_loops_known_gap`, added in #485. It currently asserts the **broken** behaviour (see "Test polarity" below).

Setup, driven through the full `ChatRoomStateV1` merge so `post_apply_cleanup` actually runs:

1. R and S both hold 100/100 messages at `max_recent_messages = 100` — 60 authored by A, 40 by X, interleaved.
2. R bans X. `post_apply_cleanup` sweeps X's 40 messages. R drops to 60.
3. R is now BELOW its cap, so its retention horizon regresses `OldestRetained(..)` -> `Open`.
4. S has never seen the ban and still holds all 100.

Written first as `..._does_not_loop`, it failed:

```
thread 'ban_sweep_reopening_the_horizon_does_not_loop' panicked at
common/tests/retention_loop_test.rs:673:5:
S was still offering payloads to R after 10 rounds. R holds 60 messages and
1 members; the offer carries 40 messages and true members.
```

## Two coupled channels, not one

- **Messages** — S re-offers X's 40 messages every round. R's author-must-be-a-member retain in `MessagesV1::apply_delta` drops all 40 on arrival.
- **Members** — S re-offers X, and also the member R's inactivity-prune removed (R ends at 1 member, S still has 3). `post_apply_cleanup` re-prunes them on arrival.

Every round is a complete no-op for R — the test now asserts that explicitly, messages and members both byte-identical before and after. R's summary therefore never changes, so S's `delta` is deterministic and identical forever. R's *reopened* horizon makes it invite MORE, not less.

## Root cause

[...]

PR #502, opened 2026-07-26
fix(ui): keep the DM rail stable under signal contention (#499)

### LINKED ISSUE #499: fix(ui): DM rail collapses to empty or shows archived threads under signal contention
Reported by Ian on his Framework laptop (2026-07-26): entries in the DIRECT MESSAGES rail appear and disappear intermittently while "Archived (7)" stays fixed.

Code trace (HEAD 5ac8b9f1) found four UI-side defects in `ui/src/components/room_list/dm_rail_section.rs`, each producing a distinct flavor of the symptom.

## 1. Whole-list collapse on `DM_LAST_SEEN` contention

`build_view()` bails with `?` on `DM_LAST_SEEN.try_read()` Err (`dm_rail_section.rs:624`), and the memo renders `.unwrap_or_default()` (`:88`), so the entire active list renders empty for that pass. `current_archived_count()` (`:593`) does not read `DM_LAST_SEEN`, so "Archived (N)" persists, which is exactly the reported asymmetry. The `ROOMS` subscription registered at `:620` means the memo recovers on the next `ROOMS` write, so this is an intermittent whole-list blink.

Aggravator: `DmThreadModalBody` calls `mark_thread_read` from its render body (`dm_thread_modal.rs:320`), writing `DM_LAST_SEEN` on every render while a thread is open. `with_mut` notifies subscribers even when the value is unchanged, so an open DM thread produces a continuous write pulse that widens the contention window.

## 2. Archive filter fails open on `HIDDEN_DM_THREADS` contention

On `HIDDEN_DM_THREADS.try_read()` Err (`:632`), the filter is skipped entirely (`:724-728`: `None => entries`), so every archived thread flashes INTO the active list for a render, while the separately-memoized archived count is unchanged. This is the "entries appearing" half of the report.

## 3. Permanent-freeze variant: fallible first read registers no subscription

`build_view()`'s first read is the fallible `ROOMS.try_read()` (`:620`). In Dioxus 0.7.9 an Err `try_read` registers no subscription (verified in vendored `dioxus-signals-0.7.9/src/signal.rs:405-417`: the borrow is taken before `reactive_context.subscribe`), and each memo run clears the prior subscription set (`dioxus-core-0.7.9/src/reactive_context.rs:195-198`). One contended poll therefore leaves the memo with zero subscriptions, frozen at empty for the rest of the session, since `DmRailSection` never unmounts (`app.rs:484` mounts `RoomList` unconditionally; mobile switching is a class swap).

The repo already has the canonical guard for this in the sibling memo, `room_list.rs:153-161` (infallible `CURRENT_ROOM.read()` before the fallible `ROOMS.try_read()`), with a comment explaining exactly this failure. None of the rail's three builders (`build_view` `:619`, `current_archived_count` `:593`, `build_archived_view` `:516`) has it.

## 4. Sort is not a total order

`:731-735` sorts by `(unread desc, last_any_ts desc)` with no tiebreak, over entries accumulated in a fresh `HashMap` per call (`:672`), so threads tied on both keys shuffle position between renders.

## Why "Archived stayed at 7" while everything else flapped

`count_currently_archived` (`:496-507`) falls back to `last_any_ts = 0` for pairs missing from room state, and `0 <= hidden_at_ts` always holds, so the archived count is structurally insensitive to DM state loss and to everything in `build_view`. It only moves when `HIDDEN_DM_THREADS` itself changes.

## Suggested fix (small, self-contained, ui/ only)

- Register an infallible signal read before the first fallible one in all three builders, mirroring the `room_list.rs:153-161` guard.
- Degrade `DM_LAST_SEEN` contention to "unread = 0" for that pass instead of emptying the list.
- Degrade `HIDDEN_DM_THREADS` contention closed (bail and keep the previous list) instead of skipping the archive filter.
- Add a stable `(room, peer)` tiebreak to the sort.
- Make `mark_thread_read` skip the write when the value is unchanged, killing the render-body write pulse.

## Related, tracked separately

Individual rows genuinely leaving and re-entering contract state is a different mechanism: the `post_apply_cleanup` membership sweep deletes every DM whose counterparty is momentarily absent from `members` (`common/src/room_state.rs:420-426` -> `direct_messages.rs:744-756`), and peer re-offers restore them. That is #490 and is not fixed by #485. Its signature is individual rows flapping in sync with member-list flicker, rather than the whole list blinking.

[AI-assisted - Claude]

sanity:
## Problem

Entries in the DIRECT MESSAGES rail appear and disappear intermittently while "Archived (7)" stays fixed (reported on Ian's Framework laptop, 2026-07-26). Issue #499's trace found four UI-side mechanisms in the rail's builders, plus an aggravator:

1. **Whole-list collapse**: `build_view()` bailed with `?` when `DM_LAST_SEEN.try_read()` was contended, so the memo rendered the entire active list empty for that pass — while the separately-memoized archived count (which never reads `DM_LAST_SEEN`) stayed put. Exactly the reported asymmetry.
2. **Archive filter failed open**: a contended `HIDDEN_DM_THREADS.try_read()` skipped the filter entirely, flashing every archived thread INTO the active rail for a render.
3. **Permanent-freeze variant**: all three fallible builders took their FIRST signal read fallibly. In Dioxus 0.7 an Err `try_read` registers no subscription and each memo run clears the previous subscription set, so one contended poll left the memo with zero subscriptions — frozen for the session, since `DmRailSection` never unmounts.
4. **Non-total sort order**: entries tied on `(unread, last_any_ts)` accumulate through a fresh `HashMap` per pass and shuffled position between renders.
5. **Aggravator**: `mark_thread_read` ran `DM_LAST_SEEN.with_mut` unconditionally from `DmThreadModalBody`'s render body; `with_mut` notifies subscribers even when nothing changed, so an open DM thread produced a continuous write pulse that widened every contention window above.

## Approach

All changes are `ui/`-only (no WASM key impact).

- **`ui/src/components/room_list/dm_rail_section.rs`**
  - Each builder (`build_view`, `current_archived_count`, `build_archived_view`) now registers an infallible `OPEN_DM_THREAD.read()` BEFORE its first fallible `try_read`, mirroring the `CURRENT_ROOM.read()` guard in `room_list.rs`. `OPEN_DM_THREAD` is the DM surface's own view-state signal and is only written through `defer` from clean contexts, so the infallible read cannot hit a live write borrow.
  - `build_view` degrades a contended `DM_LAST_SEEN` to "unread = 0 for this pass" (recency and row set unaffected) instead of aborting; the per-peer scan is extracted into the pure `accumulate_peer_activity` helper, now shared by all three builders so their `last_any_ts` scans can't drift.
  - The hidden-filter degrade now fails **closed** via the pure `resolve_active_entries` helper: on contention the pass returns the last successfully-filtered rail (a module `thread_local` cache updated only on clean-hidden passes) rather than either the unfiltered candidates (old behaviour — archived threads flashing in) or an empty list (whole-rail blink). Deliberate choice, documented in the helper's doc-comment: (1) archived threads never appear in the active rail; (2) the rail collapses as rarely as possible. A contended `ROOMS` read reuses the same cache instead of blanking.
  - `sort_rail_entries` adds a `(room key bytes, peer id)` tiebreak after `(unread desc, last_any_ts desc)`, making the order total and render-stable.
- **`ui/src/components/direct_messages.rs`**: `mark_thread_read` now `try_peek`s first and only runs `with_mut` when the stored cutoff would actually advance (pure decision helper `thread_read_needs_write`), killing the render-body write pulse.

## Testing

10 new unit tests (all verified to FAIL against the pre-fix logic by temporarily reverting each specific hunk — mutation-checked):

- `accumulate_peer_activity_contended_last_seen_degrades_unread_to_zero` / `_applies_cutoffs_and_direction` — contended `DM_LAST_SEEN` produces entries with unread = 0 and intact recency; extraction preserves the old loop's cutoff/direction semantics.
- `resolve_active_entries_contended_hidden_never_yields_archived` / `_with_no_history_shows_nothing` / `_clean_read_applies_archive_filter` — contended hidden-read never yields archived threads in the active list and never falls open (both fail when reverted to the old `None => entries`).
- `sort_rail_entries_ties_are_deterministic_across_input_orders` / `_orders_unread_then_recency` — ties sort identically regardless of HashMap-derived input order (fails without the tiebreak).
- `rail_builders_read_infallible_guard_before_first_fallible_read` — source-pin (whitespace-stripped, cut at `mod tests`, per-function segments bounded by the next function head) asserting the infallible-read-before-fallible ordering in all three builders (fails when a guard is removed).
- `thread_read_needs_write_only_when_cutoff_advances` + `mark_thread_read_write_is_gated_pinned` — decision table plus a wiring pin that the `with_mut` stays gated (fails when the gate is removed).

```
cargo test -p river-ui --bins   → 675 passed; 0 failed
cargo fmt --check               → clean
cargo clippy -p river-ui --tests → no findings in touched files
cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync → clean
```

Related, deliberately out of scope: individual rows genuinely leaving/re-entering contract state via the `post_apply_cleanup` membership sweep is #490 (different mechanism, `common/` change).

## Review findings addressed (c5840769)

Three independent review lenses (reactivity, code-first, behavioral) converged on four should-fix findings; all applied:

1. **Peek-panic fixed**: `mark_thread_read`'s contended-`try_peek` fallback flipped `true` -> `false`. `try_peek` fails only while a live write borrow exists, so falling through to `with_mut` (a panicking borrow) on that same stack was a guaranteed RefCell panic; skipping self-heals on the next render. Pin extended to reject a `true` fallback.
2. **Self-nudge recovery**: `RAIL_REBUILD_TICK` (a `GlobalSignal` bumped via a storm-proofed, self-terminating one-shot `defer`) replaces `OPEN_DM_THREAD` as the builders' infallible subscription anchor; every degraded pass schedules a bump, so recovery is one macrotask away instead of parked until unrelated user action. Overclaiming comments corrected.
3. **No reorder / no cache poisoning on contended `DM_LAST_SEEN`**: unread is the primary sort key, so the unread=0 degrade was an order change, not cosmetic — `build_view` now backfills per-pair unread from `LAST_GOOD_RAIL`, and the cache write is gated on a fully-clean pass (`hidden` AND `last_seen`).
4. **Archived-count last-good**: `current_archived_count` serves the last clean count on contention instead of 0, which could combine with an empty thread list to unmount the whole DIRECT MESSAGES section via the empty-state early return.

All four are covered by new unit tests + wiring pins, each mutation-verified (revert hunk -> red -> restore -> green).

Two pre-existing nits found during review are deliberately deferred and filed: #503 (chrome-less dead-end modal when a DM thread's room disappears) and #504 (`count_unread_dms_with` duplicates the rail's unread accumulation).

Closes #499

[AI-assisted - Claude]

https://claude.ai/code/session_01NCUNSVinAJ5fQxTbEtQxXK

### DIFF
--- a/ui/src/components/direct_messages.rs
+++ b/ui/src/components/direct_messages.rs
@@ -42,10 +42,43 @@ pub static OPEN_DM_THREAD: GlobalSignal<Option<(VerifyingKey, MemberId)>> = Glob
 pub static DM_LAST_SEEN: GlobalSignal<HashMap<(VerifyingKey, MemberId), u64>> =
     Global::new(HashMap::new);
 
+/// Pure decision for [`mark_thread_read`]: would recording `up_to_ts`
+/// actually advance the stored cutoff? `current` is the existing
+/// `DM_LAST_SEEN` entry for the `(room, peer)` pair (`None` = no entry
+/// yet, which reads as cutoff 0 everywhere else).
+///
+/// Split out because the answer gates whether `mark_thread_read`
+/// touches the signal at all: `DmThreadModalBody` calls
+/// `mark_thread_read` from its render body on every render while a
+/// thread is open, and `with_mut` notifies subscribers even when the
+/// mutation changed nothing — so an unconditional write turned an open
+/// DM thread into a continuous write pulse on `DM_LAST_SEEN`, widening
+/// the contention window that blanked the DM rail (issue #499).
+/// Pinned by the `thread_read_needs_write_*` tests plus the wiring pin
+/// `mark_thread_read_write_is_gated_pinned`.
+pub(crate) fn thread_read_needs_write(current: Option<u64>, up_to_ts: u64) -> bool {
+    up_to_ts > current.unwrap_or(0)
+}
+
 /// Mark every DM from `peer` in `room` as seen up to (and including) the
 /// most recent inbound message timestamp known to the synchronizer.
 pub fn mark_thread_read(room: VerifyingKey, peer: MemberId, up_to_ts: u64) {
     crate::util::defer(move || {
+        // Skip the write when the stored cutoff would not advance —
+        // `with_mut` notifies subscribers even for a no-op mutation and
+        // this runs on every render of an open thread (issue #499
+        // write-pulse). `try_peek` registers no subscription and takes
+        // no write borrow; if it is somehow contended (it shouldn't be
+        // — we're in a `defer`'d clean context) we conservatively
+        // attempt the write rather than risk losing a genuine
+        // mark-read.
+        let needs_write = DM_LAST_SEEN
+            .try_peek()
+            .map(|seen| thread_read_needs_write(seen.get(&(room, peer)).copied(), up_to_ts))
+            .unwrap_or(true);
+        if !needs_write {
+            return;
+        }
         DM_LAST_SEEN.with_mut(|seen| {
             let entry = seen.entry((room, peer)).or_insert(0);
             if up_to_ts > *entry {
@@ -700,6 +733,65 @@ mod tests {
                  the send here makes that fix invisible to UI users."
             );
         }
+    }
+
+    /// Issue #499 write-pulse: `mark_thread_read` is called from
+    /// `DmThreadModalBody`'s render body on every render while a thread
+    /// is open, and `with_mut` notifies subscribers even when the
+    /// mutation is a no-op — so the `with_mut` MUST stay gated on
+    /// `thread_read_needs_write`. Source-scrape (the function needs a
+    /// Dioxus runtime to exercise): match whitespace-stripped source so
+    /// rustfmt reflowing can't fake a failure; cut at `mod tests` (this
+    /// file has exactly one) so these needles can't satisfy their own
+    /// check; bound the segment by the NEXT function head so a gate
+    /// elsewhere in the file can't vacuously pass.
+    #[test]
+    fn mark_thread_read_write_is_gated_pinned() {
+        let src = include_str!("direct_messages.rs");
+        let body = &src[..src.find("mod tests").unwrap_or(src.len())];
+        let stripped: String = body.chars().filter(|c| !c.is_whitespace()).collect();
+
+        let start = stripped
+            .find("pubfnmark_thread_read")
+            .expect("mark_thread_read not found");
+        let end = stripped[start..]
+            .find("pubfnopen_dm_thread")
+            .map(|i| start + i)
+            .expect("open_dm_thread must follow mark_thread_read");
+        let seg = &stripped[start..end];
+
+        let gate = seg
+            .find("if!needs_write{return;}")
+            .expect("mark_thread_read must early-return when the cutoff would not advance");
+        let decide = seg
+            .find("thread_read_needs_write(")
+            .expect("mark_thread_read must decide via thread_read_needs_write");
+        let write = seg
+            .find(".with_mut(")
+            .expect("mark_thread_read must still perform the gated write");
+        assert!(
+            decide < gate && gate < write,
+            "mark_thread_read's with_mut must come AFTER the needs-write gate \
+             (decide at {decide}, gate at {gate}, write at {write}) — an ungated \
+             with_mut notifies DM rail subscribers on every render of an open \
+             thread (issue #499 write-pulse)"
+        );
+    }
+
+    /// Decision table for the issue #499 write-pulse gate.
+    #[test]
+    fn thread_read_needs_write_only_when_cutoff_advances() {
+        // No stored entry: any positive timestamp advances the cutoff…
+        assert!(thread_read_needs_write(None, 1));
+        // …but a zero timestamp does not (absent reads as 0 everywhere).
+        assert!(!thread_read_needs_write(None, 0));
+        // Equal to the stored cutoff: no-op — this is the exact case the
+        // render body hits on every re-render while a thread is open.
+        assert!(!thread_read_needs_write(Some(100), 100));
+        // Older than the stored cutoff: no-op.
+        assert!(!thread_read_needs_write(Some(100), 99));
+        // Strictly newer: write.
+        assert!(thread_read_needs_write(Some(100), 101));
     }
 
     use super::*;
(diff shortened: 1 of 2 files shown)


sanity:
## Consolidated review record

Process per the multi-perspective review rule:

**External model**: Codex review attempted 2026-07-26; hard usage-limit error ("You've hit your usage limit ... or try again at Aug 1st, 2026 2:42 PM"), so the diverse-Claude fallback ran per policy.

**Reviewers** (three independent, mutually blind lenses on f8620333):
- *Reactivity/state semantics*: verified every framework claim against the vendored dioxus-signals/dioxus-core 0.7.9 sources (Err try_read registers no subscription; per-run subscription reset; write borrow released before notify; try_peek semantics).
- *Code-first/coverage*: independent mechanism trace before reading the PR description; per-pass subscription enumeration for all three builders; mechanical pin/needle audit (cut points, self-match, segment bounds).
- *Behavioral*: user-visible walkthrough of every contention scenario, before vs after.

**Verdicts**: unanimous no blockers; all four #499 mechanisms plus the write-pulse aggravator confirmed closed; new tests confirmed non-vacuous.

**Findings, all resolved in c5840769**:
1. The contended-peek fallback in `mark_thread_read` was a guaranteed panic path (`try_peek` fails only under a live write borrow; `with_mut` unwraps on the same stack). Flipped to skip (self-heals next render), comment corrected, polarity pinned.
2. After a contended-ROOMS pass the memo's only surviving subscription was `OPEN_DM_THREAD`, so recovery could wait on a DM open/close. Replaced the anchor with `RAIL_REBUILD_TICK` plus a deferred, latch-guarded self-nudge from every degrade arm; staleness is now bounded at one macrotask and the comments state the real mechanism.
3. A contended `DM_LAST_SEEN` pass collapsed the unread-first sort (visible reorder) and the degraded list could poison the last-good cache. Unread is now backfilled from last-good and cache writes are gated to fully-clean passes.
4. `current_archived_count` degraded to 0 and could transiently unmount the whole section. It now serves a last-good count with clean-pass-only writes.

**Delta review** on f8620333..c5840769: merge-ready; one test-robustness nit (the builder pin asserted at least one nudge call per builder rather than per degrade arm) fixed in 356b023b, test-only, mutation-verified (stripping the ROOMS-arm nudge alone now fails the pin).

**Dismissed with rationale**: defer-registration failure sticking the nudge latch (unreachable failure class shared by every `defer` user in the app); non-WASM `defer` immediacy (web-only app); `ArchivedThreadsPanel` one-macrotask blink while open (bounded by the nudge; optional polish).

**Pre-existing issues found during review, filed separately**: #503 (chrome-less dead-end DM modal for departed rooms), #504 (`count_unread_dms_with` duplicates the rail accumulation).

[AI-assisted - Claude]

PR #505, opened 2026-07-26
fix(ui): don't slide the history window under a pinned reader (#501)

### LINKED ISSUE #501: fix(ui): #498 windowed tail regressed arrival autoscroll — sliding window shifts scrollTop and the reader-intent guard blames the reader
Reported by Ian (2026-07-26): the conversation no longer auto-scrolls to show new messages when they arrive.

## Timeline: the #495 fix is merged and published, and #498 regressed it in the same publish

- 2026-07-25 19:27: #495 merged (`bfbe9793`), the settle-measured intent pin that fixed #486's latch-away.
- 2026-07-26 14:07: #498 merged (`111febba`), the windowed-tail-with-backfill rendering refactor.
- 2026-07-26 14:27: publish 30000360, the first publish containing either. There was never a published container with #495 alone.

No unmerged autoscroll work exists anywhere (branches, worktrees, and open PRs all checked; `fix/scroll-latch-and-trigger` is fully squashed into main).

## Root cause (H1): the sliding window moves scrollTop, and the reader-intent guard attributes it to the reader

#498 renders a sliding tail: `HistoryWindow::resolve(total, window_items)` with `INITIAL_WINDOW_ITEMS = 60` (`conversation.rs:1422`, slice at `:3080-3082`). `window_items` grows only via backfill; on arrival the window slides, dropping the oldest rendered group off the top in the same patch that appends the new message at the bottom.

Removing content above the viewport shifts `scrollTop` (browser scroll anchoring; `overflow-anchor` is never overridden anywhere in `ui/`, and clamping covers the other order). The guard both follow paths share:

```rust
fn reader_moved_up_since(last_top: &Rc<Cell<i32>>) -> bool {          // :1533
    let reference = last_top.get().min(max_scroll_top(&container));   // :1540
    container.scroll_top() + SCROLL_TOP_SLACK_PX < reference          // :1541
}
```

With the reader parked at the bottom (`last_top = L`), dropping height `H_old` above and adding `H_new` below yields `scroll_top = L - H_old` and `reference = min(L, L - H_old + H_new)`; the test returns true whenever `H_new > 2px` (or `H_old > 2px` in the other branch), i.e. on essentially every arrival in a windowed room. Both the arrival effect (`:2127-2128`) and the ResizeObserver path (`:2196`) stand down. The bottom-gap grows by `H_new` per arrival while scroll anchoring holds the reader visually still; once the distance passes `BOTTOM_THRESHOLD_PX = 100`, the settle handler (`:1643`) clears the pin, and the view is permanently latched away.

The `.min(max_scroll_top)` clamp only rescues content removed below; it structurally cannot rescue "removed above and added below", which is what a sliding tail does every time.

## Secondary defects in the same code

- **H2, room-open backfill cascade**: on opening a >60-item room, the `BACKFILL_LEAD_PX` sentinel can fire from `scrollTop = 0` before the opening snap, growing the window and restoring to ~60 items above the newest message.
- **H3, unconditional backfill restore**: the restore effect (`:1798-1800`) consults neither `pinned_to_bottom` nor `force_scroll`, so it can park the view at the anchor over a concurrent arrival, and it writes `last_scroll_top` so the settle reads as ours.
- **H4**: date-separator rows appearing/vanishing at the window head as it slides add further un-attributed shifts.

Related: #487 is the same family of mistake (measure the DOM after the update, attribute the result to the reader) in the DM thread modal, still open.

## Why CI stayed green

`ui/tests/conversation-autoscroll.spec.ts` genuinely covers arrival-while-at-bottom, but the example-data fixture seeds ~15-20 display items (`example_data.rs:332`: `message_count = 6`) against `INITIAL_WINDOW_ITEMS = 60`, so `has_older` is always false: the window never slides, the sentinel never renders, the restore never runs. The entire #498 windowing behavior is unreachable by the suite; the tests exercise the pre-#498 path exclusively.

## Fix direction

1. Stop the window sliding on arrival while the reader is pinned: grow the window with new arrivals (start stays fixed) and trim only when unpinned or beyond a hard ceiling. This removes the failure mode rather than compensating for it.
2. Set `overflow-anchor: none` on the scroll container so the browser stops rewriting `scrollTop` behind the pin's back, and account for any remaining render-induced shift in `reader_moved_up_since`'s reference.
3. Gate the backfill restore on `!pinned_to_bottom` and suppress the sentinel's initial fire until the opening snap has landed (closes H2/H3).
4. Close the coverage gap: seed a >60-item fixture, assert the premise (windowed render + attached sentinel) so the test fails loudly rather than going vacuous, then run the follow loop and the mirror-image stays-put guard, plus a room-open lands-at-bottom test.

[AI-assisted - Claude]

sanity:
## Problem

Reported by Ian (2026-07-26): the conversation stopped auto-scrolling to new
messages. #498's windowed tail advanced its start index on every arrival,
removing the oldest rendered rows in the same patch that appended the new
message at the bottom. Removing content above the viewport makes the browser's
scroll anchoring rewrite `scrollTop`; `reader_moved_up_since` then attributed
the browser's adjustment to the reader, both follow paths stood down, the
bottom gap grew by one message height per arrival, and once it passed
`BOTTOM_THRESHOLD_PX` the settle handler cleared the pin permanently (#501 H1).

Secondary defects in the same code: the top backfill sentinel could fire from
`scrollTop = 0` before the room-open snap landed, cascading the window over the
whole room (H2); the backfill restore consulted neither `pinned_to_bottom` nor
`force_scroll`, so it could park the view at its anchor over a concurrent
arrival — and because it writes `last_scroll_top`, the resulting settle read as
ours (H3); and date-separator churn at a sliding window head added further
unattributed shifts (H4).

The suite stayed green because the example fixture seeds ~15-20 display items
against `INITIAL_WINDOW_ITEMS = 60`: the window never slid, the sentinel never
rendered, and the entire #498 windowing path was dead code under CI.

## Approach

**Grow, don't slide — anchored by IDENTITY (H1, H4; #505 review blocker 1).**
`HistoryWindow::resolve` carries the previous render's head as a `WindowAnchor`
(the head item's key, up to 7 spare keys behind it, and the head's last index
as a relocation hint), re-located each render by `relocate_window`: the head
under its own key, else the first surviving spare (start = its position minus
its offset in the anchor), else index 0. The spares are load-bearing because
the head key alone can vanish while its neighbours survive — see the re-review
blocker below. On arrival the head stays the same ITEM and the
window grows at the bottom — nothing is removed above the viewport, for the
pinned reader and the parked-mid-history reader alike, and the frozen window
head retires H4's separator churn. Identity (not position) matters because
positions are unstable: an at-cap room (`max_recent_messages`, the steady state
of every busy room) drains its oldest message on every arrival, shifting every
index — a positional anchor slides the head in content space one row per
arrival forever.

**Bounded growth.** `WINDOW_ITEMS_CEILING` (4x the initial window = 240 items;
choice documented on the constant) caps what arrival growth can accumulate,
with one growth step of arrival grace above a reader-requested backfill. The
ceiling never caps the backfill request itself — that would dead-end paging.
The window trims back to `INITIAL_WINDOW_ITEMS` when a settle lands at the
exact bottom — the reader's or OURS (every follow snap, and the
scroll-to-latest button touch readers rely on) — because at the bottom a
top-trim is provably invisible (the `scrollTop` clamp keeps the tail glued).
The trim gate is `SCROLL_TOP_SLACK_PX`, not the 100px pin threshold, so a
reader parked slightly above the bottom is never clamped down by it. (This
deviates from the issue's "trim on the settle after they scroll up" sketch:
that trim would drop the very rows the reader is scrolling toward.)

**Parked-reader compensation (#505 review blocker 1).** When the head IS
swapped — pruned out from under the anchor, re-keyed, or dropped by a ceiling
trim — the render captures a SURVIVING row's pre-patch `offsetTop` (rows carry
`data-item-key`) and the `head_reposition` effect shifts a parked reader's
`scrollTop` by the measured post-patch difference: the one piece of scroll
anchoring reimplemented under our own control, exact including date-separator
churn, because a top-contiguous removal shifts every surviving row identically.
The probe walk is sized to `WINDOW_ANCHOR_KEYS`, since relocation landing via
spare `k` widens the window backward by `k` rows that never existed pre-patch;
a shorter walk dead-fires exactly on the largest removals.

**The reference moves RELATIVELY (#505 delta review blocker).** Both the
reposition and the backfill restore update `last_scroll_top` by the SHIFT, not
to the reader's new absolute offset. `last_scroll_top` is what
`reader_moved_up_since` measures against, and it only carries information in
the window between the reader moving and their `scrollend` landing — writing
their own freshly-compensated position into it collapses the guard to "has not
moved" while `pinned_to_bottom` is still stale-true, after which both follow
paths scroll them to the bottom and the resulting settle re-pins, permanently.
Two steady states this PR creates reach it constantly (a reader paged to the
front of an at-cap room; a parked reader at the ceiling). Relative preserves
the reader's gap in both directions, and resolves the pin honestly: the
compensation's settle reads as the READER's, so the handler re-measures and
clears the stale pin — correct, since the effect only runs for a genuinely
parked reader.

**Head-group re-keys (#505 RE-review blocker).** A message group's key is its
FIRST message's id, so an at-cap drain that consumes only that message leaves
the group alive but RE-KEYED — the norm, since consecutive same-author
messages fold into multi-message groups. Single-key anchoring then missed, and
the reposition probed the pre-patch DOM for the NEW key, which no row carried:
no compensation, and the parked reader crawled one intra-group line per
arrival. Two changes fix it: the spare keys above (the surviving neighbours
still pin the window), and `select_reposition_probe`, which walks forward from
the head (bounded at 4 rows) to the first key that HAS a pre-patch row and
measures on that. The old fixture was blind to this by construction — strictly
alternating authors never produce a multi-message group — so the capped
fixture now seeds same-author PAIRS.

**Scope, corrected:** compensated are front-contiguous removals (at-cap
drains, batched or not, re-keyed or not) and ceiling trims. NOT compensated:
bulk MID-window removals (ban purges) and late-loading media above a parked
viewport — follow-ups filed separately; the code comments say so rather than
claiming general coverage.

**No grow→trim oscillation.** At the bottom the sentinel intersects iff
`contentHeight < clientHeight + BACKFILL_LEAD_PX`, which a 60-item tail can
satisfy on a zoomed-out or tall-portrait viewport — trim would re-fire the
backfill, which would grow, snap, settle and trim again, a silent render loop.
`trim_would_rearm_backfill` estimates the retained tail's height and skips the
trim when it would land back in the strip's reach; the window then simply
stays grown, ceiling-bounded as ever.

**Backfill growth comes from the RENDERED size (#505 review blocker 2).**
Arrivals grow an anchored window past its requested size; a growth step
computed from the stale request could resolve to a start the window already
renders — zero new rows, no DOM change, the sentinel's IntersectionObserver
never re-fires, paging dead-ends. `grown_window` grows from the rendered size,
so every step reveals a full `WINDOW_GROWTH_ITEMS` beyond what is on screen.

**`overflow-anchor: none` on `#chat-scroll-container` (H1).** The pin
machinery owns this container's `scrollTop`. With anchoring off, `scrollTop`
changes only through our scrolls, the reader's, or the browser's shrink clamp,
which the guard's `.min(max_scroll_top(..))` already accounts for.

**Backfill hardening (H2, H3).** The restore effect runs synchronously
(post-patch, pre-paint — the deferred version let the browser paint one frame
of prepended rows at the wrong offset), repositions by the same MEASURED
probe-row technique rather than the raw `scrollHeight` delta (the delta also
counts an arrival appended BELOW the viewport in the same patch and
over-shifts the reader by its height; the delta survives only as the fallback
for a probe row pruned in that patch), and stands down when the view is pinned
or a forced snap is in flight — but distrusts a stale-true pin when the reader
has visibly moved (a continuous fling from the bottom into the sentinel band
settles no pin). The head reposition additionally stands down while
`last_scroll_top` already points at the bottom, so it cannot cancel the
scroll-to-latest smooth animation mid-flight. The top sentinel only mounts once the room-open snap has
landed (`opening_snap_done`, re-checked against the room so a rapid room
switch's stale snap task cannot un-gate the new room's sentinel, and gated
through the one render where the previous room's flag is still true). The
room-switch reset of the windowing state runs inline in render — including the
backfill capture, so a capture taken in the old room cannot restore into the
new one — so the new room's first frame never renders at the old room's depth.
The relocation walk compares keys without allocating (`display_item_key_matches`),
since the miss path is O(total) probes per render at the at-cap steady state.

## Testing

**Closes the CI coverage gap that let #501 ship.** `?deep-history-room=1` adds
two fixture rooms (default fixture untouched): **Deep History Room** — 188
alternating-author fillers + the 12 standard messages = 200 items, with
`max_recent_messages` raised to 300 so delivered arrivals don't prune —
and **Capped History Room** — seeded EXACTLY at its cap (148 fillers + 12
standard = 160) so every arrival drains the oldest message (the #505
blocker-1 surface), with fillers in same-author PAIRS so its display groups
hold two messages and an at-cap drain actually RE-KEYS the head group (the
re-review blocker; 74 display items still clears the 60-item window).
Alternation/pairing makes the item count a guarantee, pinned by native tests
(`deep_history_room_exceeds_the_render_window`,
`at_cap_room_is_exactly_at_its_message_cap`). The `__riverTest` hooks now
alternate append authors (a burst is N display items, not one merged group),
prune to `max_recent_messages` exactly as `MessagesV1::apply_delta` does, and
gain `appendMessages(n)` for a batched burst in one patch. **CI now also runs
`cargo test -p river-ui --bins --features example-data,no-sync`** — the
fixture tests were feature-gated and the featureless step never executed them.

Seven Playwright tests in the windowed describe, each asserting its premise
first (sentinel attached, rendered row count is a windowed tail):

* reader at bottom: six alternating arrivals, settled-at-bottom after each,
  then the settle-trim returns the window to ~initial size;
* reader parked mid-history (deep room): arrivals move neither `scrollTop`
  nor a probed visible row's rect;
* reader parked mid-history (at-cap room): four pruning arrivals do not crawl
  the probed row (#505 blocker 1);
* reader parked in the at-cap room, `appendMessages(61)` in ONE patch: the
  batch drains past the window head and re-keys the group at the boundary —
  the surviving rows stay rendered and the probed row holds within 3px (#505
  re-review blocker). No scrollTop assertion here deliberately: the
  compensation MOVES scrollTop in order to hold the content still. The test
  waits out the reader's `scrollend` before delivering — "parked" means
  settled — because delivering INSIDE that window hits #508, whose residual
  path is pre-existing (evidence below);
* backfill paging: first fire reveals a page, the probed head row stays put
  through the restore, repeated paging reaches "history filler 00";
* arrivals-then-page: a 61-message batch grows the anchored window, then the
  FIRST sentinel fire must reveal older rows (#505 blocker 2);
* opening the deep room lands settled at the bottom with the initial window
  (H2/H3; the deterministic guard is the source pin on the sentinel's mount
  gate — the browser race is timing-dependent).

**Red-run evidence.**
* Original #501 follow test vs unfixed `origin/main` (fixture + specs kept,
  `conversation.rs` reverted, rebuilt): fails at "arrival 1 in a windowed room
  was not followed" (distance 154px); guard tests pass pre-fix as expected.
* #505 blocker tests vs the pre-review PR HEAD (`6ed7239f`, same method,
  chromium `--retries=0`): **both red** — the at-cap parked test fails at
  "content crawled under a parked reader as at-cap arrivals pruned the
  history" (probed row shifted **384px** over four arrivals, expected ≤ 2px),
  and the arrivals-then-page test fails at "the first backfill after an
  arrival burst revealed no older rows" (row count stuck at 123 — the exact
  zero-new-rows dead-fire). Both green on the fixed head.
* Re-review blocker test vs the previous head (`7ade27d5`, same method,
  chromium `--retries=0`): **red** — "the probed row left the DOM — the window
  lost the surviving rows under a parked reader". Green on the fixed head.
* Delta-review blocker (the absolute-reference write): the wait-removal check
  was re-run against a **verified-fresh** build (`index.html` -> hashed `.js`
  -> hashed `.wasm`, all carrying the build's timestamp). With the wait
  removed the batched-drain test passes **12/12** across chromium, firefox and
  mobile-safari. My earlier red-run had captured a STALE bundle — I launched
  it before the rebuild had finished writing — so it was measuring the
  absolute-write bug this PR then fixed, not a residual. The wait is gone, and
  the test now delivers INSIDE the pre-settle window on purpose, since that is
  the window the relative update protects. #508's evidence has been corrected
  accordingly.
* Trim re-anchor (found by that re-run, 5/16 failing on chromium and present
  at the pre-fix head too): the trim removes rows above a view at the bottom,
  the browser clamps `scrollTop` by their height, and `last_scroll_top` still
  held the pre-trim bottom — so an arrival growing the content back before the
  clamp's settle made `reader_moved_up_since` read the clamp as reader
  movement and the follow stood down, leaving the reader a row short (a taller
  row would clear `BOTTOM_THRESHOLD_PX` and latch the pin off). A trim now
  flags itself and an effect re-anchors through `scroll_history_to_bottom`,
  which fixes the offset and the reference together. **32/32** after, against
  **5/16 failing** before.
* Unit-test mutations (applied-then-reverted, each confirmed to flip):
  gutting the render-side room reset fails the reset pin; reverting
  `grown_window` to `requested + GROWTH` fails the growth-step test; making
  `relocate_anchor` positional fails the identity-walk tests; restricting
  `relocate_window` to the head key alone fails the mid-window-deletion test;
  narrowing the reposition probe to the head row alone fails the re-key test;
  neutering `trim_would_rearm_backfill` fails the oscillation-guard test;
  un-pairing the capped fixture's authors fails its premise pin; shortening
  `REPOSITION_PROBE_ROWS` back to 4 fails the deepest-spare test; and
  reverting either reference update to absolute fails the relative-update pin.

Rust-side: `relocate_anchor` search-order/fallback tests, an at-cap
identity-walk test (40 pruning arrivals, head identity held, then the
head-pruned spare fallback), the re-key relocation + probe-walk test, a
batched-drain-past-the-head test, a mid-window head-deletion test (the case
where the spares are load-bearing and no fallback substitutes), the
oscillation-guard arithmetic, the blocker-2 growth-step test (including the
premise that the naive step really does dead-end), ceiling grace, and updated
source pins (identity resolve + write-back, rendered-size recording,
`grown_window` in the sentinel, trim-before-ours-early-out ordering, the
geometry gate on the trim, restore/reposition fling gates, the probe walk,
the measured backfill restore, sentinel mount gate, and room-reset needles
that cannot false-pass against the trim site).

Local runs on the release example build (head `ced941ab`, served-bundle
freshness — `index.html` → hashed `.js` → hashed `.wasm` — verified before
every browser run):

* `cargo test -p river-ui --bins`: **693 passed; 0 failed**
* `cargo test -p river-ui --bins --features example-data,no-sync`:
  **696 passed; 0 failed**
* `conversation-autoscroll`, `--repeat-each=4 --retries=0` across chromium +
  firefox + mobile-safari: **156 passed (1.4m)**
* the burst test alone, `--repeat-each=32 --retries=0`: **32 passed** — with
  the re-anchor guard in place, so the repair does not depend on firing in
  states it should not (it was 5/16 FAILING before the trim re-anchor existed)
* `conversation-autoscroll`, webkit + mobile-chrome `--retries=0`:
  **26 passed (24.6s)**
* full Playwright suite, chromium: **101 passed, 6 skipped (21.4s)** — the six
  are pre-existing conditional skips in `invite-via-dm-picker.spec.ts`
* `cargo fmt --check` clean; clippy warning count **34**, one BELOW `main`'s 35
  (no new warnings; clippy is off in River CI)

Related: #487 is the same family of mistake (measuring the DOM after the
update and attributing the result to the reader) in the DM thread modal. It is
NOT fixed here — different component, different mechanism, deliberately kept
separate. Also out of scope, tracked separately: parked-reader compensation
for late-loading media and for bulk mid-window removals (ban purges), and
**#508** — filed during this work from a code reading plus a reproduction that
turned out to be invalid (it ran against a stale bundle and was actually
showing this PR's own absolute-write bug, since fixed). I have posted a
correction retracting that evidence: the clamp hazard it describes is not
demonstrated by that scenario, because the arrivals grow the content faster
than the drain shrinks it, so `.min(max_scroll_top)` never bites there. It is
left for the maintainers to retitle, keep as a code-reading hazard, or close.

Closes #501

[AI-assisted - Claude]

https://claude.ai/code/session_01NCUNSVinAJ5fQxTbEtQxXK

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -188,6 +188,17 @@ jobs:
       # `cargo test -p river-ui`. This step gates them on every push.
       run: cargo test -p river-ui --bins
 
+    - name: Test river-ui (example-data fixture tests)
+      env:
+        RUST_MIN_STACK: 8388608
+        CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      # The example_data module (and its tests — the deep/at-cap windowing
+      # fixture pins for freenet/river#501/#505) only compiles behind the
+      # example-data feature, so the featureless run above never executes
+      # them. Without this step a fixture regression that vacuates the
+      # windowing Playwright specs would pass CI silently.
+      run: cargo test -p river-ui --bins --features example-data,no-sync
+
   ui-playwright-tests:
     runs-on: freenet-default-runner
 
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -1476,6 +1476,65 @@ struct HistoryWindow {
     has_older: bool,
 }
 
+/// The identity of the window's head: which display item the last render
+/// started at, remembered by KEY, with the index it held as a relocation hint.
+///
+/// The anchor is by IDENTITY, not position, because positions are not stable:
+/// `MessagesV1::apply_delta` drains the oldest message once a room exceeds
+/// `max_recent_messages` (the steady state of every busy room), so an arrival
+/// shifts every index down by one. A positional anchor then renders a head one
+/// item LATER in identity — the top rendered row is removed in the same patch
+/// that appends the arrival, which is the #501 slide reproduced in content
+/// space, and with scroll anchoring disabled it crawls a parked reader upward
+/// one row per arrival. Anchoring on the head item's key holds the rendered
+/// set fixed in identity no matter how the indices shift underneath it.
+#[derive(Clone, PartialEq, Debug)]
+struct WindowAnchor {
+    /// `display_item_key` of the item the window started at.
+    key: String,
+    /// The index that item held last render — a hint so relocation is O(shift),
+    /// not O(total), in the common case, and the positional fallback when the
+    /// item is gone entirely.
+    index: usize,
+}
+
+/// Re-locate the anchored head item after the item list changed.
+///
+/// Front prunes shift every surviving index DOWN, so the hint is checked
+/// first, then the indices below it, then (for completeness — a merge can
+/// insert older history above the head) the indices above. `None` means the
+/// key is gone entirely — the head itself was drained, or its whole group was
+/// deleted. The caller then falls back to the POSITION: for a front prune
+/// that consumed the head, the hint (having marched to ~0 with the shifts) is
+/// the nearest surviving item — everything between the new index 0 and the
+/// old head was pruned with it — and for a mid-window deletion it is off by
+/// at most the deleted item. Either way the `head_reposition` effect
+/// compensates the parked reader's offset for whatever content the swap
+/// removed.
+fn relocate_anchor(total: usize, hint: usize, is_anchor: impl Fn(usize) -> bool) -> Option<usize> {
+    if hint < total && is_anchor(hint) {
+        return Some(hint);
+    }
+    if let Some(i) = (0..hint.min(total)).rev().find(|&i| is_anchor(i)) {
+        return Some(i);
+    }
+    (hint.saturating_add(1)..total).find(|&i| is_anchor(i))
+}
+
+/// The requested window size after one backfill growth step.
+///
+/// Grows from the RENDERED size, not the requested size. Arrivals grow an
+/// anchored window past what was requested (`rendered = total - start` exceeds
+/// `requested` by the number of arrivals), and a growth step computed from the
+/// stale requested size can resolve to a start the anchor already renders —
+/// zero new rows, no DOM change, so the sentinel's IntersectionObserver never
+/// re-fires and paging dead-ends (#505 review, blocker 2). Growing from the
+/// rendered size guarantees every step reveals `WINDOW_GROWTH_ITEMS` more
+/// items than are currently on screen.
+fn grown_window(requested: usize, rendered: usize) -> usize {
+    rendered.max(requested) + WINDOW_GROWTH_ITEMS
+}
+
 impl HistoryWindow {
     /// Resolve where the rendered tail starts.
     ///
@@ -1517,7 +1576,13 @@ impl HistoryWindow {
             Some(prev) => prev.min(base),
         };
         // The ceiling on arrival growth; never on reader-requested backfill.
-        let cap = window.max(WINDOW_ITEMS_CEILING).max(1);
+        // One growth step of grace above the requested size, so a reader who
+        // parked right after backfilling to the ceiling is not hit with a
+        // ceiling slide on the first arrival — belt-and-braces on top of the
+        // `head_reposition` compensation (#505 review).
+        let cap = (window + WINDOW_GROWTH_ITEMS)
+            .max(WINDOW_ITEMS_CEILING)
+            .max(1);
         if total_items - start > cap {
             start = total_items - cap;
         }
@@ -1553,6 +1618,29 @@ fn chat_content_wrapper() -> Option<web_sys::Element> {
 #[cfg(target_arch = "wasm32")]
 fn max_scroll_top(container: &web_sys::Element) -> i32 {
     (container.scroll_height() - container.client_height()).max(0)
+}
+
+/// The `offsetTop` of the history row carrying `data-item-key == key`, if it
+/// is currently in the DOM.
+///
+/// Matched by comparing attributes rather than an attribute SELECTOR, so a key
+/// never needs CSS escaping. The scan is bounded by the rendered window (≤ a
+/// few hundred rows) and only runs on the rare head-swap paths.
+#[cfg(target_arch = "wasm32")]
+fn history_row_offset_top(key: &str) -> Option<i32> {
+    use wasm_bindgen::JsCast;
+    let container = chat_scroll_container()?;
+    let rows = container.query_selector_all("[data-item-key]").ok()?;
+    for i in 0..rows.length() {
+        let Some(node) = rows.item(i) else { continue };
+        let Some(el) = node.dyn_ref::<web_sys::HtmlElement>() else {
+            continue;
+        };
+        if el.get_attribute("data-item-key").as_deref() == Some(key) {
+            return Some(el.offset_top());
+        }
+    }
+    None
 }
 
 /// Slack for comparing one scroll offset against another.
@@ -1664,7 +1752,7 @@ fn install_scroll_pin_listeners(
     pinned: Rc<std::cell::Cell<bool>>,
     last_top: Rc<std::cell::Cell<i32>>,
     window_items: Signal<usize>,
-    window_anchor: Rc<std::cell::Cell<Option<usize>>>,
+    window_anchor: Rc<std::cell::RefCell<Option<WindowAnchor>>>,
     window_overgrown: Rc<std::cell::Cell<bool>>,
 ) -> bool {
     use wasm_bindgen::prelude::*;
@@ -1696,35 +1784,44 @@ fn install_scroll_pin_listeners(
             // for reader settles is what let a stale offset suppress the
             // follow indefinitely.
             last_top.set(settled_at);
-            if ours {
-                return;
-            }
             let distance = container.scroll_height() as f64
                 - settled_at as f64
                 - container.client_height() as f64;
-            pinned.set(distance <= BOTTOM_THRESHOLD_PX);
-            // The reader's own settle landing at the bottom is the ONE moment
-            // a window trim is provably invisible: at the bottom, removing
-            // rows above the viewport shrinks `scrollHeight` and the browser
-            // clamps `scrollTop` down with it, so the same tail stays glued to
-            // the bottom edge (`reader_moved_up_since`'s
-            // `.min(max_scroll_top(..))` already treats that clamp as not the
-            // reader). Trimming anywhere else shifts content under the reader
-            // now that scroll anchoring is off, so growth accumulated while
-            // parked is bounded by `WINDOW_ITEMS_CEILING` instead of trimmed.
+            // A settle landing AT the bottom is the ONE moment a window trim
+            // is provably invisible: removing rows above the viewport shrinks
+            // `scrollHeight` and the browser clamps `scrollTop` down with it,
+            // so the same tail stays glued to the bottom edge
+            // (`reader_moved_up_since`'s `.min(max_scroll_top(..))` already
+            // treats that clamp as not the reader). Trimming anywhere else
+            // shifts content under the reader now that scroll anchoring is
+            // off, so growth accumulated while parked is bounded by
+            // `WINDOW_ITEMS_CEILING` instead of trimmed.
+            //
+            // Checked BEFORE the `ours` early-out, deliberately: every
+            // followed arrival's snap settles as ours, and so does the
+            // scroll-to-latest button — a touch reader who only ever returns
+            // to the bottom through the button would otherwise never trim
+            // (#505 review). Gated at SCROLL_TOP_SLACK_PX, not
+            // BOTTOM_THRESHOLD_PX: a reader parked up to 100px above the
+            // bottom still counts as pinned, and a trim from there would
+            // clamp their offset to the exact bottom — a visible yank.
             //
             // Deferred: this runs from a raw JS callback with no Dioxus scope,
             // and `window_items` is a signal the render subscribes to. See
             // .claude/rules/dioxus-signal-safety.md.
-            if distance <= BOTTOM_THRESHOLD_PX && window_overgrown.get() {
+            if distance <= SCROLL_TOP_SLACK_PX as f64 && window_overgrown.get() {
                 window_overgrown.set(false);
                 let window_anchor = window_anchor.clone();
                 let mut window_items = window_items;
                 crate::util::defer(move || {
-                    window_anchor.set(None);
+                    *window_anchor.borrow_mut() = None;
                     window_items.set(INITIAL_WINDOW_ITEMS);
                 });
             }
+            if ours {
+                return;
+            }
+            pinned.set(distance <= BOTTOM_THRESHOLD_PX);
         }) as Box<dyn FnMut()>)
     };
     let settle_fn: js_sys::Function = settle.as_ref().unchecked_ref::<js_sys::Function>().clone();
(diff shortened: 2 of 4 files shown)


sanity:
## Consolidated review record

**External model**: Codex was attempted for this PR series on 2026-07-26 and returned a hard usage-limit error ("try again at Aug 1st, 2026 2:42 PM"), so the diverse-Claude fallback ran per the multi-perspective review policy. Eleven independent passes ran in total; each of the first several found a real defect, so the sequence is recorded here rather than summarized as "reviewed".

### Pass history

1. **Four blind lenses** on the first implementation (scroll geometry, code-first, testing, behavioral) → **two blockers**:
   - The window anchor was positional, so in a room at its `max_recent_messages` cap (the steady state of every busy room) each arrival pruned the oldest message, shifted every index, and slid the window anyway. With `overflow-anchor: none` removing the browser's compensation, a parked reader's text crawled upward on every arrival. Invisible to CI because the fixture stopped just short of the cap.
   - After ~60 arrivals grew an anchored window, the first page-up computed a start that revealed **zero** rows, and since the DOM did not change the IntersectionObserver never re-fired: history paging dead-ended, violating the PR's own stated constraint.
2. **Rework** → identity-based `WindowAnchor`, measured scrollTop compensation, growth from rendered size.
3. **Two re-reviews** (geometry, behavioral) → **one blocker**: a message group re-keys when its first message is pruned, so the compensation lookup missed and the crawl persisted on the dominant production message shape (consecutive same-author messages). The fixture's alternating authors made the whole suite blind to it by construction.
4. **Fix round** → spare-key relocation, bounded forward probe, paired-author capped fixture.
5. **Delta review** → **one blocker**: `head_reposition` wrote `last_scroll_top` *absolutely*, collapsing the pre-settle protection signal while the bottom-pin was still stale-true, so both follow paths yanked a parked reader to the bottom. This was a path **introduced by this PR**, not inherited (see #508).
6. **Fix round** → relative reference update at both call sites, probe rows sized to the anchor keys, narrowed stand-down.
7. **Delta review** → flagged that the without-wait test evidence was not supported by the code.
8. **Experiment re-run** → the earlier evidence was measured against a **stale bundle**; against a verified-fresh build the test passes 12/12 without the wait, so the wait was removed and the #508 reproduction retracted. The honest re-run also exposed a real pre-existing defect: a trim clamps `scrollTop` while the stored reference holds the pre-trim bottom, so an arrival growing content back reads the browser's own clamp as reader movement and the follow stands down (5/16 failures). Fixed by a self-flagging trim re-anchor: 32/32.
9. **Review of the new trim machinery** → **one should-fix**: the re-anchor was the only programmatic scroll in the file consulting neither the pin nor position, so a reader who began a gesture inside the flag-to-effect window was yanked back and the pin re-armed.
10. **Final round** → guard added (pin **and** distance, neither subsuming the other), flag set moved inside the defer so it is atomic with the write it describes, pins strengthened to cover the repair rather than just the flag.
11. **Verification of the final diff** → guard as specified, nothing extra, burst test still 32/32.

### Coverage gap this closes

#498 shipped the original regression because the Playwright fixture seeded ~18 display items against a 60-item window: the entire windowing path was unreachable under CI. This PR adds a >60-item deep fixture and a capped fixture seeded at its cap with same-author pairs, all with premise assertions (sentinel attached, rendered rows bounded) so the tests fail loudly rather than going vacuous if a constant or fixture changes. CI now also runs the feature-gated unit tests, which previously never executed.

### Deliberately not fixed here

- #508 (the pre-existing clamp path in the #495 pin machinery) — its reproduction was retracted as stale-bundle evidence; it now rests on a code reading and is rescoped there.
- #507 (parked-reader shifts from late-loading images and bulk mid-window removals) — the two content-change classes this PR's compensation deliberately does not cover.
- #487 (the same measure-after-update mistake in the DM thread modal).

### Final state

`cargo test -p river-ui --bins` 693 passing (696 with `example-data,no-sync`); autoscroll spec 156 passing across chromium/firefox/mobile-safari at `--repeat-each=4 --retries=0`; burst test 32/32 at `--repeat-each=32`; webkit + mobile-chrome 26 passing; full chromium suite 101 passed, 6 skipped (pre-existing conditional skips). Every new test mutation-verified, including two mutations that were caught failing to apply and re-run rather than reported as passes.

[AI-assisted - Claude]

PR #516, opened 2026-07-27
fix(cli): don't treat an interleaved response as the SUBSCRIBE answer

sanity:
## Problem

The node multiplexes its responses onto a single connection, so an `UpdateNotification` can arrive between our SUBSCRIBE and its acknowledgement. Both subscribe paths did a single `recv()` and matched strictly, so anything that overtook the ack was fatal:

```
Unexpected response to SUBSCRIBE request
```

The session then died and reconnected — straight into the same race.

Measured against the official room on 2026-07-27:

| | |
|---|---|
| `Unexpected response to SUBSCRIBE request` (03:00–07:00 CDT) | **299** |
| share of all reader deaths | **~95%** |
| peak reader reconnects in one hour | **184** (one every ~20s) |

Every River client is a long-lived subscriber, so this is a general reliability problem, not a moderator-specific one.

It also caused downstream harm. Each reconnect drags a catch-up batch behind it, and the moderator replayed its own backlog, scored it as a live burst, and classified a member `BanSevereHarm`/`Flooding` at 0.999. Only a dead code path (`BanAsOwnerEmergency` has no executor) stopped an automatic ban of an innocent user.

## This is a recurrence, not a new bug

`6735825d` (#194, 2026-03-24) hit the **same race on the PUT path** and fixed it by adding one match arm, described as "PUT to an existing contract triggers an UpdateNotification".

That framing reads it as a semantic quirk of PUT rather than what it actually is: responses are multiplexed and can interleave with *any* request/response pair. Because the general lesson wasn't drawn, both SUBSCRIBE sites kept the fragile shape for four months. Worth noting for review — the fix here is the shape, not the symptom.

## Approach

Read until the acknowledgement actually arrives, bounded by the original overall deadline rather than restarting a timeout per message.

- **Streaming path**: overtaking responses are **queued**, not dropped. They're real state changes, and the main loop drains `pending` before touching the socket, so they receive exactly the handling they'd have had arriving a moment later. Nothing is lost.
- **`subscribe_to_contract`**: no loop exists to hand them to, so they're logged and dropped. Deliberate and strictly better than before, which lost the same notification *and* left the caller unsubscribed. Callers re-read contract state rather than relying on notifications for correctness.

The handshake decision is extracted into `classify_subscribe_response`, and **both call sites route through it** — otherwise the tests would be vacuous.

## Testing

Four unit tests:
- an `UpdateNotification` must not answer the subscribe (the regression)
- the acknowledgement is recognised
- a refusal stays distinguishable from "keep waiting" — otherwise a refused subscribe hangs to timeout instead of reporting why
- other responses are tolerated

**Mutation-verified**: flipping `NotYet` → `Refused` fails the regression test, so it isn't vacuous.

Full suite **272 passed**, `clippy --all-targets` clean, `cargo fmt` applied.

## What this does not fix

The node-side question of *what* it sends before the ack is tracked separately in freenet/freenet-core#4970. This makes the client correct regardless — a client that dies on any interleaved message on a multiplexed async stream is wrong by construction — but it is a client-side fix, so it makes the symptom cheap rather than absent.

Refs freenet/freenet-core#4970

[AI-assisted - Claude]

### DIFF
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -451,6 +451,37 @@ pub fn compute_contract_key(owner_vk: &VerifyingKey) -> ContractKey {
 /// always thread the room `secrets` via [`message_display_text_with_secrets`];
 /// this 2-arg form is retained for the tests that exercise public content and
 /// the genuine `<encrypted>` fallback (empty secrets).
+/// What a response received while awaiting a SUBSCRIBE acknowledgement means.
+///
+/// Split out so the interleaving case is directly testable: the node multiplexes
+/// its responses onto one connection, so an `UpdateNotification` can overtake the
+/// acknowledgement. Treating the first message as the answer made that race fatal
+/// (freenet-core#4970).
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum SubscribeAck {
+    /// The node acknowledged and we are subscribed.
+    Subscribed,
+    /// The node answered, and refused.
+    Refused,
+    /// Not the acknowledgement — keep waiting, and do not discard this.
+    NotYet,
+}
+
+pub(crate) fn classify_subscribe_response(response: &HostResponse) -> SubscribeAck {
+    match response {
+        HostResponse::ContractResponse(ContractResponse::SubscribeResponse {
+            subscribed, ..
+        }) => {
+            if *subscribed {
+                SubscribeAck::Subscribed
+            } else {
+                SubscribeAck::Refused
+            }
+        }
+        _ => SubscribeAck::NotYet,
+    }
+}
+
 #[cfg(test)]
 pub(crate) fn message_display_text(
     room_state: &ChatRoomStateV1,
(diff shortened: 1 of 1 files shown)


sanity (commented):
## Review: PR #516

**Tier: Full** — this is a protocol request/response handshake path, and the same class of bug has already recurred once here (#194's narrow fix). Lenses run serially in-context (subagents unavailable this session; Codex quota-blocked until 2026-08-01, not waited on per the multi-model-review rule).

### Findings

**1. Unbounded queue fed by the node — FIXED in `4a7169d5`.** `pending` is filled from node responses during a 30s handshake window; on a busy room that is an unbounded, externally-grown collection, which is exactly the shape `.claude/rules/code-style.md` forbids as a memory amplification vector. Now capped at 16.

Collapsing beyond the cap is **lossless**, which is what makes the cap safe rather than a silent data drop: the streaming handler discards the delta (`let _ = update;`) and re-fetches authoritative full state, so N queued notifications produce the same result as one. Verified in the handler, not assumed.

**2. Behaviour change worth stating: we now wait where we used to fail fast.** A node that never acks costs the full 30s (streaming) / 5s (`subscribe_to_contract`) instead of erroring on the first non-matching message. That is intended — the old fast failure was the bug — and the deadline is computed once rather than restarted per message, so it cannot be extended indefinitely by a chatty node. No unbounded wait.

**3. Tests are non-vacuous.** Both call sites route through `classify_subscribe_response`, so the tests cover production logic rather than a parallel copy. Mutation-verified: flipping `NotYet` → `Refused` fails `an_update_notification_does_not_answer_the_subscribe`.

**4. The refusal case is deliberately kept distinct.** Folding `subscribed: false` into "keep waiting" would turn a clean refusal into a timeout hang. Covered by `a_refusal_is_an_answer_not_a_reason_to_keep_waiting`.

### Checked and clear

- Deadline arithmetic uses `saturating_duration_since`, so no underflow panic if the clock moves.
- The `web_api` guard is acquired in both branches, so the handler's `drop(web_api)` before `get_room` still holds — this was a real compile-time catch, not a latent bug.
- `subscribe_to_contract` drops rather than queues, which is documented with its rationale; its callers re-read contract state rather than depending on notifications for correctness.

### Not addressed here (deliberate)

What the node sends before the ack is freenet-core#4970. This makes the client correct regardless — dying on an interleaved message on a multiplexed async stream is wrong by construction — but it makes the symptom cheap, not absent.

### Verdict

**Ready to merge.** 273 tests pass, `clippy --all-targets` clean, `cargo fmt` applied. HEAD `4a7169d5`.

[AI-assisted - Claude]

PR #520, opened 2026-07-27
fix(ui): unread badges and totals respect per-room NotificationMode

### LINKED ISSUE #500: fix(ui): unread badges ignore per-room NotificationMode; Muted rooms still badge and count toward title/hamburger
Observed by Ian (2026-07-26): the room-row unread badge showed "3" beside a room whose per-room notification setting is "Mentions and replies only".

## Finding

`NotificationMode` (#340, resurfaced by #429's bell modal) gates exactly one behavior in the app: whether a browser notification fires (`ui/src/components/app/notifications.rs:635-659`). Every unread-count surface was built independently on `last_read_message_id` and never consults the mode:

- Room-row badge: `room_list.rs:522-530`, fed by `count_unread_in_room_data` (`document_title.rs:172-194`).
- Document-title `(N)`: `document_title.rs:204-211`.
- Mobile hamburger badge (#445): `document_title.rs:322-335`.

Consequence: a **Muted** room still paints an accent-colored count badge and inflates the tab title and hamburger totals, contradicting the modal's own "never notify" wording (`notification_modal.rs:39-41`). A MentionsAndReplies room badges all unreads, which may be defensible as an activity signal, and is neither documented nor pinned by any test. #429's diff touched zero counting code.

Two quirks that further inflate the number:

- Room-event messages (`CONTENT_TYPE_EVENT`, e.g. "X joined the room") count toward the badge; `is_action()` (`common/src/room_state/message.rs:813-816`) excludes only `CONTENT_TYPE_ACTION`. Consistent with `display_messages()`, surprising in a badge, so a "3" can be 2 messages plus a join.
- If `last_read_message_id` has aged out of the bounded message buffer, the count falls back to every other-authored message in the buffer (`document_title.rs:164-167`, deliberate).

## Suggested behavior

- **Muted**: no numeric badge (a muted-styled dot at most), and excluded from `count_total_unread_messages` / `count_unread_behind_rooms_panel` so the tab title and hamburger stop counting it.
- **MentionsAndReplies**: either keep all-unreads badging as an activity signal (and document that), or badge the mention/reply count. The detection helper `mentions_or_replies_to_self` (`notifications.rs:507-528`) already exists, is pure, and only needs `pub(crate)`.
- Pin the chosen semantics with `count_unread_*` unit tests in `document_title.rs`'s existing test module.

## Minor adjacent finding

`notifications.rs:636` and `:662` use infallible `ROOMS.read()` on the inbound-message path, contrary to `.claude/rules/dioxus-signal-safety.md`. Currently safe (runs after `with_mut` releases), fragile if the call ever moves inside a borrow.

[AI-assisted - Claude]

sanity:
## Problem

`NotificationMode` (#340, resurfaced by #429's bell modal) gated exactly one thing: whether a browser notification fires. Every unread-count surface was built independently on `last_read_message_id` and never consulted it.

So a **Muted** room still painted an accent-coloured count badge and inflated the tab title and the mobile hamburger totals — directly contradicting the modal's own "never notify for this room" wording. Ian saw a "3" beside a room set to "Mentions and replies only".

Two quirks inflated the number further: room-event messages ("X joined the room") counted toward it, so a "3" could be 2 messages plus a join.

Closes #500.

## Approach

One mode-aware per-room count, `count_unread_in_room_data_with_mode`, that all three surfaces sum — the room-row badge, the document-title `(N)`, and the mobile hamburger badge — so they cannot disagree:

- **Muted** → always 0. No badge, no contribution to either total.
- **MentionsAndReplies** → only messages that @mention the user or reply to one of their own, through the same predicate that gates browser notifications, so the badge and the popup cannot drift on what qualifies. The row's `aria-label` says "unread mention or reply" rather than "unread messages".
- **All** → unchanged.

Room events are excluded from every mode. The mode-blind counter is now private: #500 was three surfaces reaching for exactly it.

**The mention scan decrypts message bodies**, so it is memoized per room behind a fingerprint of everything it reads. Without that it runs up to three times per `ROOMS` write, for rooms the user is not even looking at — the shape of the documented `MESSAGE_HTML_CACHE` jank incident, worse placed.

**Undecryptable bodies fail safe, but only while that is transient.** `decrypt_message_content` returns a placeholder rather than an error, so a mention check against it reads as a confident "not a mention". In mentions mode a body we cannot read is therefore counted — over-reporting resolves down, a silent zero hides a real mention. That is bounded: a message sealed under a version older than every secret we hold is one the user joined too late to ever read, and counting it would put a permanent number on the badge that opening the room cannot clear.

## Testing

Rust unit tests cover the counting, the modes, the event exclusion, the memo, and each arm of the fail-safe. Three of them exist because a review found the first version of the fail-safe wrong, and each was mutation-verified against the specific bug it describes:

- `mentions_mode_counts_a_body_whose_secret_is_present_but_wrong` — a secret present at the right version but WRONG also yields a placeholder. `repopulate_secrets_from_state` exists because an inviter can supply exactly that, so this is a live path, not a corner case.
- `a_same_version_secret_overwrite_invalidates_the_memo` — that overwrite lands at the same version, so a `secrets.len()` fingerprint would keep serving the stale count after the authoritative blob arrives. The fingerprint hashes the sorted `(version, secret)` pairs.
- `mentions_mode_does_not_count_a_body_rotated_permanently_past` — the bound above.

`every_cross_room_total_is_mode_aware` is a source pin: reverting either cross-room total to a mode-blind sum left the entire Rust suite and the entire browser suite green. The title cannot be covered in the browser, because the visible→hidden transition marks every room read, so the fixture's title never carries an `(N)`.

Browser coverage seeds one example room as Muted — otherwise the muted semantics are unreachable, since the bell modal only opens from the current room's header and opening a room marks it read. `room-unread-badge.spec.ts` asserts that room never badges while the other two do, and that the hamburger total equals the sum of the rendered room badges. `test_create_example_rooms` pins the fixture's premise (that room really does have unread other-authored messages and no read marker) in `cargo test`, so the browser assertion cannot go quietly vacuous.

Adding a muted room to the fixture also silently gutted `hamburger-unread-badge.spec.ts`, whose final "the badge clears once all are read" assertion became satisfiable by muting rather than by reading. Fixed by visiting the muted room first, with a stronger assertion in its place: opening it must leave the total exactly unchanged.

719 tests pass with `--features example-data,no-sync`, 716 without; `cargo fmt` clean.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/document_title.rs
+++ b/ui/src/components/app/document_title.rs
@@ -175,6 +175,18 @@ fn get_current_room_name() -> Option<String> {
 /// invariant `MessagesV1::apply_delta` maintains — so the slice after the
 /// marker's index is exactly the set of messages newer than the marker.
 pub fn count_unread_in_room_data(room_data: &crate::room_data::RoomData) -> usize {
+    unread_candidate_messages(room_data).count()
+}
+
+/// The unread tail of `room_data` as an iterator: display messages
+/// (non-action, non-deleted, non-event) authored by other users after
+/// `last_read_message_id`. Shared core of both counting modes
+/// ([`count_unread_in_room_data`] and the MentionsAndReplies arm of
+/// [`count_unread_in_room_data_with_mode`]) so the tail-slicing and
+/// exclusion rules can't drift between them.
+fn unread_candidate_messages(
+    room_data: &crate::room_data::RoomData,
+) -> impl Iterator<Item = &river_core::room_state::message::AuthorizedMessageV1> {
     let self_member_id: MemberId = room_data.self_sk.verifying_key().into();
     let recent = &room_data.room_state.recent_messages;
 
@@ -197,8 +209,52 @@ pub fn count_unread_in_room_data(room_data: &crate::room_data::RoomData) -> usiz
                 && !m.message.content.is_event()
                 && !recent.actions_state.deleted.contains(&m.id())
         })
-        .filter(|m| m.message.author != self_member_id)
-        .count()
+        .filter(move |m| m.message.author != self_member_id)
+}
+
+/// Count the unread messages in `room_data` that its
+/// [`NotificationMode`](crate::room_data::NotificationMode) surfaces
+/// (freenet/river#500). This is THE per-room badge value: the room-list
+/// badge, the document-title `(N)` total, and the mobile hamburger badge
+/// all sum these same per-room values, so every surface agrees.
+///
+/// * `All` — every unread other-authored display message
+///   ([`count_unread_in_room_data`]).
+/// * `MentionsAndReplies` — only unread messages that @mention the user or
+///   reply to one of their messages (the same
+///   [`mentions_or_replies_to_self`] predicate that gates browser
+///   notifications). Zero qualifying messages means no badge even when
+///   other unreads exist.
+/// * `Muted` — always 0; a muted room never badges and never inflates the
+///   totals, matching the modal's "Never notify for this room" wording.
+///
+/// Performance: the mention scan decrypts message content, so it runs only
+/// over the unread tail (the `start..` slice inside
+/// [`unread_candidate_messages`], typically small) and only for rooms in
+/// MentionsAndReplies mode — All and Muted rooms never decrypt anything.
+pub fn count_unread_in_room_data_with_mode(
+    room_data: &crate::room_data::RoomData,
+    mode: crate::room_data::NotificationMode,
+) -> usize {
+    use crate::room_data::NotificationMode;
+    match mode {
+        NotificationMode::All => count_unread_in_room_data(room_data),
+        NotificationMode::Muted => 0,
+        NotificationMode::MentionsAndReplies => {
+            let self_member_id: MemberId = room_data.self_sk.verifying_key().into();
+            let recent = &room_data.room_state.recent_messages.messages;
+            unread_candidate_messages(room_data)
+                .filter(|m| {
+                    crate::components::app::notifications::mentions_or_replies_to_self(
+                        m,
+                        self_member_id,
+                        &room_data.secrets,
+                        recent,
+                    )
+                })
+                .count()
+        }
+    }
 }
 
 /// Count total unread messages across all rooms — room messages plus
@@ -209,11 +265,19 @@ pub fn count_unread_in_room_data(room_data: &crate::room_data::RoomData) -> usiz
 /// DM unread is tab-title-relevant because the issue lists "incoming DM
 /// notifications + unread tracking" as a single line item — without this
 /// the inbox badge would update but the document title wouldn't.
+///
+/// Room totals go through the same mode-aware per-room values as the
+/// room-list badge ([`count_unread_in_room_data_with_mode`], via the
+/// shared [`count_unread_excluding_room`] sum with no exclusion), so a
+/// Muted room never inflates the tab title and a MentionsAndReplies room
+/// contributes only its qualifying messages (freenet/river#500). DM
+/// counts are unaffected — DMs have no per-thread mode.
 pub fn count_total_unread_messages() -> usize {
     let Ok(rooms) = ROOMS.try_read() else {
         return 0;
     };
-    let room_unread: usize = rooms.map.values().map(count_unread_in_room_data).sum();
+    let room_unread: usize =
+        count_unread_excluding_room(&rooms.map, &rooms.notification_modes, None);
     let dm_unread: usize = count_unread_dms(&rooms);
     room_unread + dm_unread
 }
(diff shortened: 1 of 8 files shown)


sanity:
## Review rounds

Two independent blind Claude lenses on the parked branch (skeptical adversarial, testing), then a third on the fixes. Codex is quota-exhausted, so per the project rule these are the review. **Both first-round lenses independently found the same high-severity bug**, from different directions.

### Round 1: the fail-safe missed the case it was written for

`is_undecryptable` tested only whether a secret exists at the message's version. But `decrypt_message_content` has a third outcome: the secret is present and decryption *fails*, in which case it returns `"[Encrypted message: N bytes, vM]"` — no mention token, no reply target. So the counter read it as a confident "not a mention" and hid a real mention, which is exactly what the fail-safe exists to prevent.

Not hypothetical: `RoomData::repopulate_secrets_from_state` exists *because* an inviter can supply a wrong secret that the authoritative owner-signed blob later replaces.

**And the memo then cemented it.** That replacement lands at the SAME version, so `secrets.len()` was unchanged and the fingerprint did not invalidate — the stale answer survived the very event that fixes it.

Fixed by asking the real question: `try_decrypt_message_content` returns the body's text or `None` when it cannot be read at all, and `decrypt_message_content` is now the placeholder wrapper around it, so there is one definition of "what does this body say". The fingerprint hashes the sorted `(version, secret)` pairs.

### Round 2: the bound was one-sided

Round 1 also bounded the fail-safe so a permanently-unreadable body stops counting — but only on one axis. The third lens showed the other two are not reliably transient either:

- **A version newer than everything held** looks like a blob in flight, but `MessagesV1`'s own docs describe a member offline across a rotation, missing a blob at `current_version`, as a supported and unrecoverable state; a removed member's blobs are pruned outright while their client keeps ingesting messages at the new version.
- **A wrong key at a held version** is overwritten only if the contract carries an owner-signed blob for *this* member at *that* version.

Either would put a number on the badge, the title and the hamburger that no amount of reading can clear — the reported symptom in a narrower configuration.

So the fail-safe now covers exactly one state: **the secrets map is empty**. That is every cold start of an established private room (`secrets` is `#[serde(skip)]`), it is the case the original doc justified at length, and it resolves in seconds. The tradeoff is stated in the code: an under-count here is temporary and self-healing, and the user cannot act on a mention inside a message they cannot read — message and badge appear together when the secret arrives. A permanent over-count is neither.

Every one of these has a test that goes red against the specific bug it describes; each was mutation-verified rather than trusted green.

### Other findings addressed

- **The notification path claimed an invariant this change broke** — "the badge and the notification can't disagree about what qualifies". They now share one predicate for what a mention IS, and the doc states the one deliberate difference and why (a badge that over-reports resolves down; a notification cannot be taken back, and every cold start would fire a burst).
- **`notify_new_messages` re-read `ROOMS` and rebuilt the self-authored index once per message** — a K-message delta did K signal reads and K index builds, the exact cost the index was added to remove. Now one read and one index per batch. Round 2 also caught the buffer clone being unconditional, so `All` (the default) and `Muted` paid for a deep clone they never read.
- **Two infallible `ROOMS.read()` calls on the inbound-message path** (the ones the issue itself flagged) are now `try_read`. The mode read *returns* on failure rather than defaulting — the default is `All`, so `unwrap_or_default()` there would notify a MUTED room.
- **`text_mentions_or_replies_to_self` took two independent views of "self"** with nothing coupling them; it now takes it from the index only.
- **The mode-blind `count_unread_in_room_data` is private.** #500 was three surfaces reaching for exactly it.

### Test-coverage findings

- **Adding a muted room to the fixture silently gutted `hamburger-unread-badge.spec.ts`**: its "the badge clears once all are read" assertion became satisfiable by muting rather than by reading, and its last loop iteration was dead. The muted room is now visited first, with a stronger assertion in its place — opening it must leave the total exactly unchanged, which pins that muting REMOVES the room from the total rather than merely reducing it. Round 2 caught that this assertion could pass on a stale first poll; it now re-checks after a further render pass.
- **Nothing pinned that the document title is mode-aware.** Reverting it to a mode-blind sum left the whole Rust suite and the whole browser suite green, and it cannot be covered in the browser because the visible→hidden transition marks every room read, so the fixture's title never carries an `(N)`. Added a mutation-verified source pin over both cross-room totals.
- **The muted-room fixture's premise was unpinned.** "A muted room shows no badge" only means something while that room really has unread other-authored messages and no read marker; `test_create_example_rooms` now asserts that in `cargo test`.

Lenses: skeptical (adversarial), testing (coverage + vacuity), skeptical again on the fix delta. 723 tests pass with `--features example-data,no-sync`.

[AI-assisted - Claude]

Issue #425, opened 2026-07-19
perf(ui): parallelize/defer the 72-request startup legacy-migration probe burst (cold-load, delegate-queue flood)


sanity:
## Problem

On startup, whenever the current chat delegate has no rooms — i.e. **every brand-new user, and any first-time-on-this-browser load** — the app fires a burst of **72 serial delegate requests** to probe old delegate generations for migratable data: `LEGACY_DELEGATES` is 24 generations, and `fire_legacy_migration_request` (`ui/src/components/app/chat_delegate.rs:~5197/5211`) loops over all 24, sending **3 serial `api.send().await`s each** (2 fixed GETs for `rooms_data`/`outbound_dms` + 1 `ListRequest` for per-room key discovery) = 48 GETs + 24 ListRequests.

For a brand-new user every probe returns empty — pure wasted work. Two costs on the cold-load path:

1. **Gates the empty-state resolution.** The `LOAD_IDLE_MS` debounce (1500ms after #421) starts after the *last* legacy activity, so on a high-latency link the burst pushes out when "No rooms yet" appears.
2. **Floods the node's delegate queue.** The probes land in freenet-core's `FairEventQueue` `Default` queue, whose per-tier cap is `MAX_QUEUED_PER_CONTRACT = 100` (admission past it is rejected). 72 legacy probes + `RegisterDelegate` + the current-delegate `ListRequest` + the outbound-DMs load ≈ 75+ concurrent — uncomfortably near the cap. (This is the same cap #419 had to wave-bound at 16 for the per-room load.)

Also: in the hosted null-origin iframe the `mark_legacy_migration_done()` localStorage flag **can't persist**, so this burst re-runs on *every* cold load for empty-delegate users, not just once per browser. (The primary suppression for returning users is the current-delegate `PerRoom` path, which is unaffected — only the empty-delegate `ProbeLegacy` case hits the burst.)

Measured on try.freenet.org: ~190ms active burst on a fast link (fires after the shell is interactive at ~240ms, so it doesn't block *interactivity* at low latency, but it does gate the rail's resolution and scales with latency).

## Approach (options — the assignee should pick)

[...]

Issue #159, opened 2026-03-09
fix: add invitation link to Freenet room on River welcome page


kerbo7:
### Problem
New users are probably a bit clueless what to do next, after they loaded River.

> Welcome to River. Create a new room, or get invited to an existing one.

A new user doesn't know what to do now.
Especially on mobiles it may be a bit frustrating not to see the next steps.

### Solution 
Below the welcome message add a link to 

`https://freenet.org/quickstart#invite-form`

Text eg: Click here to get an invitation to channel "Freenet Official"  

<img width="642" height="788" alt="Image" src="https://github.com/user-attachments/assets/6917c4b9-0716-4c44-a62d-3abb3cee2413" />

PR #521, opened 2026-07-27
fix(ui): tell a mobile user their rooms are loading, not that they have none

### LINKED ISSUE #509: fix(ui): mobile shows "Welcome to River, create a new room" while rooms are still loading (loading/failed states are desktop-only)
Reported by Ian (2026-07-26) loading https://try.freenet.org/v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/ on Android: the rooms loading indicator is not visible, which can leave a user thinking the UI is broken.

Confirmed, and the impact is worse than a missing spinner: **on mobile the app affirmatively tells a mid-load user they have no rooms.**

## Mechanism

The loading / migrating / empty / failed states from #397 live entirely inside `RoomList` (`ui/src/components/room_list.rs:341-387`, driven by `ROOMS_LOAD_STATE`). Below the 768px breakpoint that whole panel is `display:none` rather than unmounted (`ui/src/components/app.rs:476-500`), and the default mobile panel is Chat (`app.rs:85`, `MOBILE_VIEW` defaults to `MobileView::Chat`).

`CURRENT_ROOM` starts `None` and is only set after the load resolves, so for the entire load window the conversation panel renders its no-room arm (`ui/src/components/conversation.rs:4337-4400`):

> **Welcome to River**
> Create a new room, or get invited to an existing one.

That arm is unconditional on `ROOMS_LOAD_STATE` — it renders identically whether rooms are Loading, Migrating, LoadFailed, or genuinely empty. It is the exact false-empty message #397 was written to eliminate on the desktop rail, still shipping on mobile.

`ROOMS_LOAD_STATE` has exactly one consumer in the whole UI (`room_list.rs:247`), which is the root cause: #397 never touched `app.rs` or `conversation.rs`, so it was never wired into the mobile-visible surface.

Two aggravating details:

1. **A green "Connected" pill appears on that same screen** (`conversation.rs:4392-4399` -> `members.rs:47-89`) and flips to Connected within roughly a second, while rooms are still loading. It reflects only WebSocket transport, so it actively reassures the user that nothing is pending.
2. **`LoadFailed` is invisible on mobile too**, along with its Retry button. A stalled load is pixel-identical to an empty account, so the user is advised to create a room when the correct action is retry.

## How long the window lasts

Blank page for roughly 2-5s (WASM download plus streaming compile; #424 notes ~1.38MB of that payload is embedded contract/delegate WASM and the bar is "seconds at real-world bandwidth"), then the false Welcome screen for another ~2-4s: WebSocket connect, `RegisterDelegate`, `ListRequest`, per-room GETs in waves of 16, the legacy-migration probe burst (#425: 72 serial delegate requests, and in the hosted null-origin iframe the done-flag cannot persist so it re-runs on every cold load), then a 1500ms quiescence debounce. Worst case the hard backstop is 60s.

## Evidence

Reproduced at a Pixel 5 viewport against a local build: zero loading indicators present in the DOM at any point, body text shows the Welcome copy immediately. Screenshot attached in the thread.

## Fix

**Recommended (small, low risk):** branch the conversation panel's no-room arm on the load state, reusing `room_list_display_state()` (`room_list.rs:76`) so the two surfaces can never disagree:

- Loading -> spinner + "Loading your rooms…"
- Migrating -> spinner + "Migrating your rooms… (one-time step after an update)"
- LoadFailed -> "Couldn't load your rooms" + Retry (`retry_rooms_load()`), which also fixes the invisible-error case
- Empty / List -> today's Welcome copy, unchanged

It is purely additive rendering with no state-machine change, and being viewport-independent it also improves the desktop centre panel, which shows the same false-empty text today.

Alternatives considered: auto-opening the room list on mobile while loading (fights the current-room restore path and can steal the panel mid-interaction), and a global first-load overlay (the only option that addresses the blank-page phase, but needs a custom `index.html` and risks a hard lockout if it fails to dismiss; worth pairing with #424 as a follow-up).

## Test coverage

There is none: grepping for `room-list-loading` / `room-list-migrating` / `room-list-empty` / `room-list-retry-button` across the repo returns no matches, so #397 shipped with Rust unit tests only. Note that `ui/tests/welcome-invite-link.spec.ts:43-50` currently asserts the Welcome screen as the correct 375x812 first-load state, i.e. it pins this bug — any fix must update it. Reproducing the loading state in Playwright needs a hook to hold `ROOMS_LOAD_STATE` at Loading with an empty `ROOMS`, since the example-data build always seeds rooms; `example_data::install_test_hooks` is the natural place.

[AI-assisted - Claude]

sanity:
## Problem

On Android, loading `https://try.freenet.org/v1/contract/web/raAqMhMG.../` shows no loading indicator — and the impact is worse than a missing spinner: **the app affirmatively tells a mid-load user they have no rooms.**

#397 added the loading / migrating / failed / empty states, but put them entirely inside `RoomList` (`room_list.rs:341-387`). Below the 768px breakpoint that panel is `display:none` rather than unmounted (`app.rs:476-500`), and the default mobile view is Chat (`app.rs:85`). `CURRENT_ROOM` starts `None` and is only set after the load resolves, so for the whole load window the only thing a phone user can see is the conversation panel's no-room screen — which rendered, unconditionally:

> **Welcome to River**
> Create a new room, or get invited to an existing one.

That is the exact false-empty message #397 was written to eliminate, still shipping on mobile. `ROOMS_LOAD_STATE` had exactly one consumer in the whole UI, which is the root cause: #397 never touched `app.rs` or `conversation.rs`.

Two aggravating details, both fixed here:

- **`LoadFailed` was invisible too**, along with its Retry button. A stalled load was pixel-identical to an empty account, so the user was advised to create a room when the correct action was retry.
- **A green "Connected" pill sat on that same screen** (`conversation.rs:4392-4399`), flipping within about a second while rooms were still loading. It reflects only WebSocket transport, so it actively reassured the user that nothing was pending.

The window is roughly 2-4s of false Welcome screen after the WASM loads — connect, `RegisterDelegate`, `ListRequest`, per-room GETs in waves of 16, the legacy-migration probe burst (#425), then a 1500ms quiescence debounce — with a 60s hard backstop.

Closes #509.

## Approach

Branch the no-room screen on the same state the rail uses, through a shared `current_room_list_display()`, so the two surfaces can never disagree about whether rooms are still loading. Purely additive rendering; no state-machine change.

Viewport-independent on purpose — the desktop centre panel showed the same false-empty text.

The connection pill now lives in the Welcome arm only, so it no longer appears on a still-loading screen. `Empty` (the load resolved and there really are no rooms) and `List` (rooms exist, none selected yet) both keep today's screen unchanged.

Alternatives considered and rejected in the issue: auto-opening the room list on mobile while loading (fights the current-room restore path and can steal the panel mid-interaction), and a global first-load overlay (the only option that also addresses the blank-page phase, but needs a custom `index.html` and risks a hard lockout if it fails to dismiss).

## Testing

There was none — grepping for #397's test ids across the repo returned nothing, so those states shipped with Rust unit tests only. Which is how they went unnoticed on the surface that matters.

`rooms-loading-state.spec.ts` covers all four states at a mobile AND a desktop viewport (8 cases). Reaching them needs a hook, since the example build always seeds rooms and never leaves `ROOMS_LOAD_STATE` at anything but its default: `__riverTest.setRoomsLoadState(state)` clears `ROOMS` and sets the state, gated on `example-data` + `no-sync` like its neighbours.

Every state test carries the premise that the screen opens on the Welcome copy, so "Welcome is gone" cannot pass without the fix doing something. The `loaded` case pins that a genuinely empty account still gets the Welcome screen — without it the fix could suppress that screen entirely and every other assertion would still pass.

**Mutation-verified in the browser**, not just asserted: forcing the resolved arm (the pre-fix rendering) and rebuilding turns all six state cases red while the `loaded` case stays green. `the_no_room_screen_branches_on_the_shared_load_state` pins the same wiring in Rust, and goes red under the same mutation.

Note `welcome-invite-link.spec.ts` still passes: with rooms seeded, `room_list_display_state` returns `List`, which keeps today's screen.

701 Rust tests pass; 8/8 Playwright locally against a static example-data build; `cargo fmt` clean.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation.rs
+++ b/ui/src/components/conversation.rs
@@ -2154,6 +2154,41 @@ fn install_scroll_pin_listeners(
     true
 }
 
+/// The two affordances the no-room screen must offer in EVERY load state
+/// (freenet/river#509).
+///
+/// * The #159 quickstart invite link. A brand-new user has `room_count == 0`,
+///   so they are in an unresolved state for the whole load window; putting the
+///   link only on the resolved screen would hide their one concrete next step
+///   for exactly as long as they need it.
+/// * The connection pill. On mobile the rooms rail is `display:none`, so this
+///   is the only copy of it a phone user can see (Bug #5, Ivvor 2026-05-17).
+///   It matters most in the unresolved states: a node that never connects
+///   leaves `ROOMS_LOAD_STATE` at its `Loading` default forever, because
+///   `begin_load_attempt` — which arms the 60s backstop — runs only after a
+///   successful connect. Without the pill that user watches a spinner with
+///   nothing on screen saying the socket is down. It is also what makes the
+///   failed state's "Check your connection and try again" actionable.
+///
+/// One component so the three unresolved arms and the Welcome arm cannot drift.
+#[component]
+fn NoRoomFooter() -> Element {
+    rsx! {
+        p { class: "text-text-muted mt-3",
+            a {
+                class: "text-accent hover:underline",
+                href: "https://freenet.org/quickstart#invite-form",
+                target: "_blank",
+                rel: "noopener noreferrer",
+                "Click here to get an invitation to channel \"Freenet Official\""
+            }
+        }
+        div { class: "mt-8 md:hidden",
+            crate::components::members::ConnectionStatusIndicator {}
+        }
+    }
+}
+
 #[component]
 pub fn Conversation() -> Element {
     let current_room_data = {
@@ -4415,10 +4450,23 @@ pub fn Conversation() -> Element {
                         // they had no rooms, and hid a FAILED load (and its
                         // Retry button) behind advice to create one.
                         //
-                        // Branch on the same shared state the rail uses, so the
-                        // two surfaces cannot disagree. This is viewport-
-                        // independent on purpose: the desktop centre panel had
-                        // the same false-empty text.
+                        // Branches on the SAME call the rail makes, so the two
+                        // surfaces cannot disagree. Viewport-independent on
+                        // purpose: the desktop centre panel had the same
+                        // false-empty text.
+                        //
+                        // Every arm keeps the two things that are useful in any
+                        // state: the #159 quickstart invite link (a brand-new
+                        // user has `room_count == 0`, so they are in the
+                        // unresolved states for the whole load window and would
+                        // otherwise lose their only onboarding affordance) and
+                        // the connection pill. The pill matters MOST here: a
+                        // node that never connects leaves `ROOMS_LOAD_STATE` at
+                        // its `Loading` default indefinitely — `begin_load_attempt`,
+                        // which arms the 60s backstop, runs only after a
+                        // successful connect — so without the pill that user
+                        // would watch a spinner with nothing telling them the
+                        // socket is down (#509 review).
                         match crate::components::room_list::current_room_list_display() {
                             crate::components::room_list::RoomListDisplay::Loading => rsx! {
                                 div {
@@ -4426,6 +4474,7 @@ pub fn Conversation() -> Element {
                                     "data-testid": "conversation-rooms-loading",
                                     div { class: "animate-spin w-6 h-6 border-2 border-text-muted border-t-transparent rounded-full" }
                                     span { class: "text-sm text-text-muted", "Loading your rooms…" }
+                                    NoRoomFooter {}
                                 }
                             },
                             crate::components::room_list::RoomListDisplay::Migrating => rsx! {
@@ -4435,6 +4484,7 @@ pub fn Conversation() -> Element {
                                     div { class: "animate-spin w-6 h-6 border-2 border-text-muted border-t-transparent rounded-full" }
                                     span { class: "text-sm text-text-muted", "Migrating your rooms…" }
                                     span { class: "text-xs text-text-muted opacity-70", "(one-time step after an update)" }
+                                    NoRoomFooter {}
                                 }
                             },
                             crate::components::room_list::RoomListDisplay::LoadFailed => rsx! {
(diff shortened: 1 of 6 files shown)


sanity:
## Review round

Two independent blind Claude lenses (skeptical adversarial, code-first). Codex is quota-exhausted, so per the project rule these are the review. **One blocker**, and a cluster of findings that turned out to share a single cause: the three new arms replaced the whole panel, so they dropped things useful in every state. All fixed in `1ea1ebf7`.

### Blocker: `Loading` has no time bound when the socket never connects

`ROOMS_LOAD_STATE` defaults to `Loading`, and the only thing that advances it is `set_up_chat_delegate()`, whose `begin_load_attempt()` — which arms the 60s hard backstop — runs **only after a successful connect** (`freenet_synchronizer.rs:587`). No transport failure reaches `LoadFailed`. So a user whose node is unreachable sits at `Loading` with an empty `ROOMS`, indefinitely.

Before this PR that user saw the Welcome copy **plus** the inline connection pill reading `"Error: WebSocket connection error: WebSocket connection failed or timed out"`. The first version of this fix replaced that with a bare spinner: no pill, no error text, no Retry. Strictly worse than the bug being fixed — and the reviewer found the evidence in my own local Playwright artifacts, which had captured exactly that page.

The same omission made "Couldn't load your rooms / Check your connection and try again" the only screen in the app that hides the connection status. That state is terminal until the user acts, so the "the pill flips green within a second, which reads as everything is done" argument never applied to it.

And a brand-new user has `room_count == 0`, so they are in the unresolved states for the whole load window — losing the #159 quickstart invite link, their one concrete next step, exactly when they need it.

All three are one fix: a `NoRoomFooter` component carrying the invite link and the pill, rendered by every arm, pinned at four call sites so a future arm cannot quietly omit it.

### Other findings fixed

- **"The two surfaces cannot disagree" was an overclaim.** The rail still derived its state inline and agreed only because both paths happened to funnel into `room_list_display_state` with counts that happened to match; `current_room_list_display()` had exactly one caller. The rail now calls it too, so the claim is true rather than coincidental.
- **`_ =>` became `Empty | List =>`.** A wildcard on the mobile-only surface would absorb a future variant into "you have no rooms" — the #397 → #509 story verbatim, and the catch-all pattern the project's bug-prevention rules call out. The compiler enforces it now.
- **The `loaded` control test could not fail for the reason it existed.** It asserted Welcome-visible and both state elements absent — all already true of the pre-hook DOM, so the web-first assertions resolved against the old render. It would have stayed green if the hook did nothing (it logs and returns on an unrecognised string) or if the `Empty` arm were broken. It now drives `loading` → asserts Welcome is gone → `loaded`, and checks `room-list-empty`, which proves both halves of the premise — the state reached `Loaded` **and** `ROOMS` really was cleared. `WELCOME` alone cannot, since it renders for `List` too.
- **The spec now asserts the rail agrees**, which is the literal test for this PR's central claim, and is the first browser coverage #397's rail test ids have ever had. Their absence is why #509 went unnoticed.
- **The source pin's ordering assertion is gone.** "the pill's index is after the Welcome text" was a weak proxy for "the pill is inside the Welcome arm" — hoisting it out of the match entirely would also satisfy it — and, worse, it would have **blocked** the blocker fix above. Replaced with a count of `NoRoomFooter` call sites.
- **Doc correction:** the helper claimed an unresolved read never resolves to `Empty`. True of the load-state read, false of the room-count read — which is deliberate, and now says so.

### Verified and ruled out

- **The test hook cannot exist in a production build:** `install_test_hooks` is `#[cfg(all(target_arch = "wasm32", feature = "no-sync"))]` with a no-op native twin, and its only call site additionally requires `example-data`. All three of its writes are inside one `defer`, non-overlapping.
- **`welcome-invite-link.spec.ts` is not broken by this fix**, contrary to what #509 predicted: the example build seeds `ROOMS` at `Global` init, so `room_count > 0` → `List` → the Welcome screen from the first render, with no spinner flash. Its comment is updated anyway, since it described a first-load state that now has four variants.
- **State table walked** for every `(RoomsLoadState, room_count, CURRENT_ROOM)` combination, including "rooms exist, none selected" (Welcome), "Loaded with zero rooms" (Welcome), and "user leaves their last room" (Welcome).
- **The mutation check was corroborated** against the on-disk failure artifacts: forcing the resolved arm turns exactly the six state cases red across both viewports and leaves the two control cases green.

Lenses: skeptical (adversarial), code-first (blind to the description). 701 Rust tests pass; 17/17 Playwright locally across this spec and the two it interacts with.

[AI-assisted - Claude]

sanity:
Rebased onto current `main` after #520 merged. One conflict in `ui/src/components/room_list.rs` — both sides added a function at the same offset (#520's `unread_badge_label`, this PR's `current_room_list_display`); both kept, no other change.

Per the project's per-content review rule a conflict resolution is an unreviewed judgment call, so it was re-verified rather than assumed: 724 Rust tests with `--features example-data,no-sync` (719 without), and 23/23 Playwright locally across this spec plus the four it now shares a fixture with — `welcome-invite-link`, `connection-status-indicator`, `room-unread-badge`, `hamburger-unread-badge`.

[AI-assisted - Claude]

Issue #85, opened 2026-02-09
Invitation not joining room when room already exists in delegate


sanity:
## Bug

When a user already has a room in their delegate (ROOMS map) but is not a member of that room, and then receives and accepts an invitation to that room, they don't actually join — they can't send messages and don't appear as a member from their own perspective.

## Root Cause

In `ui/src/components/app/freenet_api/response_handler/get_response.rs` lines 102-138, when processing a pending invitation:

```rust
let room_data = entry.or_insert_with(|| {
    RoomData {
        self_sk: self_sk.clone(),  // Only set for NEW entries
        ...
    }
});

if !is_new_entry {
    // Merges state but does NOT update self_sk
    room_data.room_state.merge(...);
}
```

`or_insert_with` only runs for **vacant** entries. When the room already exists, `self_sk` is never updated to the invitation's `invitee_signing_key`. The invitation's member entry gets added to `room_state.members` (via the delta applied later), but `room_data.self_sk` still holds the old key.

Since `can_send_message()` (room_data.rs:154) checks `self.self_sk.verifying_key()` against the members list, the stale key doesn't match the newly-added member entry, so the user appears to not be a member.

## Fix

In the `!is_new_entry` branch, update `self_sk` to match the invitation's signing key:

```rust
if !is_new_entry {
    room_data.self_sk = self_sk.clone();
    // ... existing merge code ...
}
```

## How to Reproduce

1. User A creates a room
2. User B somehow gets the room in their ROOMS map (e.g., previous invitation that was revoked, or room data synced without membership)
3. User A invites User B
4. User B accepts the invitation
5. User B cannot send messages — `can_send_message()` returns `UserNotMember`

[AI-assisted - Claude]

PR #524, opened 2026-07-27
fix(contract): bound direct_messages globally so DM-pinned members stop growing

### LINKED ISSUE #490: Receiver-side rejection (ban sweep / membership prune) is invisible to the sender, so the exchange never terminates

### LINKED ISSUE #519: Official room stuck at 200/200 members: the DM exemption pins members forever
## Symptom

The Freenet Official room shows **Members (200/200) — Room is full — new members will be rejected**, and real users are blocked. In-room today:

> *"The chat room is full with 200 people. It just took me 10 tries"*
> *"This was incredibly hard to join from the website requiring multiple retries across 2 days"*

## Root cause: direct messages pin members permanently

Member pruning by activity **is** implemented — not in `MembersV1`, but in `ChatRoomStateV1::post_apply_cleanup` (`common/src/room_state.rs:101-435`), which runs after every delta and retains only members in a computed `required_ids` set built from `recent_messages` authors.

The problem is that `post_apply_cleanup` also exempts **every DM sender and recipient** (`room_state.rs:233`, via `direct_messages.rs:680-690` `active_participants()`), and the DM set has **no global cap and no age bound** — only 100 per ordered (sender, recipient) pair (`direct_messages.rs:131`). DMs accumulate monotonically, so anyone who has ever exchanged a single DM is pinned as a member forever.

## Measured, from a fresh contract GET

```
members: 181     recent_messages: 200/200     bans: 200/200     direct_messages: 499

distinct message authors (members):                    52
distinct DM participants (members):                   139
members that are DM participants with ZERO messages:  129
DM timestamp range: 2026-05-16 -> now; 377 of 499 DMs are >24h old

post_apply_cleanup required_ids simulation:
  after message authors:            52
  after + DM participants:         181
  after + invite-chain ancestors:  181
  members NOT required (prunable):   0
  WITHOUT the DM exemption:         55
```

**126 of 181 members (70%) exist solely because of the DM exemption, and nothing in the room is currently prunable.** Over 26 hours, DMs went 403 → 499 and members 174 → 181. It only goes up.

`remove_excess_members` cannot drain the list: it early-returns when `len <= max_members` (`member.rs:420-422`), so it trims exactly the overflow and holds at 200, never below. The only draining path is the `required_ids` prune, which is currently a no-op here.

## Constraint on any fix

`post_apply_cleanup` must remain a pure, idempotent function of converged state (documented at `room_state.rs:87-93` — Freenet runs it a variable number of times per peer). It may **not** consult wall-clock "now", which rules out a simple time window on the DM exemption.

## Options

**Stop-gap (no contract change):** raise `max_members` via `riverctl room config`. Owner-only config update, no WASM change, no re-key, reversible, unblocks joins immediately. Buys time only — the DM-pinned set keeps growing. Cost: ~338 bytes per member, so 200 → 400 adds roughly 68 KB to a ~260 KB state.

**Real fix (changes room contract WASM):** give `direct_messages` a global deterministic retention bound so the DM-pinned member set is bounded, mirroring how `max_recent_messages` bounds the author set. Owner-tunable `max_direct_messages`, keep newest N by (timestamp, signature). Caveat: `direct_messages.rs:1258` already documents that the per-pair cap makes merge non-monotonic and handles it via `pair_horizons` in the summary; a global cap needs an analogous horizon or peers re-offer trimmed DMs forever.

**Rejected — drop the DM exemption entirely.** One line, takes required from 181 to 55, and the step-6 sweep (`room_state.rs:422`) already drops orphaned DMs in the same pass so `verify` stays consistent. But it deletes the DM history of anyone who hasn't posted recently (the one thing users can't recover) and re-opens the bug class the secrets exemption was added for (#110): an invitee arriving via a DM invite who hasn't spoken yet is pruned immediately.

**Rejected — evict by last activity instead of invite-chain length.** Doesn't help: `remove_excess_members` only trims overflow, so reordering changes who loses at the boundary, not whether the list drains. (Worth noting the invite graph is flat anyway — 170 of 174 members were invited by Invite Bot at chain length 1, so the chain-length sort has almost nothing to discriminate on.)

Any change under `common/src/room_state/` re-keys `room_contract.wasm` and needs a `common/legacy_room_contracts.toml` entry (29 generations already registered) plus a coordinated publish. Invite URLs survive via the #292 auto-migration, but this is a re-key, not a config tweak.

## Explicitly NOT established

The per-attempt mechanism behind *"took me 10 tries"* is **unverified**. One hypothesis was tested and **rejected**: that at cap the MemberId-descending tiebreak ratchets membership toward the lowest ids ever seen — current member ids are uniformly distributed (mean quantile 0.493). The proximate cause is clearly that the room sits at cap, but the retry-level behavior should not be written up as understood.

Two corrections to earlier analysis in #511, recorded so they aren't re-derived:
- "Message pruning doesn't touch the member list, so members never age out" — **wrong**. The prune exists; it lives in `post_apply_cleanup`, not `MembersV1::apply_delta`.
- "The room has ~2 bans" — **wrong**. That was a truncated `riverctl debug bans` view; bans are saturated at 200/200 (though only 4 distinct banners, so bans are not a driver of the member count).

Possibly relevant to #511's unexplained half: `4CLPTJPM` is a deputy of Invite Bot showing `in_room: false`. An invite-issuing identity that never posts, holds no DMs, no current-version secret and no surviving ban is prunable by `required_ids`; it survives only while one of its invitees qualifies and the chain walk (`room_state.rs:310-320`) pulls it back in. That is an inference from code plus the observed flag, not an observed sequence.

[AI-assisted - Claude]

sanity:
## Problem

`ChatRoomStateV1::post_apply_cleanup` exempts every DM sender and recipient from inactivity-prune (via `DirectMessagesV1::active_participants`), and `DirectMessagesV1` had **no global bound** — only 100 per ordered `(sender, recipient)` pair. So anyone who ever exchanged a single DM was pinned as a room member forever, and the member set grew without limit.

Measured on the live official room: **181 members, 499 DMs, 129 of those members held zero messages** and were kept solely by a DM. **0 members were prunable** — `remove_excess_members` early-returns under `max_members`, so it holds at the ceiling and never drains. Without the DM exemption `required_ids` would have been 55 rather than 181.

Raising `max_members` 200→400 was a stop-gap; it buys time and does not bound anything.

## Approach

An owner-tunable `Configuration::max_direct_messages` bounding the DM set globally, mirroring how `max_recent_messages` bounds `recent_messages`, defaulting to `DEFAULT_MAX_DIRECT_MESSAGES = 300`. The trim keeps the newest N by `(timestamp, signature)` — a pure function of the held set, so peers converge regardless of arrival order.

**Why 300.** Measured on the live official room: 499 DMs, 377 of them already older than 24 hours, occupying ~128 KB of a ~260 KB room state — about half the state was DMs. 300 sheds ~40% of that set and roughly halves the DM share of state, without being as aggressive as 200. The per-pair cap of 100 is unchanged; this is a global bound on top of it.

**Accepted cost** (Ian, explicitly): this deletes history on rollout, including DMs the recipient has not read. *"people are gonna lose direct messages before they've read them, but I think we just need to accept that for now. We can implement a better system later."* Deliberately no machinery to soften it — read tracking, age exemptions and grace periods are all wall-clock-dependent, which `post_apply_cleanup` forbids, and all would be discarded by the successor design.

**This bound is interim.** The intended fix is moving DMs out of the room contract into dedicated per-member contracts. So the cap is deliberately the simplest correct thing — a count, a deterministic order, and a horizon — and the trim lives wholly inside `direct_messages.rs` with `post_apply_cleanup` untouched, so extracting DMs later is a deletion rather than an unpick.

### The merge-monotonicity problem, and the horizon that solves it

Trimming makes the merge non-monotonic, so `delta` must never offer a DM the receiver would discard on arrival — that is the resend loop which made the room contract **63.7% of all byte-weighted broadcast work network-wide** on 2026-07-25, and DM ciphertexts are up to 32 KiB each.

The existing per-pair `DmPairHorizon` **cannot** describe this axis: a DM can sit comfortably inside its own pair's window and still be the oldest DM in the room. So this adds `DmRetentionHorizon` (`Open` / `OldestRetained(DmOrderKey)` / `Closed`), the whole-set analogue of `MessagesV1::RetentionHorizon`, published in `DirectMessagesSummary` and filtered on in `delta`.

**Why it terminates.** `OldestRetained` is the smallest key the peer holds, published only once it is *at* the cap. A sender offers only strictly-greater keys; applying one pushes the peer over the cap, so the trim drops at least the horizon message itself and the horizon strictly rises. A peer below the cap publishes `Open` and discards nothing.

**The two horizons compose.** An offered DM either survives both trims (the set grew), or the pair trim drops that pair's oldest (that pair's horizon rose), or the global trim drops the room's oldest (the global horizon rose). Every exchange grows a bounded set or strictly advances a bounded key, so there are no cycles.

Like `MessagesV1`, the horizon is deliberately conservative — the oldest *held* key rather than the exact post-merge cutoff. Under-stating costs one extra round; over-stating would silently withhold DMs the peer would have kept.

## Migration — the part that could have bricked production

`AuthorizedConfigurationV1::verify_signature` re-serializes `Configuration` with ciborium and checks the owner's signature over those bytes. Adding the field naively as `#[serde(default)] usize` changes those bytes, so `verify` — **which also gates the #292 migration PUT** — would fail for **every room created before the field existed**. Verified empirically:

```
old bytes            : a2616101616202
naive re-encode      : a3616101616202616300   MATCH=false   <- every room bricked
Option + skip_if     : a2616101616202         MATCH=true
```

So the field is `Option<usize>` + `skip_serializing_if = "Option::is_none"`, read through `effective_max_direct_messages()`. An old configuration decodes to `None`, re-encodes without the key, and its signature still verifies — while still being bounded by the default, so the fix is **not inert** on existing rooms.

`legacy_configuration_bytes_still_verify_after_adding_the_field` pins this against the **real pre-change CBOR encoding of `Configuration::default()` captured from `origin/main`**, not a re-derivation from the current struct.

Two further migration-safety decisions:

- **The cap is enforced in `apply_delta` only, never in `verify`.** A legacy room whose stored set exceeds the cap must still validate, or the migration PUT is rejected and the room is stranded on an unreachable key. Mirrors `MessagesV1`, which does not enforce `max_recent_messages` in `verify` either. Pinned by `verify_accepts_state_over_the_global_cap_so_the_migration_put_survives`.
- **`verify()` semantics are otherwise unchanged**, so the #292 forward PUT of old state is still accepted.

### Registry entries (added BEFORE the rebuild)

| WASM | Entry | old → new |
|---|---|---|
| room contract | `V30` | `f8cca760…` → `65cd3876…` |
| chat delegate | `V29` | `2f8c5f1d…` → `87e91a7a…` |

Both move because river-core compiles into each; the delegate's copy is **not** dead-code-eliminated here, so it needs its own entry. Verified with the CI-equivalent invocation (not the local mode, which false-fails on intermediate WASM churn):

```
scripts/check-room-contract-migration.sh --ci origin/main HEAD  -> Old hash found … migration entry exists.
scripts/check-migration.sh              --ci origin/main HEAD  -> Old hash found … migration entry exists.
```

Invite URLs survive: an `Invitation` carries the room owner's VK, and clients derive the contract key themselves, probing older generations newest-first.

## Testing

`common/tests/dm_global_cap_test.rs` — 21 tests:

- **The bound**: enforced across pairs, keeps the newest (exact equality, so raising/removing the cap fails rather than passing vacuously); unset config falls back to the default rather than being unbounded; an owner-set value overrides it; the per-pair cap still holds; the pair trim runs *before* the global trim so the budget is not wasted.
- **Determinism/idempotence**: identical result across arrival orders *and* chunkings; `apply_delta` idempotent including the empty-delta path; an over-cap state converges down; the horizon is independent of the state vector's order.
- **The loop**: a sender offers nothing a receiver would globally prune (with premise checks that no pair horizon is doing the work); gossip from a lagging peer terminates; offset windows converge and go quiet; asymmetric caps terminate both directions; a below-cap peer still gets back-fill; a zero cap suppresses rather than loops; the horizon survives the CBOR round trip the contract performs.
- **The #519 symptom**: the DM-pinned member set is bounded (21 members → 6); `post_apply_cleanup` stays idempotent.
- **Migration**: legacy configuration bytes still verify; `verify` accepts an over-cap state.

**Mutation-verified** — six mutations, each caught by the intended test:

| Mutation | Caught by |
|---|---|
| drop the global horizon filter in `delta` | 4 loop/convergence tests |
| remove the global trim entirely | 9 tests |
| swap the trim order (global before pair) | `pair_trim_runs_before_the_global_trim_…` |
| horizon takes first element instead of `min` | `global_horizon_is_independent_of_the_state_vec_order` |
| config field without `skip_serializing_if` | `legacy_configuration_bytes_still_verify_…` |
| enforce the global cap inside `verify` | `verify_accepts_state_over_the_global_cap_…` |

The swap-order mutation initially went **uncaught** and exposed a genuine flaw: the doc comment claimed a `verify` hazard that does not exist (either order leaves pairs legal). Corrected to state the real property — the swapped order wastes the global budget on DMs the pair trim then discards — and pinned.

Full `river-core` suite green (17 retention proptests included). `cmp` confirms the ui/cli WASM copies are in sync and neither carries `__wbindgen_*`/`__wbg_*` imports.

## Trap for the release ritual: comment-only edits re-key the contract

The WASM build is deterministic — a clean rebuild reproduces the hash byte for byte. But **a comment-only change to `common/src/` can still move the room-contract hash**, because `expect`/`panic!` embed `file:line:col` location strings via `core::panic::Location`. Shifting a line above a panic site shifts those strings, and the bytes change.

Confirmed in both directions on this branch:

| edit | effect on `room_contract.wasm` |
|---|---|
| doc edit that shifted lines **above** panic sites in `direct_messages.rs` | hash changed (`ffc82012…` → `c3c787d2…`) |
| doc edit **below** the panic sites in `configuration.rs` | hash unchanged (`4bb55ded…`) |

Diffing two of those builds showed exactly 7 differing bytes: `i32.const 500` → `i32.const 300` in LEB128, plus three panic-`Location` line numbers each shifted by the doc block's growth.

So a docs-only commit is **not** reliably migration-neutral. Observing one comment-only change that left the WASM byte-identical does not generalise — it only means that particular edit didn't shift a panic's line. Re-hash after any `common/src/` edit before assuming no re-key.

## Follow-ups filed

- **#525** — purge envelopes still pin members forever, so this closes only the messages half of #519. `active_participants` also returns every `purges[].recipient_id` and nothing bounds the envelope *count*. The member-bound assertion here is deliberately scoped to a purge-free fixture rather than claiming an unconditional `<= 2 * cap`.
- **#490** (existing) — commented with a simpler reproduction of the members-channel resend loop: it fires on `origin/main` with **no ban, no messages, and zero DMs**, so it is the `MembersV1` non-monotonicity that issue already names. This PR does not cause it but widens exposure, since DM participants were previously exempt from inactivity-prune (that exemption *is* #519). Characterised here by `member_prune_reoffer_loop_mechanism_predates_the_dm_cap`, which asserts the loop is present so whoever closes #490 must flip its polarity.

Fixes #519

[AI-assisted - Claude]


sanity:
## Full-tier review — two rounds, consolidated

Contract/consensus-adjacent change that re-keys production rooms, so Full tier. **Eight independent blind reviewers** across two rounds: round 1 (skeptical, merge-convergence, migration-correctness, testing) on `146ce3df..c0c07428`; round 2 re-review (skeptical, convergence+test-integrity) on `ad9e9e3c` after the round-1 fixes. Codex not attempted — quota-blocked until ~Aug 1, per the standing rule that Claude lenses *are* the review.

### Round 1 — blocking findings, all fixed

**1. `outbound_summary` did not neutralise `global_horizon`** (found independently by two lenses). `room_synchronizer.rs`'s exhaustive destructure binds only `ChatRoomStateV1Summary`'s *own* fields, so a horizon added one level down inside `DirectMessagesSummary` compiled clean and was fed to `delta` on the **send** path — a device filtering its own outgoing DMs against its own baseline, `compute_update_data` returning `None`, and `process_rooms` advancing the baseline anyway, so the DM is never retried. Exactly the bug that function's 40-line comment exists to prevent, reintroduced on a new field. Blast radius exceeds the pair horizon it mirrors: `pair_horizons` needs 100 DMs in one pair; the global horizon fires at 500 room-wide, and the official room already held 499.

Fixed, guard extended to the leaf summaries, pinned by `outbound_update_is_not_filtered_by_the_senders_own_dm_global_horizon` (mutation-verified: reverting the fix fails it, siblings stay green).

**2. Two stale value pins would have turned CI red** — `migration.rs` (29→30, fingerprint `fc9e2622d9fa7d6f`) and `chat_delegate.rs` (`323a2f640bfd7a7f`). Both missed locally because they need feature flags / a different crate. Recomputed and verified independently, not copied from the review.

**3. `min()` → `max()` in `global_retention_horizon` survived every test.** That's the *dangerous* direction — an over-stated horizon silently withholds DMs the peer would have kept. Contiguous windows converge either way; `interleaved_windows_converge_so_the_horizon_cannot_be_over_stated` exposes it as silent divergence and now kills the mutation.

**4. Correctness: `dedup_by_signature`.** Both trims key off `DmOrderKey`; duplicates share one, and `verify` has no distinct-signature check, so a hand-built full-state PUT could put a repeated key on the global cutoff and make `retain(>= cutoff)` drop nothing — peer permanently over cap. Deduping makes the "keys are unique" premise true rather than assumed.

### Round 2 — blocking finding, fixed

**`duplicate_entries_do_not_defeat_the_global_cap` was vacuous.** Confirmed by mutation: it passed with `dedup_by_signature` deleted. The duplicates sat at the *oldest* position, so the cutoff rank landed above the run — structurally unreachable for the degenerate case. Rewritten with the run mid-range; without dedup the trim now keeps 40 instead of 25.

Also: `whole_state_gossip_...` tightened to `== CAP` (it could pass on an empty set); the residual member-set divergence is now pinned with an `assert_ne!` instead of being invisible; the known-gap test renamed to `member_prune_reoffer_loop_mechanism_predates_the_dm_cap` because the old name overclaimed; and the `outbound_summary` guard comment corrected — it claimed coverage of a future `MembersSummary.horizon` that it does not have, and that overstatement is precisely what let `global_horizon` through.

### Checked and deliberately not changed

- **`chat_delegate.wasm` is current.** Two reviewers flagged it as stale; a rebuild from HEAD is byte-identical, so the change is dead-code-eliminated from the delegate. No second generation, no extra migration entry.
- **The WASM build is deterministic** (clean rebuild reproduces the hash). The room-contract hash moves on *comment-only* edits because `expect`/`panic!` embed `file:line` location strings — worth knowing for anyone touching `common/src/`.
- **A pre-existing loop, surfaced not introduced.** Writing the whole-state test exposed a members-channel resend loop. I verified on `origin/main` **with zero DMs involved** that it reproduces there — it is the deferred `MembersV1` non-monotonicity that `outbound_summary` already names alongside `BansV1`. This PR does not cause it but does widen exposure, since DM participants were previously exempt from the prune (that exemption *is* #519). Characterised in the repo's established known-gap style; it self-heals in any room with traffic.
- **The DM-pinned member set is bounded only on the messages half.** `active_participants` also pins every `purges[].recipient_id`, and the *number* of purge envelopes is unbounded. Documented at the assertion rather than papered over.

### Coverage

25 tests in `dm_global_cap_test.rs`, 4 new proptests generalising the global axis plus an anti-vacuity self-check (every pre-existing DM property ran at the default cap of 500 against a ≤200 pool, so the global trim could never fire), and a new UI pin. Nine mutations applied, each verified to actually change the file and to be caught by the intended test.

### Status

Both migration checks green. `check-wasm-sync` is **red for a reason outside this PR's scope**: the room-contract WASM changed, so it requires a `cli/Cargo.toml` version bump (riverctl 0.2.5 is already on crates.io and embeds the WASM to derive the contract key). I was scoped out of `cli/Cargo.toml` and the riverctl release train, so that bump belongs to whoever owns the release. Not merging on red.

[AI-assisted - Claude]

sanity:
## Status on `06719d0f`: all checks green, but one green is FALSE — do not read this as mergeable

Default is now **300** (Ian's number). Rebased onto `9a7c948d`; both migration checks, `build`, and `ui-playwright-tests` pass.

### `check-wasm-sync` passed for the wrong reason

It fail-opens when the crates.io lookup errors, and that is what happened:

```
⚠️  WASM files changed — checking if riverctl version was bumped...
Local riverctl version: 0.2.5
⚠️  Could not check crates.io (riverctl not found or network error), skipping
```

Verified the real state independently:

| | |
|---|---|
| published `riverctl` (crates.io API) | **0.2.5** |
| local `cli/Cargo.toml` | **0.2.5** |
| branch `room_contract.wasm` | `4bb55ded51a6d9eb…` |
| `origin/main` `room_contract.wasm` | `f8cca7600a63dac1…` |

So the guarded condition is present: the room-contract WASM moved while riverctl's version still matches what is published. riverctl `include_bytes!`s the WASM to derive the contract key, so publishing the UI without bumping and republishing riverctl leaves every riverctl user on the abandoned key — the Feb 2026 incident this check exists for. An earlier run on `1ab8806e` failed this check correctly, when the crates.io lookup succeeded.

The bump belongs to whoever owns the riverctl release train; I was scoped out of `cli/Cargo.toml`. Not merging on a green I know to be false.

### Also still outstanding

River's `freenet-stdlib` 0.8.3 → 0.8.5 bump is being deliberately held so it rides with this contract change rather than causing a second independent re-key. It has not landed on `main` yet. Merging this first would defeat that — the bump would then re-key again on its own. This should land on main (or into this branch), then this PR rebases, rebuilds the WASMs, and re-verifies the migration entries against the new base hash.

### Note for anyone touching `common/src/`

The WASM build is deterministic, but the room-contract hash moves on **comment-only** edits, because `expect`/`panic!` embed `file:line` location strings. Confirmed both directions: a doc edit that shifted panic-bearing lines changed the hash; this commit's doc edits sit below them and the hash was unchanged. Worth knowing before assuming a docs-only commit is migration-neutral.

[AI-assisted - Claude]

sanity:
## Added: freenet-stdlib 0.8.3 → 0.8.5, deliberately riding with this re-key

Pushed as `2f2054a2`. The bump is here rather than on `main` on purpose: landing it separately would re-key every room a **second** time, which is exactly what holding it was meant to avoid.

0.8.5 (published to crates.io tonight) carries three FlatBuffers client-request boundary fixes — freenet-stdlib #85, #87 and #88. None of them is River-visible: River connects with `encodingProtocol=native` and never reaches those decode paths. The bump is taken for **WASM lockstep**, not for a functional fix.

### Both WASMs move

```
room_contract   f8cca760… (main)  →  dd63bcc9…
chat_delegate   2f8c5f1d… (main)  →  6f65e45c…
```

### Migration entries verified against the right baseline

The entries already on this branch record **main's** hashes — room `V30` = `f8cca760…`, delegate `V29` = `2f8c5f1d…` — and those remain the correct immediate predecessors, because users migrate from what is published on `main` today to what this branch publishes.

CI-mode check, which is the one that matches reality:

```
$ scripts/check-room-contract-migration.sh --ci origin/main HEAD
Room-contract WASM CHANGED: f8cca760… -> dd63bcc9…
Old hash found in common/legacy_room_contracts.toml — migration entry exists.

$ scripts/check-migration.sh --ci origin/main HEAD
Committed WASM CHANGED: 2f8c5f1d… -> 6f65e45c…
Old hash found in legacy_delegates.toml — migration entry exists.
```

**Note for anyone running the guards locally:** the default HEAD-vs-worktree invocation *fails* here, complaining that `4bb55ded…` is unregistered. That is the intermediate-churn false positive — `4bb55ded…` was this branch's WASM before the stdlib bump and was **never published**, so no user can be on it and it needs no entry. Use `--ci origin/main HEAD`.

### Still outstanding before merge

- CI green on `2f2054a2` (the previous head's green did not cover the bump).
- `check-wasm-sync` should stay green: riverctl is already at 0.2.6 against 0.2.5 published. Worth re-reading its **log line** rather than the tick — it fail-opens when the crates.io lookup errors, and it did exactly that on an earlier head, going green while the bump was genuinely missing.

[AI-assisted - Claude]

Issue #414, opened 2026-07-18
Import Identity should allow overwriting an existing room identity (warn, don't refuse)


sanity:
## Problem

Importing an identity into a room that already has one is **hard-refused** — UI `ImportIdentityModal` (`ui/src/components/members.rs:980`: "You already have an identity for this room.") and CLI (`cli/src/commands/identity.rs:212`). This leaves users **stuck on whichever identity they first used** in a room, with no way to switch to or restore a different key.

## Desired behavior

Importing over an existing identity should be **possible, behind a warning/confirmation** — not refused. The room's state (messages/members) lives on the network and re-syncs; the only thing at risk is the *current* identity's signing key if it wasn't exported, so the warning should call that out.

## Approach (localized — overwrite is already mechanically supported)

- The import's `rooms.map.insert(owner_key, room_data)` (`members.rs:1057`) already overwrites, and the rejoin/tombstone-clear logic (`removed_rooms.remove`, `mark_room_rejoined`) is already in place. Only the guard at `:980` blocks it.
- **UI:** replace the hard error with a confirm step — warn that the current identity will be replaced (and lost unless exported), offer Cancel / Replace, plus an "Export current first" shortcut.
- **CLI:** add `--force`/`--overwrite` to `identity import` that bypasses the refusal (printing the warning); without it, keep refusing but mention the flag.

No auto-export / no forced export — warn only.

[AI-assisted - Claude]

PR #528, opened 2026-07-28
fix(ui): resolve delegate identity conflicts by generation, not arrival order (#527)

### LINKED ISSUE #527: Lost IDs and rooms to River Update across two nodes.
Tried to use River on phone. It said fetching rooms. Appeared to get stuck, so I reloaded the page. It found 1 room out of about 8. Kept trying to reload and got nothing no matter how long I waited. No sign of errors.

Later tried to use River from another node on another device. This time after updating my rooms appeared to rollback to previous IDs in several rooms. No sign of errors on console again.

Didn't appear to have a problem with a 3rd node.

Only factor I can think of is my IDs are old and have been through many contract/delegate changes so maybe that did something.

sanity:
## Problem

The 2026-07-27 re-key publish (V29 → V30, for the DM retention cap in #524) forced every user through legacy delegate migration. A long-time user reported rooms coming back under identities from months earlier, on two of three nodes (#527).

`fire_legacy_migration_request` probes all **26** legacy delegate generations at once and each one holding data hydrates independently, so their responses race. Three defects turned that race into a rollback:

1. **Identity was decided by arrival order.** `Rooms::merge` kept whichever copy landed FIRST and silently skipped any later copy with a different `self_sk`. Probes are dispatched oldest-generation-first, so the oldest surviving copy usually answered first and won. Only users whose identity for a room ever changed are affected — which is why this hit a long-time user and not everyone.

2. **The signing key pushed to the delegate ignored the merge outcome.** The per-room keys were read from `loaded_rooms` BEFORE the (deferred) merge, so a room whose incoming copy the merge REJECTED still had that rejected key pushed — and `migrate_signing_key` overwrites on mismatch. The delegate was left signing with an identity the UI had already discarded, invalidating every signature it produced for that room.

3. **The migration could be sealed before the fan-out finished.** The seal ends legacy migration permanently: once set, a later session whose current delegate is still empty finds it and probes nothing. The loudest door was in `freenet_synchronizer.rs`, which sealed on any "delegate not found" API error — an error the fan-out itself provokes within milliseconds on any node missing an old delegate WASM. One absent generation spoke for the other twenty-five.

## Approach

Give the fan-out the ordering it was missing. `LEGACY_DELEGATES` is generated in age order, so an entry's index is its generation: higher is newer, and the current delegate outranks all of them.

- `resolve_identity_conflict` decides by source authority. A strictly newer generation wins; equal ranks keep local, preserving the #414 stale-load guard.
- `merge_from_source` threads the rank through and folds the local state into the incoming copy before adopting it, so no messages are lost. The map is not mutated until the fold succeeds — `ChatRoomStateV1::merge` is fallible, and mutating first meant an error deleted the room outright.
- A deliberate identity choice (import, invitation-accept) is recorded authoritative synchronously at both sites, so no load can undo it.
- Signing-key migration is deferred to run after the merge and reads the merged `ROOMS`.
- Seal doors: the two where the current delegate is authoritative stay synchronous (no fan-out is in flight, and deferring weakens the #253 guard); the two where a fan-out can be in flight are quiescence-gated, as is the delegate-not-found door. One writer, pinned.

## Scope — please read before closing #527

- **This protects nodes that have not yet hit the bug. It is NOT recovery for those that have.** An already-rolled-back user's current delegate is populated, so the per-room load path marks migration done and legacy is never re-probed. Shipping this changes nothing for them; they must re-import their identity, after which the fix keeps it.
- **A wrong ranking decision is now deterministic and permanent rather than a coin flip.** For an already-affected user, the wrong identity sits in what becomes the highest-ranked legacy generation, so it would win every conflict thereafter. That is the right trade — deterministic-correct for the overwhelming majority beats a coin flip for everyone — but it should be stated rather than discovered.

## Testing

Behavioural tests at the merge layer plus source pins for the wiring. Every pin was mutation-verified rather than trusted green; two tests were found vacuous that way and rebuilt — one of them passed with the bug fully restored, because its fixture made the fold succeed.

## Review

Four independent blind lenses (code-first, skeptical/race, testing, data-loss/migration-compat), two rounds. Three reviewers independently found the same P1 (the map mutated before a fallible merge). All findings fixed or explicitly justified; final verdict non-blocking from all four.

Deliberately not fixed here, filed instead: #531 (keep-local path discards the loser's `room_state` — pre-existing, and closing it is a behaviour change deserving its own review) and #532 (interrupted-migration recovery has no seal bypass — pre-existing).

Closes #527

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -951,48 +951,88 @@ mod tests {
     /// exactly ONE writer. This pin fails any future fifth door.
     #[test]
     fn the_seal_has_exactly_one_writer() {
-        // SCOPE NOTE: this file's `mod tests` is MID-FILE, so there is no clean
-        // "production only" slice to count over — the seal machinery lives
-        // AFTER the test module. So this pins the two things it can verify
-        // exactly, which is where all four original doors actually were:
-        //   (1) no OTHER module seals inline (three of the four doors), and
-        //   (2) the region of THIS file before the test module has no door.
-        // The remaining possibility — a new inline seal added after the test
-        // module in this same file — is covered by the companion test
-        // `the_legacy_seal_is_written_only_behind_the_quiescence_debounce`,
-        // which pins that the write sits behind the gate in schedule_legacy_seal.
+        // This file's `mod tests` is MID-FILE, so a plain `split(...).next()`
+        // would cover only the ~700 lines ahead of it and miss the region where
+        // the seal machinery actually lives. Excise the test module instead —
+        // inside it every item's closing brace is indented, so the first
+        // column-0 `}` after the module header is the module's own.
         let src = include_str!("chat_delegate.rs");
-        let before_tests = src
-            .split("#[cfg(test)]\nmod tests {")
-            .next()
-            .expect("code before the test module");
+        let tests_start = src
+            .find("#[cfg(test)]\nmod tests {")
+            .expect("test module must exist");
+        let tests_end = tests_start
+            + src[tests_start..]
+                .find("\n}\n")
+                .expect("test module must terminate")
+            + 3;
+        let production = format!("{}{}", &src[..tests_start], &src[tests_end..]);
+
+        // The CALL form carries a semicolon; the definition ends in `{` and the
+        // doc-comment mentions have neither, so this counts real writers only.
         assert_eq!(
-            before_tests.matches("mark_legacy_migration_done()").count(),
-            0,
-            "no inline seal may be added ahead of the test module — every seal \
-             must route through request_legacy_seal_on_quiescence (freenet/river#527)"
+            production.matches("mark_legacy_migration_done();").count(),
+            1,
+            "exactly ONE writer of the seal — the quiescence-gated write in \
+             schedule_legacy_seal. A second is a new door onto the third cause \
+             of freenet/river#527."
+        );
+        // Guard the excision itself: if the module bounds are ever computed
+        // wrongly, the count above could pass by looking at the wrong text.
+        assert!(
+            production.contains("fn schedule_legacy_seal(")
+                && !production.contains("fn the_seal_has_exactly_one_writer("),
+            "the excision must keep production and drop the test module"
         );
 
-        // No OTHER module may call it directly.
-        for (name, other) in [
-            (
-                "response_handler.rs",
-                include_str!("freenet_api/response_handler.rs"),
-            ),
-            (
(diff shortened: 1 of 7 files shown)


sanity:
## Review summary

Full-tier review: four independent blind lenses, two full rounds plus a targeted delta check. Risk tier is Full because this touches delegate migration and per-room identity — a data-loss surface.

**Lenses run:** code-first (read the code before the description, flag intent/impl mismatch) · skeptical (adversarial race hunt, prior-fix regressions) · testing (vacuity, pin construction, coverage gaps) · data-loss / migration-compat (can this lose a room, a message, or an identity).

### Blocking findings, all fixed

**P1 — the adopt path mutated the map before a fallible merge.** Found independently by three of the four lenses. `map.remove()` ran before `room_state.merge(...)?`, so any merge error returned with the room already gone: it vanished from the UI, the rest of that merge pass was abandoned, and if that pass was the one that re-saved to the current delegate, the room was never written there and legacy is never probed again — permanently stranded. `ChatRoomStateV1::merge` genuinely errs on this path. Fixed: the map is untouched until the fold succeeds.

**P1 — nothing pinned the production wiring.** Swapping the shared rank registry for a fresh map, or the threaded rank for a constant, reverted the whole fix with every behavioural test still green, because they all supply their own ranks map. Now pinned.

**P1 — the delegate-rank ordering test was tautological.** It asserted `rank == index`, which is an identity over `position()` and holds for any ordering — while its own doc claimed it guarded against reordering. Now asserts the emitted dates are non-decreasing.

### Notable non-blocking outcomes

- A seal door in `freenet_synchronizer.rs` — outside the original diff — sealed legacy migration on any "delegate not found" error, which the 26-generation fan-out itself provokes within milliseconds. One absent generation spoke for all the others. Now gated.
- Two seal doors were initially over-gated; two lenses showed that deferring them bought nothing and weakened the #253 guard. Reverted to synchronous with recorded rationale.
- One reviewer retracted a finding with a precise reason (the scaffold macro re-clones state before each field, so an earlier field's delta *is* visible to later ones) — worth reading if you touch merge ordering.

### Test discipline

Every pin was mutation-verified rather than trusted green. Two tests were found vacuous that way and rebuilt — one passed with the bug fully restored, because its fixture made the fold succeed rather than fail.

### Deliberately not fixed

#531 (keep-local discards the loser's `room_state` — pre-existing, closing it is a behaviour change deserving its own review) and #532 (interrupted-migration recovery has no seal bypass — pre-existing, and annotated as having no traced producer).

[AI-assisted - Claude]

Issue #461, opened 2026-07-24
feat(dm): no way to leave/block a DM — the UI ✕ is a local archive that un-hides on the next message


sanity:
Reported by ofansifkapital-xmpp in the official room (2026-07-24):

> In riverctl I can explicitly leave a room and stop the message stream from coming in. Do we have something like this for DMs? [...] let's say I start (or accept) a DM with Mallory. And Mallory turns out to be malicious and starts spamming me in the DM. [...] I think in River WebUI, there's a small "X" cross button that one can use to close a DM. Does this shut down the DM channel, or is it purely visual?

They are right on both counts. There is no per-DM leave, no block, and no ignore, and the UI "✕" is purely local and self-reverting.

## Why there is no "leave a DM"

River DMs are **in-room**, not a separate channel (`common/src/room_state/direct_messages.rs`, #230 Phase 1 — this replaced the reverted per-pair inbox contract from #234/#238). A DM is an `AuthorizedDirectMessage` carried inside `ChatRoomStateV1` for the room it was sent in, ECIES-encrypted to the recipient's member key.

You receive DMs because you are subscribed to the **room** contract, not because you subscribed to a DM thread. There is no per-DM subscription to cancel, which is why no `dm leave` exists.

## What the UI "✕" actually does

It is the per-row **Archive** button in `ui/src/components/room_list/dm_rail_section.rs` (moved there by #266). It calls `hide_dm_thread`, which writes a `(room, peer, hidden_at_ts)` entry to the local chat delegate. Nothing is sent to the network.

Worse, it self-reverts. `is_thread_hidden_for` (`ui/src/components/direct_messages.rs:168`) hides the thread only while `max_message_ts <= hidden_at_ts`, so Mallory's next message pops the thread straight back into the rail. It is a "tidy up an idle thread" affordance, not a close.

So: purely visual, and it does not survive the next spam message.

## What a victim can actually do today

| Option | Effect | Limit |
|---|---|---|
| Archive (UI ✕) | Local hide | Un-hides on the next inbound DM |
| "Delete their messages" / `riverctl dm purge` | Recipient-signed tombstone; drops those messages network-wide and blocks their replay | Does not stop new messages — a fresh send has a fresh signature, so a fresh `PurgeToken` |
| Ban | Removes the member and sweeps their DMs | Ban authority is owner / invite-chain ancestor / deputy (#410). A regular member cannot ban an arbitrary co-member |
| Leave the room | The only real stop | Costs you the entire room |

[...]

PR #529, opened 2026-07-28
fix(dm): stop ordinary sync churn from destroying DM archives

### LINKED ISSUE #526: fix(dm): archiving a DM is undone by ordinary sync churn — re-landed old DMs hard-delete the archive entry
## Summary

Archiving a DM writes a `(room, peer, hidden_at_ts)` entry and the rail hides the thread while `max_message_ts <= hidden_at_ts`. A genuinely new inbound DM is *supposed* to revive it (#461, and the ✕ tooltip says so). But the revival is implemented as a **hard delete** of the archive entry, gated on a test that cannot tell a genuinely-new DM from an **old DM that merely left local state and came back**. So archives are destroyed by ordinary sync churn, with no new message involved — and the destruction is persisted, so it survives reload.

## Mechanism

`unhide_dm_thread` (`ui/src/components/app/chat_delegate.rs:4233`) removes the entry from `HIDDEN_DM_THREADS`, inserts the pair into the session-scoped `RECENTLY_UNHIDDEN` tombstone set, and **queues a delegate save**. The archive is gone from the persisted blob, and a later hydration carrying it is suppressed by the tombstone. Nothing about this is transient.

It fires from two places, both gated on `newly_landed_inbound_senders`:

- `apply_delta_inner` — `room_synchronizer.rs:503`
- `update_room_state` (full-state path) — `room_synchronizer.rs:1457`

`newly_landed_inbound_senders` is computed by diffing post-merge DM signatures against a **pre-merge snapshot of local state** (`:348` / `:1182`, checked at `:431` / `:1345`). So the predicate is *"this DM was not in my local state a moment ago"*, **not** *"this DM is newer than the archive cutoff"*. There is no comparison against `hidden_at_ts` anywhere on this path.

The comment at `:286-294` anticipates re-deliveries, but only covers the case where the message is **still in local state** and the contract's dedupe drops it. It does not cover the case where the message **left local state and returned**.

### Trigger 1 — the membership sweep (everyday case)

`post_apply_cleanup` calls `direct_messages::sweep_after_membership_change` (`common/src/room_state.rs:420-426` -> `direct_messages.rs:744-756`) on **every merge**, retaining only DMs whose sender *and* recipient are in the current `members` list. Member-list flicker therefore deletes DMs from local state; peers re-offer them; they re-land.

Each re-landing counts as "newly landed" -> `unhide_dm_thread` -> archive destroyed, for a message the user archived days ago. This is the same delete/restore cycle #499's closing note attributes to #490.

### Trigger 2 — empty pre-merge snapshot (bulk case)

On the full-state path, if local room state holds no DMs at merge time, `pre_merge_dm_sigs` is empty and **every** DM in the incoming state is "newly landed" -> every archive in that room is destroyed at once. Two ways to get there:

- the room-not-found branch returns an empty set outright (`:1189-1200`);
- a full-state GET landing before the delegate's room blob has hydrated.

Since `ROOMS_STORAGE_KEY` and `OUTBOUND_DMS_STORAGE_KEY` are requested together (`chat_delegate.rs:5754`) and the network GET is unordered relative to both, this is a startup race, not an exotic one.

## Why it looks like "archiving does nothing"

The user archives, the thread comes back with no new message, they archive again, it comes back. Trigger 2 makes whole batches of archives vanish at once, which reads as the feature being purely cosmetic.

## Discriminator vs #461

- Thread returns **only when the peer actually messages you** -> #461, working as designed.
- Thread returns **with no new message**, or **many threads un-archive together**, or **archives are gone after a reload** -> this bug.

## Suggested fix

Gate the unhide on the archive cutoff rather than on local-state membership: only fire `unhide_dm_thread` when the landed DM's `timestamp` is actually past `hidden_at_ts` for that pair. A re-landed old DM is then inert by construction, and trigger 2 disappears with it.

That alone re-opens the same-unix-second hole #267 added this call site to close, so pair it with making `archive_row` (`dm_rail_section.rs:274`) set `cutoff = max(last_any_ts, unix_now_secs())` instead of `last_any_ts`. The cutoff then covers everything up to the click, strict `>` is the correct comparison, and the same-second race is gone.

Worth pinning with a regression test: archive a thread, remove its DMs from local state, re-merge the identical DMs, assert the archive entry survives.

## Related

- #461 — no real leave/block for DMs; archive is local and self-reverting by design. This issue is the *unintended* half of that behaviour.
- #499 — archived threads flashing into the rail under signal contention (fixed). Different mechanism: that one was render-transient, this one destroys persisted state.
- #490 — the sweep/re-offer churn that supplies trigger 1.

[AI-assisted - Claude]

sanity:
## Problem

Archiving a DM writes a `(room, peer, hidden_at_ts)` entry and the rail hides the thread while the peer's newest message is at or below it. A new inbound DM is meant to revive it (#267), but the revival was implemented as a **hard delete** of the archive entry, fired whenever a DM crossed the merge dedupe.

"Crossed the dedupe" means *"was not in LOCAL state a moment ago"*, which is a different question from *"is new content"* — and the decision never consulted `hidden_at_ts` at all. Two ways it misfired:

- **Ordinary sync churn.** `post_apply_cleanup` sweeps DMs whose counterparty is momentarily absent from `members` on every merge. Peers re-offer them, they re-land with their original timestamps, and the diff can't tell them from a new message.
- **Empty pre-merge snapshot.** On the full-state path, when the room is in `ROOMS` but its DM state hasn't hydrated, every DM in the incoming state reads as newly landed — so **every archive in the room** was destroyed at once.

`unhide_dm_thread` also writes a session `RECENTLY_UNHIDDEN` tombstone and queues a delegate save, so the loss was **persisted** and a later hydration carrying the archive was suppressed. Archiving looked like it did nothing.

## Approach

`ui/`-only — no `common/`, delegate, or contract WASM change, so **no migration**.

`unhide_dm_thread_if_dm_is_newer` replaces the unconditional call at both inbound-sync sites, dropping the entry only when the DM's timestamp is strictly past the cutoff. A pair with **no** entry is inert — deliberately including "hasn't hydrated yet", so no unhide *and no tombstone*, which defuses the startup race. `fold_newly_landed_to_max_ts` reduces a merge batch to one entry per sender keeping the largest timestamp.

### The archive clock is inbound-only

A gate is only as good as its cutoff. `last_any_ts` covers our **own outbound** DMs too, and an outbound DM's timestamp is our *local wall clock*. So a thread where we replied last was archived against **our** clock while the gate compares a **sender's** timestamp — and a peer DM already in flight across the click landed at or below the cutoff and went invisible on every surface: no rail row, no unread badge (`count_unread_dms_with` applies the same filter), no notification (DMs raise none).

The clock is now `last_inbound_ts` everywhere the archive decides visibility: `filter_rail_entries`, both archived-view projections, and the unread tally. Outbound revival comes solely from the unconditional `unhide_dm_thread` in `do_send` / `send_structured_dm` — **on the sending device only**; hydration never propagates a removal and `riverctl dm send` doesn't call it, which the docs now state rather than gloss.

The cutoff is also recomputed from live `ROOMS` state, maxed with the rendered row so a contended read falls back. The row can lag, and while the membership sweep is atomic per pair, the **re-land** is not — a peer restoring only a subset would otherwise leave the cutoff low.

### Why the cutoff is NOT clamped

An earlier revision clamped it at `now + MAX_DM_FUTURE_SKEW_SECS` to blunt a forged-future timestamp. Two reviewers independently found that this **broke the load-bearing invariant** *every existing inbound DM satisfies `ts <= cutoff`*: the rail filter compares the **unclamped** observed clock, so Archive became a visible no-op that still showed a success toast, **and** the next sweep-and-re-offer saw `ts > cutoff` and destroyed the persisted archive — this bug again, narrowed to skew-violating peers. Clamping one side of a comparison was what created the hole.

The forged-timestamp concern is handled where it cannot break the invariant: `should_unhide_for_inbound_dm` clamps the **incoming** timestamp, so an implausible-future DM can never trigger an unhide.

### Accepted residuals

All defer messages from one peer the user already chose to archive; none destroys state.

- A new inbound DM in the same unix second as the peer's previous one.
- A peer whose clock moves backward relative to their own previous message.
- A peer who stamps far in the future: their later genuine DMs wait until real time passes it.
- One never-rendered DM that the live recompute may cover.

The real remedy for a hostile peer is a block/ignore list — #461.

## Testing

`cargo test -p river-ui --bins` — 768 pass; 772 with `--features example-data,no-sync`. `cargo fmt --all -- --check` clean, wasm target clean. Full Playwright suite green (605 passed, 15 skipped) across Chromium, Firefox, WebKit, mobile Chrome and mobile Safari.

Note the example-data build contains **no DMs** (populating them needs a real ECIES envelope per DM), so Playwright cannot drive the archive flow end-to-end; it covers the surfaces this PR touches against regression, and the behaviour is pinned in Rust.

**Mutation-checked**, each reverted independently against the committed fix — all caught:

| Mutation | Result |
|---|---|
| gate always `true` (pre-fix behaviour) | red |
| gate deleted from the wrapper body | red |
| sync site calls bare imported `unhide_dm_thread` | red |
| filter compares `last_any_ts` (the outbound-anchor bug) | red |
| accumulator hoists the inbound bump out of `is_self_recipient` | red |
| row passes `last_any_ts` to `archive_row` | red |
| archived-count projection reverts to `last_any_ts` | red |
| no live recompute (trust the rendered prop) | red |
| wall-clock clamp reintroduced on the cutoff | red |
| `do_send` routed through the gated unhide | red |
| Undo toast routed through the gated unhide | red |
| determinism sort dropped | red |

Two mutations initially reported *green* and both turned out to be measurement bugs worth recording: one didn't compile (so no tests ran), and one was inert because the pin's needle matched its own explanatory comment rather than the code. Both are fixed; the clamp is now pinned structurally, because `archive_cutoff` takes no injectable `now` and fixture timestamps sit far below real wall-clock time, which makes any re-introduced clamp invisible to a behavioural test.

### Repaired two vacuous pins

The existing #267 pins searched `include_str!` for literals their own assertions contained, so all three assertions in each were tautologies. They now cut at `mod tests`, match whitespace-stripped, assemble needles at runtime, and the delta-path pin is sliced to its own function.

## Review

Three full rounds, independent blind lenses each time. Every round found a real defect in my own work; all are addressed:

- **Round 1** — killed a `max(last_any_ts, now)` cutoff as unnecessary and harmful.
- **Round 2** — found that removing it wasn't enough, because `last_any_ts` includes outbound. That produced the inbound-only rework above.
- **Round 3** — found the one-sided clamp (HIGH, both reviewers independently), the unpinned inbound producer and outbound-send sites (HIGH), a dead segment bound, a stale test asserting removed behaviour, a probabilistic determinism test, and several inaccurate comments including a rules-doc citation of a renamed test. All fixed.

Also filed **#530** — a pre-existing bug two reviewers surfaced: a save that runs before the delegate blob hydrates persists a truncated store, permanently losing archives and outbound-DM plaintexts. It reproduces two of this issue's symptoms by a different mechanism, so #526 alone does not fully close the user complaint.

Closes #526

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/direct-messages.md
+++ b/.claude/rules/direct-messages.md
@@ -162,11 +162,104 @@ field with `#[serde(default)]` so old bytes still decode. **Do not add
 a second top-level delegate storage key for hide state**: a new key
 needs its own probe in `fire_legacy_migration_request` and routing in
 `response_handler.rs`, AND splits the multi-device save path into two
-writes that can race. Filter helper `chat_delegate::is_thread_hidden`
+writes that can race. Filter helper `direct_messages::is_thread_hidden_for`
 uses strict `<=`; rail-side pure helper
 `dm_rail_section::filter_rail_entries` is pinned by
 `filter_rail_entries_*` tests; the "click Archive again after revival
 must re-hide" branch is pinned by `hide_unhide_rehide_round_trip`.
+
+### Two coupled invariants — do not change one without the other (#526)
+
+**1. `hidden_at_ts` is the newest INBOUND message timestamp**, computed as
+`max(rendered row value, live recompute from ROOMS)` and NOT clamped against
+wall-clock time.
+
+*Inbound-only.* `last_any_ts` also covers our OWN outbound DMs, whose
+timestamp is our local wall clock — so a thread where we replied last would be
+archived against OUR clock while the gate below compares a SENDER's timestamp,
+and a peer DM already in flight across the click would land at or below the
+cutoff and vanish from the rail, the unread badge, and everything else (DMs
+raise no notification). Our own sends revive the thread through the
+unconditional `unhide_dm_thread` in `do_send` / `send_structured_dm` — on the
+SENDING DEVICE only: hydration never propagates a removal, and `riverctl dm
+send` does not call it, so archiving on a laptop and replying from a phone
+leaves the laptop archived until the peer writes.
+
+*Live recompute.* The rendered row can lag (`LAST_GOOD_RAIL`, or a DM landing
+between the memo's last clean pass and the click), and while the membership
+sweep is atomic per pair, the RE-LAND is not — a peer may restore only a
+subset. A cutoff below the pair's true inbound max lets the rest re-land
+looking newer than the archive. Cost: the live read can cover an inbound DM
+that was never rendered, so archiving may swallow one message until the peer
+writes again.
+
+*Not clamped.* An earlier revision clamped the cutoff at
+`now + MAX_DM_FUTURE_SKEW_SECS`. That broke the load-bearing invariant **every
+existing inbound DM satisfies `ts <= cutoff`**: the rail filter compares the
+UNCLAMPED observed clock, so Archive became a visible no-op that still showed a
+success toast, AND the next sweep-and-re-offer saw `ts > cutoff` and destroyed
+the persisted archive — #526 itself, narrowed to skew-violating peers. Clamping
+one side of a comparison is what created the hole. The forged-timestamp concern
+is handled where it cannot break the invariant instead: the GATE bounds the
+incoming timestamp. That bound is narrow — it binds only once a forgery has
+already inflated the cutoff, stopping a peer escalating to a larger one; a
+first forgery can still revive a normally-archived thread. It is
+safe-directional regardless (the bounded value is never above the raw one, so
+it can only make the gate more conservative).
+
+The inbound-only anchoring works because `sweep_after_membership_change`
+retains on `alive(sender) && alive(recipient)` and every DM in a pair shares
+the same sender/recipient set, so the sweep is **all-or-nothing per pair** — it
+can never leave a thread partially swept.
+
+The sweep is atomic; the RE-LAND is not. If a peer re-offers only part of a
+swept pair and the rail memo refreshes cleanly in between, both the row and the
+live read agree on a value below the pair's true max, and a later re-offer of
+the rest fires the gate and destroys that archive. `max(row, live)` covers a
+row stale from BEFORE the sweep, not a cutoff taken inside the partial window.
+Narrow, but a residual — do not restate this invariant as absolute.
+
+Pinned by `archive_cutoff_*`, `archive_row_uses_inbound_and_live_cutoff`,
+`rail_row_passes_the_inbound_clock_to_archive_row`,
+`rail_filter_compares_inbound_timestamp`, and
+`archived_projections_use_the_inbound_clock`.
+
+**2. Inbound-sync paths must use the cutoff-gated unhide; explicit user
+actions must not.** The two `room_synchronizer` sites (`apply_delta_inner`,
+`update_room_state_inner`) call `unhide_dm_thread_if_dm_is_newer`, which drops
+the archive entry only when the DM's timestamp is strictly past
+`hidden_at_ts`. Their "newly landed" predicate is a diff against **local
+state**, which cannot distinguish a genuinely new DM from an old one the sweep
+removed and a peer re-offered — nor, on the full-state path, from the case
+where the pre-merge snapshot is empty because DM state has not hydrated yet
+(there, *every* DM reads as newly landed). Calling the unconditional
+`unhide_dm_thread` from either site destroys the archive on ordinary sync
+churn and **persists** the destruction (it also writes a session
+`RECENTLY_UNHIDDEN` tombstone, which then suppresses the entry when the
+delegate blob hydrates).
+
+The four explicit-user-action call sites — Undo toast and Archived-panel
+Un-archive (`dm_rail_section.rs`), and the outbound sends in `do_send`
+(`dm_thread_modal.rs`) and `send_structured_dm` (`direct_messages.rs`) — keep
+the UNCONDITIONAL form. Routing any through the gated one makes it a silent
+no-op, because at that moment the inbound clock is by construction at or below
+the cutoff. Each file pins its own sites:
+`explicit_user_actions_keep_the_unconditional_unhide` (rail) and
+`outbound_send_keeps_the_unconditional_unhide` (both send files). Those pins
+were briefly lost to a `git checkout` during mutation testing and restored —
+if you touch them, confirm they still exist rather than trusting a green run.
+
+Accepted residuals — all defer messages from one peer the user already chose
+to archive, none destroys state: a new inbound DM in the same unix second as
+the peer's previous one; a peer whose clock moves backward relative to their
+own previous message; and a peer who stamps far in the future, whose later
+genuine DMs wait until real time passes that stamp. Irreducible under
+timestamp gating — the real remedy is a block/ignore list, #461.
+
+Two narrower cases can still cost an archive ENTRY (not just defer a message):
+a cutoff taken inside a partial re-land window, and a DM the live read missed
+because `ROOMS` was contended AND the rendered row lagged. Both need an archive
+clicked inside a specific race; neither is the systematic churn #526 is about.
 
 The per-row rollover ✕ in `DmRailSection` is the archive control (the
 old modal-header "Hide" button next to close ✕ was repeatedly
--- a/common/src/chat_delegate.rs
+++ b/common/src/chat_delegate.rs
@@ -346,8 +346,9 @@ pub enum CasStoreResult {
 /// thread AND no message in the thread has `timestamp > hidden_at_ts`.
 /// The strict `>` (not `>=`) on `max_message_ts` ensures that the
 /// message used to populate `hidden_at_ts` does not itself revive the
-/// thread. Any newer DM (inbound or outbound) crosses the threshold
-/// and revives.
+/// thread. Any newer INBOUND DM crosses the threshold and revives;
+/// outbound sends revive via the explicit `unhide_dm_thread` instead,
+/// since freenet/river#526 made the archive clock inbound-only.
 ///
 /// `hidden_threads` is the full slice as loaded from the delegate;
 /// the lookup is linear because the list is tiny (bounded by the
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -2518,6 +2518,175 @@ mod tests {
         assert_eq!(to_insert[0].1.hidden_at_ts, 5_000);
     }
 
+    // ===== should_unhide_for_inbound_dm (issue freenet/river#526) =====
+    //
+    // The inbound-sync paths in `room_synchronizer` used to call
+    // `unhide_dm_thread` unconditionally for every DM that crossed the
+    // merge dedupe. "Crossed the dedupe" means "was not in LOCAL state a
+    // moment ago", which is a different question from "is new content":
+    // `post_apply_cleanup`'s membership sweep removes DMs whose
+    // counterparty is momentarily absent from `members`, peers re-offer
+    // them, and they re-land. Every such re-landing destroyed the user's
+    // archive — and `unhide_dm_thread` persists the removal, so the loss
+    // survived a reload. These tests pin the replacement decision.
+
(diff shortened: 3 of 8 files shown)


sanity:
## Blocking finding from re-review — do not merge as-is

Two independent reviewers converged on the same live defect in the current commit (`ba08ddae`), and I verified the mechanism directly.

### The defect

`archive_row` anchors `hidden_at_ts` at `last_any_ts`, which `accumulate_peer_activity` computes over **both** directions — the `if timestamp > acc.last_any_ts` bump at `dm_rail_section.rs:564-566` sits outside the `is_self_recipient` branch. An outbound DM's timestamp is `unix_now()`, the local user's own wall clock (`direct_messages.rs:462`, `:556-562`).

So whenever the user replies and *then* archives, the cutoff is receiver-clock `now` — precisely the condition this revision removed the `max(last_any_ts, now)` anchor to avoid. The strict `>` gate then swallows any inbound DM stamped at or below it.

**Scenario (no clock skew needed):**
1. `t=1010` Bob sends a DM, `ts=1010`, still propagating.
2. `t=1012` Alice replies; her outbound `ts=1012` is applied to local state immediately (`dm_thread_modal.rs:539-560`).
3. `t=1013` Alice archives. `last_any_ts = 1012` → `hidden_at_ts = 1012`.
4. `t=1020` Bob's DM lands. Gate: `1010 > 1012` → false. No unhide.
5. Rail: thread max is still 1012 → `1012 <= 1012` → hidden.

Bob's message has no surface: no rail row, no unread badge (`count_unread_dms_with` applies the same filter), no notification (DMs never raise one), and the Archived panel shows no unread indicator. It stays invisible until Bob sends something stamped above 1012 — indefinitely if he doesn't.

On `main` step 4 fires the unconditional unhide and the thread reappears, so this is a **regression**, not a residual. The PR's "one second, self-healing" claim is wrong: the window is (peer clock lag + propagation delay), and backward clock skew is unbounded — `check_dm_future_skew` bounds only the future direction, only at compose time, only against the sender's own clock.

### What the fix should be

Make archive **inbound-anchored** — "hidden until *they* write again":

- `PeerDmActivity` / `DmRailEntry` gain `last_inbound_ts` (max ts where `recipient == self`).
- `archive_row` anchors the cutoff on that.
- `filter_rail_entries`, `count_currently_archived`, `build_archived_rows`, and `document_title::count_unread_dms_with` compare `last_inbound_ts` against `hidden_at_ts`.

The user's own outbound already revives the thread through the unconditional `unhide_dm_thread` in `do_send`, so nothing is lost by taking outbound out of the filter — and the semantics get *more* coherent, not less. Re-landed old inbound DMs stay inert (their ts is at or below the inbound max at archive time), which is what keeps #526 closed.

Residual after that fix: a peer whose clock moves backward relative to their own previous message. Irreducible under timestamp gating, and far narrower than the current hole.

### Other findings to fold in

- `gated_unhide_wrapper_consults_the_cutoff` pins the gate condition and the cutoff read, but not that `unhide_dm_thread(room_owner_vk, peer)` is still *inside* the `if` — deleting that line kills #267 revival entirely with a green suite.
- Same test: assertion 2 pins that the cutoff is read, not that it is *passed*; `should_unhide_for_inbound_dm(None, ts)` satisfies both and makes the wrapper a permanent no-op.
- `dm_rail_section.rs:1120` — `rest.find("fn ")` can never match, because the haystack is whitespace-stripped. The segment bound is dead code and `seg` spans the whole production half.
- `chat_delegate.rs:2236` — `rest.find("pubfn")` misses `pub async fn`, so the segment is ~110 lines rather than the ~19 the comment claims.
- `chat_delegate.rs:2152-2156` — assert message still cites `max(newest, now)` from the reverted revision, contradicting the sibling pin that now forbids it.
- Nothing pins that the explicit-user-action sites (Undo toast, Un-archive, outbound send) keep the **unconditional** unhide; routing Undo through the gated form would make it a silent no-op.
- `apply_delta_inner_revives_...` doesn't slice at the function head, so it doesn't actually scope to `apply_delta_inner` (mitigated by the `count == 2` assertion).

### Verified clean by the re-review

The all-or-nothing-per-pair sweep claim holds. `trim_pairs_to_cap` / `trim_to_global_cap` drop oldest-first so the surviving max is untouched; purge tombstones can lower the max but purged messages are gated at ingestion and re-dropped after merge, so they cannot re-land. Recipient filtering at both collection sites is correct. `fold_newly_landed_to_max_ts` is correct and deterministic. The signal-safety of `try_read` + explicit `drop` + deferred `with_mut` is correct, and the `drop` is load-bearing on non-wasm where `defer` runs synchronously. The two repaired #267 pins are genuinely non-vacuous.

Moving to draft until the inbound anchoring lands.

[AI-assisted - Claude]

sanity:
## Addendum — two further residuals from the third lens

Both are in the same family (timestamp gating leaking), and both need to be folded into the inbound-anchored rework rather than fixed separately.

### The sweep is all-or-nothing, but the RE-LAND is not

The safety argument I wrote says "every DM in the thread satisfies `ts <= cutoff`, by construction". That holds in the steady state only. It needs a second premise the comments never state: **local state must hold the pair's complete set at click time.**

The sweep removes a pair atomically, but restoration is delta-driven and unordered. A peer can re-offer a subset:

1. Thread holds ts=800 and ts=1000. Membership flicker sweeps both.
2. A peer re-offers only ts=800 (the only one it holds, or delta batching). The rail row reappears with `last_any_ts = 800`.
3. User archives → cutoff = 800.
4. ts=1000 re-lands from another peer → `1000 > 800` → gate fires → archive destroyed and persisted.

That is #526 unchanged, reached through the fixed code. Narrower window, same outcome. A second, smaller instance of the same shape: `archive_row` is fed the rendered prop (`dm_rail_section.rs:184` → `:205`), so a DM that has landed in `ROOMS` but not yet re-rendered leaves the cutoff stale.

Proper close: have `hide_dm_thread` recompute the cutoff from live room state rather than trusting the rendered prop. That also subsumes the `LAST_GOOD_RAIL` staleness path.

### A future-dated DM makes a thread permanently invisible

`check_dm_future_skew` runs only at compose time on an honest client (`direct_messages.rs:522`); `verify` deliberately does not enforce it, and nothing bounds `timestamp` on the ingestion path.

1. A peer with a broken or hostile clock sends a DM stamped `now + 1 year`. It arrives, `ts > cutoff`, thread revives.
2. User archives again → cutoff is now that forged timestamp.
3. Every subsequent genuine DM from that peer has `ts <= cutoff`: gate blocks the unhide, rail filter hides the row, unread badge skips it.

Invisible on every surface for a year. This is a regression vs `main`, where the unconditional unhide deleted the entry on the next inbound DM. Unlike the same-second case it is neither bounded nor self-healing.

Mitigation: clamp the cutoff at `now + MAX_DM_FUTURE_SKEW_SECS` when writing it in `archive_row`.

### Smaller items to fold in

- `chat_delegate.rs:4460-4462` lists three explicit-user-action call sites; there are four — `direct_messages::send_structured_dm` (invite-via-DM) is missing.
- `chat_delegate.rs:4507-4508` justifies the `drop(hidden)` with "before `unhide_dm_thread` defers a `with_mut`". On wasm the defer alone would suffice; the real reason is that `defer` runs **synchronously** on non-wasm (`util.rs:75-77`), which is how the test suite runs — so without the drop it is a re-entrant borrow panic. Worth saying, or the drop reads as deletable.
- `.claude/rules/direct-messages.md:165` refers to `chat_delegate::is_thread_hidden`; no such function exists (it is `direct_messages::is_thread_hidden_for`). Pre-existing, six lines above the new section.

### Net assessment

The gate is the right mechanism and closes the reported bug for the common case. But anchoring it on a timestamp — any timestamp — keeps leaking, in both directions: too low a cutoff destroys archives, too high a cutoff hides real messages. The inbound anchoring plus a future-skew clamp plus recomputing from live state closes the ones found so far. Whether that is the right stopping point, versus recording arrival identity, is worth deciding deliberately before the next revision.

[AI-assisted - Claude]

PR #540, opened 2026-07-29
feat(cli): show whether each ban still enforces in `debug bans`

### LINKED ISSUE #472: riverctl: `debug bans` does not show whether a ban still enforces

sanity:
## Problem

`riverctl debug bans <room>` printed every ban with `banned_user_id`, `banned_by_id` and `banned_at_secs`, and nothing about whether the contract still **enforces** it.

Since deputy ban authority (#410), storage and enforcement are separate. A ban goes inert without being removed when the banner leaves or is pruned, when their deputy grant is revoked, or when the step-4 deputizer guardrail trips. A moderator reading `debug bans` saw the ban listed and reasonably concluded the member was kept out. They may well be back in the room.

## Approach

Add `enforcing: bool` to `BanInfo`, computed from `BansV1::ban_is_enforcing` (already `pub`), and surface it in both output branches.

Two small pure helpers carry it so the behaviour is unit testable without a live node:

- `collect_ban_infos(&ChatRoomStateV1, &VerifyingKey) -> Vec<BanInfo>`
- `ban_list_lines(&[BanInfo]) -> Vec<String>` for the human branch

Human output now reads:

```
Ban List (2 bans, 1 enforcing)
=========
  7XSOGJTK banned by PMAQEUP5 at 1700000000
  4KDLM2NQ banned by 9WTZR6HB at 1700000042  [NOT ENFORCING]

Note: 1 ban(s) are stored in room state but are NOT keeping their target out, so
those users are in the room. A ban stops enforcing when its banner leaves the
room or loses the deputy authority it banned under. Check the banner with
`riverctl member deputized-by <room> <banned_by_id>`.
```

The JSON change is additive: `banned_user_id`, `banned_by_id` and `banned_at_secs` keep their names and types, with `enforcing` alongside. No consumer outside this file reads that shape.

### What `enforcing` claims, and what it does not

Worth stating precisely, because the review went down a wrong path here and the field docs now carry the warning.

`enforcing: false` is the actionable case and it is unambiguous: the target is a **current member** that this ban fails to exclude, so that person is in the room right now. That is exactly the situation the issue describes.

`enforcing: true` for a target who is already gone means "not in the room". It is deliberately **not** a promise that the ban would hold if a different member re-invited them. `ban_is_enforcing` skips the authority check for an absent target on purpose: once a ban has done its job the target is gone, so there is no invite chain left to walk and the ancestor grant that authorized the ban cannot be re-derived.

The middle commit on this branch tried to tighten that by computing the flag from a bare `is_ban_authorized` instead, to close the narrow case of an unauthorized banner with an absent target. Testing an ancestor ban (owner -> bob -> alice, bob bans alice) showed why that is wrong: while alice is present both agree the ban enforces, but after the contract removes her the replacement reports NOT ENFORCING. Since every successfully enforced ban ends with its target removed, that flagged essentially **every working ban** as inert, which is far more misleading than the edge case it tightened. It is reverted, and `legitimate_ancestor_ban_still_enforcing_after_its_target_is_removed` pins it so the same wrong fix cannot land again.

The optional reason-string enhancement the issue floats ("banner absent / authority revoked / guardrail") is **not** included. The note points at `member deputized-by`, which answers the same question with a command that already exists.

## Testing

Ten new tests in `cli/src/commands/debug.rs`, built on the fixed-seed key fixtures used by `cli/src/deputies.rs` (the cli crate's dalek build has no `rand` `generate` helper).

Enforcement classification:
- `owner_ban_of_a_present_member_reports_enforcing` (positive baseline, so the negative cases cannot pass against a hardcoded `false`)
- `deputy_ban_reports_enforcing_while_the_grant_stands`
- `deputy_ban_reports_not_enforcing_after_the_grant_is_revoked` (the issue's headline case: revocation via a higher-version `member_info` record, exercising `canonical`)
- `ban_reports_not_enforcing_once_its_banner_is_no_longer_a_member`
- `legitimate_ancestor_ban_still_enforcing_after_its_target_is_removed` (guards the wrong fix described above)
- `a_present_target_agrees_with_the_contracts_excluded_set` ties the flag to `MembersV1::banned_member_ids` across authorized and unauthorized banners, and asserts the target is present so the equivalence it claims actually holds

The target stays a current member in the negative cases on purpose: `ban_is_enforcing` short-circuits to `true` for an absent target, so a fixture that removed the target would have passed for the wrong reason.

Output shape:
- `ban_json_carries_the_enforcing_flag_alongside_the_pre_existing_shape`
- `human_output_marks_only_the_non_enforcing_bans`
- `human_output_stays_quiet_when_every_ban_enforces` (no false alarm on an all-live room)
- `human_output_handles_an_empty_ban_list`

Mutation-tested against the final content rather than assumed:

| Mutation | Tests killed |
|---|---|
| `enforcing` hardcoded `true` | 3 |
| `enforcing` hardcoded `false` | 6 |
| `[NOT ENFORCING]` marker removed | 1 |
| flag computed from a bare `is_ban_authorized` (the reverted approach) | 1, and only that one |

Full `riverctl` lib suite: 285 passed, 0 failed. `cargo fmt --check` clean; `cargo clippy` introduces no new warnings.

Bans have no expiry or TTL in River, so there is no expired-ban case to report on.

Closes #472

[AI-assisted - Claude]

### DIFF
--- a/cli/src/commands/debug.rs
+++ b/cli/src/commands/debug.rs
@@ -5,7 +5,7 @@ use anyhow::{anyhow, Result};
 use clap::Subcommand;
 use ed25519_dalek::VerifyingKey;
 use river_core::room_state::ban::BansV1;
-use river_core::room_state::member::{MemberId, MembersV1};
+use river_core::room_state::member::MemberId;
 use river_core::room_state::ChatRoomStateV1;
 use serde::Serialize;
 
@@ -45,23 +45,26 @@ struct BanInfo {
     banned_user_id: String,
     banned_by_id: String,
     banned_at_secs: u64,
-    /// Whether the contract currently EXCLUDES this ban's target, as opposed to
-    /// the ban being present in state but inert. Since deputy ban authority
-    /// (freenet/river#410) a ban can stop enforcing without being removed: the
-    /// banner left or was pruned, their deputy authority was revoked, or (once
-    /// no absolute grant applies) the target has since deputized the banner.
+    /// Whether this ban is currently keeping its target out of the room, as
+    /// opposed to sitting in state while the target walks around inside. Since
+    /// deputy ban authority (freenet/river#410) a ban can go inert without being
+    /// removed: the banner left or was pruned, their deputy authority was
+    /// revoked, or (once no absolute grant applies) the target has since
+    /// deputized the banner.
     ///
-    /// This mirrors the test `MembersV1::banned_member_ids` applies when it
-    /// builds the excluded set in `post_apply_cleanup`, so it answers the
-    /// question a moderator is actually asking: is this person kept out?
+    /// `false` is the case worth acting on, and it is unambiguous: the target is
+    /// a CURRENT member that this ban fails to exclude, so that person is in the
+    /// room right now.
     ///
-    /// Deliberately NOT `BansV1::ban_is_enforcing`, despite the name. That
-    /// predicate answers a different question — whether a ban is worth KEEPING
-    /// under `max_user_bans` eviction pressure — and to do so it returns `true`
-    /// for any signature-verified member's ban of an ABSENT target without ever
-    /// consulting `is_ban_authorized`. Reporting that as "enforcing" would tell a
-    /// moderator an unauthorized ban keeps someone out when rejoining would in
-    /// fact readmit them, which is the very confusion #472 exists to remove.
+    /// Scope, deliberately not overclaimed: for a target who is already gone this
+    /// reports `true` without re-deriving the banner's authority, because an
+    /// absent member has no invite chain to walk and `is_ban_authorized` cannot
+    /// be evaluated against a hypothetical future one. So `true` means "not in
+    /// the room", not "guaranteed to hold if they are re-invited by someone new".
+    /// Do NOT switch this to a bare `is_ban_authorized` to tighten that: every
+    /// successfully enforced ban ends with its target removed, so doing so
+    /// reports NOT ENFORCING for essentially every ban that is working. Pinned by
+    /// `legitimate_ancestor_ban_still_enforcing_after_its_target_is_removed`.
     enforcing: bool,
 }
 
@@ -82,18 +85,13 @@ fn collect_ban_infos(room_state: &ChatRoomStateV1, owner_vk: &VerifyingKey) -> V
                 .duration_since(std::time::UNIX_EPOCH)
                 .map(|d| d.as_secs())
                 .unwrap_or(0);
-            // The same two gates, in the same order, that `banned_member_ids`
-            // applies per ban. `ban_excludes_its_target_matches_banned_member_ids`
-            // pins the agreement so this cannot drift from the contract.
-            let enforcing =
-                BansV1::ban_signature_matches_current_key(ban, &members_by_id, owner_id, owner_vk)
-                    && MembersV1::is_ban_authorized(
-                        ban.banned_by,
-                        ban.ban.banned_user,
-                        &members_by_id,
-                        &room_state.member_info,
-                        owner_id,
-                    );
+            let enforcing = BansV1::ban_is_enforcing(
+                ban,
+                &members_by_id,
+                &room_state.member_info,
+                owner_id,
+                owner_vk,
+            );
             BanInfo {
                 banned_user_id: ban.ban.banned_user.to_string(),
                 banned_by_id: ban.banned_by.to_string(),
@@ -144,10 +142,11 @@ fn ban_list_lines(bans: &[BanInfo]) -> Vec<String> {
     if inert_count > 0 {
         lines.push(String::new());
         lines.push(format!(
-            "Note: {} ban(s) are stored in room state but NOT enforced, so those \
-             users are not kept out. A ban stops enforcing when its banner leaves \
-             the room or loses the deputy authority it banned under. Check the \
-             banner with `riverctl member deputized-by <room> <banned_by_id>`.",
+            "Note: {} ban(s) are stored in room state but are NOT keeping their \
+             target out, so those users are in the room. A ban stops enforcing \
+             when its banner leaves the room or loses the deputy authority it \
+             banned under. Check the banner with `riverctl member deputized-by \
+             <room> <banned_by_id>`.",
             inert_count
         ));
     }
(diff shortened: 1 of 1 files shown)


sanity:
## Review (2 lenses, run serially and blind to each other's notes)

Tier: **Light+**. Small self-contained diff in a read-only debug command, no wire-format or protocol change, but it reports on ban authority, so the second lens was aimed squarely at whether the reported value is actually true.

Both lenses read the checked-out code, not the diff text. Claude lenses only; no external models (opt-in per the review rule, not run here).

---

### Lens 1 — code-first (read the code before the description; intent vs implementation)

**C1 [BLOCKING, and it turned out to be half-right] — the flag was computed from a predicate that answers a different question.**

Reading `collect_ban_infos` before the issue text, `BansV1::ban_is_enforcing` looked like the obvious source, but its own rustdoc says it answers "is this ban worth KEEPING under `max_user_bans` eviction pressure?", and `post_apply_cleanup` uses it only to order cap eviction. The path that actually excludes members is `MembersV1::banned_member_ids`.

Probed rather than assumed. For a member banner with no authority over an **absent** target: `ban_is_enforcing = true` while `is_ban_authorized = false`. So an unauthorized ban would have been reported as enforcing.

Acted on in 60d382f7, then **reverted in 3ae47318 after lens 2 falsified the fix** — see T1. The finding was real but the obvious correction was worse. Net outcome: `ban_is_enforcing` retained, with the scope of the claim now documented on the field and pinned by a test.

**C2 [minor, fixed]** — the field doc said a ban goes inert when "the target has since deputized the banner", stated unconditionally. The step-4 guardrail only applies once the absolute grants (owner, strict ancestor, owner-appointed moderator) are ruled out. Reworded to "once no absolute grant applies".

**C3 [dismissed, with reason]** — `debug room-state` still reports a bare `Bans: N / max_user_bans` that counts inert bans too. Not changed: that display is a **capacity** reading, and stored bans consume the cap regardless of enforcement (`post_apply_cleanup` caps on `self.bans.0.len()`). Reporting only enforcing bans there would make the capacity number wrong.

**C4 [dismissed]** — `"1 bans"` reads awkwardly. Pre-existing phrasing from the original `"Ban List ({} bans)"`; not a regression and not worth churn in this PR.

---

### Lens 2 — skeptical + testing

**T1 [BLOCKING — falsified lens 1's fix].** Asked what the "corrected" predicate reports for a *legitimate* ban whose target has already been removed, and tested it: owner -> bob -> alice, bob bans alice as her strict ancestor.

| | alice present | after the contract removes her |
|---|---|---|
| `ban_is_enforcing` | true | true |
| bare `is_ban_authorized` (the 60d382f7 fix) | true | **false** |

Once the target is gone there is no invite chain to walk, so the ancestor grant that authorized the ban cannot be re-derived. Since **every successfully enforced ban ends with its target removed**, 60d382f7 would have marked essentially every working ban `[NOT ENFORCING]` — much worse than the narrow case it closed. Reverted, with `legitimate_ancestor_ban_still_enforcing_after_its_target_is_removed` added so the same wrong fix cannot land again, and a `do NOT switch this` warning on the field doc.

**T2 — would the tests fail if the fix were reverted?** Mutation-tested against the final content, not assumed:

| Mutation | Killed by |
|---|---|
| `enforcing` hardcoded `true` | 3 tests |
| `enforcing` hardcoded `false` | 6 tests |
| `[NOT ENFORCING]` marker removed | 1 test |
| bare `is_ban_authorized` (the reverted approach) | 1 test, and only that one |

The last row matters: the guard is precisely targeted rather than incidentally tripping other assertions.

**T3 — is it right for deputy-authority bans (#410)?** Covered in three shapes: grant standing, grant revoked via a higher-version `member_info` record (exercising `canonical`, so a revocation lingering in a duplicate record cannot report stale), and banner no longer a member.

**T4 — expired bans?** No such case exists. `UserBan` carries `owner_member_id`, `banned_at`, `banned_user` and no TTL or expiry field, so there is nothing to report.

**T5 — superseded / duplicate bans.** Two bans on the same target are possible (nothing dedupes by target; `verify`'s own tests construct six bans of one user). The per-ban flag is the right granularity, but the human note originally read "so those users are not kept out", which overclaims when another ban covers the same person. Reworded to "are NOT keeping their target out".

**T6 — vacuous-test check.** `a_present_target_agrees_with_the_contracts_excluded_set` asserts both outcomes actually occur across its cases, and asserts the target is present in each, so the equivalence it claims genuinely holds rather than being true by construction. The negative-case fixtures deliberately keep the target a member, since `ban_is_enforcing` short-circuits to `true` for an absent target and would otherwise have passed for the wrong reason.

**T7 — JSON compatibility.** `banned_user_id` / `banned_by_id` / `banned_at_secs` keep their names and types; `enforcing` is additive. Grepped for consumers of that shape: none outside this file.

---

### Result

Both blocking findings are resolved in the branch. C2, T5 fixed; C3, C4 dismissed with specific reasons above. Full `riverctl` lib suite 285 passed / 0 failed, `cargo fmt --check` clean, no new clippy warnings.

The honest summary of this review: lens 1 correctly identified that the predicate answers a different question, and the fix it motivated was wrong in a way only testing the opposite case exposed. The branch keeps the original predicate and the test coverage from both sides.

[AI-assisted - Claude]

PR #542, opened 2026-07-29
fix(ui): handle notification_status from the shell and offer a manual retry

### LINKED ISSUE #510: fix(ui): River discards every notification_status from the shell, so a blocked permission fails silently with no retry
Found while investigating Ian's report (2026-07-26) that clicking **Enable** on the gateway's notification prompt at try.freenet.org appeared to do nothing. The prompt itself belongs to freenet-core's shell page and the primary defect is tracked in freenet/freenet-core#4966. This issue covers River's half: even when the shell tells us what happened, we throw the answer away and never offer a retry.

## River discards every notification_status message

`install_shell_notification_listener` (`ui/src/components/app/notifications.rs:248-311`) handles exactly one message type:

```rust
if kind != "notification_click" {
    return;
}
```

The shell sends `notification_status` with `granted` / `denied` / `dismissed` / `undeliverable` / `unsupported` (`shell_bridge.js`, `notifyStatusToIframe`), and a repo-wide grep confirms `notification_status` appears nowhere in `ui/src`. All five outcomes are discarded identically, so River cannot tell a granted permission from a blocked one and never surfaces anything to the user.

## We ask once per session and there is no way to retry

`request_enable_via_shell` (`notifications.rs:220-225`) is latched by `ENABLE_PROMPT_SENT: AtomicBool`, and the shell independently sets its own shown-flag. So a prompt that was suppressed, embargoed, or mis-clicked cannot be re-requested for the rest of the session. There is no manual "Enable notifications" control anywhere in the UI to recover: the bell modal (`ui/src/components/room_list/notification_modal.rs`) only chooses All / Mentions / Muted and never touches browser permission.

Net effect for the user: notifications silently never work, with no explanation and no path to fix it.

## Suggested fix

- Handle `notification_status` alongside `notification_click`, storing the result in a global. Signal writes from the JS `MessageEvent` callback must go through `crate::util::defer`, exactly like the existing `notification_click` arm at `notifications.rs:295`.
- On `denied` / `undeliverable` / `unsupported`, surface a brief inline explanation rather than failing silently.
- On `default` (prompt dismissed or suppressed without a decision), reset `ENABLE_PROMPT_SENT` so a later genuine user gesture can ask again.
- Add an explicit "Enable notifications" affordance so the user has a manual retry that carries a fresh user gesture. Natural home is the bell modal, next to the per-room mode choices.

## Latent, separate: the non-gateway path loses user activation

`notifications.rs:350-357` (reached only when `is_in_shell_iframe()` is false, i.e. `dx serve` / local dev) calls `request_permission()` inside `safe_spawn_local`, which is `setTimeout(0)` (`ui/src/util.rs:20-33`), from a call site already several awaits deep in the send pipeline (`conversation.rs:3639`). Transient user activation is long gone by then, so on Firefox and Safari this path fails with `NotAllowedError`, swallowed by the `Err(e) => { warn!(...); false }` arm at `notifications.rs:113-121`. Not the reported bug (gateway users take the shell path), but it will bite local-dev and any future non-gateway deployment.

## Also: #408 is stale

#408 ("browser notifications never fire in the gateway deployment") describes River being unable to call the API at all from the opaque-origin iframe. Its suggested fix (postMessage to the shell) has since shipped: freenet-core#4801 built the proxy and River's client half exists (`is_in_shell_iframe`, `request_enable_via_shell`, `post_notification_to_shell`). #408 should be updated or closed, with this issue carrying the remaining client-side gap.

## Test coverage

`ui/tests/notification-bell.spec.ts` covers only the per-room mode modal. Nothing covers permission, the shell bridge, or `notification_status` handling, which is what #408 already flagged and is still true.

[AI-assisted - Claude]

sanity:
## Problem

River's shell-bridge listener (`ui/src/components/app/notifications.rs`) dropped
every message whose `type` was not `notification_click`:

```rust
if kind != "notification_click" {
    return;
}
```

The gateway shell reports the outcome of its permission offer as
`notification_status` (`notifyStatusToIframe` in freenet-core's
`shell_bridge.js`), with `granted` / `denied` / `dismissed` / `undeliverable` /
`unsupported` / `default`. All six were discarded identically, and
`notification_status` appeared nowhere in `ui/src`. River could not tell a
granted permission from a blocked one and surfaced nothing to the user.

Worse, `request_enable_via_shell` was latched by a `ENABLE_PROMPT_SENT`
`AtomicBool` with no reset, and the bell modal only chose All / Mentions /
Muted — so a prompt that was blocked, snoozed, or mis-clicked could not be
re-requested for the rest of the session. Net effect: notifications silently
never worked, with no explanation and no path to fix it.

This is River's half only. The shell-side defect is freenet/freenet-core#4966;
nothing here depends on it — the shell already sends `notification_status` on
`main`.

## Approach

**Consume the status.** A new `notification_status` arm parses the shell's
`status` string into a `NotificationStatus` enum and records it in a
`NOTIFICATION_STATUS` global. An unrecognised string is ignored rather than
guessed at, so a newer shell adding a status can never overwrite a known one
with a wrong reading. The signal write goes through `crate::util::defer`, as
required for a write originating in a JS `MessageEvent` callback
(`.claude/rules/dioxus-signal-safety.md`), and `record_notification_status` is
the only writer so no path can bypass the re-arm below.

**Re-arm the ask, bounded.** A status of `default` (the prompt closed with no
answer) clears `ENABLE_PROMPT_SENT` so a later gesture can offer again. The
re-arm is capped at one, because the shell re-shows its affordance every time
River asks: uncapped, a user who keeps closing the browser dialog would get the
bar back on every message they send. `dismissed` deliberately does NOT re-arm —
the shell also snoozes it for 24h, so River re-asking would just bounce off that
snooze.

**Surface it, and offer a retry.** The bell modal grows a "Browser permission"
section with a short explanation and an "Enable notifications" button. The
button is offered exactly where asking again can change the answer (`dismissed`,
`default`, or nothing known yet) — offering it against a browser-level block or
an already-granted permission would just move the silent dead end. Its click
calls `request_enable_now()` **directly**, not through `crate::util::defer`,
because the browser only honours a permission request made inside the click's
transient user activation and `defer` is a `setTimeout(0)`.

The manual path is deliberately not latched: the once-per-session latch exists
so sending messages doesn't nag, and an explicit click is exactly the retry it
must not swallow.

Served top-level (dev / `no-sync`) River has a real origin, so the same button
calls the Notifications API directly. `Notification::request_permission()` is
invoked synchronously inside the click for the activation reason above; only the
awaiting of the already-created promise is deferred. This incidentally avoids
the lost-activation problem the issue notes for the existing auto path, which is
otherwise unchanged.

**One bug found while writing this:** `current_notification_status` reads
`Notification.permission` on every modal render, and `web_sys`'s binding for it
is not `Result`-returning — reading `.permission` off an undefined `Notification`
raises a JS `ReferenceError` that aborts the WASM module. Guarded by
`notification_api_available()`. Playwright's mobile-safari project has no
Notifications API at all, so this was reachable.

## Testing

`cargo test -p river-ui --bins`: 785 pass (10 new). New coverage:

- The six shell status strings parse to distinct variants (a cross-repo wire
  contract — a rename on either side reverts the fix exactly), and unrecognised
  strings are ignored.
- Only an unanswered prompt re-arms the automatic ask, and the re-arm is bounded.
- The Enable button is offered exactly where asking again can help, and every
  state has its own non-empty explanation.
- Source pins for the parts no native test can reach (a JS closure and rsx): the
  listener consumes `notification_status`; the status write stays inside
  `defer` and has exactly one writer; `request_permission()` is called before the
  spawn; the Enable click is not deferred.

Every new test was mutation-tested — the fix reverted or subtly broken in eight
ways, each caught by a named test. One of the eight (raising the re-arm cap to
`usize::MAX`) initially passed, because the bound assertion was phrased against
the constant it was testing and so held for any cap at all; that vacuous test is
fixed in its own commit.

Playwright `notification-bell.spec.ts`: 30 pass across chromium, firefox, webkit,
mobile-chrome and mobile-safari. The new spec asserts the explanation always
renders and that the button appears iff the browser state is one where asking
again helps. The five projects cover all three branches — Chromium reports
`denied`, Firefox and desktop WebKit `default`, mobile Safari has no API. Both
new `data-testid`s are absent on `main`, so the spec fails there.

Manually verified in Firefox that clicking the button leaves the app alive and
the modal rendered, with no new console error.

Closes #510

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -108,7 +108,9 @@ pub fn App() -> Element {
 
     // In the gateway iframe, notifications are shown by the shell page; listen
     // for its `notification_click` replies to route to the clicked room
-    // (freenet/river#408). No-op when served as a top-level page. Installs once.
+    // (freenet/river#408) and its `notification_status` replies to learn whether
+    // the browser will show anything at all (freenet/river#510). No-op when
+    // served as a top-level page. Installs once.
     crate::components::app::notifications::install_shell_notification_listener();
 
     // Install the document-level click interceptor that catches
--- a/ui/src/components/app/notifications.rs
+++ b/ui/src/components/app/notifications.rs
@@ -1363,13 +1363,24 @@ mod notify_gate_tests {
     /// isn't shown the shell's bar again on every message they send.
     #[test]
     fn the_automatic_rearm_is_bounded() {
+        // Concrete numbers, NOT `MAX_ENABLE_PROMPT_REARMS` itself: phrased
+        // against the constant this test passes for ANY cap, including no cap
+        // at all — it stayed green under a mutation that raised the cap to
+        // `usize::MAX`.
         assert!(
-            status_rearms_auto_prompt(NotificationStatus::Undecided, MAX_ENABLE_PROMPT_REARMS - 1),
-            "the last permitted re-arm must still be allowed"
+            MAX_ENABLE_PROMPT_REARMS <= 2,
+            "more than a couple of automatic re-asks per session is a nag; the \
+             manual button is the unbounded path"
         );
         assert!(
-            !status_rearms_auto_prompt(NotificationStatus::Undecided, MAX_ENABLE_PROMPT_REARMS),
-            "re-arming past the cap would re-prompt on every message"
+            status_rearms_auto_prompt(NotificationStatus::Undecided, 0),
+            "the first re-arm must be allowed, or an unanswered prompt is final \
+             and #510's no-retry complaint stands"
+        );
+        assert!(
+            !status_rearms_auto_prompt(NotificationStatus::Undecided, 3),
+            "the automatic ask must stop re-arming; otherwise the shell shows \
+             its bar again on every message the user sends"
         );
     }
 
--- a/ui/src/components/app/notifications.rs
+++ b/ui/src/components/app/notifications.rs
@@ -16,7 +16,7 @@ use river_core::room_state::member::MemberId;
 use river_core::room_state::member_info::MemberInfoV1;
 use river_core::room_state::message::{AuthorizedMessageV1, MessageId, RoomMessageBody};
 use std::collections::HashSet;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use wasm_bindgen::prelude::*;
 use wasm_bindgen_futures::JsFuture;
 use web_sys::{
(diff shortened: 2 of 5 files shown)


sanity:
## Review

Two independent adversarial lenses, run serially against the checked-out code (no subagents available in this session). Per `~/.claude/rules/multi-model-review.md` as amended 2026-07-26, external non-Claude models are opt-in and were not invoked.

### Lens 1 — code-first (read the code before the description; intent vs implementation)

Two findings, both fixed in c8e01269:

1. **Stale permission on the top-level path (medium).** `current_notification_status` preferred the value River remembered from its own last request over the browser's own answer. A user who denied once and then allowed notifications in site settings would be told "your browser is blocking notifications" for the rest of the session, with the Enable button withheld — a stale silent failure of exactly the shape this PR removes. The browser is authoritative when River is served top-level and stays authoritative, so the remembered value is now ignored there. It is still *read*, because the read is what subscribes the render to the signal; dropping the read would stop the modal updating when a request completes. The precedence is now a pure `resolve_status(framed, stored, live)` with tests.

2. **`dismissed` copy promised a retry that doesn't work yet (low).** That status also covers "inside the gateway's snooze window from an earlier decline", during which asking again provably shows nothing for up to a day. The old wording ("You can ask for it again") would have put the user back in front of a control that silently does nothing. Reworded to say the prompt may not reappear straight away.

Checked and found correct: all six `notifyStatusToIframe` strings are covered; the source/`__freenet_shell__` checks still gate the new arm; `record_notification_status` is the only writer, so no path can skip the re-arm; `read_browser_permission` is invoked lazily via `bool::then`, so the framed path never reads a meaningless opaque-origin permission.

### Lens 2 — skeptical + testing (assume bugs exist; does the status actually reach the UI?)

No new defects.

**Does it silently no-op when the shell sends nothing?** No. Framed with no status ever reported, `resolve_status` returns `None`, which is its own state: the user sees "Desktop notifications are not set up yet for this browser." plus a working Enable button. Both orderings work — a status arriving before the modal opens is read from the signal on open; one arriving while it is open re-renders it through the `try_read` subscription. `defer` is safe here because `capture_runtime()` runs earlier in `App()` than `install_shell_notification_listener()`.

**Against an older gateway shell** that predates `notifyStatusToIframe`, no status is ever sent, so the modal shows the "not set up yet" state and the Enable button posts `notification_enable_prompt` — a message old shells already handle. Degrades to today's behaviour plus a manual trigger, never to a crash.

**Would the tests fail if the fix were reverted?** Verified by mutation, not by inspection. Nine mutations, each caught by a named test:

| Mutation | Caught by |
|---|---|
| Delete the `notification_status` arm | `the_shell_listener_consumes_notification_status` |
| Move `request_permission()` inside the spawn | `the_direct_permission_request_stays_inside_the_click_gesture` |
| Drop the `defer` around the signal write | `the_status_write_is_deferred_and_has_one_writer` |
| Wrap the Enable click in `defer` | `the_enable_click_is_not_deferred` |
| Let `dismissed` re-arm the auto ask | `only_an_unanswered_prompt_rearms_the_automatic_ask` |
| Offer Enable on `denied` | `enable_is_offered_exactly_where_asking_again_can_help` |
| Rename the `"default"` token | `every_shell_status_string_parses` |
| Change the re-arm cap constant | `the_automatic_rearm_is_bounded` |
| ~~Break the re-arm counter~~ | ~~covered~~ **NOT COVERED — see correction below** |
| Prefer the remembered status over the live one | `the_browser_answer_wins_over_a_remembered_one_when_unframed` |

The re-arm-cap mutation initially passed: the assertion was phrased against `MAX_ENABLE_PROMPT_REARMS` itself, so it held for any cap including none. Fixed in 423ce609 to assert concrete numbers.

### Known gaps, accepted

- **No native test drives the real atomics.** Only the pure `status_rearms_auto_prompt` predicate is tested. The statics are process-global and native tests run in parallel threads, so a test mutating them would be flaky and could corrupt a sibling. The predicate plus the one-writer pin is the trade.
- **No browser test of the framed branch.** Simulating the shell would mean standing up a fake shell page posting `notification_status`. The decision is covered by `resolve_status`'s unit tests and the wire contract by the status-string test.
- **The latent activation problem on the automatic path is not fixed here**, deliberately. As the issue notes, `request_permission_on_first_message` is reached several awaits deep in the send pipeline, so activation is already gone before `safe_spawn_local` is involved — removing the spawn would not fix it. A real fix means moving the trigger to the send button's click handler, which changes when River prompts. The new manual button gives affected users a working path in the meantime.

[AI-assisted - Claude]


---

## Correction (2026-07-29)

**The mutation table above overstated one row, and I am correcting it in place rather than leaving the wrong claim standing.**

I originally listed "Remove the re-arm cap" as covered by `the_automatic_rearm_is_bounded`. That is true only for *changing the constant*. It is false for *breaking the counter*: that test asserts `MAX_ENABLE_PROMPT_REARMS <= 2` and calls the predicate with a caller-supplied `rearms_used`, so it passes with `ENABLE_PROMPT_REARMS.fetch_add(1, ...)` deleted.

Independent review confirmed by execution that **both** statements of the re-arm could be deleted with the whole suite green:

```
baseline                                        787 passed
delete ENABLE_PROMPT_REARMS.fetch_add(1, ...)   787 passed   <-- survived
delete ENABLE_PROMPT_SENT.store(false, ...)     787 passed   <-- survived
```

So the "reset the flag" half of #510 shipped here with no coverage at all. Dropping the `SENT` reset means an unanswered prompt never re-arms; dropping the `REARMS` increment leaves the counter at 0, making the re-arm unbounded so the shell's affordance bar returns on every message sent.

My "known gaps, accepted" note above correctly ruled out a *behavioural* test (process-global statics vs parallel test threads) but stopped there, missing that a **source pin** has no parallelism problem — an idiom this same file already uses four times.

Fixed forward in **#550**, along with a vacuous browser assertion, a `debug!` that is compiled out in release, and the bell-level visibility gap. That PR's pin is verified by watching it fail with each line deleted.

[AI-assisted - Claude]

sanity:
## Review follow-up: the framed-branch gap is now closed

The review above listed "no browser test of the framed branch" as an accepted gap. It turned out to be cheap to close: `invitation-sandbox.spec.ts` already establishes a fake-gateway-shell fixture, and `notification-shell-status.spec.ts` (2501da3c) reuses that shape. A minimal shell parent page posts `notification_status` down to a River iframe and counts the `notification_enable_prompt` messages River posts up.

That covers, in a real browser and a real cross-frame postMessage, the whole path no native test in a WASM crate can reach: framing detection, the listener arm, the deferred signal write, the framed branch of `resolve_status`, the notice mapping, and — the half of #510 the one-way latch made impossible — that the manual retry actually reaches the shell, twice in a row.

**Mutation-verified rather than assumed.** I deleted the `notification_status` arm, rebuilt the UI, and re-ran the spec: both tests fail, on `toContainText` — the modal renders but the status never changes what it says, which is #510 exactly. Restoring the arm and rebuilding returns all 10 (2 tests x 5 browser projects) to green.

Local runs on the final code: `cargo test -p river-ui --bins` 787 pass; Playwright `notification-bell` + `notification-shell-status` 40 pass across chromium, firefox, webkit, mobile-chrome and mobile-safari.

Two gaps remain accepted, unchanged: no native test drives the real atomics (process-global statics vs parallel test threads), and the latent activation problem on the automatic prompt path is deliberately out of scope, since the activation is already lost before `safe_spawn_local` is reached and a real fix would change when River prompts.

[AI-assisted - Claude]

PR #543, opened 2026-07-29
perf(ui): stop the composer autosize from forcing a layout every keystroke

### LINKED ISSUE #468: perf(ui): composer textarea auto-resize forces ~3 layouts per keystroke

sanity:
## Problem

`auto_resize_message_input` runs on every `oninput`, and always did:

```rust
el.style().set_property("height", "auto").ok();      // write
let new_height = el.scroll_height().min(168);        // read  -> forced sync layout
el.style().set_property("height", &format!(...)).ok(); // write
```

The write immediately before the read forces a synchronous layout, and both
writes invalidate style. Confirmed by instrumenting `CSSStyleDeclaration.setProperty`
and the `scrollHeight`/`clientHeight` getters in a Chromium page and typing into a
real composer — the recorded trace is exactly `write height=auto` /
`read scrollHeight` / `write height=44px`, repeated once per keystroke, with no
variation. CDP `Performance.getMetrics` over 26 keystrokes reproduces the numbers
from #468: **79 layouts, 80 style recalcs** (~3 per keystroke).

## Approach

`scrollHeight` reports `max(content height, clientHeight)`. So it *is* the content
height while the content overflows the box, and saturates at the box height once
the content fits. The composer is sized to its content's padding-box height while
`box-sizing: border-box` counts the textarea's 1px borders inside that number, so a
correctly-sized composer sits 2px short of its own content and keeps reporting
`scrollHeight > clientHeight`. That makes the common case — typing, growing —
measurable in place:

- Measure first. If `scrollHeight > clientHeight` the content overflows, so
  `scrollHeight` is already the answer: no collapse.
- Only when `scrollHeight == clientHeight` (the content stopped overflowing, i.e.
  it shrank below the box) is the measurement uninformative and the collapse paid for.
- Write the height only when it actually changes.

The asymmetry is what makes the shortcut safe: the fast path is taken *only* when
the content overflows, i.e. only when the box is too **short**, so it can never
leave a stale too-tall composer behind. Shrinking always goes through the collapse.

Rejected the hidden-mirror-element alternative from the issue: it needs a second
node kept in sync with the textarea's font, padding, and width, still costs a
forced layout to read, and buys nothing over measuring in place.

Also promoted the bare `168` to `MAX_COMPOSER_HEIGHT` and interpolated it into the
textarea's `max-height`, so the clamp and the CSS ceiling can't drift apart.

## Testing

New Playwright test `Composer auto-resize cost (#468)` in `ui/tests/message-layout.spec.ts`.
It records inline-`height` writes on the composer rather than timing anything, so it
asserts the DOM access pattern (which is what the issue is about) and is not
timing-dependent:

- typing 19 characters within a line produces **zero** height writes, and the height is unchanged
- growing to a second line still resizes, in exactly one write and with no `height: auto` collapse
- `fill()` with 4-line and 30-line (clamped) content, then back to one character, returns to the exact single-line height — this covers the paste and shrink-from-clamped paths

Verified non-vacuous three ways, each by rebuilding the wasm and re-running:

| change | result |
|---|---|
| revert the fix entirely | FAILS — 38 writes recorded (`auto`/`44px` per keystroke) |
| keep the fix but never collapse (fast path always) | FAILS — height stuck at 114px instead of 44px; **also** fails the existing #221 test (138px vs `<= 48`) |
| keep the fix but always write | FAILS — 21 writes where 0 expected |

Measured improvement on the same 26-keystroke CDP probe:

| | before | after |
|---|---|---|
| LayoutCount | 79 | **27** |
| RecalcStyleCount | 80 | **28** |
| LayoutDuration | 6.94 ms | **3.89 ms** |
| RecalcStyleDuration | 10.14 ms | **1.36 ms** |

Combined layout + style time drops from 17.1 ms to 5.2 ms (-70%).

Regression coverage requested on the issue passes on all five browser projects
(chromium, firefox, webkit, mobile-chrome, mobile-safari): `textarea height resets
after sending a multi-line message` (#221) and `edit container fits within viewport
at narrow widths` (#205). Full Playwright suite: 687 passed, 23 skipped, 0 failed.
`cargo fmt --check` clean, `cargo clippy` adds no warnings in the changed file,
`cargo test -p river-ui` 780 passed.

Closes #468

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/conversation/message_input.rs
+++ b/ui/src/components/conversation/message_input.rs
@@ -29,16 +29,65 @@ fn measure_draft(text: &str, reply: Option<&ReplyContext>, is_private: bool) -> 
     }
 }
 
+/// Height ceiling for the composer, in px (~7 lines). Kept in step with the
+/// textarea's `max-height`: this clamp is what stops the growth, the CSS is
+/// what leaves the overflow scrollable.
+const MAX_COMPOSER_HEIGHT: i32 = 168;
+
+/// Size the composer to its content, clamped to [`MAX_COMPOSER_HEIGHT`].
+///
+/// This runs on every keystroke, so the DOM access pattern is the whole cost.
+/// Collapsing to `height: auto` before every measurement — the obvious
+/// implementation — puts a style write immediately before a `scrollHeight`
+/// read, which forces a synchronous layout, and then writes again; measured at
+/// ~3 layouts per keystroke and ~6.6 ms of layout + style per keystroke under a
+/// 6x CPU throttle (freenet/river#468). The measuring read is unavoidable. Both
+/// writes usually are not:
+///
+/// `scrollHeight` reports `max(content height, clientHeight)`, so it is exactly
+/// the content height while the content OVERFLOWS the box, and saturates at the
+/// box height once it fits. We set `height` to the content's padding-box height
+/// while `box-sizing: border-box` counts the textarea's borders inside that
+/// number, so a correctly-sized composer sits its border-width short of its own
+/// content and keeps reporting `scrollHeight > clientHeight`. Typing therefore
+/// measures in place: no collapse, and no write at all while the height is
+/// unchanged.
+///
+/// `scrollHeight == clientHeight` means the content stopped overflowing, i.e.
+/// it shrank below the box. That is the one case where the measurement carries
+/// no information and the collapse has to be paid for.
+///
+/// Note the asymmetry that makes the fast path safe: it is taken only when the
+/// content overflows, i.e. only when the box is too SHORT, so it can never
+/// leave a stale too-tall composer behind.
+///
+/// The "no writes while typing within a line" property is pinned by a Playwright
+/// test (`Composer auto-resize cost (#468)`). If the textarea's border or
+/// box-sizing ever changes such that the box fits its content exactly, the fast
+/// path stops firing — that test goes red rather than the cost silently
+/// returning.
 fn auto_resize_message_input() {
     let Some(el) = get_message_textarea() else {
         return;
     };
-    // Reset height to auto to measure scrollHeight correctly, then clamp to ~7 lines.
-    el.style().set_property("height", "auto").ok();
-    let new_height = el.scroll_height().min(168);
-    el.style()
-        .set_property("height", &format!("{}px", new_height))
-        .ok();
+    let style = el.style();
+
+    // Both reads come off one layout; no write separates them.
+    let scroll_height = el.scroll_height();
+    let content_height = if scroll_height > el.client_height() {
+        scroll_height
+    } else {
+        style.set_property("height", "auto").ok();
+        el.scroll_height()
+    };
+
+    let target = format!("{}px", content_height.min(MAX_COMPOSER_HEIGHT));
+    // Skip a no-op write: its style invalidation is the bulk of the
+    // per-keystroke cost. After a collapse the inline value is `auto`, so that
+    // branch always writes.
+    if style.get_property_value("height").unwrap_or_default() != target {
+        style.set_property("height", &target).ok();
+    }
 }
 
 fn get_message_textarea() -> Option<web_sys::HtmlTextAreaElement> {
@@ -189,7 +238,7 @@ pub fn MessageInput(
                             id: "message-input",
                             "data-testid": "message-input",
                             class: "w-full px-4 py-2.5 bg-surface border border-border rounded-xl text-text placeholder-text-muted focus:outline-none focus:ring-2 focus:ring-accent/50 focus:border-accent transition-colors resize-none min-h-[44px] overflow-y-auto",
-                            style: "max-height: 168px;",
+                            style: "max-height: {MAX_COMPOSER_HEIGHT}px;",
                             placeholder: "Type your message...",
                             value: "{message_text}",
                             rows: "1",
--- a/ui/tests/message-layout.spec.ts
+++ b/ui/tests/message-layout.spec.ts
@@ -442,6 +442,109 @@ test.describe("Message input auto-resize (#221)", () => {
   });
 });
 
+// #468: the composer resize runs on every keystroke, and the original
+// implementation always did collapse-to-`auto` → read `scrollHeight` → write
+// the result. The write immediately before the read forces a synchronous
+// layout, and both writes invalidate style, measured on the issue at ~3 layouts
+// per keystroke (79 layouts over 26 keystrokes) and ~6.6 ms of layout + style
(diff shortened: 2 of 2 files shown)


sanity:
## Review — 2 independent lenses, run serially against the checked-out code

Risk tier: **Light** per `multi-model-review` (small self-contained UI diff, no
concurrency / crypto / protocol / migration surface), so two distinct lenses
rather than the full 3-5. Both read the code before this description. No
external models (Claude lenses are the standard review).

### Lens 1 — code-first (intent vs. implementation)

Read `auto_resize_message_input` and its three call sites cold, then checked
each claim the code makes against the running app rather than accepting it.

| # | Claim under test | Result |
|---|---|---|
| 1 | `style.get_property_value("height")` must not force layout, or the fix defeats itself | **Clean.** It reads the inline declaration, not computed style. Confirmed empirically: post-fix `LayoutCount` fell 79 -> 27 over 26 keystrokes, which is impossible if the guard read flushed layout. |
| 2 | The doc comment promises the pinning test goes red if the border/box-sizing coupling breaks | **Clean.** If the fast path stops firing, the collapse branch writes on every keystroke and `expect(typingWrites).toEqual([])` fails. Demonstrated by the full-revert mutation. |
| 3 | `style: "max-height: {MAX_COMPOSER_HEIGHT}px;"` renders real CSS, not a literal brace | **Clean.** At runtime: computed `max-height: 168px`, attribute `max-height: 168px;`. Filling 40 lines clamps to rect 168 / inline `168px` / scroll 980 / client 166 with `overflow-y: auto`. |
| 4 | A Dioxus re-render could clobber the inline height, which the fast path would then mis-read | **Clean.** Inline `68px` survives emoji-picker open/close and two viewport changes. |
| 5 | `unwrap_or_default()` on the guard read | **Clean**, fails safe: an error yields `""`, which never equals the target, so it writes. |

**One pre-existing gap found, not introduced here** (reported below, not fixed
in this PR): `handle_emoji_select` sets `message_text` without calling the
resize, and the picker closes after each pick, so an emoji that pushes content
onto a new line doesn't grow the composer until the next keystroke. Identical
before and after this change — the old code also only ran on `oninput`.

### Lens 2 — skeptical, targeting THE failure mode: does the shortcut ever leave a stale height?

Attacked the "skip when unchanged" path directly. The nastiest case is not a
text edit at all: a **rewrap**, where the content height changes with no
`oninput` to react to it, leaving the box the wrong size — precisely where a
naive in-place measurement would wrongly conclude "nothing changed".

| # | Attack | Result |
|---|---|---|
| A | Type wrapping text at 640px, **widen** to 1400px. Content shrinks 2 lines -> 1 with no input event, so the box is left too tall. | **Clean.** At that moment `scrollHeight 66 == clientHeight 66`, so the next keystroke takes the collapse path and lands on **44px** — ground truth (manual collapse-and-measure) is also **44**. |
| B | Same in reverse: **narrow** 1280 -> 600. | **Clean.** Recovers to **65px**; ground truth **65**. |
| C | `@mention` selection (a call site that passes the resize as a callback, replacing `@` with a 33-char token). | **Clean.** 44px, ground truth 44. |
| D | Send path (`defer(auto_resize_message_input)` after clearing). | **Clean.** Existing #221 test green on all 5 projects, and proven non-vacuous — the never-collapse mutation fails it at 138px vs `<= 48`. |
| E | Shrink straight from the clamped ceiling (30 lines -> 1 char). | **Clean**, asserted in the new test. |

The structural argument behind those results: the fast path is taken **iff the
content overflows the box**, and `scrollHeight == max(content, clientHeight)`,
so `scrollHeight > clientHeight` implies `scrollHeight == content height`. It
therefore fires only when the box is too **short**, and can never leave a stale
too-**tall** composer. A stale height is not merely untested — it is
unreachable through this branch.

Sub-pixel rounding checked too: both metrics are integer-rounded, so a hidden
<1px overflow sends us down the (correct, slower) collapse path, and a
fabricated one sets the box to the rounded content height. Both directions fail
safe.

### Test quality

- **Would it fail if the fix were reverted?** Yes — 38 writes recorded instead of 0.
- **Two further mutations, each rebuilt and re-run:** fast-path-always leaves the height at 114px (expected 44) *and* breaks #221; write-always records 21 writes where 0 are expected. So both halves of the fix are independently pinned.
- **Determinism:** 6 consecutive runs x 5 browser projects = 30/30 at `--retries=0`. The test records style writes rather than timing anything, so it has no timing dependency.

### Findings to address

Nothing blocking. Two observations, both **pre-existing and unchanged by this
diff**, so deliberately not folded into a perf PR — filed separately instead:
the composer height is only recomputed on input-ish events, so (a) emoji
insertion and (b) a window resize that rewraps a draft both leave the height
stale until the next keystroke.

[AI-assisted - Claude]

Issue #538, opened 2026-07-29
Flaky: invitation-sandbox 'no premature suppression' on mobile-safari under full-suite load


sanity:
## Problem

`invitation-sandbox.spec.ts:197` — *"invite modal still appears on reload
BEFORE any action (no premature suppression)"* — is flaky on the
**mobile-safari** project when the full Playwright suite runs. It failed once
and passed on retry during a local full-suite run (685 tests, 5 projects).

Per the repo's "flaky tests are broken tests" rule, filing rather than
ignoring. Found incidentally while working on #537; unrelated to that change.

## Evidence

- Full suite (685 tests): 661 passed, 23 skipped, **1 flaky** — this test.
- Standalone, same build, same server: **6/6 passed**, ~2.5s per run.

So it does not reproduce in isolation; it only fails under the parallel load
of the full suite.

## Hypothesis (not yet confirmed)

This is one of the heaviest test shapes in the suite: it boots the full WASM
app inside a sandboxed iframe **twice** (initial load, then
`window.__shellReload()`), each gated on
`expect(modalHeading(frame)).toBeVisible({ timeout: 15_000 })`.

Under full-suite parallelism, two sequential WASM boots plus iframe setup
plausibly exceed the 15s budget on a loaded machine. That would make this a
resource-contention timeout rather than a logic error.

If that is right, the fix is *not* simply a longer timeout — that hides the
cost. Better options are to wait on a deterministic readiness signal rather
than the modal's visibility, or to reduce the double-boot.

I did not confirm the hypothesis: the flake did not reproduce standalone, and
the retry succeeded so no failure artifact was retained.

## Reproduction

```bash
cargo make build-ui-example-no-sync
cd target/dx/river-ui/release/web/public && python3 -m http.server 8101 &
cd ui/tests && PLAYWRIGHT_BASE_URL=http://127.0.0.1:8101 npx playwright test
```

Watch for the `1 flaky` line naming this test.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-08-22T16:44:40.019541140+00:00
Issue hash: `b20933e3`

## Value Assessment

[...]

PR #545, opened 2026-07-29
fix(ui): defer event-handler global-signal writes; retain traces for flakes

### LINKED ISSUE #539: edit_room_modal closes via an undeferred GlobalSignal write in onclick (dioxus-signal-safety violation)
## Problem

`ui/src/components/room_list/edit_room_modal.rs` closes the room-details panel
by writing a **GlobalSignal directly inside an `onclick`**, with no
`crate::util::defer()`:

- `edit_room_modal.rs:66` — the backdrop overlay
- `edit_room_modal.rs:375` — the ✕ close button

```rust
onclick: move |_| {
    EDIT_ROOM_MODAL.write().room = None;
},
```

`.claude/rules/dioxus-signal-safety.md` names this shape explicitly as wrong:

> **ALSO WRONG — onclick handlers trigger the same RefCell panic**
> ```rust
> onclick: move |_| {
>     ROOMS.write().map.remove(&key);
> };
> ```

The rationale is that a signal write's Drop fires subscriber notifications
synchronously, and Firefox mobile runs those during Drop, so a memo doing
`try_read()` on the same signal can hit a re-entrant `RefCell already borrowed`
panic. `defer()` also supplies the Dioxus runtime + root scope.

The leave-room path in the **same file** (line 333) already does this correctly,
inside a `crate::util::defer()` block — so this is an inconsistency within one
component, not a deliberate exemption.

## Evidence / severity

No symptom reproduced. An instrumented 25-cycle open/copy/close/reopen loop
against both WebKit and Firefox (alternating close-button and backdrop routes)
produced no panic and no stale render. So this is filed as a latent
rule violation, **not** a known-live bug — severity low, correctness risk real
but unobserved.

Deliberately not fixed in #536 (the Firefox selection fix): making the close
deferred changes modal-close timing app-wide, which is a bigger blast radius
than that PR warrants without evidence of harm. Noting it here so it is not
silently lost.

## Suggested fix

Wrap both writes:

```rust
onclick: move |_| {
    crate::util::defer(move || {
        EDIT_ROOM_MODAL.write().room = None;
    });
},
```

Then confirm `copy-clipboard-feedback.spec.ts` (which closes this class of
modal via the backdrop) and the room-details specs still pass — the close
becomes asynchronous, so any assertion that races it would surface there.

Worth grepping for the same shape in the other modals while in there.

[AI-assisted - Claude]

sanity:
## Problem

Two follow-ups from #536.

**#539 — signal-safety violation.** Seven event handlers across four files
mutated a `GlobalSignal` directly inside `onclick`:

| file | handlers |
|---|---|
| `room_list.rs` | open the create-room modal |
| `create_room_modal.rs` | backdrop, Cancel |
| `edit_room_modal.rs` | backdrop, close |
| `receive_invitation_modal.rs` | Retry, Dismiss |

`.claude/rules/dioxus-signal-safety.md` names this exact shape as wrong: a
signal write's Drop fires subscriber notifications synchronously, and Firefox
mobile runs them during Drop, so a memo doing `try_read()` on the same signal
can hit a re-entrant `RefCell already borrowed` panic. `defer` also supplies
the Dioxus runtime + root scope a bare handler lacks. `notification_modal.rs`
already did this correctly — this makes the rest match it.

**#538 — a flake that destroyed its own evidence.** `retries: 2` means an
intermittent failure is reported as "flaky" and the job still goes green, and
the artifact upload was gated on `if: failure()`. So a flake left nothing
behind but the test's name.

## Approach

**#539** — wrap all seven in `crate::util::defer`. The Dismiss handler needed
care: its two statements are order-dependent and
`dismiss_invitation_persistently` itself writes a signal, so both go inside the
defer (code after a `defer()` runs BEFORE the deferred closure), and it needs a
clone because an `onclick` is `FnMut` and the invitation is not `Copy`.

Guarded by a **source scan**, not per-site pins: the deferred and undeferred
shapes are indistinguishable from the outside — the modal closes either way —
and pinning seven sites would not stop the eighth.

**#538 — I could not reproduce it, and the hypothesis I filed it with is
wrong.** I am deliberately not shipping a speculative fix.

I filed it guessing "timeout under parallel load". Measuring the phases on a
machine at load average 86 (where it was originally seen):

| phase | time | budget |
|---|---|---|
| first boot → `.app-root` | ~1.0s | 30s |
| → modal visible | ~40ms | 15s |
| reload boot → `.app-root` | ~0.8s | 30s |
| → modal visible | **~28ms** | **15s** |

That is a ~500x margin on the assertion that actually failed, so a timeout
does not explain it. Nor is it reproducible: **58 runs all passed** — 6
standalone, 12 through Playwright at 6 workers, and 40 across concurrent
browser contexts with generous timeouts and page-error capture, which would
have caught a genuine "modal never appears" race.

Bumping the timeout would have been the obvious move and would have been
cargo-culting a fix onto a disproven theory. What this PR fixes instead is the
reason the flake was undiagnosable: keep a trace `on-first-retry`, and upload
`test-results/` even when the job goes green.

**#538 therefore stays open** — this makes the next occurrence produce a real
artifact rather than another guess.

## Testing

- `cargo test -p river-ui --bins` — **781 passed** (780 + the new scan).
- Full Playwright suite — **682 passed, 23 skipped, 0 flaky**, across all five
  projects. Deferring modal open/close changes their timing, so this was the
  real risk; the create-room, edit-room, receive-invitation, notification and
  rooms-loading specs all pass.
- `cargo fmt` clean. No delegate/contract WASM rebuilt or committed.

**Both parser branches of the new scan are mutation-tested.** Un-deferring a
braced handler makes it fail with `edit_room_modal.rs:381`; rewriting one as an
expression-bodied handler (`onclick: move |_| MODAL.write().show = true,`)
makes it fail with `room_list.rs:395`. The first draft of the scan had a false
positive — for a brace-less handler it grabbed the next unrelated `{` block —
which the mutation test caught and which is now fixed and commented.

The trace change is verified too: a deliberately failing probe produced
`test-results/<test>-retry1/trace.zip`.

Closes #539

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -311,12 +311,20 @@ jobs:
       run: npx playwright test
       working-directory: ./ui/tests
 
+    # `always()`, not `failure()`: with `retries: 2` an intermittent test is
+    # reported as "flaky" and the job still succeeds, so a `failure()` guard
+    # threw away the one artifact that could explain it (see freenet/river#538,
+    # where the trace was gone and the attached hypothesis turned out to be
+    # wrong). The playwright config keeps a trace `on-first-retry`, so this
+    # uploads the flake evidence too. On a clean run the directory is empty and
+    # this is a no-op.
     - name: Upload test results
-      if: failure()
+      if: always()
       uses: actions/upload-artifact@v4
       with:
         name: playwright-report
         path: ui/tests/test-results/
+        if-no-files-found: ignore
 
   claude-ci-analysis:
     name: Claude CI Analysis
--- a/ui/src/components/room_list.rs
+++ b/ui/src/components/room_list.rs
@@ -389,8 +389,13 @@ pub fn RoomList() -> Element {
                         "data-testid": "create-room-button",
                         class: "p-1.5 rounded-md text-text-muted hover:text-accent hover:bg-surface transition-colors",
                         title: "Create Room",
+                        // Signal mutation from an event handler must be deferred
+                        // (dioxus-signal-safety: direct writes here are the
+                        // Firefox mobile RefCell re-entrancy crash path).
                         onclick: move |_| {
-                            CREATE_ROOM_MODAL.write().show = true;
+                            crate::util::defer(move || {
+                                CREATE_ROOM_MODAL.write().show = true;
+                            });
                         },
                         Icon { width: 14, height: 14, icon: FaPlus }
                     }
--- a/ui/src/components/room_list/create_room_modal.rs
+++ b/ui/src/components/room_list/create_room_modal.rs
@@ -136,9 +136,14 @@ pub fn CreateRoomModal() -> Element {
         // Backdrop
         div {
             class: "fixed inset-0 bg-black/50 z-40",
+            // Signal mutation from an event handler must be deferred
+            // (dioxus-signal-safety: direct writes here are the Firefox
+            // mobile RefCell re-entrancy crash path).
             onclick: move |_| {
-                CREATE_ROOM_MODAL.with_mut(|modal| {
-                    modal.show = false;
+                crate::util::defer(move || {
+                    CREATE_ROOM_MODAL.with_mut(|modal| {
+                        modal.show = false;
+                    });
                 });
             }
         }
@@ -211,9 +216,13 @@ pub fn CreateRoomModal() -> Element {
                     button {
                         "data-testid": "create-room-cancel-button",
                         class: "px-4 py-2 text-sm text-text-muted hover:text-text hover:bg-surface rounded-lg transition-colors",
+                        // Deferred close — same signal-safety rule as the
+                        // backdrop handler above.
                         onclick: move |_| {
-                            CREATE_ROOM_MODAL.with_mut(|modal| {
-                                modal.show = false;
+                            crate::util::defer(move || {
+                                CREATE_ROOM_MODAL.with_mut(|modal| {
+                                    modal.show = false;
+                                });
                             });
                         },
                         "Cancel"
--- a/ui/src/components/room_list/edit_room_modal.rs
+++ b/ui/src/components/room_list/edit_room_modal.rs
@@ -62,8 +62,13 @@ pub fn EditRoomModal() -> Element {
                 // Overlay
                 div {
                     class: "absolute inset-0 bg-black/50",
+                    // Signal mutation from an event handler must be deferred
+                    // (dioxus-signal-safety: direct writes here are the Firefox
+                    // mobile RefCell re-entrancy crash path).
                     onclick: move |_| {
-                        EDIT_ROOM_MODAL.write().room = None;
+                        crate::util::defer(move || {
+                            EDIT_ROOM_MODAL.write().room = None;
+                        });
                     }
                 }
                 // Modal content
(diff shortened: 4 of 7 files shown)


sanity:
## Review (Full tier)

Tier: **Full** — the diff touches `.github/workflows/build.yml`, which
`~/.claude/rules/multi-model-review.md` puts on the always-Full list
(CI configuration), and it changes signal-handling semantics across four UI
files.

Lenses run sequentially in-session (no `pr-review` skill in this session,
subagents not requested): **code-first**, **testing**, **skeptical**,
**big-picture**, plus the change-specific lens, **defer-ordering semantics**.
External non-Claude reviewers not invoked — opt-in per the rule.

### Finding, fixed in 9658c2f

**The new source scan had a hole (testing lens).** It only inspected INLINE
closures, so a handler passed by name — `onclick: handle_close,` — was
invisible to it. There are 20+ such handlers in the UI. A guard that misses the
most common alternative spelling is not a guard.

Extended to resolve each named handler to its `let <name> = move |…| { … }`
binding and scan that body. Checked all of them: **no current violations**, so
this adds coverage rather than fixes. Mutation-tested — removing the `defer`
from `dm_thread_modal.rs`'s `close` makes it fail with
``dm_thread_modal.rs:411 (handler `close`)``.

An earlier draft also had a **false positive**: for a brace-less handler
(`onclick: move |_| show_confirm.set(true),`) it took "the next `{`", swallowing
an unrelated block. Caught by mutation testing, fixed by parsing the closure
properly, and both body shapes now have their own mutation test.

### Checked, no change needed

- **Defer ordering.** `defer()` runs its closure on a later tick, so code after
  it runs FIRST. Only one of the seven sites has more than one statement — the
  Dismiss handler — and both of its statements go inside the defer, preserving
  order. It also needs a `clone()`, since an `onclick` is `FnMut` and the
  invitation is not `Copy`. The other six are single writes with no captures.
- **`dismiss_invitation_persistently` writes a signal too** (`invitation.set(None)`).
  Deferring the whole handler fixes that write as well, rather than leaving it
  as an undeferred write one call deeper.
- **Local `use_signal` handles are deliberately NOT flagged.** They have no
  external subscribers, and the existing copy buttons set them directly in
  `onclick`. Flagging them would fight the codebase for no safety gain.
- **Deferring the create-room modal's OPEN** (not just closes) — `show = true`
  is idempotent, so a double click is harmless, and the create-room specs pass.
- **`if: always()` on the artifact upload** — pairs with
  `if-no-files-found: ignore`, so a clean run is a no-op rather than a warning.

### On #538 specifically

I filed #538 myself with a timeout hypothesis. **That hypothesis is wrong**, and
this PR says so rather than shipping a fix built on it. Measured margin is ~500x
on the assertion that failed, and 58 runs could not reproduce it. Bumping the
timeout would have looked like a fix and fixed nothing. #538 stays open with the
measurements recorded; the trace change means the next occurrence produces an
artifact instead of another guess.

### Verification

- `cargo test -p river-ui --bins` — 781 passed.
- Full Playwright suite — 682 passed, 23 skipped, **0 flaky**, all five projects.
  Deferring modal open/close changes their timing, so this was the real risk.
- All three scan branches (braced, expression-bodied, named) mutation-tested;
  each fails with the exact offending file:line.
- Source confirmed free of mutation residue before commit.

[AI-assisted - Claude]

Issue #557, opened 2026-07-29
fix(ui): a background invitation-accept steals CURRENT_ROOM, silently switching the room the reader is viewing


sanity:
## Symptom

The room you are reading silently changes to a different room, with no click and no reload. Messages keep arriving in your original room off-screen, so it reads as "the chat stopped and no new messages loaded" — and the unread badge on the room you *were* in starts climbing.

## Root cause

`get_response.rs` (pending-invite branch, ~line 766) sets `CURRENT_ROOM` unconditionally when an invitation-accept's PUT+subscribe completes:

```rust
crate::util::defer(move || {
    PENDING_INVITES.with_mut(|pending| { ... status = PendingRoomStatus::Subscribed; ... });
    CURRENT_ROOM.with_mut(|current_room| {
        current_room.owner_key = Some(owner_vk);   // no check of what the reader is viewing
    });
});
```

Focusing a room the user just chose to join is reasonable. The problem is that this path is **also reached without any user action**: #218 wires a click-free auto-resume at startup (`app.rs:346-382`). If localStorage still holds an invitation plus nickname — i.e. a previous session clicked Accept but the accept never completed — the app silently re-issues it on load. `receive_invitation_modal.rs:960-978` documents why: *"an accept whose room-contract GET never completed (intermittently, e.g. freenet-core #4345's large multi-fragment GET)... Real users hit exactly this on the official Freenet River room."*

So a stale accept against a large, slow room can land minutes into the session and yank the reader out of whatever they were reading. `DocumentTitleUpdater`'s `use_effect` on `CURRENT_ROOM` (`document_title.rs:611-624`) then marks only the *new* current room as read, so the abandoned room starts accruing an unread badge while its messages arrive invisibly.

Every other `CURRENT_ROOM` writer is either a direct user action (room click, create, import identity, leave, notification click) or already gated: the delegate-restore path checks `CURRENT_ROOM.read().owner_key.is_some()` via `decide_current_room_restore` (`response_handler.rs:1948-1966`, `SkipUserAlreadySelected`). This site is the only unguarded one.

## Fix

Focus the newly-joined room only when nothing is currently selected, mirroring the `SkipUserAlreadySelected` discipline already used by the delegate-restore path. A background accept should still join the room and update its unread state; it should not steal the view.

## Also worth hardening (not fixed here)

[...]

PR #558, opened 2026-07-29
fix(ui): stop three mechanisms that silently freeze the conversation

### LINKED ISSUE #555: fix(ui): contended ROOMS.try_read permanently latches the conversation memo, rendering "No messages yet" until reload
## Symptom

A River tab that has been open a while stops showing new messages, and the fix is a page reload. Reported on Matrix 2026-07-29 ("the chat stopped and no new messages were loaded until i reloaded the site"), and seen before by @sanity and others. This is the symptom river#382 was closed for on 2026-07-06; #383's WebSocket watchdog did not stop it.

In the worst form the conversation does not merely stop updating — it renders **"No messages yet. Start the conversation!"** in a room full of messages.

## Root cause

`message_groups` (`ui/src/components/conversation.rs:2563`) is a `use_memo` whose closure early-returns when the `ROOMS` borrow is contended:

```rust
let message_groups = use_memo(move || {
    let current_room = CURRENT_ROOM.read();          // subscribes
    if let Some(key) = current_room.owner_key {
        let Ok(rooms) = ROOMS.try_read() else { return None; };   // <-- no subscription on Err
        ...
    }
    None
});
```

Three facts from the Dioxus 0.7.9 source make this a **permanent latch**, not a one-frame blip:

1. **A memo clears its dependencies on every recompute.** `Memo::new_with_location` recomputes via `rc.reset_and_run_in(&mut f)` (`dioxus-signals-0.7.9/src/memo.rs:62`), and `reset_and_run_in` is `clear_subscribers()` then `run_in(f)` (`dioxus-core-0.7.9/src/reactive_context.rs:195-198`). The dependency set is rebuilt from whatever the closure actually read during that run.

2. **`try_read()` does not subscribe on the `Err` path.** `Signal::try_read_unchecked` propagates the borrow failure with `?` *before* reaching `reactive_context.subscribe(...)`:
   ```rust
   let inner = self.inner.try_read_unchecked()?;                 // signal.rs:409 — returns here
   if let Some(reactive_context) = ReactiveContext::current() {
       reactive_context.subscribe(inner.subscribers.clone());    // signal.rs:413 — never reached
   }
   ```

3. So a contended recompute leaves the memo subscribed **only to `CURRENT_ROOM`** (read via `.read()` before the early return). `RECEIVE_TIMES.read()` sits after the `try_read` and is never reached on that path.

Net effect: one contended recompute and the memo is deaf to `ROOMS` forever. It re-evaluates only when `CURRENT_ROOM` changes (switch rooms) or the component remounts (reload) — which is exactly the reported "reload fixes it".

The `None` is then rendered as the empty state at `conversation.rs:4229-4232`, so a populated room reads as an empty one.

## Why contention happens routinely

Per `.claude/rules/dioxus-signal-safety.md`, a signal write's Drop handler fires subscriber notifications **synchronously**, and those notifications poll memos that `try_read()` the signal still being written. That is the documented hazard the `try_read()` idiom exists for, and it is what produced river#499 ("DM rail collapses to empty or shows archived threads under signal contention") and is the same family as river#397 ("room list shows blank ... reads as 'all my rooms are gone'").

#499 was fixed for the DM rail with a `LAST_GOOD_RAIL` cache (`dm_rail_section.rs:726`). The conversation never got the equivalent, and a last-good cache alone does not clear the latch — it stops the blank render but leaves the memo unable to wake.

## Proposed fix

Two parts, both needed:

1. **Never render the empty state on contention.** Fall back to the last successfully computed groups, mirroring `LAST_GOOD_RAIL`.
2. **Restore a wake-up path.** Because `try_read()` cannot subscribe on `Err`, the memo needs a dependency it always reads before any early return, plus a deferred retry that bumps it, so a contended pass re-evaluates instead of latching.

## Testing

The repo currently has no test that builds a `VirtualDom`, so this reactive behaviour is untested. A regression test should reproduce the latch (contended recompute, then assert the memo still observes a later `ROOMS` change) rather than only pinning the fix shape.

Also worth auditing: every other `use_memo` in `ui/src/` that early-returns before reading a signal it depends on has the same latch shape.

[AI-assisted - Claude]

### LINKED ISSUE #556: fix(ui): after a WebSocket reconnect the UI never re-subscribes rooms it already believes are Subscribed
## Symptom

A River tab reconnects after a transient network interruption (laptop sleep/wake, WiFi blip, or the #383 liveness watchdog declaring the socket dead), shows **Connected**, and then never receives another room update for the rest of the session. Only a page reload restores it.

This is the reported symptom of river#382, which #383's watchdog did not stop — and the watchdog is implicated rather than protective here, because *it is what triggers the reconnect*.

## Root cause

A client subscription lives on the node against a `ClientId`, and the node tears down **every** subscription belonging to that client on each socket exit path — `websocket.rs::notify_disconnect` (12 call sites) → `client_events.rs::remove_client_from_all_subscriptions`. A reconnect opens a new socket and therefore gets a **fresh `ClientId`**, so immediately after a reconnect the node holds no client subscription for any room.

The UI does not re-subscribe them:

- `SYNC_INFO` is a page-lifetime `GlobalSignal` (`ui/src/components/app/sync_info.rs`) that survives the reconnect and still reports every room as `Subscribed`.
- `rooms_awaiting_subscription()` (`sync_info.rs:225-293`) only returns rooms whose status is `Disconnected`, or a `Subscribing` room that has timed out. A `Subscribed` room is invisible to it.
- Nothing on the reconnect path resets room status. The `Connect` handler (`freenet_synchronizer.rs`) calls `reset_ensure_subscription_dedup()`, but that re-arms the **chat delegate's** subscription (owner-side private-room secret rotation), not this client's own per-room subscriptions. The only sites that set `Disconnected` are `subscribe_response.rs:49`, `get_response.rs:252`, `get_response.rs:1114` and `room_synchronizer.rs:653` — none on the reconnect path.

So `process_rooms()` runs after the reconnect and finds nothing to do. The tab is connected, believes it is subscribed, and is subscribed to nothing. A reload cures it only because it starts with an empty `SYNC_INFO`.

The comments at `freenet_synchronizer.rs:283` and `connection_watchdog.rs:15,38` describing the reconnect as a "re-subscribe" path are accurate only for rooms *not* already marked `Subscribed`.

## Fix

On a genuinely new socket, flip rooms that believe they are `Subscribed` back to `Disconnected` so the existing `process_rooms()` PUT+subscribe path re-runs.

Deliberately scoped:
- Only on `Connect` (a new socket ⇒ new `ClientId`), **not** on the `PageBecameVisible`/`RefreshAllRooms` path, which runs on an apparently-live socket where the node's client subscriptions are still intact. Re-subscribing there would ship full room state on every tab focus.
- Only `Subscribed` is flipped: `Subscribing` keeps its timeout bookkeeping, and terminal `Error` is not silently revived past its #290 retry budget.

## Related but out of scope

Two adjacent routes to the same "no live subscription" state, neither fixed here:

- **Node-side, deliberate:** `ring/hosting.rs::teardown_evicted_in_use_contract` clears `client_subscriptions` on last-resort eviction and its own comment says this *"silently STRANDS the local client... there is deliberately no client-notification surface"*. It also leaves `RuntimePool::shared_notifications` in place, so the client's channel lingers with nothing feeding it. Needs a freenet-core issue.
- **A periodic UI re-subscribe would NOT work** for the renewal-failure route, and is worth recording so nobody builds it: `is_receiving_updates` is `is_subscribed() || has_client_subscriptions()` (`hosting.rs:2348`), so while the client is still registered the GET handler takes `should_serve_local_copy` and answers from the node's own possibly-stale copy, and the re-subscribe kick is gated `if !is_subscribed` and never fires. The node already retries that route itself every 30s (`recover_orphaned_subscriptions`, branch 2 of `contracts_needing_renewal`).

[AI-assisted - Claude]

sanity:
## Problem

A River tab left open stops showing new messages and only a page reload fixes it. Reported on Matrix 2026-07-29 ("the chat stopped and no new messages were loaded until i reloaded the site"), and seen before by @sanity and others. This is the symptom river#382 was closed for on 2026-07-06; #383's WebSocket liveness watchdog did not stop it.

Investigating produced **three independent mechanisms**, each of which on its own presents as "the chat stopped". **Two are fixed here (#555, #556).** The third (#557) was implemented, then dropped during review and left open -- see "Dropped from this PR".

### 1. A contended `try_read()` permanently latches a memo (#555)

`message_groups` (`conversation.rs`) early-returns when the `ROOMS` borrow is contended. Two facts from the Dioxus 0.7.9 source make that permanent rather than transient:

- A memo **rebuilds** its dependency set every pass: `Memo` recomputes via `rc.reset_and_run_in(&mut f)` (`dioxus-signals/src/memo.rs:62`), which is `clear_subscribers()` then `run_in(f)` (`dioxus-core/src/reactive_context.rs:195-198`).
- `try_read()` **does not subscribe on `Err`**: `Signal::try_read_unchecked` propagates the borrow failure with `?` at `signal.rs:409`, before reaching `reactive_context.subscribe(...)` at `:413`.

So one contended pass leaves the memo subscribed only to whatever it *did* read — and it then never reacts to `ROOMS` again. The conversation memo kept its `CURRENT_ROOM` subscription, so it recovered on a room switch; worse, it rendered **"No messages yet. Start the conversation!"** into a room full of history while latched.

Two sites were strictly worse — `member_info_modal.rs` (`self_member_id`) and `dm_thread_modal.rs` (`view`) read `ROOMS` *first*, so a contended pass left them with **zero** subscriptions, and both are mounted for the whole app session.

**What holds the borrow is not diagnosed.** An earlier revision of this PR repeated the explanation from `.claude/rules/dioxus-signal-safety.md` (a write's Drop firing notifications that synchronously poll memos). Three reviewers independently disproved it for 0.7.9: `WriteLock`'s borrow field drops before its `SignalSubscriberDrop` (`write.rs:164`), `update_subscribers` takes a fresh read (`signal.rs:260`, which would panic on every write otherwise), and `mark_dirty` on a memo only sets a flag and sends on a channel (`memo.rs:53`). So the notification path is not the contending writer.

Contention is still real and observed -- `signal.rs:262` notes "mark_dirty can run user code", and #397/#499 are the user-visible instances -- and a single `Err` latches permanently, so the guard is worth having regardless. But the trigger is **inferred**, and naming the contending writer remains open work (#560).

### 2. A reconnect leaves the tab subscribed to nothing (#556)

Client subscriptions live on the node against a `ClientId`, and the node tears down every subscription for that client on each socket exit (`websocket.rs::notify_disconnect` → `remove_client_from_all_subscriptions`). A reconnect gets a fresh `ClientId`, so afterwards the node holds no client subscription for any room.

The UI never re-subscribes: `SYNC_INFO` is a page-lifetime signal that still says `Subscribed`, and `rooms_awaiting_subscription()` only returns `Disconnected` (or timed-out `Subscribing`) rooms. `reset_ensure_subscription_dedup()` on the `Connect` path re-arms the chat *delegate's* subscription, not this client's per-room ones.

Net: the tab reconnects, shows **Connected**, and receives nothing for the rest of the session. The #383 watchdog is implicated rather than protective, since declaring the socket dead is precisely what triggers the reconnect.

### 3. A background invitation-accept steals the view (#557) — DIAGNOSIS ONLY, fix dropped

`get_response.rs` set `CURRENT_ROOM` unconditionally when an accept's PUT+subscribe completed. #218 also wires a **click-free auto-resume** at startup, so a stale stored accept against a large room can land minutes in and yank the reader elsewhere. Their original room keeps receiving messages off-screen, and its unread badge climbs. Every other `CURRENT_ROOM` writer is a user action or already gated by `SkipUserAlreadySelected`.

## Approach

**#555** adds `util::signal_guard`, generalising the three-part pattern `dm_rail_section.rs` grew for #499: an infallible `anchor()` read before any fallible read (so a memo can never come out of a contended pass with zero subscriptions), and a deferred `schedule_nudge()` on every degraded branch (so it re-polls a macrotask later instead of waiting for an unrelated signal write). Applied to all **12** fallible memos across 8 files (`conversation.rs` alone has 4; an earlier revision of this description said eight, conflating memos with files). `dm_rail_section.rs` keeps its own working `RAIL_REBUILD_TICK` — rewiring a working recovery path for tidiness would be a needless risk.

Deliberately **not** included: a last-good cache for `message_groups`. The nudge bounds the empty state to one macrotask instead of a whole session, which is the bulk of the harm. Reviewers correctly pushed back on my stated reason (deep-clone cost): an `Rc`-wrapped memo value makes the cache cheap, so the honest position is that this is deferred scope, not an impossibility. And "one macrotask" bounds duration, not visibility — the `None` render commits before the `setTimeout(0)` fires, so a brief flash of "No messages yet" (and of an empty room list, #397's symptom) is possible. Filed as #562.

**#556** flips `Subscribed` → `Disconnected` on a genuinely new socket so the existing `process_rooms()` path re-subscribes, and arms an independent `ProcessRooms` retry.

The retry is load-bearing, not belt-and-braces: `rooms_awaiting_subscription()` returns an empty map when `ROOMS.try_read()` is contended, on the assumption it will be called again — and on the reconnect path nothing calls it again (the `process_rooms()` call discards its result, every other sender is driven by a user edit or a response, and with nothing subscribed no response arrives). The watchdog cannot rescue it either, since its probe GET succeeds on a healthy socket. Without the retry this fix could fail exactly like the bug it fixes.

**Two costs of #556, disclosed because reviewers rightly asked:** re-subscribing goes through `ContractRequest::Put { subscribe: true }`, which ships **full room state per room, per reconnect** — the same cost this PR cites to justify excluding the `PageBecameVisible` path. `ContractRequest::Subscribe` would be the semantically matching repair, but a bare Subscribe is rejected when the contract WASM is not cached locally (pinned by `issue_287_loaded_rooms_do_not_bare_subscribe`), so switching needs its own analysis: filed as #561. Separately, `needs_to_send_update()` only returns `Subscribed` rooms, so outbound UPDATEs are suspended for the reconnect window — also #561. Scoped tightly: only on `Connect`, never on `PageBecameVisible`/`RefreshAllRooms` (that runs on a live socket where the node's subscriptions are intact, and re-subscribing there would ship full room state on every tab focus). Only `Subscribed` is flipped, so `Subscribing` keeps its timeout bookkeeping and terminal `Error` is not revived past its #290 retry budget.

### Rejected approach, recorded so nobody rebuilds it

Making the watchdog probe re-subscribe periodically **cannot work**. `is_receiving_updates` is `is_subscribed() || has_client_subscriptions()` (`hosting.rs:2348`), so while the client is still registered the node takes `should_serve_local_copy` and answers from its own possibly-stale copy, and the re-subscribe kick is gated `if !is_subscribed` and never fires. The node already retries that route every 30s itself (`recover_orphaned_subscriptions`).

## Testing

- **#556** is covered by real behavioural unit tests in `sync_info.rs`: a `Subscribed` room is re-armed, `Subscribing`/`Error`/`Disconnected` are left alone (including that a `Subscribing` room keeps its timeout stamp and a terminal `Error` room is not revived), and re-arming spends no #290 retry budget.
- **#555** is covered by source-scrape pins in `util/signal_guard.rs`, matching the idiom `dm_rail_section.rs` already uses for this exact class (`rail_builders_read_infallible_guard_before_first_fallible_read`) — exercising the real behaviour needs a Dioxus runtime and a rendered tree, and no test in this repo builds a `VirtualDom`. The pins assert, **per memo body** (not per file, so one anchor cannot appear to cover a file with several memos), that every fallible memo reads the anchor *before* its first `try_read` and nudges on contention; that the count is **exactly 12**, so the matcher going blind fails loudly rather than silently; that `anchor()` uses `read()` not `peek()` (a `peek` anchor would compile, look right, and protect nothing); and that `schedule_nudge` writes the very signal `anchor` reads — deleting that one line reverted all 12 guards to permanent latches with a fully green suite before that pin existed. All four are mutation-verified.

**Why CI did not catch these:** reactive-subscription behaviour has no test harness in this repo at all, so the whole class was invisible. The `signal_guard` pins close it for the pattern going forward; a `VirtualDom`-based harness that could test the behaviour directly is worth building separately.

## Dropped from this PR

**#557 (invitation-accept stealing `CURRENT_ROOM`) was implemented and then removed.** All three reviewers found the guard regresses the *deliberate* Accept, not just the background re-issue: both callers funnel through `accept_invitation` and are indistinguishable at the GET handler, and the delegate-restore path writes `CURRENT_ROOM` at startup, so any returning user has `owner_key.is_some()` before an invitation lands. Since `get_response.rs:766` is the only path that navigates to a newly accepted room, the result would have been: click Accept, room joins, modal closes, you stay in the old room. It also had no test, and the underlying diagnosis was explicitly unattributed. It needs a `user_initiated` flag threaded from the modal plus the pure-decision-function treatment `decide_current_room_restore` already has.

## Known gaps, filed not hidden

- **#559** — `use_effect` has the identical latch and 7 sites are unguarded, 4 of them single-dependency (so zero-subscription on contention). One is 24 lines from a memo this PR fixes. #555 explicitly asked for this audit.
- **#560** — name the contending writer, and replace the source-scrape pins with a native `VirtualDom` test. Reviewers showed such a test IS feasible (`dioxus_core` is not wasm-gated; `defer` runs inline on native), so "no test in this repo builds a VirtualDom" is a weaker claim than I made.
- **#561** — #556's full-state PUT per reconnect, and the suspended-outbound-UPDATE window.
- **#562** — `Rc` last-good caches to remove the one-frame blank.
- **#563** — `GUARDED_MEMO_SITES` is a hardcoded 8-file list that rots silently; walk `ui/src` at test time instead. Also consolidate `dm_rail_section`'s parallel tick, and collapse the per-site boilerplate into a `guarded_read` helper that makes the ordering structural rather than pin-enforced.

Closes #555
Closes #556

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/dioxus-signal-safety.md
+++ b/.claude/rules/dioxus-signal-safety.md
@@ -22,16 +22,61 @@ let Ok(rooms) = ROOMS.try_read() else { return; };
 ```
 
 **IMPORTANT:** In Dioxus 0.7.x, `try_read()` does NOT register signal
-subscriptions when it returns `Err`. The subscription is registered only
-on the success path (after the borrow succeeds). This means a `use_memo`
-that hits `try_read() -> Err` will NOT be notified of future signal
-changes — it permanently stops re-evaluating.
+subscriptions when it returns `Err` — `Signal::try_read_unchecked`
+propagates the borrow error with `?` (`signal.rs:409`) before reaching
+`reactive_context.subscribe(...)` (`:413`). And a memo/effect REBUILDS its
+dependency set on every pass: `reset_and_run_in` is `clear_subscribers()`
+then `run_in(f)` (`dioxus-core` `reactive_context.rs:195`). So a pass that
+early-returns on `Err` ends subscribed only to what it actually read — and
+if that is nothing, to nothing at all. It then never re-evaluates.
 
-To mitigate: ensure signal mutations happen in clean execution contexts
-(via `crate::util::defer()`) so `try_read()` never encounters a
-concurrent borrow. Also, memos that read multiple signals (e.g.,
-`CURRENT_ROOM.read()` + `ROOMS.try_read()`) get a backup subscription
-from the non-try signal.
+### Required: anchor + nudge for every fallible read (freenet/river#555)
+
+The two mitigations previously listed here are **not sufficient**, and
+relying on them is what produced #397, #499 and #555:
+
+- `defer()`-ing mutations does not prevent contention. It was already used
+  throughout when all three bugs happened.
+- The "backup subscription from the non-try signal" only downgrades the
+  failure from *permanent* to *stuck until that other signal changes*. For
+  `message_groups` the backup was `CURRENT_ROOM`, so the conversation froze
+  until the reader switched rooms or reloaded.
+
+Use `crate::util::signal_guard` instead. Read the anchor FIRST, before any
+fallible read, and nudge on every degraded branch:
+
+```rust
+let thing = use_memo(move || {
+    crate::util::signal_guard::anchor();          // before any try_read
+    let Ok(rooms) = ROOMS.try_read() else {
+        crate::util::signal_guard::schedule_nudge();
+        return None;
+    };
+    ...
+});
+```
+
+This applies to **`use_effect` as well as `use_memo`** — `use_effect` uses
+the same `reset_and_run_in` primitive (`dioxus-hooks` `use_effect.rs:34`),
+so an effect whose only read is a contended `try_read` is dead for the
+session. A memo/effect with a SECOND fallible read needs a nudge on that
+branch too; the anchor keeps the memo alive but does not re-subscribe the
+signal that failed.
+
+Still read an infallible signal before the fallible one where you can. It
+costs nothing and leaves a backup if the nudge channel is ever broken.
+
+`signal_guard`'s pin test enforces the ordering for the memos it lists, but
+it only scans `use_memo(` bodies in an explicit file list — it cannot see
+effects, helper functions, or unlisted files. Do not treat a green pin as
+proof your new site is covered.
+
+What actually holds a borrow across a recompute is **not diagnosed**. The
+explanation this file used to give (a write's Drop firing notifications that
+synchronously poll memos) does not hold in 0.7.9: the borrow is released
+before subscribers are marked, and `mark_dirty` on a memo only sets a flag
+and sends on a channel. Contention is nonetheless observed, and one `Err` is
+enough to latch, so the guard is required regardless.
 
 ## Never call `spawn_local` inside a polled future
 
--- a/.claude/rules/dioxus-signal-safety.md
+++ b/.claude/rules/dioxus-signal-safety.md
@@ -6,10 +6,17 @@ globs:
 
 # Dioxus WASM Signal Safety Rules
 
-The UI runs as single-threaded WASM. Firefox mobile runs Dioxus signal
-subscriber notifications synchronously during Drop, causing
-`RefCell already borrowed` panics. These rules prevent re-entrant borrow
-crashes.
+The UI runs as single-threaded WASM, where a re-entrant signal borrow is a
+`RefCell already borrowed` panic rather than a benign contention. These rules
+prevent that, and prevent the subscription latch below.
+
+Note on the mechanism: earlier revisions of this file said Dioxus fires
+subscriber notifications synchronously during a write guard's Drop, with the
+borrow still held. That is not what dioxus 0.7.9 does — `WriteLock`'s borrow
+field drops before its `SignalSubscriberDrop` (`write.rs:164`),
+`update_subscribers` takes a fresh read (`signal.rs:260`), and `mark_dirty` on
+a memo only sets a flag and sends on a channel (`memo.rs:53`). The rules still
+hold, but for the reasons stated at each rule, not that one.
 
 ## Always use `try_read()` for reactive signal reads
 
@@ -33,7 +40,9 @@ if that is nothing, to nothing at all. It then never re-evaluates.
 ### Required: anchor + nudge for every fallible read (freenet/river#555)
 
 The two mitigations previously listed here are **not sufficient**, and
-relying on them is what produced #397, #499 and #555:
+relying on them is what produced #499 and #555 (#397 is the same *blank
+render* family but a different cause — a missing loading state — so it is
+not evidence for this one):
 
 - `defer()`-ing mutations does not prevent contention. It was already used
   throughout when all three bugs happened.
@@ -98,11 +107,15 @@ Signal mutations (`ROOMS.with_mut()`, `ROOMS.write()`,
 synchronous event handlers (`onclick`, etc.). This is required for TWO
 reasons:
 
-1. **RefCell re-entrancy**: Signal write Drop handlers fire subscriber
-   notifications synchronously. Those notifications poll memos that call
-   `try_read()` on the same signal — panics if the write guard's
-   RefCell borrow is still held. `setTimeout(0)` breaks the call stack
-   so no borrows are active.
+1. **RefCell re-entrancy**: a mutation run from a handler or a polled
+   future can be re-entered while an outer borrow of the same signal is
+   still live — `mark_dirty` explicitly "can run user code"
+   (`signal.rs:262`), and this repo has an observed instance
+   (`process_rooms()` being driven during a write-guard drop, see
+   `sync_info.rs::rooms_awaiting_subscription`). A re-entrant borrow is a
+   panic here, not a soft failure. `setTimeout(0)` breaks the call stack so
+   no borrows are active. (This is NOT the debunked "Drop notifies
+   synchronously while holding the borrow" story — see the note at the top.)
 
 2. **Missing Dioxus scope**: `wasm_bindgen_futures::spawn_local` tasks
    run without a Dioxus scope on the `scope_stack`. Signal subscriber
--- a/ui/src/components/app/freenet_api/freenet_synchronizer.rs
+++ b/ui/src/components/app/freenet_api/freenet_synchronizer.rs
@@ -607,6 +607,40 @@ impl FreenetSynchronizer {
                                              subscription the node dropped with the old socket",
                                             restaled
                                         );
+                                        // The flip alone is not enough. Only
+                                        // `process_rooms()` acts on it, and its
+                                        // `rooms_awaiting_subscription()` returns an EMPTY
+                                        // map when `ROOMS.try_read()` is contended, on the
+                                        // assumption that "we'll be called again on the
+                                        // next ProcessRooms". On the reconnect path there
+                                        // is no next one: the `process_rooms()` call below
+                                        // discards its result, and every other
+                                        // `ProcessRooms` sender is driven by a user edit or
+                                        // by a response — and with nothing subscribed, no
+                                        // response arrives. The watchdog cannot rescue it
+                                        // either, because its probe GET succeeds on a
+                                        // healthy socket so it never declares death.
+                                        //
+                                        // Losing that one read would leave the tab
+                                        // connected and subscribed to nothing: the exact
+                                        // bug this fix is for, wearing a different status
+                                        // label. So arm an independent retry.
+                                        let retry_tx = message_tx.clone();
+                                        crate::util::safe_spawn_local(async move {
+                                            crate::util::sleep(std::time::Duration::from_millis(
+                                                super::constants::REPUT_DELAY_MS + 1000,
+                                            ))
+                                            .await;
+                                            if let Err(e) = retry_tx
+                                                .unbounded_send(SynchronizerMessage::ProcessRooms)
+                                            {
+                                                warn!(
+                                                    "Failed to arm post-reconnect \
+                                                     re-subscribe retry: {}",
+                                                    e
+                                                );
+                                            }
+                                        });
                                     }
 
                                     // Set up the chat delegate to load rooms from storage
--- a/ui/src/components/app/freenet_api/freenet_synchronizer.rs
+++ b/ui/src/components/app/freenet_api/freenet_synchronizer.rs
@@ -583,6 +583,32 @@ impl FreenetSynchronizer {
                                     // load-rooms path, gated on a non-Failed signing-key
                                     // migration).
                                     reset_ensure_subscription_dedup();
+
+                                    // A new socket means a new ClientId, and the node
+                                    // dropped every client subscription belonging to the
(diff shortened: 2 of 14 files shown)


sanity:
## Review summary

Ran the Full tier: three independent Claude lenses, each blind to the others, each reading the checked-out code. No external (non-Claude) model pass — per the standing rule that is opt-in only and was not requested for this change.

Lenses: **skeptical** (adversarial bug hunt), **testing + big-picture** (coverage, goal alignment, anti-patterns, stale docs), **code-first + simplification** (read code before description, then diff against it).

All three independently verified the two Dioxus facts the fix rests on, to the exact line — `reactive_context.rs:195` clearing subscribers per pass, and `signal.rs:409` returning before `:413`'s `subscribe`. Both hold.

### Acted on

| Finding | Reviewers | Action |
|---|---|---|
| #557 guard regresses the deliberate user-clicked Accept | 3/3 (blocking) | **Fix dropped from this PR.** #557 stays open. |
| Stated contention mechanism is false for 0.7.9 | 3/3 | Corrected in module doc and PR body; trigger now labelled inferred. #560 filed. |
| Deleting `schedule_nudge`'s tick write left the suite green | 2/3 | New mutation-verified pin. |
| Vacuity floor `>= 8` vs 12 real memos | 3/3 | Now `assert_eq!(checked, 12)`. |
| #556 one-shot: a contended read strands it, same symptom | 1/3 | Independent `ProcessRooms` retry armed. |
| Two secondary fallible reads without nudges | 1/3 | Nudges added (`OUTBOUND_DMS`, `SYNC_INFO`). |
| `member_info_modal` read-order inversion dropped a backup subscription | 3/3 | `CURRENT_ROOM`-first restored. |
| `dioxus-signal-safety.md` stale and actively misleading | 1/3 | Rewritten: anchor+nudge required, `use_effect` named, pin limits stated. |
| #556 full-state PUT cost undisclosed | 3/3 | Disclosed in the PR; #561 filed. |
| Wrong comment; PR miscounts memos; PR invents a pin mechanism | 3/3 | All corrected. |

### Deferred, with issues rather than silence

#559 `use_effect` has the identical latch (7 sites, 4 zero-subscription) · #560 name the contending writer + a real `VirtualDom` test (reviewers showed it IS feasible, so my "not attempted" was too strong a claim) · #561 full-state PUT + suspended outbound UPDATEs · #562 `Rc` last-good to remove the one-frame blank · #563 make the invariant structural (`guarded_read`, directory-walking pin, consolidate `dm_rail_section`'s parallel tick)

### One reviewer finding rejected

The two `.syncthing.*.tmp` files under `room_list/` were reported as untracked strays. They are **tracked on `origin/main`**. Deleting tracked files is out of scope here.

### Status

The diff changed materially after review, so **this content has not been re-reviewed** and auto-merge is deliberately not set. Re-review of `3f47ab9e` should focus on the #556 retry (new control flow in the message loop) and the dropped #557 revert.

PR #565, opened 2026-07-30
fix(ui): stop re-renders wiping in-progress typing in room dialogs

### LINKED ISSUE #564: fix(ui): editable fields are wiped mid-typing by unrelated re-renders (volatile `value` + no `oninput`)
Reported by Ian: editing the **Room Description** in the Room Details dialog, the field cleared itself every so often (roughly every 30 seconds) while typing.

## Mechanism

Three facts combine:

1. `value` on `input` / `textarea` / `select` is declared **volatile** in dioxus-html (`dioxus-html-0.7.9/src/elements.rs:1488,1571,1594`).
2. dioxus-core writes volatile attributes to the DOM on **every** re-render, even when the rendered string has not changed: `if volatile || attribute_changed { self.write_attribute(...) }` (`dioxus-core-0.7.9/src/diff/node.rs:463`).
3. The interpreter then assigns the value whenever the *live DOM* value differs from the VDOM value: `case "value": ... else if (node.value !== value) node.value = value` (`dioxus-interpreter-js-0.7.9/src/ts/set_attribute.ts:31-33`).

So for any field whose bound signal is updated only in `onchange` (which fires on blur/commit, not per keystroke), the signal still holds the **pre-typing** text while the user types. Any re-render in that window sees `node.value !== value` and resets the DOM node, discarding the in-progress edit.

`RoomDescriptionField` reads `CURRENT_ROOM` and `ROOMS` in its render body, so it is subscribed to both and re-renders on every room-state write. That is what supplies the periodic trigger the report describes.

The sibling `RoomNameField` was never affected because it already has `oninput` (added for the Enter-to-commit handler in #21), which keeps the signal equal to the DOM value and makes the re-write a no-op.

## Affected fields

Audited every `input` / `textarea` / `select` in `ui/src`. Editable fields binding `value:` with no `oninput:`:

- `ui/src/components/room_list/edit_room_modal.rs` `RoomDescriptionField` (Room Description) — the reported one
- `ui/src/components/room_list/edit_room_modal.rs` `NumericConfigField` — renders 6 fields (Max Recent Messages, Max Message Size, Max User Bans, Max Nickname Size, Max Room Name, Max Room Description)
- `ui/src/components/room_list/edit_room_modal.rs` `MaxMembersField` — also re-renders on its `member_count` prop, so a member joining mid-edit was enough
- `ui/src/components/room_list/create_room_modal.rs` Room Name (its sibling nickname input already used `oninput`)

Read-only display fields that bind `value:` without `oninput` (invite link, invite code, public key, contract ID, member ID, export token) are not affected: nothing types into them.

The private-room checkbox binds `checked:`, which is **not** volatile, and is updated by `onchange` on click, so it stays in sync.

## Fix

Add `oninput` to each affected field so the bound signal tracks the live value, and add a source-scrape pin test so a new field cannot reintroduce the class.

[AI-assisted - Claude]

sanity:
Closes #564

## Problem

Editing the **Room Description** in the Room Details dialog, the field cleared itself every few seconds while typing (reported by Ian).

Three facts combine:

1. `value` on `input` / `textarea` / `select` is declared **volatile** in dioxus-html (`dioxus-html-0.7.9/src/elements.rs:1488,1571,1594`).
2. dioxus-core writes volatile attributes to the DOM on **every** re-render, even when the rendered string has not changed: `if volatile || attribute_changed { self.write_attribute(...) }` (`dioxus-core-0.7.9/src/diff/node.rs:463`).
3. The interpreter then assigns the value whenever the *live DOM* value differs from the VDOM value: `case "value": ... else if (node.value !== value) node.value = value;` (`dioxus-interpreter-js-0.7.9/src/ts/set_attribute.ts:31-33`).

`onchange` fires on blur, not per keystroke. So a field bound only through `onchange` still held the **pre-typing** text while the user typed, and any re-render in that window saw `node.value !== value` and reset the control.

`RoomDescriptionField` reads `CURRENT_ROOM` and `ROOMS` in its render body, so it is subscribed to both and re-renders on every room-state write. That is what supplied the periodic trigger in the report.

The sibling `RoomNameField` was never affected because it already has `oninput` (added for the Enter-to-commit handler in #21), which keeps the signal equal to the DOM value and makes the re-write a no-op.

## Approach

Ian asked me to check whether this affected more than the reported field, so I audited **every** `input` / `textarea` / `select` in `ui/src` rather than patching the one report. Added `oninput` to each editable control that bound `value:` without it:

| Field | Why it re-rendered |
|---|---|
| `RoomDescriptionField` (Room Description) | reads `CURRENT_ROOM` + `ROOMS` in its body, so every room-state write |
| `NumericConfigField` — renders **6** fields (Max Recent Messages, Max Message Size, Max User Bans, Max Nickname Size, Max Room Name, Max Room Description) | its `config` prop changes |
| `MaxMembersField` | its `config` prop, and also `member_count`, so a member joining mid-edit was enough |
| Create Room "Room Name" | its sibling nickname input already used `oninput`, so this was a plain oversight |

**Not affected, and deliberately left alone:**

- Read-only display controls (invite link, invite code, room public key, contract ID, member ID, export token) bind `value:` with no `oninput`, but nothing types into them. They declare `readonly: true`.
- The private-room checkbox binds `checked:`, which dioxus does **not** mark volatile, and is updated by `onchange` on click.
- No `contenteditable` and no real `select`/`option` elements exist in the tree.

Alternative considered and rejected: stopping `RoomDescriptionField` from subscribing to `ROOMS`/`CURRENT_ROOM` (e.g. via `peek()`). That removes one trigger but not the class, and it would stop the field from picking up a room secret that arrives after the dialog opens. `oninput` makes the volatile re-write a no-op regardless of what triggers the render, which is the actual invariant.

## Testing

**`components.rs::volatile_value_binding_audit`** — source-scrape audit over the whole `ui/src` tree, so a newly-added field cannot reintroduce the class.

- **Verified it fails on the pre-fix source**, reporting exactly `ui/src/components/room_list/edit_room_modal.rs:593 <textarea>`.
- Anti-vacuity floors on how many editable and display-only bindings it must find, so a refactor that stops the scanner matching this tree fails loudly instead of going green.
- Cuts each file at `#[cfg(test)]` so it cannot match its own fixtures (`feedback_source_pin_needle_matches_itself`).
- A self-test runs the scanner over the pre-fix *shape* of the field and asserts it is flagged, plus that the fixed shape passes.
- A third test asserts a nested child's `oninput` cannot satisfy its parent's requirement.
- Treats a literal `readonly: true` as display-only but **not** a conditional `readonly: !is_owner`, since that field is editable for the owner, who is exactly who loses their typing.

**`ui/tests/room-details-typing-survives-rerender.spec.ts`** — browser regression test.

- Types into the description with real keystrokes, then drives a genuine room-state write (a max-members change: sign config, apply delta to `ROOMS`, mark for sync) from inside the open dialog **without moving focus**, so the draft stays uncommitted. Clicking or tabbing would blur the textarea and fire its `onchange`, which would make the test pass against the broken build too.
- Gates on the `Members (n/NEW)` label reaching the new cap **before** asserting anything, so a write that never landed fails loudly rather than passing vacuously.
- **Verified it fails against the pre-fix build** with `Expected: "A description I am still in the middle of typing" / Received: ""` — the exact reported symptom.
- A second case asserts `oninput` did not break the `onchange` save path: blur to commit, close, reopen (which remounts the field and reseeds it from the stored configuration), and confirm the description round-tripped.

**Suites:**

- `cargo test -p river-ui --bins` — 801 passed
- `cargo test -p river-ui --bins --features example-data,no-sync` — 806 passed
- Full Playwright suite on chromium — 144 passed, 6 skipped, 0 failed
- `cargo fmt` and `cargo clippy` clean (no findings in changed files)

Also documented the hazard in `.claude/rules/dioxus-signal-safety.md`, since nothing about the broken call site looks wrong and reasoning about the component in isolation will not reveal it.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/dioxus-signal-safety.md
+++ b/.claude/rules/dioxus-signal-safety.md
@@ -190,6 +190,56 @@ Signal clears that the effect subscribes to must be synchronous.
 Deferring causes an infinite loop (set remains non-empty → effect
 re-runs → defers clear → effect re-runs...).
 
+## Bind `oninput` on every EDITABLE `value:` field
+
+`value` on `input` / `textarea` / `select` is a **volatile** attribute in
+dioxus-html (`elements.rs:1488,1571,1594`). dioxus-core re-writes volatile
+attributes to the DOM on every re-render, even when the rendered string has
+not changed:
+
+```text
+// dioxus-core-0.7.9/src/diff/node.rs:463
+if volatile || attribute_changed { self.write_attribute(...) }
+```
+
+and the interpreter then assigns the value whenever the LIVE DOM value differs
+from the VDOM value:
+
+```text
+// dioxus-interpreter-js-0.7.9/src/ts/set_attribute.ts:31-33
+case "value": ... else if (node.value !== value) node.value = value;
+```
+
+`onchange` fires on blur, not per keystroke. So a field bound only through
+`onchange` still holds the PRE-TYPING text for as long as the user is typing,
+and the next re-render resets the control and discards the edit.
+
+```rust
+// WRONG — any re-render wipes what the user has typed so far
+textarea { value: "{description}", onchange: save }
+
+// RIGHT — the signal tracks the DOM, so the re-write is a no-op
+textarea {
+    value: "{description}",
+    oninput: move |evt| description.set(evt.value().to_string()),
+    onchange: save,
+}
+```
+
+Nothing about the broken call site looks wrong, and reasoning about the
+component in isolation will not reveal it: the trigger is whatever unrelated
+signal that component happens to read. `RoomDescriptionField` reads
+`CURRENT_ROOM` and `ROOMS` in its render body, so every room-state write
+re-rendered it, and the owner's half-typed room description vanished every few
+seconds (freenet/river#564).
+
+Display-only controls may bind `value:` with no `oninput`, but must declare
+themselves with a literal `readonly: true`. A conditional
+`readonly: !is_owner` does NOT count: the field is editable for somebody, and
+that somebody is exactly who loses their typing.
+`components.rs::volatile_value_binding_audit` enforces this across the whole
+`ui/src` tree, so a new field cannot reintroduce the class.
+
 ## Don't `use_memo` against non-signal values in an always-mounted component
 
 The modals in `app.rs` (`MemberInfoModal`, `DmThreadModal`,
--- a/.claude/rules/dioxus-signal-safety.md
+++ b/.claude/rules/dioxus-signal-safety.md
@@ -230,15 +230,42 @@ Nothing about the broken call site looks wrong, and reasoning about the
 component in isolation will not reveal it: the trigger is whatever unrelated
 signal that component happens to read. `RoomDescriptionField` reads
 `CURRENT_ROOM` and `ROOMS` in its render body, so every room-state write
-re-rendered it, and the owner's half-typed room description vanished every few
-seconds (freenet/river#564).
+re-rendered it, and the owner's half-typed room description vanished
+(freenet/river#564). There is no single timer behind that cadence: the
+recurring ROOMS writers are the ~60 s idle liveness probe whose GET reply
+merges unconditionally, the ~21 s `ProcessRooms` loop while a room awaits
+subscription, and every subscription `UpdateNotification`. Do not go looking
+for one interval to blame.
+
+**This is the documented exception to the `defer()` rule above.** A controlled
+input's bound signal must be written SYNCHRONOUSLY from `oninput`. Deferring it
+lags the DOM by a `setTimeout(0)`, so a re-render landing in that window writes
+the pre-keystroke value back and drops characters, which is the very bug this
+section exists to prevent. The rule above targets GLOBAL signals (`ROOMS`,
+`CURRENT_ROOM`, ...) with external subscribers; `util.rs`'s
+`event_handlers_never_write_a_global_signal_without_defer` pin is scoped to
+those deliberately and does not match `oninput`. Local `use_signal` handles
+have no external subscribers and are set directly. See also the note at
+`members.rs`'s import-token `oninput`.
 
 Display-only controls may bind `value:` with no `oninput`, but must declare
-themselves with a literal `readonly: true`. A conditional
-`readonly: !is_owner` does NOT count: the field is editable for somebody, and
-that somebody is exactly who loses their typing.
-`components.rs::volatile_value_binding_audit` enforces this across the whole
-`ui/src` tree, so a new field cannot reintroduce the class.
+themselves with a literal `readonly: true` (or `readonly: "true"`). A
+conditional `readonly: !is_owner` does NOT count: the field is editable for
+somebody, and that somebody is exactly who loses their typing.
+`type="checkbox"` and `type="radio"` are exempt too, for the opposite reason:
+dioxus reports their `evt.value()` as `"true"`/`"false"`, never the DOM's
+`"on"`, so tracking a `value:` binding there would CAUSE a rewrite every
+render. Bind `checked:`, which is not volatile.
+
+`components.rs::volatile_value_binding_audit` scans every `.rs` file under
+`ui/src` for this and pins the exact set of bindings it finds, so a field that
+drops out of the scan fails loudly rather than quietly losing coverage. Two
+limits worth knowing: it proves `oninput` is present and not a no-op, but not
+that the handler writes the signal the `value:` binding reads; and it says
+nothing about the OTHER route to the same symptom, where a parent unmounts the
+field (a contended `try_read` memo returning `None`) and the remount re-seeds
+`use_signal`. That second route is the #499/#555 contention family and
+`oninput` does not help there.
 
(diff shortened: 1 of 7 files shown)


sanity:
## Multi-perspective review (round 1)

Full tier: four independent blind reviewers, each reading the checked-out code with a distinct lens, none seeing the others' findings. Lenses: **code-first** (read code before description, flag intent/impl mismatch), **testing** (coverage gaps and vacuous tests), **skeptical** (adversarial attack on the root-cause diagnosis itself), and **UI failure-mode** (every surface and state these fields appear in). External non-Claude models not run, per the opt-in policy.

### Diagnosis: independently confirmed

All four verified the causal chain against the vendored dioxus 0.7.9 sources rather than taking it on trust. Every citation checks out (`elements.rs:1488/1536/1571/1594`, `diff/node.rs:463`, `set_attribute.ts:31-33`). The skeptical lens specifically hunted for a guard that would prevent the clobber — focus/`activeElement` checks, dirty-value tracking, template memoization, `initial_value` handling — and found none, and confirmed props memoization cannot save it because the component subscribes to `ROOMS` directly in its render body.

Strongest corroboration, found independently: `RoomNameField` already had `oninput` **before this branch**. Two sibling fields in the same dialog, identical re-render exposure, identical volatile binding, and only the one lacking `oninput` was reported broken. The hypothesis predicts exactly that asymmetry.

The scope audit was re-done from scratch by two reviewers who did not trust the PR's own: same four fields, same eight exempt read-only controls, `checked:` correctly out of scope.

### Blocking findings, both in the guard rather than the fix

**1. The audit was blind to 41% of `ui/src`.** Cutting at the first `#[cfg(test)]` lands on a test-only *helper* at `conversation.rs:1156` of 9007, so 87% of the largest UI file went unscanned — including the message-edit textarea at `conversation.rs:4937`, a `value:`-bound editable control in exactly the audited class. This repo documents that trap three times in `conversation.rs` and once in `members.rs`. Those pins survive it because they assert a needle is PRESENT (over-short slice fails loudly); this one asserts a needle is ABSENT (over-short slice passes silently). Fixed by removing each `#[cfg(test)]` item individually, which also handles this repo's mid-file `mod tests` in `chat_delegate.rs` that the `"#[cfg(test)]\nmod tests {"` needle would have truncated.

**2. Anti-vacuity floors could not detect losing the fixed fields.** `editable_checked >= 8` against a true count of 13: all four fields this PR fixes could drop out of the scan with the test still green. Replaced with an exact pinned set of all 14 bindings.

### Also fixed

- Character-based attribute parsing over a real lexer. The line-based version could not see attributes on an element's opening line (single-line elements silently skipped, `readonly:` on the opening line misread), and a `//` inside a string literal corrupted brace depth for the rest of a block. A comment above an attribute was being glued onto its name — that bug produced six false positives in an early prototype.
- `option { selected: }` brought into scope; `type="checkbox"`/`"radio"` explicitly exempted, since dioxus reports their `evt.value()` as `"true"`/`"false"` and tracking a `value:` binding there would *cause* the rewrite.
- No-op `oninput` handlers now rejected. The stronger "handler writes the bound signal" check was tried and **rejected**: it false-positives on `nickname_field.rs` (named handler) and `invite_via_dm_picker_modal.rs` (value binds a derived local). The module docs now state plainly what the audit does and does not prove, rather than overclaiming.
- Per-keystroke ECIES decrypt (flagged by three lenses): `oninput` re-renders per keystroke, and the render body cloned the room's whole secret set and ran an unseal. Extracted to `stored_description()`, called only at mount and on the privacy revert.
- `create_room_modal`: `onchange` restored alongside `oninput`. Dropping it would lose the value from anything that sets `.value` and fires only `change`, and `create_room` fails silently on an empty name. The comment no longer claims a reproduced bug there — that component reads no room-state signal, so the change is defensive.
- Rules doc now records that controlled inputs are the documented **exception** to the `defer()` rule (deferring lags the DOM and drops keystrokes), resolving a contradiction with the section above it.

### Known and deliberately not closed

There is a second route to the same symptom: if a contended `ROOMS.try_read()` makes `EditRoomModal` render `rsx!{}`, the field unmounts and the remount re-seeds `use_signal`, losing the draft regardless of `oninput`. That is the #499/#555 contention family, tracked separately. It does not fit this report (it would reseed the Room Name field too and destroy focus, neither of which was reported), but the audit no longer claims to close every route.

### Verification

Every claim mutation-tested rather than assumed:

- Audit fails on the pre-fix source, naming `edit_room_modal.rs:593 <textarea>`.
- Playwright test 1 fails against the pre-fix build with `Expected: "A description I am still in the middle of typing" / Received: ""` — the exact reported symptom.
- Re-ran the mutation **after** the `stored_description` refactor, since changing when the component re-renders could have made the browser tests vacuous. Removing only the max-members `oninput` fails test 2 (typed `242` reverts to stored `200`) while tests 1 and 3 pass, so each test discriminates for its own field.
- A vacuous gate I introduced in the new test 2 (asserting a DOM value the test itself had written) was caught and re-pointed at the room heading, which renders from room state.

[AI-assisted - Claude]

Issue #513, opened 2026-07-27
Rooms have no public address — the only shareable URL is a one-shot invite containing a private key


sanity:
## Problem

There is currently no way to point at a River room publicly.

The only room-bearing URL the UI accepts is `?invitation=` (`ui/src/components/app.rs`), and its payload is:

```rust
pub struct Invitation {
    pub room: VerifyingKey,
    pub invitee_signing_key: SigningKey,   // <-- a private key
    pub invitee: AuthorizedMember,
}
```

So every room reference in circulation is a **secret bearer token, personalised and one-shot**. You cannot list a room on Atlas, link one from a blog post, paste one into another room's config, or read one out loud. Compare Delta, where ten public characters in a URL are enough for anyone to find and read a site.

This is a real adoption ceiling: rooms are undiscoverable by construction, and the only growth path is someone personally minting you a token.

## The fix is nearly free

`ChatRoomParametersV1 { owner: VerifyingKey }` means the owner key already fully determines the contract address, and `ui/src/util.rs::owner_vk_to_contract_key` already computes it. So a public address form needs:

- **no contract change**
- **no re-key**
- **no migration**
- **no state format change**

Just a `?room=<handle>` URL parameter alongside `?invitation=`, resolving through the existing derivation, plus somewhere in the UI to copy a room's public address.

## Scope note

A public address gets you **read**. Joining still requires a member with invite authority to sign an `AuthorizedMember`, so "find a room and join it" additionally needs a self-serve admission path (tracked separately). The address is necessary, not sufficient — but it is the piece everything else waits on, and it is the cheap one.

## Handle format

Shipping the full owner key as the handle works today at about 44 base58 characters. A shorter 22-character form is proposed separately and would change the parameters, so this issue should not block on it: ship the address now, and the handle format can change underneath it later since both derive from the same key.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-07-29T16:08:49.428369464+00:00
Issue hash: `9c2e7263`

## Value Assessment

[...]

Issue #506, opened 2026-07-26
feat(dm): invite button inside the DM thread modal (invite the counterparty without hunting the member list)


sanity:
Requested by Ivvor in the official room (2026-07-26):

> When in a DM with somebody it'd be nice to have an invite button to send an invite from the DM window. Currently I have to find them in the members list and invite from there. Which can be tricky given the strange and mysterious member ordering there. Also a problem if the name is used by several people.

## Request

An "Invite to room" button inside the DM thread modal (`ui/src/components/direct_messages/dm_thread_modal.rs`) that starts the invite flow with the current DM counterparty as the implicit target, instead of requiring the user to relocate that person in the member list.

## Why the existing flow falls short

- The member list's ordering is opaque to users, so finding the same person again is harder than it should be (related: #493, search member by name).
- Nicknames are not unique, so picking the right entry from the member list is ambiguous when several members share a name. The DM thread is already bound to a specific MemberId, which removes the ambiguity entirely.

## Implementation sketch

Most of the machinery exists. `INVITE_VIA_DM_PICKER` opens `invite_via_dm_picker_modal.rs`, which lists every other room the local user is in, signs an `Invitation` against the CANDIDATE room's key, and hands the composed body to `DmThreadModalBody` via `DM_DRAFT` (appending, never overwriting). Today that picker is entered from elsewhere; this feature adds an entry point from the open DM thread itself, pre-filling the target as the thread's counterparty and returning the draft into the already-open thread.

Notes from the existing design (`.claude/rules/direct-messages.md`):
- The cross-room "is the target already a member of the candidate room" filter was deliberately deferred because per-room identities make it structurally infeasible without a global identity layer; the button should not block on that (same behavior as the current picker).
- `invite_member_modal::get_invitation_base_url()` is the single source of the URL format; reuse it via the picker rather than composing URLs anywhere new.

## Scheduling

Queued behind the 2026-07-26 fix pipeline (#499/#501/#500) and its publish; pick up after that ships.

[AI-assisted - Claude]

sanity:
Ian's response to Ivvor in the room: clicking the person's name in the message list also works — it opens the member-info modal, which has the invite path, and since you click a specific message author it avoids both the member-list ordering and the duplicate-name ambiguity.

[...]

PR #567, opened 2026-07-30
feat(ui): make Share Invite the primary invite action, demote invite links

### LINKED ISSUE #566: ux: invite links are the discoverable path but break when shared — promote Share Invite instead
## Problem

The members panel's primary action was **Invite Member**, which mints an invite link. That link embeds `invitee_signing_key` — a private key — and is good for **exactly one person**. In practice people share one link with several people. Everyone who used it then holds the same identity, and none of them work correctly.

Ian, 2026-07-29:

> Invite Member button is a bad idea, gives people links, they don't understand that they can't share with multiple people and it's ruined because private key in link. instead we deprioritise this and direct users to use "Share Invite" button, maybe make it more discoverable

River already had a route with none of that failure mode: **Share invite**, which DMs an invitation card the recipient accepts in one click, with no credential travelling through an outside channel (#252, #457). But it was effectively undiscoverable, and worse, it pointed the wrong way:

- It lived behind a member's card in the member list, three clicks deep.
- It starts from a **person in the room you are viewing** and asks which of your **other** rooms to invite them to. Someone standing in room X wanting to invite Alice to room X had to first navigate to some other room where Alice is a member, click her name there, and pick room X from a list. Nobody finds that.

So the only *discoverable* way to invite anyone was the one that breaks when shared.

## Change

1. **"Share Invite" becomes the primary action** in the members panel (accent-filled, first). It opens a new contact picker that runs in the direction users actually arrive from: the room is fixed (the one being viewed) and you pick **who** to send it to, from the members of your other rooms. Search filters by person or by room; people you have DM history with sort first.
2. **The link flow is demoted, not removed** — a small secondary "Invite by link" button beside "Export ID". It is still the only way to invite somebody who is not on River yet.
3. **Copy is blunter** about what the link is ("contains a private key", "one person only", "the most common way invites go wrong") and points at Share Invite by its actual label.

Both pickers now share one send path (`invite_dm.rs::compose_and_send_invite_dm`) so an invitation minted by one cannot drift from the other.

## Not in scope

- The invite-from-inside-a-DM-thread entry point (#506) — a third entry point into the same machinery, still open.
- A public, non-secret room address (#513), which is the underlying reason every room reference in circulation is a one-shot bearer token.

[AI-assisted - Claude]

sanity:
## Problem

The members panel's primary action was **Invite Member**, which mints an invite link. That link embeds `invitee_signing_key` — a private key — and is good for **exactly one person**. In practice people share one link with several people; everyone who used it then holds the same identity, and none of them work.

Ian, 2026-07-29:

> Invite Member button is a bad idea, gives people links, they don't understand that they can't share with multiple people and it's ruined because private key in link. instead we deprioritise this and direct users to use "Share Invite" button, maybe make it more discoverable

River already had a route with none of that failure mode — DM an invitation card the recipient accepts in one click, no credential leaving River (#252, #457). But it was undiscoverable **and pointed the wrong way**: it starts from a *person in the room you are viewing* and asks which of your *other* rooms to invite them to. Somebody standing in room X wanting to invite Alice to room X had to first navigate to another room where Alice is a member, click her name there, and pick room X from a list. So the only *discoverable* invite path was the one that breaks when shared.

## Approach

**1. "Share Invite" becomes the primary action** (accent-filled, first in the panel). It opens a new contact picker running the direction users actually arrive from: the room is fixed (the one being viewed) and you pick *who*, from the members of your other rooms. Search filters by person **or by room** ("someone from the team room" is how people remember a contact); people you have DM history with sort first; the list caps at 40 rows and states how many it withheld.

**2. The link flow is demoted, not removed** — a small secondary "Invite by link" button beside "Export ID". It is still the only way to invite somebody who is not on River yet. Its `data-testid` is unchanged so existing specs keep pointing at the link flow they were written for.

**3. Copy is blunter** about what the link is ("contains a private key", "one person only", "the most common way invites go wrong") and names Share Invite by the label it actually carries.

Why a second picker rather than a mode on the existing one: the two run in opposite directions and each has non-trivial per-open invariants (session-tagged selection, generation-gated watchdog). Two components with one shared send path keeps each set of invariants local. The shared path is `invite_dm.rs::compose_and_send_invite_dm`, so an invitation minted from one entry point cannot drift from the other.

### Impersonation warning (second commit)

Promoting the picker created a route that **bypasses the member list**. Until now every path to the DM-invite flow went through that list, so the user had already been shown the ⚠ before choosing anyone. The picker lists members of rooms the user may never have opened, so without the badge the primary invite action became the one surface where an imitator's name renders clean — and picking the wrong "Finn Clarke" sends a stranger an invitation to your room.

Each row now carries the same warning, from the same helpers, with the checker and badge map built **once per candidate room** rather than per peer. The row's `aria-label` repeats the warning, because an explicit `aria-label` replaces element content for naming — a badge inside a labelled button is invisible to a screen reader otherwise. Member ID is on hover, which is the remedy the tooltip names.

## Testing

**Unit (`cargo test -p river-ui --bins`, 805 pass):** 5 tests for the candidate sort/filter, including that the comparator is *total* — two members of one room can share a nickname, and a non-total comparator lets rows swap between renders.

**Source pins:**
- `share_invite_is_the_primary_invite_action_and_the_link_flow_is_secondary` — order, which handler each button opens, and that the link button never gets `bg-accent` back.
- `impersonation_warning_is_wired_into_every_render_surface` extended to a fourth surface, including the argument pin (`self_id` for `peer` compiles and flags the genuine owner and every real moderator).
- `each_picker_passes_the_carrier_room_first_and_the_target_room_third` — the two pickers hand the same two rooms to `compose_and_send_invite_dm` in opposite positions and a swap compiles. A swap signs the invitation against the room the DM is merely travelling through and sends the DM into the room the recipient is not in yet, so the user is told "the recipient is no longer a member of this room" about a room they can see them in. Nothing else can reach it — the send needs the chat delegate, which is absent under `no-sync` — so the argument names are pinned.

**Browser (`invite-priority.spec.ts`, 7 cases × 5 projects):** button order and that the two really differ in computed background; the picker's header names the target room and no candidate row is from it; Send arms only after a pick; filtering by person and by room, and the no-matches state; the ⚠ badge with its tooltip, aria text and member-ID hover; close; and that the link flow still opens.

**Full suite: 752 passed, 23 skipped, 0 failed** across chromium, firefox, webkit, mobile-chrome, mobile-safari.

**Mutation-tested** (11 mutations, each caught by the test named): dropping the DM-history sort priority, dropping the member-id tiebreak, filtering on person only, reporting 0 withheld, giving the link button the accent fill back, pointing Share Invite at the link modal, removing Share Invite's accent, swapping carrier/target at either call site, computing the warning from `self_id`, and — the case that justifies the browser spec — suppressing the badge while every source pin still matches.

### Incidental fix

`responsive-layout.spec.ts`'s two sandboxed-iframe tests hardcoded `http://localhost:8082`, silently ignoring `PLAYWRIGHT_BASE_URL`. Running the suite on any other port — which `AGENTS.md` tells you to do, since 8082 is shared between worktrees — pointed the iframe at whatever build happened to be on 8082, or at nothing, and failed those two for reasons unrelated to the change under test.

## Not in scope

- **#506** — invite button inside the DM thread modal. A third entry point into the same machinery; still open.
- **#513** — a public, non-secret room address. The underlying reason every room reference in circulation is a one-shot bearer token.

No WASM, contract, or delegate change — UI only, so no migration.

Closes #566

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components.rs
+++ b/ui/src/components.rs
@@ -97,6 +97,8 @@ mod volatile_value_binding_audit {
         r#"components/conversation.rs <textarea> "{edit_text}""#,
         r#"components/conversation/message_input.rs <textarea> "{message_text}""#,
         r#"components/direct_messages/dm_thread_modal.rs <textarea> "{draft.read()}""#,
+        r#"components/direct_messages/invite_contact_picker_modal.rs <input> "{query_value}""#,
+        r#"components/direct_messages/invite_contact_picker_modal.rs <textarea> "{personal_message_value}""#,
         r#"components/direct_messages/invite_via_dm_picker_modal.rs <textarea> "{personal_message_value}""#,
         r#"components/members.rs <textarea> "{token_input}""#,
         r#"components/members/member_info_modal/nickname_field.rs <input> "{temp_nickname}""#,
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -10,7 +10,9 @@ use crate::components::app::document_title::DocumentTitleUpdater;
 use crate::components::app::freenet_api::freenet_synchronizer::SynchronizerMessage;
 use crate::components::app::freenet_api::freenet_synchronizer::SynchronizerStatus;
 use crate::components::app::freenet_api::FreenetSynchronizer;
-use crate::components::direct_messages::{DmThreadModal, InviteViaDmPickerModal};
+use crate::components::direct_messages::{
+    DmThreadModal, InviteContactPickerModal, InviteViaDmPickerModal,
+};
 use crate::components::members::member_info_modal::MemberInfoModal;
 use crate::components::members::Invitation;
 use crate::components::room_list::create_room_modal::CreateRoomModal;
@@ -506,6 +508,7 @@ pub fn App() -> Element {
         CreateRoomModal {}
         DmThreadModal {}
         InviteViaDmPickerModal {}
+        InviteContactPickerModal {}
         ReceiveInvitationModal {
             invitation: receive_invitation
         }
--- a/ui/src/components/direct_messages.rs
+++ b/ui/src/components/direct_messages.rs
@@ -19,9 +19,14 @@
 //! back as unread on every page load.
 
 mod dm_thread_modal;
+mod invite_contact_picker_modal;
+mod invite_dm;
 mod invite_via_dm_picker_modal;
 
 pub use dm_thread_modal::DmThreadModal;
+pub use invite_contact_picker_modal::{
+    has_invitable_contacts, open_invite_contact_picker_for_current_room, InviteContactPickerModal,
+};
 pub use invite_via_dm_picker_modal::InviteViaDmPickerModal;
 
 use dioxus::prelude::*;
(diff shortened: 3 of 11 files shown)


sanity:
Note for whoever rebases this: **#565 and this PR will collide in one specific place.**

#565 adds `components.rs::volatile_value_binding_audit`, which pins the *exact set* of `value:` bindings across `ui/src`. This PR adds two new ones — the contact picker's search `input` and its personal-message `textarea`. Both already satisfy #565's rule (each binds `oninput` writing the signal its `value:` reads, and both are local `use_signal` handles, which is the documented exception), so this is an inventory update, not a fix: after rebasing onto #565, the two new bindings have to be added to that audit's pinned list or it fails.

Landing order doesn't matter for correctness, only for who does the inventory edit. #565 is the smaller diff and is already out of draft, so it going first seems right.

[AI-assisted - Claude]

Issue #569, opened 2026-07-30
perf(ui): release build logs ~23 MB to the console during startup, costing ~29% of startup blocking


sanity:
## Symptom

Users report the UI feels slow/laggy. Profiling on a real NATed peer (framework, against the published UI at web-container 30000369) puts the cost in **startup**, and about a quarter to a third of it is the app logging to the browser console.

## Measurement

Headless Chromium, published UI, real rooms (8 rooms including the 200-member Official room). Long tasks from an in-page `PerformanceObserver`. A/B: identical runs, the only difference being whether the app's own `console.*` output is emitted. The counting shim is present in **both** arms, so its cost is controlled; only the emit/format is toggled.

| | main-thread blocking | time to rooms hydrated | long tasks |
|---|---|---|---|
| logging (as shipped) | **3371 ms** | 1924 ms | 22.7 |
| console suppressed | **2405 ms** | 1438 ms | 20.7 |

Mean of 3 paired runs. Direction was consistent (suppressed ≤ logging in every pair). **Suppressing the app's own logging removed ~966 ms (~29%) of startup blocking and ~486 ms (~25%) of time-to-rooms.**

Volume, measured in-frame: **~1,066 console calls totalling ~23 MB** during startup alone.

Two caveats, stated so the numbers are not over-read:
- Absolute figures here are inflated by the counting shim. A cleaner run without any shim or sampler measures **~1.75 s** blocking over ~12 long tasks (worst ~270 ms). The **A/B difference** is the reliable part, not the absolutes.
- An earlier run of mine reported 6.8 s / 75.8% CPU. That was inflated roughly 4x by my own CPU sampler (200 µs) plus a document-wide `MutationObserver`. Do not cite it.

## Cause

`ui/Cargo.toml` sets `tracing = { features = [..., "release_max_level_info"] }`, so `debug!`/`trace!` compile out but **`info!` ships to users**. The startup path is heavily narrated at `info!` — `chat_delegate.rs`, `response_handler.rs`, `freenet_synchronizer.rs` — and several sites log whole payloads as byte arrays, e.g.

```
INFO response_handler.rs:167 ApplicationMessage payload: [161, 116, 71, 101, 116, 80, 117, ...]
INFO freenet_synchronizer.rs:686 Delegate response with key: DelegateKey { key: [195, 98, 79, ...] }
```

[...]

Issue #554, opened 2026-07-29
UI: enforced_banned_member_ids does 200 Ed25519 verifications per call on the render path


sanity:
## Problem

`RoomData::enforced_banned_member_ids` (`ui/src/room_data.rs:598`) calls the **uncached** `MembersV1::banned_member_ids`, which performs one Ed25519 verification per stored ban. On the live Freenet Official room that is **200 verifications per call**.

It is called from the render path (`ui/src/room_data.rs:1288`), and again at `:641` from `is_self_enforced_banned`.

## Cost

Ed25519 verification in the browser's wasm32 build is in the same range as the contract's (~0.26–0.42 ms/verification measured under wasmtime at `OptLevel::None`; the browser figure will differ but not by orders of magnitude). At 200 bans that is plausibly **tens of milliseconds per call on the render path**, repeated per render.

## Relationship to #548

freenet/river#548 introduced `BanSignatureCache` for exactly this redundancy on the contract side, and `MembersV1::banned_member_ids_cached` is `pub`. The UI can take the same fix by hoisting one cache across the calls it makes per render, rather than constructing a fresh one per call (which is what the uncached entry point does internally).

Not bundled into #548 because that PR is scoped to the contract WASM and its output-identity argument; this is a pure UI render-path concern with different review surface and no migration implications.

## Suggested fix

Hold a `BanSignatureCache` for the duration of a render pass (or memoize `enforced_banned_member_ids` itself against the room state's ban+member generation) and call `banned_member_ids_cached`. Worth checking with the conversation-render profiling already done for #458 whether this shows up measurably before optimising.

Related: #548, #422.

[AI-assisted - Claude]

sanity:
# Issue Priority Assessment

Generated: 2026-08-22T16:46:07.241745145+00:00
Issue hash: `1a30eae8`

## Value Assessment

Unassigned, still open, and its prerequisite (#548, `BanSignatureCache`) is unmerged — so this issue is still fully relevant; nothing in the code has changed to invalidate it.

## Assessment

[...]

Issue #523, opened 2026-07-27
perf(ui): every message group re-renders on every Conversation render (inline closures defeat memoization)


sanity:
## Problem

A user reported the chat interface has "overactive reactivity" with "a lot of reflow/repaint". Profiling (Chromium CDP `Performance.getMetrics` + a `MutationObserver`, example-data build padded to 90 messages) said the reactivity half is right and the layout half is not: sending one message cost **1466 ms** of main-thread time at 6x CPU throttle, of which layout + style recalc was **24 ms (1.6%)** and scripting was **1187 ms (81%)** — while only 2-8 DOM nodes actually changed. The browser was doing almost no layout work; the app was computing a render it then threw away.

## Root cause

Dioxus decides whether a child component can be skipped by comparing **all** of its props with `PartialEq` — event handlers included. The derive excludes callbacks from the *move* logic, not from the *equality* check:

```rust
let equal = self == new;   // derived PartialEq over every field
#move_event_handlers
if !equal { /* move regular fields */ }
equal
```

A closure written inline in `rsx!` becomes a fresh `Callback` every render, and `Callback`'s `PartialEq` is pointer identity. So the four inline handlers passed to `MessageGroupComponent` (`on_react`, `on_request_delete`, `on_edit`, `on_reply`) made it **permanently un-memoizable**: every re-render of `Conversation` re-rendered every message group.

Instrumented group-render counts on an 85-group room, before/after a prototype fix:

| interaction | before | after |
|---|---|---|
| send one message | 340 (4 full passes) | 4 |
| scroll the history | 85 (1 full pass) | 0 |
| type a character | 0 | 0 |

Scrolling re-rendered the entire history because crossing the bottom sentinel flips `is_at_bottom`, which `Conversation` reads in render.

## Two independent pieces

**1. Stable callback identities.** `use_callback` allocates the handle once and swaps the inner closure each render, so identity is stable (props compare equal → the group memoizes) while captured state stays current. Freshness does not depend on the memoize path: `use_callback` replaces the inner closure on every `Conversation` render, and Dioxus additionally re-points the old handle at the newest closure inside `memoize`, which is called unconditionally before the equality result is used.

Considered and rejected in the prototype: hand-writing the props struct with a manual `PartialEq` that skips the handler fields. That is the documented opt-out, but it means restating 12 fields and re-deriving `Props` by hand, and it silently breaks if a field is added later.

[...]

PR #570, opened 2026-07-30
perf(ui): stop dumping delegate payloads to the console at startup

sanity:
## Problem

Users reported the UI feeling slow/laggy. I profiled the published UI (web-container 30000369) on a real NATed peer with Playwright, measuring per interaction rather than in aggregate.

**The cost is startup.** Everything else measured clean:

| phase | result |
|---|---|
| startup | **~1.75s main-thread blocking**, ~12 long tasks, worst ~270ms; rooms usable at ~1.5-2.4s |
| idle 30s | **0.2% CPU**, 0 long tasks, 0 DOM mutations |
| scroll | **0 janky frames of 170**, p95 16.7ms |
| room switch | ~0.1-0.7s |

Then a census of console output grouped by source site found **one line is 98.5% of all bytes River emits during startup**:

```
response_handler.rs  "Successfully deserialized as ChatDelegateResponseMsg: {:?}"
```

`ChatDelegateResponseMsg`'s `Debug` prints a `ListResponse`'s keys as decimal byte arrays. On a real account: **39 lines averaging ~401 KB each, 14.93 MB of the 15.15 MB total** (1,034 messages).

## Approach

Summarise instead of Debug-dumping. `describe_delegate_response` keeps everything the logs were actually used for — variant, key names, sizes, ok/err — and prints keys as the text they are (`outbound_dms`, `rooms_meta`, `room:<vk>`), falling back to a length for genuinely binary keys. Plus:

- payload logged as `{} bytes` rather than up to 100 decimal bytes
- `DelegateKey` logged as a 4-byte fingerprint rather than a 32-byte decimal array, at both sites

**No log levels changed.** My first instinct was to drop `release_max_level_info`, and the census is why I didn't: that would delete every field diagnostic to fix what turned out to be four call sites — including the `Timeout waiting for delegate response` warnings that are worth keeping (see below).

## Expected effect — a prediction, not a measurement

An A/B that suppressed the app's console **entirely** removed ~29% of startup blocking (3371ms → 2405ms, mean of 3 paired runs, the counting shim present in both arms so its cost is controlled).

This change removes ~98.5% of the *bytes* but only 39 of ~1,034 *calls*. So:
- if the cost is dominated by bytes and formatting, the gain should be most of that 29%
- if it is dominated by per-call overhead, considerably less

The 39 lines also cost roughly **15.6 MB of `String` building inside WASM**, which no console setting avoids, so I expect the former. I will re-run the census and the A/B after publish and post the actual numbers here rather than leave the prediction standing.

## Testing

Five new tests:
- key names present and byte arrays absent in a `ListResponse` summary
- the summary grows with key-name length, not byte-array length
- values reported as sizes, never contents
- binary (non-UTF-8) keys fall back to a length
- a regression pin that fails if either the `Debug` dump or the byte-array payload log returns, **mutation-verified**: restoring `{:?}` on the response makes it red

## Also found while profiling, not fixed here

Startup consistently logs `Failed to store signing key in delegate: Timeout waiting for delegate response - using local signing` and `Skipping EnsureRoomSubscription`. Delegate round-trips are timing out during load and the app falls back to local signing. That is correctness-adjacent, not just slow, and is consistent with **#425**, whose burst is now **78 serial** requests (26 legacy generations x 3) and whose localStorage "done" gate cannot persist in the sandboxed gateway iframe (opaque origin), so hosted users may re-arm it on every load.

Interaction-time cost is real but tracked separately and **not** addressed here: **#523** (inline closures defeat `MessageGroupComponent` memoization), **#554** (~200 Ed25519 verifies per render via `can_participate()`), plus an unconditional ~343 KB `RoomData` deep clone at the top of every `Conversation` render and no `Intl.DateTimeFormat` cache. Those are correctness-adjacent enough to deserve their own review rather than riding along with a logging change.

## Corrections to my own earlier numbers

- An initial profiler run reported 6.8s blocking / 75.8% CPU. **Wrong** — inflated roughly 4x by my own CPU sampler (200µs) plus a document-wide `MutationObserver`. The ~1.75s figure above is from a clean run.
- An initial "~3.1s room switch" was a fixed settle-sleep in my harness, not app cost.

Profilers: `framework:~/river-freeze-monitor/{profile_ui.py,ab_startup.py,log_census.py}`.

Refs #569

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/freenet_api/freenet_synchronizer.rs
+++ b/ui/src/components/app/freenet_api/freenet_synchronizer.rs
@@ -683,9 +683,18 @@ impl FreenetSynchronizer {
                                 // Log more details based on response type
                                 match &host_response {
                                     HostResponse::DelegateResponse { key, values } => {
+                                        // Key fingerprint, not the full Debug: that prints
+                                        // the 32-byte key as a decimal array on every
+                                        // delegate response (freenet/river#569).
+                                        let key_head: String = key
+                                            .bytes()
+                                            .iter()
+                                            .take(4)
+                                            .map(|b| format!("{b:02x}"))
+                                            .collect();
                                         info!(
-                                            "Delegate response with key: {:?}, values count: {}",
-                                            key,
+                                            "Delegate response with key: {}.., values count: {}",
+                                            key_head,
                                             values.len()
                                         );
                                         for (i, v) in values.iter().enumerate() {
--- a/ui/src/components/app/freenet_api/response_handler.rs
+++ b/ui/src/components/app/freenet_api/response_handler.rs
@@ -144,9 +144,12 @@ impl ResponseHandler {
                     info!("Received response from LEGACY delegate - checking for migration data");
                 }
 
+                // Log a short key fingerprint, not the full `DelegateKey` Debug: that
+                // prints the 32-byte key as a decimal byte array (~450 B a line, on
+                // every delegate response). freenet/river#569.
                 info!(
-                    "Received delegate response from API with key: {:?} containing {} values",
-                    key,
+                    "Received delegate response from API with key: {} containing {} values",
+                    delegate_key_fp(&key),
                     values.len()
                 );
                 for (i, v) in values.iter().enumerate() {
@@ -158,13 +161,14 @@ impl ResponseHandler {
                                 app_msg.processed
                             );
 
-                            // Log the raw payload for debugging
-                            let payload_str = if app_msg.payload.len() < 100 {
-                                format!("{:?}", app_msg.payload)
-                            } else {
-                                format!("{:?}... (truncated)", &app_msg.payload[..100])
-                            };
-                            info!("ApplicationMessage payload: {}", payload_str);
+                            // Size only. This used to format up to 100 payload bytes as a
+                            // decimal array on every delegate response; the byte values
+                            // were never actually used for triage, the size and the
+                            // decoded variant below are. freenet/river#569.
+                            info!(
+                                "ApplicationMessage payload: {} bytes",
+                                app_msg.payload.len()
+                            );
 
                             // Try to deserialize as a response
                             let deserialization_result = from_reader::<ChatDelegateResponseMsg, _>(
@@ -181,9 +185,14 @@ impl ResponseHandler {
                             );
 
                             if let Ok(response) = deserialization_result {
+                                // Summary, NOT the full Debug. This single line was 98.5%
+                                // of all console bytes River emits during startup: a
+                                // `ListResponse` Debug-prints every key as a decimal byte
+                                // array, measured at ~401 KB per line across 39 lines
+                                // (14.93 MB of 15.15 MB total). freenet/river#569.
                                 info!(
-                                    "Successfully deserialized as ChatDelegateResponseMsg: {:?}",
-                                    response
+                                    "Successfully deserialized as ChatDelegateResponseMsg: {}",
+                                    describe_delegate_response(&response)
                                 );
 
                                 // For a response from a LEGACY delegate, the awaiting request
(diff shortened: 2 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #570

### Summary
- **PR Title:** perf(ui): stop dumping delegate payloads to the console at startup
- **Type:** perf (logging/observability only — no protocol, wire-format, or control-flow change)
- **CI Status:** all checks green on `9cb0caabdccdf7b6f580c5535768430ae13a70ae` (build, check-delegate-migration, check-room-contract-migration, ui-playwright-tests, license/cla)
- **Linked Issues:** river#569
- **Review tier:** Light — no high-risk surface (no concurrency, crypto/auth, migration, wire-format, routing, transport, or delegate-WASM-logic change); diff is self-contained to two UI files. Escalated to a second lens because the new summarization/truncation code is real logic, not mechanical churn.
- **Reviewers run:** two independent Claude lenses (skeptical/adversarial bug-hunt; code-first intent-vs-implementation) plus a testing pass and a big-picture pass, run directly against the checked-out PR code — **plus `codex review --base origin/main`**, which ran long (background, ~4+ min) but did complete and returned two findings, one of which independently corroborates a bug found by the Claude skeptical pass.

---

### Code-First Analysis
**Independent Understanding:** Replaces `{:?}`-Debug-dumping of decoded `ChatDelegateResponseMsg` values (and raw `DelegateKey`/payload bytes) in River's delegate-response handling with a bounded, human-readable summary (`describe_delegate_response`, `delegate_storage_key_str`, `delegate_key_fp`, `room_key_fp`). No control flow, error handling, or protocol logic changes — purely what gets written to the console.

**Stated Intent:** Same, per PR body: eliminate ~15MB/startup of Debug-dumped bytes (98.5% of River's startup console output, per the author's own census) without deleting any diagnostic that was actually used for triage.

**Alignment:** Matches. One minor discrepancy: the PR body says "No log levels changed," but the diff demotes three call sites from `info!` to `debug!` (`"Processing delegate response value #{}"`, the "processed flag" line, and "Deserialization as request result"). Read in context this means "the global `release_max_level_info` compile-time filter is untouched" (the very next sentence), not "no individual call site changed level" — but as written it's inaccurate at the per-call-site level. Cosmetic; doesn't affect correctness. **Consider.**

---

### Testing Assessment
**Coverage Level:** Adequate for the stated goal, one real gap (see below).

| Unit | ✅ | 6 new tests: key names preserved without byte-array dump; summary scales with names not bytes (relative to `Debug` length, not an absolute bound); `GetResponse` reports size not contents; `CasStoreResponse::Conflict` reports blob size not contents (added after an earlier round of this same PR reintroduced the defect — the test comment says so directly, a good sign of real review iteration already having happened); binary/non-UTF-8 key falls back to a length; source-scrape regression pin against the `{:?}` dump and the raw payload format returning, anchored past `mod tests {` so it can't match itself |
| Integration | N/A | Not needed for a pure logging change |
| Simulation | N/A | Not applicable — River UI, not core routing |
| E2E | ✅ | `ui-playwright-tests` green on this head SHA |

**Regression Test:** Present and valid — the PR description states it was mutation-verified (reintroducing `{:?}` on the response makes it fail); the pin's anchoring avoids the vacuous-pass failure mode this repo has hit before.

**Missing Tests:** No test exercises the `MAX`-length truncation branch of `delegate_storage_key_str` with a multi-byte UTF-8 key — a test like that would have caught Finding #1 below. No test exercises `ListResponse` with a large key count (Finding #2).

---

### Skeptical Findings
**Risk Level:** Low — both issues below are real but require inputs the current codebase does not produce.

| Char-boundary panic in defensive truncation | Should Fix (non-blocking) | `ui/src/components/app/freenet_api/response_handler.rs:2047` — `delegate_storage_key_str` | `&s[..MAX]` slices a `&str` at a fixed byte offset (128). Rust panics if that offset isn't on a UTF-8 char boundary. Every real key today (`outbound_dms`, `rooms_meta`, `room:<base58 vk>`) is ASCII and well under 128 bytes, so this is unreachable with current invariants — but it's exactly the case the function's own comment says it exists to guard against ("bounded by construction rather than by assumption about who builds keys later"): a non-ASCII UTF-8 key over 128 bytes whose 128th byte lands mid-character crashes this log call. **Found independently by both the Claude skeptical pass and `codex review`** (same file:line), which is corroborating rather than duplicate signal. Fix: truncate on a char boundary (e.g. `s.char_indices().take_while(|(i,_)| *i < MAX).last()` or similar) instead of a raw byte index. |
| `ListResponse` key-name join is uncapped | Should Fix (non-blocking) | `ui/src/components/app/freenet_api/response_handler.rs:2071-2075` — `describe_delegate_response`, `R::ListResponse` arm | Joins *every* key name (each already bounded to 128 chars by `delegate_storage_key_str`) with no cap on key count. The code comment explicitly reasons about this ("tens of them, not thousands") and today's real-account census (~40 keys) supports that. Flagged by `codex review`: if a legacy delegate store or migration path (the PR's own "Also found" section notes legacy-generation multiplication in the #425 discussion) ever produces a much larger key set, this line grows unbounded again — a smaller-magnitude echo of the exact problem this PR fixes. Fix: cap the rendered names (e.g. first 20 + "and N more") the same way the per-key length is already capped. |

No concurrency, auth, or protocol-surface issues — this is single-threaded synchronous formatting code in the UI's response-handling path, and no message sent over the wire changes.

---

### Big Picture Assessment
**Goal Alignment:** Yes — directly fixes river#569, backed by real profiling data (per-line byte census) rather than a guess.
**Anti-Patterns Detected:** None from this repo's bug-prevention list (freenet-core-specific: `biased;` select, spawn/channel patterns, etc. — none apply to synchronous UI logging code).
**Removed Code Concerns:** None — no tests removed, no behavior deleted beyond the intended Debug-dump removal.
**Scope Assessment:** Focused. The PR body explicitly calls out several adjacent perf/correctness issues found while profiling (#425, #523, #554) and deliberately does *not* fix them here — good scope discipline, matches this repo's "one logical change per PR" rule.

---

### Documentation
- Code docs: Complete — every new function/branch carries a comment explaining *why*, including the deliberate "no catch-all match arm" design choice (forces a compile error on new response variants instead of silently Debug-dumping them).
- Architecture docs: N/A
- User docs: N/A

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important, non-blocking here)
1. `delegate_storage_key_str`'s `&s[..MAX]` can panic on a non-char-boundary byte offset for a >128-byte multi-byte-UTF-8 key. Unreachable with today's key formats (independently confirmed by two reviewers), so not blocking, but recommend a fast-follow: char-boundary-safe truncation plus a test with a multi-byte key >128 bytes.
2. `ListResponse` summary joins an uncapped number of key names. Fine at today's real-world key counts (~40), but worth capping defensively given this PR exists specifically because an "it's always small" assumption about a different field turned out false at scale.

#### Consider (Suggestions)
1. Reconcile "No log levels changed" in the PR description with the three `info!`→`debug!` demotions in the diff (almost certainly means "the global `release_max_level_info` filter is unchanged" — just worth clarifying in the description).
2. De-duplicate the 4-byte hex fingerprint logic between `freenet_synchronizer.rs`'s new `key_head` block and `delegate_key_fp` in `response_handler.rs`.

---

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `9cb0caabdccdf7b6f580c5535768430ae13a70ae`

No blocking findings. Both Should-Fix items are real defects but require inputs the current codebase cannot produce (ASCII-only, short delegate keys; small key counts), and both are logging-only code paths — not the wire protocol, not contract/delegate state. This fix has been open since 2026-07-30 with green CI and zero reviews; merging now and tracking the two Should-Fix items as a fast-follow is the right tradeoff against leaving a validated fix to rot further. CI is green on the current head SHA and `mergeStateStatus` is CLEAN.

[AI-assisted - Claude]

Issue #167, opened 2026-03-13
fix: member list should include current user even if they haven't posted a message


sanity:
## Problem

When a user joins a room (e.g., via invitation acceptance), they don't appear in `riverctl member list` until they've posted a message. This means:

1. They can't see themselves in the member list
2. They can't rename themselves (since `set-nickname` requires knowing they're a member)
3. They can't see other member info that might be gated on membership

This was discovered when the Invite Bot joined the Freenet Official room — it showed as "Member" (default nickname) but didn't appear in the member list at all until after a nickname change was forced through.

## Expected Behavior

`member list` should include the current user (identified by their signing key in `rooms.json`) even if they haven't yet posted a message to the room. The user's membership is established at invitation acceptance time, not at first message.

## Steps to Reproduce

1. Create an invitation: `riverctl invite create <ROOM_VK>`
2. Accept it with a separate config: `riverctl --config-dir /tmp/bot invite accept "<CODE>" --nickname "Bot"`
3. List members from the accepting config: `riverctl --config-dir /tmp/bot member list <ROOM_VK>`
4. The newly joined user may not appear in the list

[AI-assisted - Claude]

PR #585, opened 2026-08-03
fix(ui): pin your own row to the top of the member list

### LINKED ISSUE #584: fix(ui): your own name isn't pinned to the top of the room member list
## Problem

The right-hand member list renders as an invite tree (owner → invitees), with
siblings sorted by `MemberId` (`ui/src/components/members.rs:512`). The local
user gets no special treatment, so they land wherever their inviter chain and
`MemberId` happen to put them — arbitrary, and in a large room, far down and
off-screen.

Finding yourself in the list is a routine action (checking your own nickname,
opening your own member-info), and today it's a scroll-and-hunt.

## Expected

The local user's own entry is pinned to the top of the member list, visually
distinct from the tree below it.

## Notes

The tree structure is load-bearing (it shows who invited whom), so this
probably wants self rendered as a pinned row above the tree rather than
re-sorting the tree itself.

[AI-assisted - Claude]

skandragon:
Closes #584.

Your own row now sorts to index 0 of the member list, so finding yourself isn't a scroll-and-hunt in a large room.

The rest of the list is a viewer-scoped invite/deputy tree whose order is load-bearing (it shows who invited and who deputized whom), so self is **hoisted** out of it rather than the tree being re-sorted — every other row keeps its relative position. `rotate_right(1)` on the prefix is the stable form of "move this element to the front".

No new styling: the ⭐ tag `member_display_parts` already emits is what makes the pinned row distinct.

## Test coverage

Three unit tests in `ui/src/components/members.rs`:

- `pin_self_to_top_hoists_self_and_preserves_the_rest` — self moves to index 0 and every other row keeps its relative tree order.
- `pin_self_to_top_is_a_noop_when_self_is_first_or_absent` — the owner (already the tree root at index 0) and a viewer not yet ingested into the list (the #167 shape) both leave the order untouched.
- `pin_self_to_top_is_wired_into_the_member_list` — source-grep pin on the call site. The hoist is only reachable through `MemberList`'s memo, which no unit test can drive, so deleting the call would otherwise leave the two behavioural tests green.

`cargo test -p river-ui --bins` passes. No contract/delegate/WASM changes, so no migration or redeploy steps beyond the normal UI publish.

### DIFF
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -331,6 +331,24 @@ fn member_display_parts(member: &MemberDisplay) -> MemberDisplayParts {
     }
 }
 
+/// Lift the viewer's own row to the top of the member list (freenet/river#584).
+///
+/// The rest of the list is a viewer-scoped invite/deputy tree whose order is
+/// load-bearing — it shows who invited and who deputized whom — so self is
+/// HOISTED out of it rather than the tree being re-sorted: every other row
+/// keeps its relative position. A `rotate_right` on the prefix is the stable
+/// form of "move this element to the front".
+///
+/// No-op when the viewer isn't in the list at all (they may have joined and
+/// not yet been ingested — the freenet/river#167 shape), or is the owner
+/// (already at index 0). The ⭐ tag from `member_display_parts` is what makes
+/// the pinned row visually distinct; no extra styling is needed here.
+fn pin_self_to_top<T>(rows: &mut [(T, MemberId)], self_id: MemberId) {
+    if let Some(pos) = rows.iter().position(|(_, id)| *id == self_id) {
+        rows[..=pos].rotate_right(1);
+    }
+}
+
 /// Order member IDs by DFS pre-order traversal of the invite tree.
 /// Owner is the root; within siblings, order matches `members.members`
 /// (sorted by MemberId after CRDT convergence).
@@ -1490,6 +1508,11 @@ pub fn MemberList() -> Element {
 
             all_members.push((member_display_parts(&member_display), member_id));
         }
+
+        // freenet/river#584: your own row goes to the top, so finding yourself
+        // isn't a scroll-and-hunt in a large room. Everything below keeps its
+        // tree order.
+        pin_self_to_top(&mut all_members, self_member_id);
 
         Some(all_members)
     })()
@@ -3538,6 +3561,37 @@ mod tests {
         let icons: Vec<&str> = parts.tags.iter().map(|t| t.glyph).collect();
         assert!(icons.contains(&"👑"));
         assert!(icons.contains(&"⭐"));
+    }
+
+    fn mid(n: i64) -> MemberId {
+        MemberId(freenet_scaffold::util::FastHash(n))
+    }
+
+    fn ids(rows: &[((), MemberId)]) -> Vec<MemberId> {
+        rows.iter().map(|(_, id)| *id).collect()
+    }
+
+    /// freenet/river#584: self moves to index 0, and every other row keeps its
+    /// relative tree order (this is a hoist, not a re-sort).
+    #[test]
+    fn pin_self_to_top_hoists_self_and_preserves_the_rest() {
+        let mut rows: Vec<((), MemberId)> = (0..5).map(|n| ((), mid(n))).collect();
+        pin_self_to_top(&mut rows, mid(3));
+        assert_eq!(ids(&rows), vec![mid(3), mid(0), mid(1), mid(2), mid(4)]);
+    }
+
+    /// The owner viewing their own room is already at index 0 (tree root), and
+    /// a viewer not yet present in the list must not reorder anything.
+    #[test]
+    fn pin_self_to_top_is_a_noop_when_self_is_first_or_absent() {
+        let mut rows: Vec<((), MemberId)> = (0..3).map(|n| ((), mid(n))).collect();
+        let before = ids(&rows);
+
+        pin_self_to_top(&mut rows, mid(0));
+        assert_eq!(ids(&rows), before);
+
+        pin_self_to_top(&mut rows, mid(99));
+        assert_eq!(ids(&rows), before);
     }
 
     /// The 🛡 deputy shield renders exactly when `deputized_by` is non-empty,
--- a/ui/src/components/members.rs
+++ b/ui/src/components/members.rs
@@ -333,11 +333,22 @@ fn member_display_parts(member: &MemberDisplay) -> MemberDisplayParts {
 
 /// Lift the viewer's own row to the top of the member list (freenet/river#584).
 ///
-/// The rest of the list is a viewer-scoped invite/deputy tree whose order is
-/// load-bearing — it shows who invited and who deputized whom — so self is
-/// HOISTED out of it rather than the tree being re-sorted: every other row
-/// keeps its relative position. A `rotate_right` on the prefix is the stable
-/// form of "move this element to the front".
+/// The rest of the list is a viewer-scoped invite/deputy tree
+/// (`deputy_display_order`, #410), so self is HOISTED out of it rather than
+/// the tree being re-sorted: every other row keeps its relative position, and
+/// `rotate_right` on the prefix is the stable form of "move this element to
+/// the front".
+///
+/// What the hoist DOES cost, stated plainly so nobody re-derives it as a bug:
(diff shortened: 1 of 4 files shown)


sanity:
## Review — 3 independent Claude lenses (skeptical bug-hunt, supply-chain/security, big-picture + test quality)

Full tier, because the PR is from a first-time external contributor and River's UI publishes straight to a shared Freenet contract key that every user loads. Reviewers read the checked-out code blind, not just the diff.

**Verdict: the behaviour is correct and safe. The tests were not.** I've pushed one commit (b8f99f6) fixing the findings rather than bouncing it back — details below.

### Security / supply-chain: clean

- Branch touches exactly one file (`git diff main...HEAD --numstat`), `--summary` empty: no new files, submodules, symlinks, mode changes, or binary blobs. Nothing under `.github/`, `Cargo.toml`/`Cargo.lock`, `Makefile.toml`, `package.json`, `.cargo/`, `build.rs`, `contracts/`, `delegates/`, `common/`, or the publish scripts. CI's `check-delegate-migration` and `check-room-contract-migration` independently confirm the WASM is byte-identical, so "no contract/delegate/WASM changes" is machine-verified, not just eyeballed.
- Trojan Source scan of added lines (bidi/zero-width/invisible codepoints): zero hits.
- `pin_self_to_top` takes only a `MemberId` derived from the viewer's own **public** verifying key. No key material, no network call, no `unsafe`, no `dangerous_inner_html`, no panic path (`pos` comes from `position()`, so `rows[..=pos]` is in-bounds by construction).
- The hoist cannot be used to displace a trust signal: badges travel with the row, the impersonation verdict is computed before the hoist and never takes position as an input, and the "self" match is against the viewer's own key on the viewer's own device — an attacker cannot make their row the pinned one.
- `include_str!` is inside the single `#[cfg(test)]` module, so it cannot reach the shipped WASM.

### Blocking finding (fixed in b8f99f6): the call-site pin was vacuous

`pin_self_to_top_is_wired_into_the_member_list` grepped the raw `include_str!("members.rs")` for `pin_self_to_top(&mut all_members, self_member_id)`. That needle occurs **twice** — the real call, and the test's own string literal. So the assertion matched itself:

```
needle occurrences in PRODUCTION : 1
needle occurrences in TEST module: 1
after deleting the prod call, include_str!(..).contains(needle) == True
```

Two reviewers confirmed it by compiling and running the test with the call site deleted: it passed. Its own doc comment and the PR body both claim it guards exactly that deletion, so the PR's stated coverage of its own behaviour change was zero.

This is the trap `members.rs` already has two helpers for — `production_source()` (line ~4937) cuts at the first `#[cfg(test)]` and backs ten sibling pins; `prod_source()` documents the hazard verbatim ("a needle could match inside test code and the pin goes silently vacuous"). The technique is fine and well-established here; this one instance just bypassed the guard.

**Fixed:** now uses `production_source()`. Mutation-verified both directions — deleting the call passed before the fix, fails after.

### Also in b8f99f6

- **New Playwright spec** `ui/tests/member-list-self-pinned.spec.ts`. A source-grep proves the call exists; it cannot prove the row lands first in the DOM, and the hoist runs inside a `use_memo` no unit test can drive. The spec asserts row 0 carries `member-list-self` and **not** `member-list-owner` — the owner is the invite-tree root and is index 0 without the pin, which is what makes it non-vacuous rather than a fixture coincidence. Verified by rebuilding the UI with the call site deleted and watching it go red. Passes on all 5 browser projects.
- **Self at the last index** — the maximal rotation, where an off-by-one in the `..=pos` bound would show and at an interior index would not.
- **Doc comment corrected.** It claimed the tree order "is load-bearing — it shows who invited and who deputized whom" and that the hoist preserves it. Both overstated: the list renders as a flat `ul` with no indentation, so ordering never conveyed parentage (the 🔑 🌐 🎪 🔭 🛡 badges do), and the hoist genuinely does detach self's own invitees/deputies from self. The design is fine — but for the opposite reason to the one given, so the comment now says so. This is the one place the PR's rationale was self-defeating: "the order is load-bearing, therefore hoist" — hoisting is what detaches a subtree.
- **Stale comment** in `invite-via-dm-picker.spec.ts` ("the 'You' entry can appear at any position") updated; that helper doesn't depend on the position either way.

### Non-blocking, not changed

- **Second half of #584 not delivered.** The issue asks for pinned *and* "visually distinct from the tree below it". The ⭐ already marked your row before this change, so the row gains no *pinned*-ness signal — a non-owner now simply appears above the owner with no separator. Left as-is: it's a design call for a follow-up, not a defect.
- **`deputy_display_order`'s doc** ("a global mod rises to the top in every view") is now not literally true of the rendered list for a non-owner viewer. The function's own output is unchanged; documentation only.

### Verification

- `cargo test -p river-ui --bins`: 817 passed, 0 failed. `cargo fmt --check` clean.
- Local Playwright: 727 passed. The 10 failures are `responsive-layout.spec.ts`'s two sandboxed-iframe tests × 5 browsers, which hardcode `src="http://localhost:8082"` and so ignore `PLAYWRIGHT_BASE_URL`; port 8082 was held by another worktree's `dx serve`, so I ran on 8399. Environmental, unrelated to this change, and green in CI.
- Rendered in a real browser: `Ben Rodriguez (You) ⭐` at index 0, owner below.

Lenses run: skeptical bug-hunt, supply-chain/security, big-picture + test quality. Every finding above is either fixed or has a stated reason for not being.

[AI-assisted - Claude]

sanity:
## Follow-up: my own fix commit went through an independent adversarial review too

Per the "reviews are per-code-content" rule, b8f99f6 changed the code, so it got its own blind review. No blocking findings, but two worth acting on — pushed as 2a11180.

**The pin is now scoped to `MemberList`'s memo**, not just to production code. `production_source()` closed the self-matching hole but was the weaker of the two helpers this file offers, and left two other ways to rot. It now uses `prod_source` + the `pub fn MemberList()` / `let handle_member_click` anchors — the same pair, for the same reason, as `impersonation_warning_is_wired_into_every_render_surface`, whose comment reads "Scoping is the entire point — do not widen these ranges."

Mutation-verified against all three failure modes:

| mutation | original | b8f99f6 | 2a11180 |
|---|---|---|---|
| call site deleted | **passed** ❌ | FAILED ✅ | FAILED ✅ |
| call site commented out | **passed** ❌ | **passed** ❌ | FAILED ✅ |
| call moved out of `MemberList` into a production helper | **passed** ❌ | **passed** ❌ | FAILED ✅ |

The comment-out case is closed by `prod_source` stripping line comments (which `production_source` does not — that difference is now documented on both). The moved-out case is closed by the anchors.

Also in 2a11180, all from that review:

- The last-index test's justifying comment was **wrong**. I'd written that an off-by-one in the `..=pos` bound "is invisible at an interior index" — it isn't; both `..pos` and `..=pos + 1` are caught by the interior assertion two lines above. What self-at-last uniquely adds is the out-of-bounds direction: `..=pos + 1` panics there and merely permutes at an interior index. Comment corrected; the assertion itself was right.
- `production_source()`'s doc said "used by the two source-grep pins below" — there are ten, and b8f99f6 added the first one *above* it. Fixed, and it now points at `prod_source` for pins that need comment-stripping.
- The new spec's `expect(await rows.count()).toBeGreaterThan(1)` was the file's one non-retrying assertion, on a list with a documented flake of exactly that shape. Now `await expect(rows.nth(1)).toBeVisible()`, and it asserts the room click actually landed so a miss reports that instead of timing out later.
- b8f99f6 fixed the stale "the (You) entry can appear at any position" comment in `invite-via-dm-picker.spec.ts` but not the identical sentence in its twin `dm-thread-modal.spec.ts`. Both fixed now.

Reviewer independently confirmed the spec's non-vacuity from the rendered DOM rather than taking my word: pre-pin order in Team Chat Room recovers as `[owner, deputy, self, impostor]`, so index 0 without the pin genuinely is the owner. It also noted the `toHaveCount(0)` owner assertion earns its place for a reason I hadn't thought of — in "Your Private Room" the local user *is* the owner and row 0 carries both testids, so if the fixture ever flipped Team Chat Room to `SelfIs::Owner`, "row 0 is self" would pass vacuously and that assertion catches it.

Verification on 2a11180: `cargo test -p river-ui --bins` 817 passed / 0 failed, `cargo fmt` clean, and the new plus both edited specs pass 45/45 across all five browser projects locally.

[AI-assisted - Claude]

PR #589, opened 2026-08-03
fix(ui): stop rendering quoted text black inside own message bubbles

sanity:
## Problem

A `> quoted` line inside the local user's **own** message bubble rendered near-black while every other word in the same bubble was white.

`.bg-accent` is the sent-message bubble — solid brand blue with `text-white`. The blockquote override painted the quote `rgba(0, 0, 0, 0.8)`:

```css
.bg-accent .prose blockquote {
  border-left: 3px solid var(--color-accent-hover);
  color: rgba(0, 0, 0, 0.8);
}
```

That composites to **3.52:1** against the bubble — under the 4.5:1 WCAG AA floor, and *inverted* relative to the 5.23:1 white text right beside it. `--color-accent` is the same blue in light and dark mode, so this was broken in **both** schemes, not only the dark mode it was reported against. The border made it worse: `--color-accent-hover` on `--color-accent` is **1.29:1**, effectively invisible, so a quote in a sent bubble had no marker beyond the indent.

Measuring the other quote surfaces turned up two more sub-AA cases, both from `--color-text-muted` — and both in **light** mode:

| surface | before | after |
|---|---|---|
| sent bubble, dark | **3.52:1** | 5.23:1 (matches its normal text) |
| sent bubble, light | **3.52:1** | 5.23:1 (matches its normal text) |
| outgoing DM bubble, light | **3.96:1** ← worst in the app | 4.97:1 |
| received bubble, light | **4.47:1** | 5.61:1 |
| outgoing DM bubble, dark | 5.56:1 | 7.55:1 |
| received bubble, dark | 5.93:1 | 8.05:1 |
| sent-bubble quote bar | **1.29:1** | 3.34:1 |

## Approach

**Sent bubbles take the bubble's own text colour** (`color: inherit`) instead of a hardcoded one, so the quote can't drift from the bubble's palette again, plus a translucent-white bar for the affordance. Dimming is deliberately *not* how the quote is marked here: white on the accent blue is only 5.23:1 to begin with, so any meaningful reduction falls under AA. This is the same convention `.bg-accent .prose a` (white + underline) and the reply strip (`bg-white/25 text-white/90`) already use.

**A dedicated `--color-text-quote` token** for the non-accent surfaces, rather than nudging `--color-text-muted`. The muted token is tuned for glanceable metadata (timestamps, field labels) and changing it would repaint every timestamp, placeholder and label in the app; a quote is body text and needs to clear AA. This part goes beyond the literal report — it is the "any similar issue in light mode" half — and is easy to drop if you'd rather keep the muted colour.

## Testing

`ui/tests/blockquote-contrast.spec.ts` (new) measures three real surfaces × both colour schemes: the sent bubble, the received bubble, and the outgoing DM bubble (which has no example data, so the test sends a real DM to create one). It composites the colours the browser computed and measures WCAG contrast. Beyond the AA floor it asserts:

- the quote must not sit on the opposite side of the background's luminance from the normal text in the same bubble — the reported symptom, stated directly;
- outside the accent bubble, the quote must stay visibly *secondary* to its normal text, so a change that makes it merely legible by collapsing it into the body colour also fails.

**Every assertion is mutation-verified** against the served stylesheet:

| mutation | caught by | observed |
|---|---|---|
| revert the sent-bubble rule | accent tests, both schemes | quote `rgb(0,20,46)` vs normal `rgb(255,255,255)` over `rgb(2,101,231)` |
| revert the base rule to `--color-text-muted` | light received + light DM | 4.470 and 3.961 vs the 4.5 floor |
| delete the **light** `--color-text-quote` | light received + light DM | quote collapses to body colour, 14.60 vs a 11.68 ceiling |
| delete the **dark** `--color-text-quote` | dark received + dark DM | 2.487 and 2.331 vs the 4.5 floor |
| revert the border only | accent tests, both schemes | 1.285 vs the 3.0 floor |
| `color: inherit; color: #111;` (cascade) | Rust sent-bubble pin | winning declaration is `#111`, not `inherit` |

The Rust pins in `conversation.rs` cover the declarations themselves — parsing the *winning* `color` declaration rather than substring-matching, since CSS is last-one-wins — and that `> …` still reaches the stylesheet as a `<blockquote>` through `message_to_html`'s hard-break rewrite.

Full suite: `cargo test -p river-ui --bins` (817) and `--features example-data,no-sync` (822) pass; Playwright passes across chromium / firefox / webkit / mobile-chrome / mobile-safari.

## A wrong number in the first revision of this PR

The original description claimed `--color-text-muted` gave **4.40:1 on a dark-mode DM bubble**. That was wrong in both the scheme and the value, and review caught it. Two compounding causes, both mine:

1. I measured a synthetic page that put the DM bubble on `--color-bg`, not the modal's panel background.
2. My colour helper divided by alpha to "un-premultiply" a `getImageData` readback. `getImageData` already returns straight alpha (verified: `rgba(255,0,0,0.2)` reads back as `[255,0,0,51]`, not `[51,0,0,51]` — in chromium, firefox and webkit alike), so every translucent colour was inflated and clamped. `bg-accent/20` resolved to `rgb(204,255,255)` instead of `rgb(204,224,250)`.

The real DM modal measures **5.56:1 dark** (already fine) and **3.96:1 light** — the worst quote surface in the app, which the wrong number hid entirely. The fix helps it either way, but the justification named the wrong scheme. The spec now measures that surface directly, so the number can't drift from reality again.

## Drive-by: two Playwright tests were measuring the wrong build

`responsive-layout.spec.ts`'s two sandboxed-iframe tests hardcoded `src="http://localhost:8082"` and ignored `PLAYWRIGHT_BASE_URL`, so running the suite on any other port silently pointed them at whatever build occupied 8082 — the shared-port trap AGENTS.md warns about. They failed for exactly that reason during this work (10 failures across all 5 projects) and pass once pointed at the configured base URL, which is what the other specs already do. With the variable unset the fallback is byte-identical to the old literal, so CI behaviour is unchanged.

## Not changed, but measured

- The quote bar in a **dark-mode DM bubble** is 2.36:1 (`--color-accent` on a `bg-accent/20` tint). Decorative, and the quote is identifiable without it, so I left the received/DM bar alone rather than restyling a surface this bug isn't about.
- `.river-mention` in dark mode is ~3.2:1 (accent text on a dark surface). Separate element, separate call — flagging it, not touching it here.
- Two pre-existing flakes surfaced during full-suite runs under heavy CPU load, each passing on retry and each a different test (`invitation-sandbox.spec.ts:197` and `message-layout.spec.ts:485`, both mobile-safari). Neither reproduced on focused repeats (8/8 and clean respectively); unrelated to this diff.

[AI-assisted - Claude]

### DIFF
--- a/ui/assets/tailwind.css
+++ b/ui/assets/tailwind.css
@@ -35,6 +35,7 @@
   --color-border-hover: #d0d0d4;
   --color-text: #1a1a1a;
   --color-text-muted: #6b6b6b;
+  --color-text-quote: #5c5c5c;
   --color-text-inverse: #ffffff;
   --color-success-bg: #f0fdf4;
   --color-warning-bg: #fefce8;
@@ -53,6 +54,7 @@
     --color-border-hover: #3a3a3e;
     --color-text: #f0f0f2;
     --color-text-muted: #9a9a9e;
+    --color-text-quote: #b4b4b9;
     --color-text-inverse: #1a1a1a;
     --color-success-bg: #052e16;
     --color-warning-bg: #422006;
@@ -181,19 +183,49 @@ html, body {
   padding: 0;
 }
 
-/* Blockquotes in prose */
+/* Blockquotes in prose (received bubbles, DM bubbles, room descriptions).
+ *
+ * `--color-text-quote` rather than `--color-text-muted`: a quote must stay
+ * READABLE while reading as secondary, and the muted token — which is tuned
+ * for timestamps and field labels, i.e. glanceable metadata — lands just under
+ * the 4.5:1 WCAG AA floor on two of the backgrounds a quote actually appears
+ * over (4.47:1 on light-mode `--color-surface`, 4.40:1 on a dark-mode
+ * `bg-accent/20` DM bubble). The quote token is one step further from the
+ * background in each mode, which clears AA on every quote surface while
+ * staying well clear of `--color-text` so the muted look survives. It is
+ * deliberately a SEPARATE token: nudging `--color-text-muted` itself would
+ * repaint every timestamp, placeholder and label in the app. */
 .prose blockquote {
   border-left: 3px solid var(--color-accent);
   padding-left: 1rem;
   margin-left: 0;
-  color: var(--color-text-muted);
+  color: var(--color-text-quote);
   font-style: italic;
 }
 
-/* Blockquotes in own message bubbles (blue bg) */
+/* Blockquotes in own message bubbles (solid brand-blue bg, white text).
+ *
+ * This rule used to paint the quote `rgba(0, 0, 0, 0.8)` — near-black text in
+ * a bubble whose every other word is white. That is the same failure the
+ * `.bg-accent .river-mention` / `.bg-accent .prose a` overrides in main.css
+ * exist to prevent, just for a different element: `.bg-accent` is the SAME
+ * brand blue in light and dark mode, so the quote read as black-on-blue in
+ * BOTH (3.52:1, under the 4.5:1 WCAG AA floor, against 5.23:1 for the normal
+ * text right beside it). The `--color-accent-hover` border made it worse — at
+ * 1.29:1 against the accent background it was invisible, so a quote in a sent
+ * bubble had no marker at all beyond the indent.
+ *
+ * Take the bubble's own text colour (`inherit` → white) rather than a
+ * hardcoded one, so the quote can never drift from the bubble's palette again.
+ * Dimming is deliberately NOT how the quote is distinguished here: white on
+ * the accent blue is only 5.23:1 to begin with, so any meaningful reduction
+ * drops under AA. The affordance is carried by the border + the inherited
+ * `font-style: italic` instead — the same convention `.bg-accent .prose a`
+ * (white + underline) and the reply strip (`bg-white/25 text-white/90`) use
+ * for sent bubbles. */
 .bg-accent .prose blockquote {
-  border-left: 3px solid var(--color-accent-hover);
-  color: rgba(0, 0, 0, 0.8);
+  border-left-color: rgba(255, 255, 255, 0.7);
+  color: inherit;
 }
 
 /* Lists in prose (Tailwind reset strips list-style) */
--- a/ui/assets/tailwind.css
+++ b/ui/assets/tailwind.css
@@ -187,14 +187,25 @@ html, body {
  *
  * `--color-text-quote` rather than `--color-text-muted`: a quote must stay
  * READABLE while reading as secondary, and the muted token — which is tuned
- * for timestamps and field labels, i.e. glanceable metadata — lands just under
- * the 4.5:1 WCAG AA floor on two of the backgrounds a quote actually appears
- * over (4.47:1 on light-mode `--color-surface`, 4.40:1 on a dark-mode
- * `bg-accent/20` DM bubble). The quote token is one step further from the
- * background in each mode, which clears AA on every quote surface while
- * staying well clear of `--color-text` so the muted look survives. It is
- * deliberately a SEPARATE token: nudging `--color-text-muted` itself would
- * repaint every timestamp, placeholder and label in the app. */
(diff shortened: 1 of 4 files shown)


sanity:
## Review record

Risk tier: **Light** (stylesheet + tests; no concurrency, crypto, auth, migration, wire-format or CI surface). Two independent lenses, both blind to the author's reasoning.

### Lens 1 — testing (subagent)

Ran 12 mutations, 10 caught, 2 not. Reproduced the headline figures exactly (4.47:1, 1.29:1, 5.23:1). Three findings, all addressed in ee31997:

| finding | disposition |
|---|---|
| The DM-bubble evidence is wrong: real modal is 5.56:1 **dark** and 3.96:1 **light**, not "4.40:1 dark" | **Fixed.** Confirmed independently before acting — the cause was `resolve()` dividing by alpha on an already-straight-alpha `getImageData`, plus measuring a synthetic page instead of the modal. Both comments corrected; the spec now measures the real modal. |
| `resolve()` double-un-premultiplies alpha | **Fixed.** Verified the premise directly in all three engines: `rgba(255,0,0,0.2)` → `[255,0,0,51]`. Latent for opaque colours, live for the DM bubble. |
| Rust pin passes with `color: inherit; color: #111;` | **Fixed.** Now parses the winning declaration. Re-ran the mutation: previously 3 passed, now FAILED. |
| No coverage of any `bg-accent/20` DM bubble | **Fixed.** Added, both schemes; it is the tightest post-fix margin in the app at 4.97:1. |
| No ceiling assertion — deleting the light token collapses the quote into body colour and every floor-only test passes | **Fixed.** Added; that mutation now fails at 14.60 vs an 11.68 ceiling. |
| Doc comment states the `var()` failure mode backwards | **Fixed.** Dropping the *dark* value leaves `:root`'s light value resolving at 2.49:1; dropping the *light* value is what invalidates it. |
| 2 of 4 browser tests not load-bearing (light/dark duplication on the accent bubble) | **Kept deliberately**, now with a comment saying why: `.bg-accent` is the same blue in both schemes, and the duplication is what pins that. |

Cleared on inspection: bubble selection cannot confuse a DM bubble for a sent one (`bg-accent/20` is a distinct class token); determinism 20/20 across projects; `oklab()`/`color-mix()` parse correctly in all three engines; nested quotes do not compound.

### Lens 2 — adversarial / reach (run in-session)

The second subagent went idle without reporting, so this lens was run directly. Findings, with what was checked:

- **`.prose blockquote` reach** — exactly three `.prose` containers exist app-wide: the room message body (`conversation.rs:5206`), the room header description (`conversation.rs:3895`), the DM bubble (`dm_thread_modal.rs:1126`). Two are measured by the new spec; the header description on `bg-panel` computes to 6.69:1 light / 8.91:1 dark, better than the muted token it replaces. No regression on any of the three.
- **`.bg-accent .prose blockquote` reach** — `.bg-accent` appears on two elements: the sent bubble (`bg-accent text-white`, `conversation.rs:5078`) and the edit container (`bg-accent`, no `text-white`, `conversation.rs:4884`). The edit container holds a textarea, not a `.prose`, so the rule cannot match there. `bg-accent/20` is a separate class token and is correctly not matched — confirmed by measurement, the DM quote resolves via the base rule.
- **Cascade** — `.bg-accent .prose blockquote` (0,2,1) beats `.prose blockquote` (0,1,1), so the `border-left-color` longhand overrides the shorthand's colour component. Verified empirically: 3.34:1 in place, 1.285:1 when reverted.
- **Nested content in a sent-bubble quote** — inline `code` goes through `.bg-accent .prose code { color: inherit }`, so it now inherits white instead of near-black (an incidental improvement). Links and `.river-mention` have their own `.bg-accent` overrides and are unaffected.

**Nothing blocking from either lens.** Every finding is fixed rather than dismissed.

### Known, not changed

- Dark-mode DM quote bar: 2.36:1 (`--color-accent` on an accent tint). Decorative; the quote is identifiable without it.
- `.river-mention` in dark mode: ~3.2:1. Separate element, separate call.

[AI-assisted - Claude]

Issue #592, opened 2026-08-03
chat-delegate: DeleteRequest rewrites the whole key index unconditionally, and an unparseable index silently becomes empty


sanity:
Found while reviewing #587. **Delegate-WASM change — deliberately not fixed there**, because re-keying the delegate to fix a latent handler bug mints another generation, which is the exact failure mode #586 is about. This should land with the next delegate bump that happens for other reasons.

## 1. `handle_delete_request` rewrites the key index unconditionally

`delegates/chat-delegate/src/handlers.rs::handle_delete_request` calls `set_key_index` even when the key was not in the index. `handle_store_request` and the CAS-store path both rewrite only when adding a genuinely new key. So a no-op delete still triggers a whole-index rewrite.

## 2. `get_key_index` turns an unparseable index into an empty one

```rust
…from_reader(...).ok().unwrap_or_default()
```

An index secret that exists but fails to parse becomes an **empty** `KeyIndex`. Combined with (1), one unreadable index read during a delete writes an empty index over the real one: every room key vanishes from `ListResponse`, the values remain on disk but are unreachable, and there is no error. Silent, total, permanent.

The exposure is not new — `handle_store_request` uses the same `get_key_index` when adding a key — but until #587 the client had never sent a `DeleteRequest` at all, so the delete path has never executed in production.

## Suggested fixes

- Make the delete a no-op on the index when the key is not present, matching Store's shape.
- Make `get_key_index` distinguish *absent* (legitimately empty) from *corrupt* (refuse rather than silently truncate). A corrupt index should fail the request loudly, not erase the user's keys.

## Client-side mitigation already in place

#587's branch only sends the `DeleteRequest` when the marker is believed present, reading the delegate-seeded tri-state (so it works in the deployed app, where localStorage throws). That reduces the unconditional rewrite from once per migrating user to once per genuinely-interrupted migration. It does not address either bug above.

[AI-assisted - Claude]

sanity:
The identity-destruction consequence of this issue's empty-index behaviour is written up on #588: an unparseable index makes rooms invisible to `ListRequest` while they stay readable by direct key, which routes the loader into the legacy probe and ends with an older delegate generation's identity CAS-written over the stored one. `self_sk` is unrecoverable.

[...]

Issue #588, opened 2026-08-03
perf(ui): a user with no rooms re-fires all 81 legacy-delegate probes on every page load


sanity:
Split out of #586 / #587.

A user with genuinely **no** rooms anywhere never reaches a completion point that calls `mark_legacy_migration_done()`, so `LoadPlan::ProbeLegacy` fires the whole legacy fan-out on *every* page load, forever: 27 generations × {`rooms_data` GET, `outbound_dms` GET, `ListRequest`} = **81 delegate requests**.

Measured in-browser against try.freenet.org with a fresh profile: 27 probes dispatched, and every one of the 27 legacy delegates answered. Node-side, `crates/core/src/contract/executor/runtime/delegates.rs:433` ("Delegate not found in store (expected for migration probes)") is rate-limited at 30/s and still *drops* 300–600 events/hour.

#587 does not change this. It fixes the case where the migration never converges (the quota exhaustion) and makes the seal durable, but it deliberately leaves the "nothing responded with data" case unsealed.

## Why it is deliberately unsealed today

The empty-`rooms_data` legacy branch must NOT mark migration done — a legacy delegate returning no single blob may still hold dynamic `room:<vk>` keys that `migrate_legacy_per_room` has yet to discover, and an early mark-done permanently strands them. That is the V24→V25 data loss, pinned by `legacy_empty_rooms_data_does_not_mark_migration_done`.

## What a fix would need

A completion point that is genuinely safe: the fan-out has gone quiescent, **every responding generation reported nothing**, and `SAW_FETCH_FAILURE` is clear (so a connection that died mid-fan-out can never be mistaken for "nothing to migrate"). A non-responding generation can never yield data later either — River only ever registers the current delegate, so a generation absent from the node's delegate store stays absent.

That is a seal-on-absence rule, and getting it wrong strands rooms rather than merely wasting requests, so it wants its own change and its own review rather than riding along with #587.

[AI-assisted - Claude]

sanity:
## Repairing the seal SWAPS which data-loss route is armed — net exposure probably goes up

Investigating a separate question (whether `reconcile_room_present`'s rank-blind diverged-identity branch is reachable), a review lens established that the localStorage breakage is **not simply suppressing this issue's probe storm**. The same failure that makes `mark_legacy_migration_in_progress` unsettable also makes `mark_legacy_migration_done` unsettable — and the second is what would otherwise stop the legacy probe firing at all.

So a seal that works has two effects, not one.

[...]

Issue #420, opened 2026-07-19
Persisted identity-generation counter — structural fix for identity-overwrite async-race reversals (multi-tab + in-flight writers)


sanity:
## Context

PR #415 adds an "Import Identity over an existing one" escape hatch (issue #414) — a user stuck on a room identity they can no longer use can import a different one over it. It ships as a get-unstuck tool, with a **known multi-tab limitation documented here**.

## The limitation

When a room already has an identity and the user imports a replacement, the UI overwrite path updates `ROOMS[owner].self_sk` and re-saves the per-room delegate slot (`room:<owner_vk>` → `RoomSlot::Present`). But if a **second tab or device** for the same room is still open holding the OLD identity, that session will, on its next per-room save, write the OLD identity back as `RoomSlot::Present`.

The reconciler `chat_delegate::reconcile_room_present` is **local-authoritative on save**: when the delegate slot and the local copy disagree on `self_sk` (`local.self_sk != remote.self_sk`), it keeps the *saving tab's* local identity and does not merge the other identity's state. There is no way to tell which of the two identities is *newer*, so it is effectively last-writer-wins. On the next cold load, whichever identity is in the `Present` slot wins — so a stale tab can silently undo the replacement.

## Why it's out of scope for #414

#414 is a get-unstuck escape hatch, not full multi-tab identity coordination. The confirm dialog now tells the user to close other tabs/devices for the room before replacing, which avoids the reversal in practice.

## Proper fix

Add a **persisted, monotonically-increasing identity-generation counter** per room (alongside `self_sk` in `RoomData` / the per-room slot). An overwrite bumps the generation. `reconcile_room_present` then adopts the slot with the **higher generation** on an identity conflict (instead of last-writer-wins), so a stale `Present` writer holding an older generation loses and adopts the newer identity. This is a delegate/persisted-state change (new field on the per-room slot; `#[serde(default)]` for back-compat) and needs its own design + migration consideration.

## References

- PR #415 (issue #414) — the overwrite escape hatch.
- `ui/src/components/app/chat_delegate.rs::reconcile_room_present` — the diverged-identity branch (local-wins).
- `ui/src/components/members.rs::complete_identity_import` — the overwrite site.

[AI-assisted - Claude]

---

## Broader scope (from the #415 design review, 2026-07-19)

[...]

PR #595, opened 2026-08-04
fix(ui): stop a legacy-sourced copy overwriting the stored room identity

sanity:
## Problem

`reconcile_room_present`'s diverged-identity arm keeps `local` unconditionally and never consults rank, while the merge path (#527, #590) does. The save path contains **zero** rank reads. So an identity loaded from an older delegate generation can be CAS-written over the current delegate's slot, destroying `self_sk` — which cannot be re-derived from the network.

**Reachability** (freenet/river#588 route 2): #592's `get_key_index` turns an unparseable index into an **empty** one, while `handle_get_request` and `handle_get_versioned_request` read secrets *directly* and never consult the index. A room therefore becomes invisible to `ListRequest` while remaining fully readable by key — which routes the loader into `ProbeLegacy`, whose only guard is a localStorage flag that reads false forever in the sandboxed opaque-origin iframe. A legacy generation then answers with the old identity and the re-save overwrites the stored one.

No code path blocks that chain, and the terminal step was executed rather than reasoned about (`stored_self_sk_survived=false`). Whether the index actually corrupts **in the field** is unestablished — quota exhaustion was investigated and cleanly ruled out (the quota check precedes the write; the write is tmp+rename atomic). So this is **cheap insurance against an unrecoverable loss**, not a response to an observed incident.

### Why not "keep the higher-ranked identity"

That phrasing does not survive contact with the code, and the reason is #590's session boundary again: **there is only ONE rank in existence here.** `MergeRanks` is never persisted and `RoomSlot::Present` carries no provenance, so there is nothing to compare the stored slot against. Any implementation quietly becomes "compare local's rank to a constant", and the constant is then the entire fix.

What *is* decidable is whether the copy we hold is legacy-sourced. If it is, the stored slot was written by the current delegate, which outranks every legacy generation by construction, so refusing is always right. Refusing costs `room_state`, which the network can re-derive. Writing costs `self_sk`, which nothing can.

### The default IS the fix

An **absent** rank entry does not mean "unknown, assume the worst". `record_identity_source` deliberately records nothing for a room that is already present and unranked — exactly how an in-session create or import looks — so absent must read as `SOURCE_RANK_AUTHORITATIVE`. `unwrap_or(0)` would make every locally created room overwritable by any legacy copy: strictly worse than having no guard.

It is extracted into `identity_rank_for_save` and pinned, because **the first mutation run showed `unwrap_or(0)` SURVIVING** — the behavioural tests call `reconcile_room_present` directly with an explicit rank and never exercised the caller's default. Same untested-half pattern that produced four findings in #593.

The rank is read **once before** the per-room loop, not inside the CAS closure: the closure re-runs per retry and this is the same non-reentrant mutex that produced a near-deadlock in #590's drain.

## Testing

Four-case matrix driving the real function against real slot bytes:

| case | expected |
|---|---|
| legacy rank + differing identity | `Ok(None)`, stored `self_sk` intact |
| current-delegate rank + differing | writes local (**#420 unchanged**) |
| absent rank + differing | writes local (**#414 unchanged**) |
| legacy rank + **matching** identity | normal merge, still persists |

The last case is the one that catches an over-broad guard that would silently stop persisting every migrated room. Plus a test that `Absent`/`Tombstone` paths never consult rank, so round-9 rejoin behaviour is untouched.

**Mutations verified red:** `unwrap_or(0)`; `<=` for `<`; comparison reversed; compared against `SOURCE_RANK_AUTHORITATIVE` rather than `current_delegate_source_rank()`; `ranks.tombstone` substituted for `ranks.identity` (compiles — both are `HashMap<RoomKey, u32>`); guard deleted; helper ignoring the registry entirely.

839 tests green, clippy clean. Also drops an unused import introduced by #593.

## Residual — NOT closed

The design rests on "the stored slot was written by the current delegate, so it outranks any legacy generation". True for slots written by this client, but a slot written by an **older River build** that itself held a legacy-sourced identity violates it, and detecting that requires persisted provenance — a much larger change. This closes route 2; it does not make the branch correct in general.

## Sequencing

This is a **precondition for #588's seal**, not a follow-up. Repairing the seal closes route 2 but opens route 1 (an interrupted migration can then set the in-progress flag), and route 1's trigger — a tab closed mid-migration — is far more common than index corruption. Landing the seal first would trade a rare trigger for a common one.

Design and test matrix were specified by an independent review lens **before** implementation.

Refs #588, #592

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -1538,7 +1538,8 @@ mod tests {
             rejoined,
             crate::room_data::SOURCE_RANK_AUTHORITATIVE,
         )
-        .expect("reconcile must succeed");
+        .expect("reconcile must succeed")
+        .written();
         let bytes = out.expect(
             "a ranked resurrection MUST overwrite the delegate's tombstone — \
              without it the room lives one session and is then gone for good",
@@ -1576,9 +1577,17 @@ mod tests {
         // 1. legacy rank + differing identity -> refuse, stored survives.
         let out =
             reconcile_room_present(Some(&stored_bytes), &vk, &local, false, legacy_rank).unwrap();
-        assert!(
-            out.is_none(),
-            "a legacy-sourced copy must not be written over the stored identity"
+        // The VARIANT matters, not merely that nothing was written. The caller
+        // uses it to decide what the delegate now holds, and the two aborts imply
+        // opposite answers: `AdoptedTombstone` means a Tombstone is there,
+        // `RefusedIdentityOverwrite` means a Present we declined to touch. Assert
+        // the specific one or the caller's ROOM_SLOT_STATE bookkeeping is unpinned.
+        assert_eq!(
+            out,
+            ReconcileOutcome::RefusedIdentityOverwrite,
+            "refusing an identity overwrite must be DISTINGUISHABLE from adopting \
+             a tombstone — recording Tombstone here would be a lie about a slot \
+             that actually holds Present"
         );
         match ciborium::de::from_reader::<RoomSlot, _>(&stored_bytes[..]).unwrap() {
             RoomSlot::Present(r) => assert_eq!(
@@ -1601,7 +1610,7 @@ mod tests {
         )
         .unwrap();
         assert!(
-            out.is_some(),
+            out.written().is_some(),
             "a current-delegate-ranked copy still wins (#420 unchanged)"
         );
 
@@ -1616,7 +1625,7 @@ mod tests {
         )
         .unwrap();
         assert!(
-            out.is_some(),
+            out.written().is_some(),
             "an in-session identity must win over any delegate copy (#414)"
         );
 
@@ -1626,7 +1635,7 @@ mod tests {
         let matching = present_slot_bytes(&local);
         let out = reconcile_room_present(Some(&matching), &vk, &local, false, legacy_rank).unwrap();
         assert!(
-            out.is_some(),
+            out.written().is_some(),
             "a legacy-sourced copy with a MATCHING identity must still merge and persist"
         );
     }
@@ -1643,6 +1652,7 @@ mod tests {
         assert!(
             reconcile_room_present(None, &vk, &local, false, legacy_rank)
                 .unwrap()
+                .written()
                 .is_some(),
             "Absent -> StoreFresh, regardless of rank"
         );
@@ -1655,6 +1665,7 @@ mod tests {
                 legacy_rank
             )
             .unwrap()
+            .written()
             .is_some(),
             "Tombstone + explicit rejoin -> StoreFresh, regardless of rank"
         );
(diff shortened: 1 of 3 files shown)


sanity:
## Review result: blocking regression found — parking this PR

Two independent blind reviewers read this branch (adversarial lens + code-first; a data-loss lens was attempted repeatedly and never produced a report, so it is **not** counted as a clearance). One blocking finding, which I then verified against the code myself.

### The guard's stated premise does not hold during the legacy fan-out

The comment at `ui/src/components/app/chat_delegate.rs:4944` justifies the refusal with:

> "If it did, the stored slot was written by the current delegate, which outranks every legacy generation by construction, so refusing is always right."

That conflates *which delegate key the bytes live under* with *which generation supplied the identity inside them*. During migration, `ui/src/components/app/freenet_api/response_handler.rs:1891` fires a `save_rooms_to_delegate()` for **every** responding legacy generation, so the current delegate's slot is routinely written **by the legacy migration itself**. `RoomSlot::Present` carries no provenance, so the guard cannot tell the two apart.

Concrete sequence (32 legacy generations, so `current_delegate_source_rank() == 32`):

1. Generations are probed together. `LEGACY_DELEGATES` is oldest-first age order (`chat_delegate.rs:747`) and the oldest copy usually answers first (`:269`). Gen 2 answers with room R carrying identity **A**; the slot is empty, so this is a `StoreFresh` and the slot becomes `Present(A)`.
2. Gen 20 answers with identity **B**. The ranked merge correctly prefers B in memory and records `identity[R] = 20`.
3. Gen 20's save runs: `identity_rank_for_save(R) == 20 < 32` and `local.self_sk (B) != remote.self_sk (A)` → **`RefusedIdentityOverwrite`**. Nothing is written.
4. `request_legacy_seal_on_quiescence()` fires and no later session re-probes gen 20.

Net: memory holds B, the delegate keeps A, and B becomes unreachable for good.

On `origin/main` step 3 writes B and the rollback does not occur, so **this is a regression in the migration path** — the scenario this workstream exists to fix. It is also the same hazard `response_handler.rs:1913-1921` already describes in its own words as what the #527 seal-delay was written to prevent, reintroduced from the save side.

### Second finding, same root

The deferred signing-key migration (`response_handler.rs:1640-1710`) is unconditional and knows nothing about the refusal. So a refused pass can leave `room:<vk> → Present(self_sk = A)` while `__signing_key:<room> → B` — the delegate signing with an identity the stored room does not carry, which is the split #527 fixed, re-created from the other direction. The refusal cannot be evaluated in isolation from the rest of the migration.

### Lower-severity (real, but holes in a new guard rather than regressions)

- **TOCTOU** — `rooms` is snapshotted once at `chat_delegate.rs:5079` but `identity_rank` is read per-room later at `:5149`, after awaited round-trips during which deferred merges can mutate both. The guard can then pass on a rank that does not describe the snapshot it is about to write.
- **Source pin satisfiable by an unrelated match** — `chat_delegate.rs:1765` searches the whole extracted body for `SavedSlot::Tombstone`, which also occurs in the tombstone loop (`:5225`, `:5231`), so deleting the bookkeeping the assertion protects leaves the pin green.
- **`chat_delegate.rs:1739`** — measured body length 9,215 against a 10,000-char guard; the next ~15 lines trip it, with a message that misdirects.

Note the adversarial reviewer did **not** run mutation testing (the brief forbade modifying the worktree), so its non-vacuity claims are reasoning rather than measurement.

### Status

Parking this as a draft. It fixes a bug that is real in code but has no established field trigger, while currently introducing one that fires in ordinary migration — net-negative as written, and there is no incident forcing a rushed reshape.

The available direction: `ROOM_SLOT_STATE` already records which rooms *this session* wrote, which is exactly the missing provenance — if this session wrote the slot, overwriting it is safe. That is a reshape of the guard rather than a tweak, and would need a fresh review round.

[AI-assisted - Claude]

PR #600, opened 2026-08-05
feat(ui): invite to a room from inside the DM conversation

sanity:
Sharing an invite via DM could only be started from the Member Info
dialog, which is the wrong place for it — the DM thread is where you are
already talking to the person you want to invite.

Adds an "Invite to a room" control to the DM thread's footer. It reuses
`open_invite_via_dm_picker`, so the picker, the invitation signing, and
the send path are all unchanged; this is a second entry point, not a
second implementation. The picker already overlays the still-mounted
thread modal, so no layout changes were needed.

Prominence: sending a message is what this modal is for, so Send keeps
the only filled `bg-accent` treatment and the new control is a bare text
button matching the "Delete their messages" convention already in that
row. Both secondary actions now sit together on the right.

Disabled when the local user belongs to no other room — the same gate the
Member Info button uses, since the picker would otherwise open on an empty
list. The flag rides on the existing `view` memo's `ROOMS` read rather
than adding a second fallible read needing its own anchor/nudge.

The accessible name deliberately avoids "Send", "Send invite" and "Share
an invite": the Playwright specs match those exactly, and the picker
overlays this still-mounted modal, so a collision would make their
selectors ambiguous rather than fail loudly.

Also fixes how ALREADY-SENT invites appear in the thread. The sender
cannot decrypt their own outbound DM, so their bubble renders from the
`OUTBOUND_DMS` plaintext cache — which stored the bare string
"[Invitation]". Two invitations to different rooms were therefore
indistinguishable in the sender's own thread, while the recipient saw a
card naming each room. The cache now records the invited room's name
(threaded down from the picker, which has already unsealed it) and the
sender's bubble renders as the outbound twin of the recipient's card:
same chrome, right-aligned, no Accept button, since accepting your own
invitation is meaningless.

Legacy cache entries still parse and render, minus the room name. One
documented residual, pinned by a test: a legacy note beginning "to " is
read back as a room name, because the old format put the note where the
name now goes.

UI-only — no delegate, contract or `common/` change, so no WASM migration.

Tests: 845 pass. Six unit tests for the summary round-trip (including the
newline-injection case, the legacy shapes, and text that merely mentions
an invitation), plus three source pins for the render path, which ends in
`rsx!` and a GlobalSignal write and so cannot be exercised natively: the
button opens the picker for THIS thread's (room, peer), it does not take
Send's filled treatment, and the sent card carries no Accept button.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/direct_messages.rs
+++ b/ui/src/components/direct_messages.rs
@@ -530,9 +530,9 @@ pub async fn send_structured_dm(
             // who had scrolled up: the picker's success banner lives inside
             // the picker and is torn down in the same defer block that closes
             // it, so it renders for zero frames, and the thread simply looked
-            // unchanged. This send also grows `message_count`, which is what
-            // supplies the auto-scroll effect's subscription — the condition
-            // `OUTBOUND_SEND_COUNTER`'s docs require of any writer.
+            // unchanged. This send adds a message, so a fresh bubble mounts
+            // and re-runs the auto-scroll effect — the condition
+            // `note_outbound_send`'s docs require of any caller.
             dm_thread_modal::note_outbound_send();
         }
         let _ = tx.send(outcome);
--- a/ui/src/components/direct_messages.rs
+++ b/ui/src/components/direct_messages.rs
@@ -340,7 +340,6 @@ pub async fn send_structured_dm(
     room: VerifyingKey,
     peer: MemberId,
     body: river_core::room_state::dm_body::DirectMessageBody,
-    invited_room_label: Option<String>,
 ) -> SendDmOutcome {
     use crate::components::app::chat_delegate::{save_outbound_dm, unhide_dm_thread};
     use crate::components::app::{mark_needs_sync, ROOMS};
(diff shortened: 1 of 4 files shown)


sanity:
## Behaviour change on already-stored data

Worth calling out explicitly for anyone reviewing this: an existing outbound **text** DM whose body begins with the literal string `[Invitation]` will now re-render as an invitation card rather than as its literal text.

This is the documented, accepted false positive. The sender cannot decrypt their own outbound DM, so the bubble renders from the `OUTBOUND_DMS` plaintext cache, which stores a bare `String` with no body-kind field to consult — the `[Invitation]` sentinel predates this change and is the only signal available. Adding a body-kind field means changing `OutboundDmEntry` in `river-core`, which re-keys the chat delegate and needs a migration.

The cost is bounded: the card carries the rest of the text through verbatim, so nothing the user wrote is hidden, and the recipient always sees the true body. Pinned by `text_that_looks_like_an_invite_keeps_the_users_words`, which asserts the bounding property rather than denying the false positive.

## Deferred, deliberately

Naming the invited room on the **sender's** side. The recipient's card names the room (it comes from the decrypted payload); the sender's cannot, so two invitations to different rooms still look alike in the sender's own thread. An earlier revision of this branch encoded the room name into the cached summary string and was reverted, because that one `String` is shared: `riverctl dm list` prints it as a single line per message, and re-parsing a richer format mangled legacy entries whose personal message could itself contain newlines. Doing it properly needs structured storage in `OutboundDmEntry` → delegate migration.

[AI-assisted - Claude]

PR #604, opened 2026-08-06
fix(ui): tolerate an absent self_sk, persist self_vk, stop persisting invite keys

sanity:
## Problem

River persists each room's private signing key inside the `RoomSlot` blob it stores in the chat delegate (`RoomData.self_sk`, `ui/src/room_data.rs`). That blob is reachable from the app API, so it is the wrong place for key material — but it cannot simply be moved yet, and this PR does **not** move it. It removes the obstacles that make moving it later a data-loss event, and closes one exposure that is independent of the delegate entirely.

Three specific problems:

1. **`self_sk` was a REQUIRED CBOR field.** The moment any future River stops writing it, an older client reading that blob fails to deserialize `RoomSlot` *as a whole*. A `RoomSlot` that fails to decode does not degrade — the room **vanishes** from the user's list. Because River clients update at their own pace, the tolerant reader has to be in the field well before the writer changes, or the writer change strands whoever has not upgraded.

2. **Roughly 35 sites reached for the private key when they only needed the public half** — `MemberId` derivation, owner checks, "is this me" comparisons. Nothing could stop needing the private key while those sites existed.

3. **The pending invitation was written to `localStorage`.** An `Invitation` carries `invitee_signing_key` — a full ed25519 private key — and `room_secrets`, the room's plaintext symmetric secrets. It is a bearer credential, it was durable, it was readable by any script in the origin, and the flow has **no terminal state for an abandoned invite**: opening an invite link and never acting on it parked that key material on disk indefinitely, with no path that ever cleared it.

## Approach

Deliberately the subset that requires **no delegate change and therefore no delegate re-key**.

**1. Reader tolerance.** `self_sk` becomes `Option<SigningKey>`; the writer is unchanged and still always stores `Some(..)`. Reads go through four accessors — `signing_key()`, `self_verifying_key()`, `self_member_id()`, `is_self_owner()` — and every one of ~160 call sites was audited rather than unwrapped. There is no `unwrap`/`expect`/`panic!` on the key anywhere in production code, and a test pins that.

The degradation policy: a site needing only the public half uses `self_verifying_key()`; a site that genuinely signs or decrypts refuses through whatever failure channel it already has (`None`, `Err`, `false`, or a logged no-op). `SendMessageError` gains an `IdentityUnavailable` variant so "this device cannot sign here" is not mislabelled as "you are not a member".

**2. Persist `self_vk`.** Public-half work keeps functioning with no private key in hand. `self_verifying_key()` derives from `self_sk` **in preference to** the stored field, so a stale or tampered `self_vk` can never redirect the local identity while the real key is held. The field is stamped onto the blob at the single save chokepoint (`reconcile_room_present`), so it does not depend on which of the many `RoomData` construction sites built the in-memory copy.

Two identity comparisons that previously compared raw key bytes now compare *identities*, because a copy that merely lacks the private key is not a different identity — and both of those branches are lossy (they discard the other copy's state).

**3. Pending invitation.** Held in process memory (`PENDING_INVITATION`) for the pending-invite window, which is entirely within one page load. The nickname and its BLAKE3 fingerprint stay in `localStorage` — neither is a secret — so the #333 binding is intact. Any invitation an older River persisted is actively **erased** at startup: ceasing to write the key does nothing about copies already on users' disks, and the users most exposed are exactly the ones who never open another invite.

*Cost, stated plainly:* a mid-flow reload no longer recovers the invitation from storage. For the URL-bar path — the common one — nothing changes in practice, because the `?invitation=` parameter survives the reload and `App` re-parses it before the recovery block runs. For the DM-card, paste-a-code, and click-interceptor paths a reload drops the pending invite and the user re-opens it. In the deployed gateway this reload-resume has never worked anyway: the iframe runs on an opaque origin, where `localStorage` is unavailable (freenet/river#219).

### What is NOT in this PR, and why

**The relocation itself.** Secrets in the delegate's `signing_key:` namespace do **not** survive a delegate re-key — `fire_legacy_migration_request` recovers old generations via `GetRequest`/`ListRequest`, and that namespace is structurally unreachable by both. Signing keys survive a re-key today *only* because `self_sk` rides inside the migratable blob. River has re-keyed 28 recorded times since 2026-01-15, roughly one every 6-7 days. Shipping "stop writing `self_sk` to the blob" before a re-key-survival mechanism exists would lock every user out of every room at the next re-key. That mechanism is being designed separately, and the writer flip must not land before it.

**Two other findings, both delegate-side, both excluded:**

- The `String::from_utf8_lossy` key aliasing in `create_origin_key` (`delegates/chat-delegate/src/utils.rs`), where distinct non-UTF-8 keys collapse onto one slot.
- The cross-origin write primitive via `EnsureRoomSubscription` (`delegates/chat-delegate/src/handlers.rs` → `subscription.rs`), where the `room_sub:` namespaces are not origin-scoped.

Both live entirely inside the delegate. Neither has a UI-side mitigation — the UI cannot stop another origin calling the delegate, and it already base58-encodes its own keys, so it does not trigger the aliasing itself. Fixing either changes the delegate WASM and therefore forces a re-key, which is precisely what this PR is scoped to avoid. They should be batched into whatever release next re-keys the delegate.

## Testing

The change alters how data already sitting in every existing user's store is decoded, so the round-trip tests carry the weight. Each was written first and confirmed to fail (or be meaningfully exercised) before being confirmed to pass.

**Backward compatibility — protects current users:**
- `roomslot_from_a_current_writer_still_round_trips` — a `RoomSlot` in today's exact shape (required `self_sk`, no `self_vk`) decodes with the key intact byte-for-byte.
- `roomdata_decodes_from_minimal_legacy_blob` — extended to assert the legacy required field lands in `Some(..)` and the accessor still derives the public half from it.

**Forward compatibility — makes the later relocation safe:**
- `roomdata_decodes_from_a_blob_with_no_self_sk` and `roomslot_decodes_from_a_blob_with_no_self_sk` — a blob that OMITS the field (not one writing `null`, which is what a future writer would actually emit) decodes rather than erroring, at both the struct and the `RoomSlot` enum level.
- The first carries a **differential oracle**: the same bytes are decoded into the pre-change shape with a required `self_sk`, and that must fail. Without it the test could pass while proving nothing. Worth recording: `#[serde(default)]` turned out **not** to be the load-bearing part — removing it did not break the tests, because `Option` alone is what makes the field optional. The oracle is what establishes the property is real.

**Behaviour when the key is absent:**
- `read_paths_degrade_gracefully_with_no_private_key` — asserts a precondition (the member CAN act with the key present, so the test is not passing on an inert room), then removes the key and checks each path's specific answer. **This test found a real design error:** `can_send_message` was gated on the public half, so it returned `Ok` for a copy that could not sign — showing the user a composer whose every send would be silently rejected. Both it and `can_participate` now require the private key.
- `an_owner_without_a_private_key_is_still_the_owner_but_cannot_rotate` — ownership is a property of the public half and survives; rotation derives and signs, so it does not. Separated so a future refactor cannot collapse the two checks.
- `self_sk_outranks_a_disagreeing_self_vk` — the held private key wins over a disagreeing `self_vk`, and backfill only ever fills an empty slot.

**Wiring:**
- `a_saved_slot_always_carries_self_vk`, `a_missing_private_key_is_not_a_diverged_identity`, `self_sk_and_self_vk_are_assigned_together` (the repo's paired-`Option`-fields pattern; the type-system remedy is unavailable because both names are wire-format field names, so it is a scrape — with a non-vacuity floor on the number of sites found).
- `the_pending_invitation_is_never_persisted` — a source scrape, because the write goes through `web_sys`, which does not exist under `cargo test`; a behavioural test there could not fail. It also asserts the nickname binding is still persisted and the legacy purge still present, so the pin cannot be satisfied by deleting the storage code wholesale.
- `pending_invitation_round_trips_in_memory_and_clears`.
- `the_send_path_handles_a_missing_local_signing_key` replaces `the_send_path_always_has_a_local_signing_key`. That test was a deliberate canary whose comment said "if it ever becomes optional... the message path needs rethinking rather than an `unwrap`". This is that event, so the successor pins the obligation it named: the accessor is fallible (compile time) and no signing path unwraps it (source scrape).

**Mutation-verified.** Every new test was confirmed to fail under a targeted mutation before being trusted: preferring `self_vk` over `self_sk` in the accessor; deleting the save-path backfill; reverting the identity comparison to a raw `Option` compare; dropping or mis-deriving the paired `self_vk` write; reintroducing the `localStorage` persist; restoring the nickname sweep; reordering the startup adopt hook; re-introducing the conversation blank-render bail; planting an `unwrap` in a file the earlier allow-list pin did not scan; and refactoring a production literal out of the paired-field scan's needle.

Two guards were found to be weaker than they looked and were replaced rather than kept: an ordering pin that stayed green under the defect it was meant to catch (ordering was only ever a proxy for the outcome), and a `#[cfg(test)]` split that silently truncated production code following a test helper — so the scans were quietly narrowing rather than refining. Both now read whole files and assert outcomes.

### What the review rounds changed, and two things not to over-read

Three rounds of independent review followed. The substantive outcomes, with their reachability stated honestly:

- **A user-visible bug in the first cut.** The guard in `repopulate_secrets_from_state` was scoped too widely and skipped the `invitation_secrets` fold and the `current_secret_version` alignment, neither of which needs a key. That would have left a private room rendering "[Encrypted message — secret vN not available]" over content it could decrypt. Narrowed, and the test that had *ratified* the bug now pins the intended behaviour.
- **A regression I introduced while addressing review feedback.** A tidy-up swept the nickname binding inside the same purge that erases the legacy invitation artifact, which the startup adopt hook calls immediately after reading it — turning the #218 auto-resume into a re-prompt on a join already in flight. The purge now takes only the artifact.
- **`self_vk` erasure on save: LATENT, not live, and pre-empted.** `reconcile_room_present` builds from `local.clone()`, and the backfill can only derive `self_vk` from a held `self_sk`, so a copy carrying neither half would write `self_vk: None` over a stored identity. The test fails without the fix — but against a state the *test* constructs. Every production `self_sk` literal and assignment outside tests writes `Some`, so the precondition is unreachable in this build. The guard is in because the relocation is exactly what makes it reachable, and it is cheaper to add now than to remember later. It is not a bug being fixed in shipped behaviour.
- **The `localStorage` tests do not cover the deployed path.** Proving the invitation regression needed the storage layer to be drivable under `cargo test`, so access goes through three key/value functions backed by an in-process map off-wasm. In the production gateway the iframe has an opaque origin, `localStorage` throws, and all three are no-ops — so every one of those tests exercises a branch that cannot occur there, as does the legacy artifact they read. Their real coverage is `dx serve` and non-sandboxed deployments. The logic is worth getting right for those and worth testing rather than scraping, but this is not coverage of the gateway.

**Suite:** 891 `river-ui` unit tests pass (896 with `example-data,no-sync`), 252 `river-core` lib tests, plus the migration, room-contract-migration and delegate-key suites. `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync` is clean.

### No re-key — verified

Only `ui/src/` is touched; nothing under `delegates/`, `contracts/`, or `common/`, and no `.wasm` byte changed.

```
cargo make check-migration
  Committed WASM matches HEAD — no migration needed.
cargo make check-room-contract-migration
  Committed room-contract WASM matches HEAD — no migration entry needed.
```

Delegate WASM BLAKE3 is `a44c64014d60fd245fe8fb5172f8fd7397039b248b3b6a8e95e89a0bb539fb5e` before and after; the room-contract and web-container artifacts are likewise unchanged. No `legacy_delegates.toml` entry is required.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -17,10 +17,11 @@ use crate::components::room_list::create_room_modal::CreateRoomModal;
 use crate::components::room_list::edit_room_modal::EditRoomModal;
 use crate::components::room_list::notification_modal::NotificationModal;
 use crate::components::room_list::receive_invitation_modal::{
-    accept_invitation, clear_invitation_from_storage, decide_recovered_invitation,
-    is_invitation_processed, load_invitation_from_storage, load_invitation_nickname_from_storage,
-    purge_legacy_persisted_invitation_once, save_invitation_to_storage, take_resume_once,
-    ReceiveInvitationModal, RecoveredInvitationAction, PRESENT_INVITATION_REQUEST,
+    accept_invitation, adopt_and_purge_legacy_persisted_invitation_once,
+    clear_invitation_from_storage, decide_recovered_invitation, is_invitation_processed,
+    load_invitation_from_storage, load_invitation_nickname_from_storage,
+    save_invitation_to_storage, take_resume_once, ReceiveInvitationModal,
+    RecoveredInvitationAction, PRESENT_INVITATION_REQUEST,
 };
 use crate::invites::PendingInvites;
 use crate::room_data::{CurrentRoom, Rooms};
@@ -325,13 +326,17 @@ pub fn App() -> Element {
         receive_invitation.set(Some(inv));
     });
 
-    // Retire any invitation an OLDER River persisted to localStorage. That
-    // blob carries a full ed25519 private key plus the room's plaintext
-    // secrets, and the old flow had no terminal state for an abandoned invite,
-    // so it could sit there indefinitely. River no longer writes it (the
-    // pending invitation is held in memory for the page's lifetime), but not
-    // writing it does nothing about what is already on disk — this does.
-    purge_legacy_persisted_invitation_once();
+    // Retire any invitation an OLDER River persisted to localStorage. That blob
+    // carries a full ed25519 private key plus the room's plaintext secrets, and
+    // the old flow had no terminal state for an abandoned invite, so it could
+    // sit there indefinitely. River no longer writes it (the pending invitation
+    // is held in memory for the page's lifetime), but not writing it does
+    // nothing about what is already on disk — this does.
+    //
+    // It ADOPTS the artifact into memory before erasing it, so an in-flight
+    // join left by the older build still completes on this one. Runs before the
+    // recovery block below, which is what consumes it.
+    adopt_and_purge_legacy_persisted_invitation_once();
 
     // Recover the pending invitation if it was not found in the URL. Note this
     // now only recovers WITHIN a page load, not across a reload: the artifact
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -19,8 +19,8 @@ use crate::components::room_list::notification_modal::NotificationModal;
 use crate::components::room_list::receive_invitation_modal::{
     accept_invitation, clear_invitation_from_storage, decide_recovered_invitation,
     is_invitation_processed, load_invitation_from_storage, load_invitation_nickname_from_storage,
-    save_invitation_to_storage, take_resume_once, ReceiveInvitationModal,
-    RecoveredInvitationAction, PRESENT_INVITATION_REQUEST,
+    purge_legacy_persisted_invitation_once, save_invitation_to_storage, take_resume_once,
+    ReceiveInvitationModal, RecoveredInvitationAction, PRESENT_INVITATION_REQUEST,
 };
 use crate::invites::PendingInvites;
 use crate::room_data::{CurrentRoom, Rooms};
@@ -325,9 +325,21 @@ pub fn App() -> Element {
         receive_invitation.set(Some(inv));
     });
 
-    // Recover invitation from localStorage if not found in URL (e.g. after
-    // page reload before subscription completed). The invitation artifact
-    // stays in storage from the moment it is seen until the room is
+    // Retire any invitation an OLDER River persisted to localStorage. That
+    // blob carries a full ed25519 private key plus the room's plaintext
+    // secrets, and the old flow had no terminal state for an abandoned invite,
+    // so it could sit there indefinitely. River no longer writes it (the
+    // pending invitation is held in memory for the page's lifetime), but not
+    // writing it does nothing about what is already on disk — this does.
+    purge_legacy_persisted_invitation_once();
+
+    // Recover the pending invitation if it was not found in the URL. Note this
+    // now only recovers WITHIN a page load, not across a reload: the artifact
+    // is a bearer credential and is deliberately no longer persisted. See
+    // `PENDING_INVITATION`. The three cases below are unchanged in shape; case
+    // 1 (auto-resume) is reachable only when the invitation is still in memory.
+    //
+    // The artifact is held from the moment it is seen until the room is
     // subscribed OR the user dismisses it (`clear_invitation_from_storage`).
     //
     // Three cases, in priority order:
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -194,6 +194,22 @@ pub fn App() -> Element {
         }
     });
 
+    // Retire any invitation an OLDER River persisted to localStorage, adopting
+    // it into memory first. That blob carries a full ed25519 private key plus
+    // the room's plaintext secrets, and the old flow had no terminal state for
+    // an abandoned invite, so it could sit there indefinitely. River no longer
+    // writes it (the pending invitation is held in memory for the page's
+    // lifetime), but not writing it does nothing about what is already on disk.
+    //
+    // MUST STAY ABOVE the URL-parse block below. That block calls
+    // `save_invitation_to_storage`, which itself purges the legacy artifact —
+    // so with the order reversed this hook would find nothing left to adopt,
+    // and an in-flight join left by the older build would be destroyed unread.
+    // Running first makes the ordering structural rather than a coincidence of
+    // which path happens to be taken; pinned by
+    // `the_startup_purge_is_wired_in_and_adopts_before_erasing`.
+    adopt_and_purge_legacy_persisted_invitation_once();
+
     // Get auth token from window global (injected by Freenet gateway)
     // This is synchronous - no network request needed
     get_auth_token_from_window();
@@ -325,18 +341,6 @@ pub fn App() -> Element {
         );
         receive_invitation.set(Some(inv));
     });
-
-    // Retire any invitation an OLDER River persisted to localStorage. That blob
-    // carries a full ed25519 private key plus the room's plaintext secrets, and
-    // the old flow had no terminal state for an abandoned invite, so it could
-    // sit there indefinitely. River no longer writes it (the pending invitation
-    // is held in memory for the page's lifetime), but not writing it does
-    // nothing about what is already on disk — this does.
-    //
-    // It ADOPTS the artifact into memory before erasing it, so an in-flight
-    // join left by the older build still completes on this one. Runs before the
-    // recovery block below, which is what consumes it.
-    adopt_and_purge_legacy_persisted_invitation_once();
 
     // Recover the pending invitation if it was not found in the URL. Note this
     // now only recovers WITHIN a page load, not across a reload: the artifact
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -1649,6 +1649,11 @@ mod tests {
     /// is how a future River will know who the local user is once `self_sk` is
     /// no longer in the blob, and it can only rely on that if today's build has
     /// been writing it all along.
+    ///
+    /// Writing it here does NOT make it durable — an older tab or build saving
+    /// the same room strips it again, and the backfill never runs on decode.
+    /// See the `RoomData::self_vk` field docs: the relocation writer must check
+    /// per-room that the STORED blob already carries it, not assume it.
     #[test]
     fn a_saved_slot_always_carries_self_vk() {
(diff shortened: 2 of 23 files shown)


sanity:
Review addressed. All eight MUST-FIX items and R9-R14; R15 folded into the doc wording.

**The two ratifying tests were the important catch.** Both now pin the intended behaviour, and I verified each FAILS against the old behaviour before passing against the new (surgical mutation, then restore):

- `read_paths_degrade_gracefully_with_no_private_key` — now asserts `repopulate_secrets_from_state()` returns **1**, that the secret lands in `secrets`, and that `get_secret()` is `Some`. Against the pre-fix wide bail it fails on the count.
- `self_sk_outranks_a_disagreeing_self_vk` — now asserts backfill **overwrites** a disagreeing `self_vk`, plus a second case pinning that with no private key there is nothing to correct against. Against fill-only it fails.

**R1** — guard moved to wrap only the ECIES loop. Your trace was exactly right: the `invitation_secrets` fold and the `current_secret_version` alignment need no key, and skipping the alignment is what makes `get_secret()` return `None` and turns readable content into "[Encrypted message — secret vN not available]".

**R2** — `backfill_self_vk` now repairs unconditionally when `self_sk` is held. Agreed there is no case where a `self_vk` contradicting a held key is correct to persist.

**R3** — renamed to `adopt_and_purge_legacy_persisted_invitation_once`: reads the stored artifact into `PENDING_INVITATION` (without clobbering a fresher one from the URL parse), then removes it. Pinned by an ordering test that asserts the read appears before the purge in the source, mutation-verified by swapping them.

**R4** — `edit_room_modal` now splits `user_is_owner` (public half; still drives the leave warning and whether the owner-only row exists) from a new `user_can_edit = user_is_owner && signing_key().is_some()` that gates every editing affordance. Key-less owner gets a visible notice; Rotate stays visible but disabled with an explanatory `title`. `is_self_owner()` itself untouched.

**R5** — conversation history, the member roster and DM threads now carry `Option<MemberId>` and degrade cosmetics to "not me", following the `document_title` / `room_synchronizer` pattern. Added `messages_still_render_when_the_local_identity_is_unknown`, mutation-verified against a re-introduced bail. One trap found while doing it: the self-mention highlight had to be `self_member_id.is_some_and(|me| resolved == Some(me))`, because `resolved == self_member_id` would call an *unresolved* mention a mention of you via `None == None`.

DM direction: with no identity at all the thread renders **empty** rather than guessing, since every message there involves the peer and any default would show someone else's message as the user's own. Note this is only reachable with neither `self_sk` nor `self_vk`; with `self_vk` present direction still works.

**R6** — the durability claim is corrected rather than papered over. The field docs now state both holes (an older tab strips it; the backfill never runs on decode) and say explicitly that the relocation writer **must verify `self_vk` is present in the STORED blob per room** before dropping `self_sk`. Added `decoding_alone_does_not_backfill_self_vk`, which documents the gap and is written to fail if someone adds a decode-time backfill without relaxing the docs.

**R7** — the pin now walks the crate from `CARGO_MANIFEST_DIR` instead of four hand-listed files. Verified by planting `.signing_key().unwrap()` in `ban_button.rs`, a previously unscanned file: it now fails.

While there I found the split I was using made both scans weaker, not stronger: splitting at the first `#[cfg(test)]` truncates production code that FOLLOWS a test helper, which `room_data.rs` does (`test_minimal_room_data` sits above `create_new_room_with_name`). Nothing in the crate unwraps the key even in tests, so the unwrap scan now reads whole files, and the paired-field scan splits at `mod tests {`.

**R8** — one rule, stated in both places: a *conflict* test treats an unknown identity as no-conflict; a *change* test (`swap_room_identity_in_place`) treats it as changed, because it asks a different question and an import always supplies an identity. Also recorded the gap you identified — a local copy naming no identity can never ACQUIRE one through merge — as a known gap the relocation must close.

**R9** ✅ **R10** ✅ (also fixed a false positive it had: the literal matcher was matching `self_sk: &SigningKey` *parameters*; now anchored on `Some(`/`None`, with the non-vacuity floor raised) **R12** ✅ **R13** ✅ (startup-purge wiring + decode gap; the remaining two — cross-tab end-to-end, and the guard-scoping fix — I did not add, see below).

**R11** — I did not force these to match, because the difference turns out to be principled, and I have stated the rule in both: *degrade toward the answer that stays approximately true.* In the room badge the identity only REFINES the count (excluding your own messages), so dropping it over-counts mildly and stays roughly right. In the DM badge the identity is what makes the count mean anything — which threads are even yours, and which way each message went — so a number computed without it would be about other people's conversations, and zero is the honest answer. Happy to force them to match if you'd rather have uniformity than this distinction.

**R14 — checked, and it is clear.** `invitation-sandbox.spec.ts` exercises the hash-bridge dedup (#215/#217), not localStorage resume; nothing in the suite asserts reload-resume. Its own header comment independently states that the gateway iframe cannot use `localStorage`, which corroborates your verification from a third direction.

**R15** — accepted and folded into the `PENDING_INVITATION` docs: on a non-sandboxed deployment the strip succeeds, so a mid-flow reload loses the invite on the URL path too.

### Not done, deliberately

- **R13's cross-tab end-to-end and a test for the guard-scoping fix.** The first needs two concurrent WASM clients against a live delegate, which this suite cannot express; the encoding half is covered by the round-trip tests. The second sits inside a large GET handler with no seam to test through — I fixed the code and commented why an early return is wrong there, but a test would have to be a source scrape asserting the absence of `return`, which is a weak guard. Both are accepted risk rather than oversight; say the word if you'd rather I build the seam.

### Verification

880 river-ui unit tests (885 with `example-data,no-sync`), 252 river-core, wasm32 target clean. `cargo make check-migration` and `check-room-contract-migration` both report no migration needed; delegate WASM stays `a44c6401…`, byte-identical to `main`. Only `ui/src/` is touched.

Not merged.

[AI-assisted - Claude]

PR #605, opened 2026-08-07
fix(ui): make the add-reaction + button visible on touch devices

sanity:
## Problem

On mobile (Android reported), the inline add-reaction "+" button next
to a message is invisible — but tapping roughly where it should be
still opens the reaction picker. If a message already has reactions,
the "+" is (dimly) visible.

## Approach

`.add-reaction-btn` in `ui/assets/main.css` reveals via `.group:hover`,
a real CSS `:hover`, which touch devices can't reliably trigger:

```css
.add-reaction-btn { opacity: 0; }
.group:hover .add-reaction-btn { opacity: 0.4; }
.add-reaction-btn.has-reactions { opacity: 0.2; }
```

So on touch it stays at `opacity: 0` (or `0.2` once it has reactions,
matching the report), while the click handler still fires because an
`opacity: 0` element is still tappable.

This is the same failure mode already fixed twice in this file —
freenet/river#402 (the hover message-action bar) and #462 (the DM rail
archive ✕) — so the fix mirrors #462's pointer-capability media query,
gating on `(hover: none), (any-pointer: coarse)` rather than viewport
width (a touchscreen laptop reports `hover: hover` while a finger is on
the glass, which `any-pointer: coarse` still catches).

I checked for other instances of the same pattern
(`group-hover:opacity`/`.group:hover` reveal) in the UI: the hover
message-action bar and the DM archive button both already have their
touch-visible counterparts (#402/#462). `MessageActions` /
`QuickReactionButton` in `message_actions.rs` have the same
`opacity-0 group-hover:opacity-100` pattern but are dead code (not
referenced anywhere in the app), so left untouched.

## Testing

- Added `ui/tests/add-reaction-touch.spec.ts`, a cascade probe against
  the shipped stylesheet (mirrors `dm-archive-touch.spec.ts`).
  Verified it fails without the fix (checked via mutation: stripped
  the new CSS rule from the built output, reran on `mobile-chrome`,
  got the expected red) and passes with it.
- Ran the full Playwright suite (`npx playwright test`, all 5 browser
  projects): 787 passed, 23 pre-existing skips, no regressions.

No delegate/contract/common WASM touched — CSS-only change, no
migration needed.

[AI-assisted - Claude]

### DIFF
--- a/ui/assets/main.css
+++ b/ui/assets/main.css
@@ -32,10 +32,20 @@
  * Gate on pointer CAPABILITY, not viewport width — see the #462 comment
  * above for why `(hover: none), (any-pointer: coarse)` is the correct query
  * (`any-pointer: coarse` also catches a touchscreen laptop, which reports
- * `hover: hover` while a finger is on the glass). */
+ * `hover: hover` while a finger is on the glass).
+ *
+ * Also widen the tap target to >=44px (WCAG 2.5.8 / Apple HIG), matching
+ * #462's fix for the same bug class: at rest the button is just a bare "+"
+ * glyph (`text-xl leading-none`, no padding), well under the minimum. Unlike
+ * #462's absolutely-positioned archive ✕, this button sits inline in a
+ * `flex flex-wrap` reaction row (`conversation.rs`), so growing it needs no
+ * companion padding fix elsewhere — the row simply wraps if a message has
+ * many reactions and the row is tight. */
 @media (hover: none), (any-pointer: coarse) {
     .add-reaction-btn {
         opacity: 0.4;
+        min-width: 2.75rem;
+        min-height: 2.75rem;
     }
 
     .add-reaction-btn.has-reactions {
--- a/ui/assets/main.css
+++ b/ui/assets/main.css
@@ -21,6 +21,28 @@
     opacity: 0.4;
 }
 
+/* Touch reveal for the inline add-reaction "+" button (same failure mode as
+ * #402/#462 above): `.add-reaction-btn` reveals via `.group:hover`, a real
+ * CSS `:hover`, which touch devices can't reliably trigger — so on a phone
+ * the button was invisible (opacity: 0) yet still tappable, unless the
+ * message already had reactions (`.has-reactions` gives it opacity: 0.2 at
+ * rest regardless of hover), which is why the bug only showed on
+ * reaction-less messages.
+ *
+ * Gate on pointer CAPABILITY, not viewport width — see the #462 comment
+ * above for why `(hover: none), (any-pointer: coarse)` is the correct query
+ * (`any-pointer: coarse` also catches a touchscreen laptop, which reports
+ * `hover: hover` while a finger is on the glass). */
+@media (hover: none), (any-pointer: coarse) {
+    .add-reaction-btn {
+        opacity: 0.4;
+    }
+
+    .add-reaction-btn.has-reactions {
+        opacity: 0.4;
+    }
+}
+
 /* @mention chips. Rendered as raw <span class="river-mention"> inside message
  * bodies (see conversation::message_to_html_with_mentions). Clickable via the
  * document-level mention click interceptor.
--- a/ui/tests/add-reaction-touch.spec.ts
+++ b/ui/tests/add-reaction-touch.spec.ts
@@ -1,0 +1,75 @@
+import { test, expect } from "@playwright/test";
+
+// Regression coverage for the inline add-reaction "+" button being invisible
+// (but still tappable) on touch devices.
+//
+// `.add-reaction-btn` (ui/assets/main.css) reveals via `.group:hover`, a real
+// CSS `:hover`, which touch devices can't reliably trigger — so on a phone
+// the button rendered at `opacity: 0` yet still received taps (the reported
+// symptom: invisible, but tapping roughly where it should be opens the
+// picker). Same failure mode as freenet/river#402 and #462; the fix mirrors
+// #462's pointer-capability media query. This probes the cascade directly
+// against the shipped stylesheets, since that's the layer where the bug
+// actually lived (Tailwind's utility layer vs. main.css's plain rules).
+
+const PROBE_ID = "add-reaction-cascade-probe";
+
+async function measureProbe(page: import("@playwright/test").Page, hasReactions: boolean) {
+  return page.evaluate(
+    ({ id, hasReactions }) => {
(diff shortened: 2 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #605

### Summary
- **PR Title:** fix(ui): make the add-reaction + button visible on touch devices
- **Type:** fix
- **CI Status:** passing (on commit `55070f0c`; re-confirming on current HEAD `03e4b7b6` before merge)
- **Linked Issues:** none filed separately — fixes a bug reported directly in chat
- **Review tier:** Light (small, self-contained CSS + test diff; no high-risk surface)
- **Reviewers run:** `freenet:skeptical-reviewer` + `freenet:code-first-reviewer` (first round, on `55070f0c`); `freenet:skeptical-reviewer` again (focused re-check, on the `03e4b7b6` tap-target addition). No external model attempted, per the standing "just use claude" directive (`~/.claude/rules/multi-model-review.md`).

---

### Code-First Analysis
**Independent Understanding:** `.add-reaction-btn` (`ui/assets/main.css`) revealed only via `.group:hover`, a real CSS `:hover`, which touch devices can't reliably trigger — leaving the button at `opacity: 0` (or `0.2` once a message has reactions) but still tappable. The fix adds a `@media (hover: none), (any-pointer: coarse)` rule forcing `opacity: 0.4` in both cases, plus (added in round 2) `min-width`/`min-height: 2.75rem` for a 44px tap target.

**Stated Intent:** Matches exactly — the PR description states this precisely, including the opacity values and the explicit mirroring of the #402/#462 precedents already in the same file.

**Alignment:** Full match. Independently re-derived the CSS cascade specificity/source-order math for both the "no reactions" and "has reactions" cases and confirmed the new rule wins on a coarse pointer without shadowing the existing desktop hover rules — verified twice, by two independent reviewers, with matching conclusions.

**Gaps:** None in the final (post round-2) version. Round 1 found one real gap (tap-target sizing), which was fixed and re-verified.

---

### Testing Assessment
**Coverage Level:** Adequate

| Unit | N/A | Pure CSS change, no Rust logic |
| Integration | N/A | — |
| Simulation | N/A | — |
| E2E (Playwright) | ✅ | `add-reaction-touch.spec.ts`, mirrors the established `dm-archive-touch.spec.ts` pattern; cascade-probes the shipped stylesheet across all 5 browser projects (desktop + mobile-chrome + mobile-safari), asserting both the touch branch (`opacity >= 0.4`, `min(width,height) >= 44`) and that the desktop hover-gated branch is unchanged |

**Regression Test:** Present and valid. Verified via mutation testing (stripped the new CSS rule from the built output, reran on `mobile-chrome`, confirmed red; restored, confirmed green) — this is real evidence the test isn't vacuous, not just an assumption.

**Missing Tests:** None blocking. One noted coverage gap (see Skeptical Findings) — the test's synthetic probe element doesn't reproduce the real button's exact flex ancestry (`inline-flex items-center justify-center` inside a `flex flex-wrap` row), so it wouldn't catch a hypothetical *future* regression where a `flex-shrink-0`/`min-w-0` utility gets added to the real markup and starts collapsing the min-size. Verified today's markup does NOT have that problem (empirically measured `flexShrink: "1"` with the explicit `min-width` holding as a hard floor), so this is a latent gap, not a live bug.

---

### Skeptical Findings
**Risk Level:** Low

| Reaction row's flex line height inflates to 44px on any line containing the button, even without wrapping, visually mismatched against the ~19-24px reaction pills sharing that line | Low | `ui/assets/main.css:44-49`, `ui/src/components/conversation.rs:5601` | Verified empirically (reproduced the real row markup under coarse-pointer emulation). Not a functional bug or WCAG violation — visually confirmed via a real Playwright screenshot at 375px with existing reaction pills present, and it reads fine (items-center keeps everything aligned; the extra height is whitespace, not visual clutter). Noted for awareness, not blocking. |
| Playwright probe doesn't reproduce the real button's flex ancestry | Low | `ui/tests/add-reaction-touch.spec.ts:22-33` | See Testing Assessment above — coverage gap, not a live bug. |

No High or Medium findings remain. The one Medium finding from round 1 (missing tap-target sizing, mirroring the #462 precedent's 44px minimum) was fixed in the second commit (`03e4b7b6`) and re-verified: `min-width`/`min-height` genuinely holds under the real flex context (empirically measured, not just reasoned about — explicit `min-width` overrides the flex item's default `auto` floor and survives `flex-shrink: 1`).

Also verified independently (both rounds): no CSS specificity/cascade shadowing; no `@layer`/stylesheet-ordering risk (this button, unlike the DM-archive button, carries no competing Tailwind opacity utility at all, so that risk class doesn't even apply here); single call site for `.add-reaction-btn` in the whole `ui/src` tree; the `MessageActions`/`QuickReactionButton` dead-code claim in the PR description is accurate (grepped, no references anywhere outside their own file).

---

### Big Picture Assessment
**Goal Alignment:** Yes — fixes exactly the reported bug (invisible-but-tappable + button on Android), matches the established fix pattern already used twice in the same file for the identical failure mode.

**Removed Code Concerns:** None — no tests or code removed.

**Scope Assessment:** Focused. Two files changed (`main.css`, one test file), no unrelated changes. The PR correctly identified and left alone the dead `message_actions.rs` component that shares the same buggy pattern but isn't reachable from any live code path.

---

### Documentation
- Code docs: complete — the new CSS carries an explanatory comment matching the style of the #402/#462 precedents it extends.
- Architecture docs: N/A
- User docs: N/A

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None remaining — the one Should-Fix from round 1 (tap-target sizing) is resolved in `03e4b7b6`.

#### Consider (Suggestions)
1. If a future change touches this button's wrapper classes, double-check `flex-shrink`/`min-w-0` aren't added without re-verifying the 44px floor still holds (the test's synthetic probe won't catch that regression on its own).

---

### Verdict
**State:** Ready to Merge (pending final CI confirmation on current HEAD SHA)
**HEAD SHA reviewed:** `03e4b7b6` (round 2 fix verified against this commit's diff; round 1's base cascade logic was reviewed against `55070f0c` and is unchanged by round 2)

This is a re-review: round 1 (on `55070f0c`) found the core fix correct but flagged a missing tap-target size (mirroring the #462 precedent). Round 2 (on the incremental `03e4b7b6` diff) confirms that gap is closed, verified empirically (not just reasoned), with only two Low-severity, non-blocking notes. No further re-review needed before merge — the diff since round 1 is a small, well-scoped addition and both findings from round 1 are resolved.

[AI-assisted - Claude]

PR #606, opened 2026-08-07
fix(ui): stop the add-reaction button from growing the reaction row

sanity:
## Problem

Follow-up to #605. That PR fixed the reaction "+" button being
invisible-but-tappable on touch devices, and additionally widened its
tap target to 44px (mirroring the #462 archive-button fix), reasoning
it was under the WCAG 2.5.8 minimum.

User feedback in River chat afterward:

> Am i imagining it or did the react button not used to be below the
> message text. It feels like it's taking up more space than it used
> to. Also kinda redundant on touch devices as react is in the side
> menu anyway

Confirmed: an `opacity: 0` element still occupies its layout box, so
the button already had space reserved at its natural size
(`text-xl leading-none`, ~20px tall, no padding). Forcing
`min-width`/`min-height: 2.75rem` grew that to 44px on *every*
message's reaction row, not just ones with existing reactions —
a real, user-visible regression.

## Approach

Revert the tap-target sizing, keep the visibility fix (opacity via
the `(hover: none), (any-pointer: coarse)` media query — that part
of #605 is correct and unaffected).

Left the sizing out rather than shrinking it to something in between,
because:
- This button is lower-stakes than the DM-archive ✕ the sizing
  mirrored — a mis-tap just opens the emoji picker, not a destructive
  action.
- Touch users already have a second path to the same action: the
  message kebab menu's "React" entry (`.touch-actions` in
  `conversation.rs`) — a full menu row (~33-36px tall), comfortably
  above the WCAG 2.5.8 AA 24px floor though, like this button, short
  of the 44px AAA/HIG target. (Corrected after review — an earlier
  version of this description overstated this as "already >=44px".)

## Testing

- Extended `ui/tests/add-reaction-touch.spec.ts` to assert the button
  has **no** forced minimum size on a coarse pointer (previously it
  asserted `>= 44px`; now asserts `< 44px`) — this pins the decision,
  so a future re-add of `min-width`/`min-height` fails the test.
- Verified via mutation testing: reintroduced the min-size rule in the
  built CSS output, confirmed the test goes red; restored, confirmed
  green.
- Verified visually via a Playwright screenshot at a 375px viewport on
  a room with existing reactions — row spacing matches the original
  (pre-#605) compact layout.
- Full local Playwright suite (`npx playwright test`, all 5 browser
  projects): 787 passed, 23 pre-existing skips, no regressions.

No delegate/contract/common WASM touched — CSS-only change (plus the
test), no migration needed.

[AI-assisted - Claude]

### DIFF
--- a/ui/assets/main.css
+++ b/ui/assets/main.css
@@ -40,10 +40,11 @@
  * which reads as "the composer now takes up more room" — a real visible
  * regression, not just a WCAG nice-to-have. This button is also a
  * lower-stakes target than the DM-archive ✕ (a mis-tap just opens the
- * emoji picker, not a destructive action) and touch users have a second,
- * already-sized path to the same action via the message kebab menu's
- * "React" entry (`conversation.rs`, `.touch-actions`). So: fix visibility
- * only, leave sizing alone. */
+ * emoji picker, not a destructive action) and touch users have a second
+ * path to the same action via the message kebab menu's "React" entry
+ * (`conversation.rs`, `.touch-actions`) — a full menu row, comfortably
+ * above the WCAG 2.5.8 AA 24px floor though still short of the 44px AAA
+ * target itself. So: fix visibility only, leave sizing alone. */
 @media (hover: none), (any-pointer: coarse) {
     .add-reaction-btn {
         opacity: 0.4;
--- a/ui/assets/main.css
+++ b/ui/assets/main.css
@@ -34,18 +34,19 @@
  * (`any-pointer: coarse` also catches a touchscreen laptop, which reports
  * `hover: hover` while a finger is on the glass).
  *
- * Also widen the tap target to >=44px (WCAG 2.5.8 / Apple HIG), matching
- * #462's fix for the same bug class: at rest the button is just a bare "+"
- * glyph (`text-xl leading-none`, no padding), well under the minimum. Unlike
- * #462's absolutely-positioned archive ✕, this button sits inline in a
- * `flex flex-wrap` reaction row (`conversation.rs`), so growing it needs no
- * companion padding fix elsewhere — the row simply wraps if a message has
- * many reactions and the row is tight. */
+ * Deliberately does NOT widen the tap target the way #462 did: a min-size
+ * here (tried in #605, reverted after user feedback) grows the reaction
+ * row's line height on every message, even ones without reactions yet,
+ * which reads as "the composer now takes up more room" — a real visible
+ * regression, not just a WCAG nice-to-have. This button is also a
+ * lower-stakes target than the DM-archive ✕ (a mis-tap just opens the
+ * emoji picker, not a destructive action) and touch users have a second,
+ * already-sized path to the same action via the message kebab menu's
+ * "React" entry (`conversation.rs`, `.touch-actions`). So: fix visibility
+ * only, leave sizing alone. */
 @media (hover: none), (any-pointer: coarse) {
     .add-reaction-btn {
         opacity: 0.4;
-        min-width: 2.75rem;
-        min-height: 2.75rem;
     }
 
     .add-reaction-btn.has-reactions {
--- a/ui/tests/add-reaction-touch.spec.ts
+++ b/ui/tests/add-reaction-touch.spec.ts
@@ -11,6 +11,13 @@ import { test, expect } from "@playwright/test";
 // #462's pointer-capability media query. This probes the cascade directly
 // against the shipped stylesheets, since that's the layer where the bug
 // actually lived (Tailwind's utility layer vs. main.css's plain rules).
+//
+// Deliberately does NOT assert a minimum tap-target size: an earlier version
+// of this fix added `min-width`/`min-height: 2.75rem` (mirroring #462), but
+// that grew the reaction row's height on every message and read as a visible
+// regression (river chat feedback, 2026-08-07) — reverted. The size
+// assertion below pins that decision: it fails if `min-width`/`min-height`
+// gets re-added to `.add-reaction-btn` without deliberately revisiting this.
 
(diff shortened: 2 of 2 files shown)


sanity (commented):
## Comprehensive PR Review: #606

### Summary
- **PR Title:** fix(ui): stop the add-reaction button from growing the reaction row
- **Type:** fix (follow-up to #605, already merged and published)
- **CI Status:** confirming on current HEAD before merge
- **Linked Issues:** none filed — follow-up to direct user feedback in River chat after #605 shipped
- **Review tier:** Light (small, self-contained CSS revert + test-assertion flip; the follow-up comment-only commit is Skip-tier and wasn't re-reviewed separately)
- **Reviewers run:** `freenet:skeptical-reviewer` + `freenet:code-first-reviewer`, both on `271ba6d9`. No external model attempted, per the standing "just use claude" directive.

---

### Code-First Analysis
**Independent Understanding:** #605 (merged, published) fixed the reaction "+" button being invisible-but-tappable on touch, and separately added `min-width`/`min-height: 2.75rem` to widen its tap target, mirroring an earlier fix (#462) for a different button. A user then reported the reaction row visibly grew taller. Root cause, confirmed by both reviewers independently: an `opacity: 0` element still occupies its layout box at its natural size — the button already had space reserved (~20px), and forcing a 44px minimum grew that on every message, not just ones with reactions. This PR removes just the two sizing lines, keeps the opacity/visibility fix, and flips the regression test's assertion to pin "no forced minimum size" as the considered decision.

**Stated Intent:** Matches exactly.

**Alignment:** Full match, verified independently by both reviewers via a full cascade trace of `.add-reaction-btn` in `main.css` (not just the diff hunk) — the opacity rules that fixed the original #605 bug are byte-identical; only the sizing lines were removed. Both reviewers also confirmed the sibling `.dm-archive-btn` rule (#462) still has its own `min-width`/`min-height`, so this PR didn't accidentally touch that.

**Gaps:** None in the code. One documentation accuracy gap, found independently by both reviewers (see below) — fixed in a follow-up commit.

---

### Testing Assessment
**Coverage Level:** Adequate

| E2E (Playwright) | ✅ | `add-reaction-touch.spec.ts`'s coarse-pointer assertion flipped from `>= 44` to `< 44`, verified by both reviewers to probe the same unchanged synthetic element construction — a pure assertion-direction change, not a different test. Mouse-only branch confirmed byte-for-byte unchanged. |

**Regression Test:** Present and valid. Verified via mutation testing (reintroduced the min-size rule in the built CSS output, confirmed the test goes red on `mobile-chrome`; restored, confirmed green).

**Missing Tests:** None blocking. Code-first reviewer noted a pre-existing (not introduced by this PR) test-fidelity gap: the synthetic probe doesn't carry the real button's Tailwind utility classes, so it pins "no min-size rule in main.css" rather than "the real rendered button is under 44px" — a distinction that would only matter if a future change added sizing via a different mechanism (e.g. padding utility classes on the button itself rather than a CSS rule). Worth a comment for future editors, not blocking here.

---

### Skeptical Findings
**Risk Level:** Low

| PR description and CSS comment overstated the kebab-menu "React" fallback as "already-sized" / "≥44px" | Low | `ui/assets/main.css:37-46` (fixed in `fb91ec04`), PR body (fixed via `gh pr edit`) | Both reviewers independently measured the actual kebab toggle (32px) and "React" menu row (~33-36px) and found neither reaches 44px — comfortably above WCAG 2.5.8 AA's 24px floor, but short of the AAA/HIG 44px target. Not a functional bug (the kebab path genuinely works and is wired to the same `on_react` handler, confirmed by the skeptical reviewer tracing `open_emoji_picker`/`on_react.call` through both the inline button and kebab-menu paths) — purely a documentation-accuracy issue. Fixed: corrected the code comment (comment-only commit, no functional change) and edited the PR description. |

No Medium or High findings. Both reviewers explicitly stated the change is otherwise clean, minimal, and correctly scoped.

Also verified independently (both reviewers): scope is exactly the two expected files (confirmed via `git diff origin/main --stat`); the kebab "React" alternate path is real, wired to the same reaction-adding flow, and available on both own and others' messages; no other CSS rule or test references the removed sizing.

---

### Big Picture Assessment
**Goal Alignment:** Yes — directly addresses real user feedback with a minimal, well-reasoned revert rather than a speculative "maybe this is fine" guess. Good instance of treating production feedback as real evidence (not N=1 noise) since it's a concrete, reproducible visual regression a user actually reported.

**Removed Code Concerns:** None — the regression test is strengthened (flipped to pin the new decision), not weakened or deleted.

**Scope Assessment:** Focused. Two files (three commits: the revert, then a comment-accuracy correction after review).

---

### Documentation
- Code docs: complete — the `main.css` comment explains the WHY (user feedback, opacity-0-still-occupies-layout-space mechanism, kebab-menu alternative with accurate sizing) in the same guard-comment style as the adjacent #402/#462 precedents.
- PR description: corrected post-review to match the code comment's accuracy.

---

### Recommendations

#### Must Fix (Blocking)
None.

#### Should Fix (Important)
None remaining — the one finding (overstated kebab sizing claim) is resolved in `fb91ec04` and the edited PR description.

#### Consider (Suggestions)
1. If the synthetic test probe in `add-reaction-touch.spec.ts` is ever extended, consider adding the real button's Tailwind utility classes to close the test-fidelity gap noted above (not blocking; the current probe is adequate for what it pins).

---

### Verdict
**State:** Ready to Merge (pending final CI confirmation on current HEAD SHA)
**HEAD SHA reviewed:** `271ba6d9` (functional diff); `fb91ec04` is a comment-only, Skip-tier follow-up correcting review feedback with zero CSS behavior change — re-verified balanced braces, no re-review needed per the mechanical-change carve-out.

[AI-assisted - Claude]

PR #624, opened 2026-08-18
feat: publish pointer records so integrators stop pinning keys that move

sanity:
## Problem

River's room contract and chat delegate **re-key on any WASM change** — the delegate roughly weekly, with 30 recorded generations in `legacy_delegates.toml`. Our own users are carried across that by the migration registries. A **third party** is not: their reference is a build-time constant, and after a re-key it silently addresses an empty namespace.

Every read then comes back looking like *"this user has nothing stored"*, which at the protocol level is indistinguishable from the truth. That is not hypothetical — it is [freenet/ghostkeys#21](https://github.com/freenet/ghostkeys/issues/21), where an integrating app told a user to buy a Ghost Key they already owned.

## Approach

Publish a **pointer record** for each artifact: a contract at a *fixed* address whose state names that artifact's current code hash, signed by River's author key. Integrators resolve it instead of pinning. The address never moves, so their constant never goes stale.

| `app_id` | Pointer key (fixed) |
|---|---|
| `river.room-contract` | `Ai4VLoC2jGdhpcB2UU8VPo3efUoxjm1Ju9VKXqRC63Az` |
| `river.chat-delegate` | `6qF2H5JRPBxbKC45UtPnzdDzyfsejYFW1UwDLGDU66mu` |

Implements the convention in [freenet-core#5194](https://github.com/freenet/freenet-core/issues/5194). River is the first publisher in the ecosystem. Signing and verification use the shared `pointer-record` tool from [freenet-migrate#26](https://github.com/freenet/freenet-migrate/pull/26), pinned by merge SHA — no signing code is added here.

## The CI gate matters more than the launch

**Once a pointer exists, a stale pointer is worse than no pointer.** Before one existed, an integrator pinning our key knew they were pinning it. After, they resolve, get a confident answer, and derive a dead key — the failure gets *quieter* exactly as more people depend on it.

So `check-pointer-freshness` runs on every PR and fails the build when a pointed-at WASM changed and no new record was signed. It deliberately asks a **different question** from `check-delegate-migration`: not *"does an entry exist?"* — which passes forever after the first record is committed — but *"does the record name **these** bytes?"*, which is meaningful on every commit.

Three checks per record, each able to fail alone:

1. `code_hash` == BLAKE3 of the committed WASM at `wasm_path`.
2. The 100-byte record **verifies** under the author key published in `FREENET.md`, and carries that same hash and version. Catches a hand-edited hash that left its signature behind, which check 1 alone would accept.
3. The version strictly increased if the record changed. A repeat version is a silent no-op on the network — the contract treats a stale update as success by design, so nothing downstream would tell us.

Plus a fourth, structural: **no record may vanish.** Deleting a record's block would otherwise pass, since every check is per-record and the gate happily verifies what remains.

### Every check was watched going red

A guard nobody has seen fail is not evidence it can.

| Fault injected | Result |
|---|---|
| append a byte to `chat_delegate.wasm`, sign nothing (**the real failure**) | RED — *"Any integrator resolving river.chat-delegate would derive a DEAD key"* |
| hand-edit `code_hash` | RED |
| hand-bump `version` without re-signing | RED |
| `FREENET.md` and the TOML disagree on the author key | RED |
| hand-edit `pointer_key` | RED |
| truncate a record | RED |
| mistyped flag falling through to local mode | RED |
| delete a whole record block (CI mode) | RED — names the missing `app_id` |
| zero records / missing WASM / empty TOML | RED |

## Verified against the live network, not just the repo

The record for each artifact was checked against the bytes **users are actually running**: the live web container was fetched from the network, its webapp archive unpacked, and the two WASMs inside hashed. Both match the committed copies.

```
LIVE (network)   e765339b…  room_contract.wasm      a44c6401…  chat_delegate.wasm
REPO (committed) e765339b…  room_contract.wasm      a44c6401…  chat_delegate.wasm
```

`publish-pointer-records.sh` repeats that check on every publish, because those two **do** diverge in practice — the live web container's version counter is currently two ahead of the committed `contract-version.txt`, which means at least one publish happened without the counter being committed. Worth a separate look; it does not affect the records here.

## The signing key

The same key that signs the web container, per Ian's decision of 2026-08-18. Safe against cross-app replay because the pointer contract signs over the **whole params blob** (`author_vk ‖ app_id`), so a record for one `app_id` cannot be replayed into another.

Two consequences are written down in `FREENET.md` and `pointer-records.toml` rather than left to be discovered: River does **not** keep this key offline, contrary to the pointer contract's own recommendation, because the same key signs every UI publish; and rotating it would move the web container's address **and** both pointer addresses at once, so it is a coordinated flag day rather than routine key hygiene.

## Scope, stated everywhere it could be misread

**A pointer solves addressing only.** It says nothing about whether state or secrets under the old key survived. This matters most for the delegate: secrets move only when River's own UI has run on that user's node, so an integrator can resolve `river.chat-delegate` perfectly and still find an **empty namespace** — which looks like *"this user has no data"* rather than like an error. That is precisely the confusion the pointer exists to remove, so conflating the two would reintroduce it one level up.

## Testing

Beyond the fault-injection table above: the gate was run in both local and CI mode, and the whole toolchain was exercised through the exact path CI takes — `cargo install --git … --rev 5e1759c --features publish --locked`, then the gate against that binary. Re-signing is deterministic (same inputs reproduce byte-identical records), verified by round-tripping the version counter.

No WASM was rebuilt (`.claude/rules/wasm-safety.md`) and **nothing is published to the network by this PR** — publishing is a separate step from `main` after merge.

## Follow-up

`cargo make publish-pointer-records` from `main` once this lands, which runs the five pre-flight gates and then re-reads each record back off the network and resolves it through the real resolver. "The PUT returned OK" is not evidence: a publish at an already-used version is a no-op *success*.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/delegate-migration.md
+++ b/.claude/rules/delegate-migration.md
@@ -63,17 +63,28 @@ cargo make add-room-contract-migration
 # 2. Build new WASMs and copy to all committed locations
 cargo make sync-wasm
 
+# 2b. Re-sign the pointer records against the WASM you just built.
+#     Step 1 carries OUR users forward; this carries THIRD PARTIES forward
+#     (see "The other half of a re-key" below). Same trigger, and CI's
+#     check-pointer-freshness fails the PR if you skip it.
+cargo make sign-pointer-records
+
 # 3. Run migration tests
 cargo test -p river-core --test migration_test
 cargo test -p river-core --test room_contract_migration_test
+cargo make check-pointer-freshness
 
 # 4. Verify UI compiles with new generated code
 cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync
 
 # 5. Commit everything
 git add legacy_delegates.toml common/legacy_room_contracts.toml \
-    ui/public/contracts/ cli/contracts/
+    pointer-records.toml ui/public/contracts/ cli/contracts/
 git commit -m "fix: update WASMs with delegate migration entry"
+
+# 6. AFTER the PR merges, from main: publish the re-signed records.
+#    Signing is offline and belongs in the PR; the network write does not.
+#    See .claude/rules/river-publish.md Step 7.
 ```
 
 ## The other half of a re-key: pointer records
--- a/.claude/rules/delegate-migration.md
+++ b/.claude/rules/delegate-migration.md
@@ -76,6 +76,32 @@ git add legacy_delegates.toml common/legacy_room_contracts.toml \
 git commit -m "fix: update WASMs with delegate migration entry"
 ```
 
+## The other half of a re-key: pointer records
+
+The registries above carry **our own users'** data across a re-key. They do
+nothing for a **third party** integrating with River, whose reference to our
+contract or delegate key is a build-time constant that silently goes stale —
+every read comes back looking like "this user has nothing stored".
+
+That half is `pointer-records.toml`: a record at a FIXED address naming the
+artifact's current code hash, signed by River's author key, which integrators
+resolve instead of pinning. Same trigger as a migration entry, different
+beneficiary.
+
+```bash
+cargo make sign-pointer-records      # after sync-wasm, BEFORE committing
+cargo make check-pointer-freshness   # what CI runs
+# then, from main after the PR merges:
+cargo make publish-pointer-records
+```
+
+CI's `check-pointer-freshness` fails the PR if a pointed-at WASM changed and no
+new record was signed. So in practice: whenever you run `add-migration`, you
+also need `sign-pointer-records`. See FREENET.md for the integrator-facing side,
+including the scope boundary — **a pointer solves addressing only** and says
+nothing about whether secrets survived the re-key, which is exactly what the
+rest of this file is about.
+
 ## Validation
 
 - **`cargo make check-migration`** — local check: builds delegate WASM and verifies migration entry exists if hash changed
@@ -83,6 +109,7 @@ git commit -m "fix: update WASMs with delegate migration entry"
 - **CI `check-delegate-migration` workflow** — builds base and PR WASMs, verifies old hash is in `legacy_delegates.toml`
 - **CI `check-room-contract-migration` workflow** — verifies a changed room-contract WASM's old hash is in `common/legacy_room_contracts.toml`
 - **CI `check-cli-wasm` workflow** — verifies `ui/public/contracts/` and `cli/contracts/` WASMs are in sync
+- **CI `check-pointer-freshness` workflow** — verifies every record in `pointer-records.toml` still names the committed WASM and is signed by the author key published in `FREENET.md`
 
 ## Key Files
 
@@ -96,6 +123,8 @@ git commit -m "fix: update WASMs with delegate migration entry"
 | `ui/src/components/app/chat_delegate.rs` | Uses generated `LEGACY_DELEGATES` for runtime migration |
 | `scripts/check-migration.sh` / `scripts/add-migration.sh` | Delegate migration validation / entry |
 | `scripts/check-room-contract-migration.sh` / `scripts/add-room-contract-migration.sh` | Room-contract registry validation / entry |
+| `pointer-records.toml` | Signed pointer records — third-party stable identity (addressing only) |
+| `scripts/check-pointer-freshness.sh` / `scripts/sign-pointer-records.sh` / `scripts/publish-pointer-records.sh` | Pointer gate / re-sign / network publish |
 | `scripts/sync-wasm.sh` | Builds all WASMs and copies to committed locations |
 | `common/tests/migration_test.rs` / `common/tests/room_contract_migration_test.rs` | Validate TOML entries are well-formed |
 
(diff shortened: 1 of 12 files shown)


sanity:
## Review: 3 lenses, 9 findings, all fixed

Full tier — this touches a production signing key and gates a release. Lenses: an external non-Claude model (Codex `gpt-5.6-sol`) plus two independent Claude reviewers reading blind (adversarial, big-picture). Every finding below was **reproduced before being fixed**, and the fixes re-verified.

### The two that let something through the gate

**1. A renamed `app_id` passed the vanished-record check.** The base commit's app_id was interpolated straight into a `grep -E` pattern. Both real app_ids contain a `.`, which is a regex wildcard, so `river.room-contract` matched `riverXroom-contract` — and renaming a record *is* the deletion this check exists to catch. Confirmed directly. Now compared as exact strings via `grep -qxF`. (The signing script already did this correctly with `re.escape`; the gate did not.)

**2. Reordering two records skipped the version check for both.** Base and head records were matched by *positional index*, with a guard that skipped the comparison when the app at that index differed — so swapping the blocks made every index mismatch and every record dodge the version gate at once, in a diff where nothing signals "skip me". Now looked up by `app_id`. Proven with the **full** bypass rather than half of it: reorder the blocks *and* regress one record from v2 back to its v1 bytes. That now fails with `the record changed but version did not increase: base version 2, head version 1`.

### The publish path

**3. A mid-loop failure skipped verification for records that had already published.** Publish and verify were separate loops; under `set -e` a failed `fdev` on record 2 killed the script before verification ran at all, including for record 1, already live. The script's own header argues re-reading is the only thing that proves a publish worked — and it was skipped in exactly the partial-failure case where that matters. Now one loop, with an explicit three-way report: published-and-verified / not published / **wrote but did not verify, treat as unknown state on the network**.

**4. Absence was inferred from a failed GET.** A timeout and a genuine "nothing there" both took the PUT branch, which *skips* the version-monotonicity check. `fdev` does distinguish them, and all three branches were verified against the live network before relying on it:

```
derived-but-unpublished pointer key -> PUT    (network said NOT FOUND)
River's live web container          -> UPDATE (1083217 bytes)
unreachable node port               -> REFUSE (ambiguous)
```

**5. The version rule `== onnet + 1` broke the recovery it prescribes.** After a partial publish the failure path says "re-run" — but the record that already succeeded is *at* the desired version, so the exact `+1` aborted before reaching the failed one. Same rule refused a legitimate local-v3-vs-network-v1 case. The contract's rule is monotonicity, not contiguity: byte-identical state is now "already done, verify only", and any strictly-greater version is accepted.

**6. The peer check counted only 17-character peer ids.** A peer id is base58 of 12 bytes, which is 16 characters ~18% of the time — measured 71 of 400 random samples, and 21 of the ~92 peers this node was actually connected to. The count was always low, and for a node whose peers all happened to be short it would be **zero**, aborting every publish with "no connected peers".

### The gate's own correctness

**7. It compared two different git trees.** On `pull_request`, `actions/checkout` provides the synthetic *merge* commit, so the registry and `FREENET.md` were read from base+head merged while the WASM was hashed from `head.sha`. If the base branch lands a new WASM and its record after this branch forked, the merged registry correctly carries base's new record and was then checked against head's **old** WASM — failing as stale on a PR that is fine. CI mode now materializes every head-side file from `$HEAD_SHA`, mirroring `check-migration.sh`.

**8. Three near-identical awk parsers.** A parsing fix to one would not reach the others, and the signer and the gate disagreeing about what a record says is the worst failure available here. Now one `scripts/pointer-toml-lib.sh`, which also fixed trailing-`# comment` handling (the old `gsub(/^"|"$/, "")` mangled `"abc" # note` into `abc" # note`).

### The instruction surface

**9. Pointer records were documented where an agent would not read them.** I traced the flows and found `AGENTS.md`, `river-publish.md` — the canonical publish checklist — and `wasm-safety.md` all had **zero** mentions, while all three tell you to run `cargo make sync-wasm`. Worse, `delegate-migration.md`'s numbered workflow ends in a `git add` line people copy verbatim, and it omitted `pointer-records.toml`. CI would have caught the omission, but a rule whose only expression is a failing build teaches nothing and lands on whoever is least equipped to read it. Now: Step 3b/Step 7 in the publish checklist, step 2b and a corrected `git add` in the migration workflow, an entry in wasm-safety, and the coupling in AGENTS.md.

### Verification

The full fault-injection battery re-run after every change — all eight still red, including the real one (append a byte to `chat_delegate.wasm`, sign nothing). `shellcheck` clean at every severity on all four scripts. The whole toolchain exercised through the exact path CI takes: `cargo install --git … --rev 5e1759c --features publish --locked`, then the gate against that binary. `check-pointer-freshness` is passing in real CI on this PR.

### The pattern

Not one of the nine was in signing or verification — that delegates wholly to freenet-migrate's `pointer-record` tool and was never in question. Every one was in shell argument handling, git-tree selection, output parsing, or documentation. Two of them were bypasses of the gate itself, which is the thing this PR exists to add; a gate that can be walked around is worse than none, because it is reassuring.

[AI-assisted - Claude]

PR #634, opened 2026-08-25
fix(publish): read the network before signing the web container

sanity:
## Problem

River's web-container publish path signed at `local counter + 1` and judged the result by `fdev`'s exit code. Neither is evidence, and on **2026-08-04 (commit `1032d373`)** the combination put two different archives on the network at the same version.

`cargo make publish-river` signed version 30000377 and uploaded it. `fdev network publish` reported `put timed out after 1 peer attempt(s)`, so the rollback arm in `[tasks.publish-river]` restored the counter to 30000376. The operator retried; `sign-webapp` rebuilt the UI — producing a **different archive**, because the build is not reproducible — signed it as 30000377 again, and published it. Only then did the network say `New state version 30000377 must be higher than current version 30000377`: the first PUT had landed all along.

**Why that forks the site rather than resolving.** The container is correctly gated — `update_state` rejects `version <= current_version` — which makes *differing* versions converge and does nothing for two states at the **same** version: it rejects in both directions, so each peer keeps whichever it saw first. And `summarize_state` emits only the `u32` version, so anti-entropy between two split peers compares equal summaries and never attempts to heal. The split is silent. The second archive reaches peers because a peer **new** to the contract takes the initial-state bypass, which runs `validate_state` only (signature, `version != 0`), not the gate.

River self-heals at the next successful publish, so exposure is bounded at one release cycle. Bounded, not harmless: for that window two populations of users are on different builds with no signal that they are.

## Approach

Ported, not invented. `scripts/publish-pointer-records.sh` already does the right thing for pointer records; this brings the same discipline to the web container, in `scripts/publish-web-container.sh`.

**1. Read before signing.** The version signed is `max(local counter, on-network version) + 1`, with the on-network version fetched and verified under our own contract parameters before use. The counter becomes bookkeeping; the network is the authority. A counter that fell behind — a rollback, a discarded working-tree change, a fresh clone — can no longer reissue a version the network has already seen. Strictly greater, never "exactly one more": gaps are fine, the contract enforces monotonicity, not contiguity.

**2. Prove absence, don't infer it.** `known` / `absent` / `unknown` are three separate states. This is **stricter than the pointer script on one point**, deliberately: that script treats `Contract not found` as proof of a first publish, which is right for a record that may never have been published. For the web container it is not proof of anything — a Freenet GET can dead-end and report NotFound for a contract that exists, and `published-contract/` says this one is live. So NotFound refuses too, behind its own `RIVER_WC_ALLOW_FIRST_PUBLISH=1` escape.

**3. Re-read after publishing.** `fdev`'s exit code is not evidence in either direction: a zero exit says the node we published *through* accepted the PUT, not that the bytes users fetch are ours, and a publish that reports a timeout may have landed anyway (2026-08-04). The script fetches the state back and compares the archive **byte for byte** against what it published. Signature validity cannot do this — both forked archives were validly signed by the same key — so the byte comparison is the only check that can tell "my state is live" from "a state at my version is live".

This also turns the 2026-08-04 case into a clean success: a publish that reports a timeout but whose state IS live now prints `PUBLISHED — despite fdev exiting N` and tells the operator explicitly not to retry.

**4. The counter no longer rolls back.** The rollback is what handed the retry a version that had, in fact, been used. Forward-only costs a counter gap and nothing else, and the read-before-sign floor recomputes from the network anyway, so a bumped-but-unpublished counter is harmless. A burned version is never reissued.

**5. The double bump.** `[tasks.publish-all]` ran `sign-webapp` and then `publish-river`, which re-ran `sign-webapp` as a dependency in a fresh `cargo make` process: two bumps and two UI builds for one publish, with the `.prev` snapshot overwritten so a failure rolled back only one of them. Those are the gaps visible in the counter's history. `publish-river` now owns the whole sequence and `publish-all` calls it once. Separately, `update-published-contract` now uses `export-parameters` instead of depending on `sign-webapp`, so refreshing the committed contract files no longer burns a counter value it never publishes.

**6. `flock`.** `[tasks.sign-webapp]` has carried a comment saying "wrap in `flock` if you need that" since the counter was introduced; nothing ever did. The lock is now taken by the script itself, across the whole sign to publish to verify sequence rather than the counter write alone — the gap between "sign at N" and "publish N" is where the fork lived. A **missing** `flock` warns rather than refusing: blocking a release because a lock utility is absent is a worse failure than the race it prevents.

### The could-not-determine decision, and why

**Confirmed lower version on the network: publish.** **Could not determine: refuse, with an explicit per-run override.**

Refusing when the network cannot be read may block a release that would have been fine. That failure is annoying but **loud and self-limiting**: the operator sees it, retries, or overrides with `RIVER_WC_ALLOW_UNVERIFIED=1`. Signing blind is **silent and lasts a release cycle**, because two states at one version never converge and anti-entropy cannot even see the difference. Loud-and-recoverable beats silent-and-forked, so ambiguity refuses by default and every escape is explicit.

The override is not a bypass of the whole fix: it skips only the pre-flight floor. The post-publish read-back still runs and still decides the exit code, so an override that publishes into a stale version is reported as `NOT PUBLISHED`, not as success.

## Testing

**What I tested.**

- `web-container-tool inspect` (new subcommand — parses the packed state, verifies the signature against the contract parameters, prints `version=N`, optionally extracts the archive) is unit-tested for the sign/pack/parse/verify round trip, foreign signatures, tampered archives, truncated and lying length fields, `--expect-version` mismatch, and the case where two different archives are both validly signed at one version. `cargo test -p web-container-tool`: 7 passed.
- `scripts/tests/web-container-publish-lib-test.sh` — every decision branch with stubbed inputs: the version floor (including the exact 30000376 with network-at-30000377 case), the three network statuses and their overrides, and the outcome classification including "rc != 0 but our bytes are live" and "rc == 0 but the network is still behind". 27 checks.
- `scripts/tests/web-container-publish-rehearsal.sh` — the **real** publish script and the **real** `web-container-tool` against a throwaway repo layout, a throwaway signing key, and an `fdev` that is a shell script modelling the container's own version gate, plus stubbed `gh` and `cargo`. 36 scenarios, 138 checks, including the 2026-08-04 shape end to end and the retry that used to fork the site. This covers the wiring a pure function cannot: `fdev` argument order, the `Contract not found` string absence is keyed off, the `version=` line the script parses, and that the verdict comes from the read-back.
- **Mutation-tested, not just green.** Removing the network floor fails scenario 3 and 2 lib checks; making the verdict trust `PUB_RC` fails scenario 2; removing the byte comparison fails scenario 10 and 1 lib check. Each was applied and confirmed red before being reverted.
- `shellcheck` clean on all three scripts; `cargo fmt --check` clean; `cargo clippy -p web-container-tool` clean (the 5 warnings in `river-core` are pre-existing toolchain drift, and `clippy.yml` is disabled).
- `cargo make --list-all-steps` loads the rewritten `Makefile.toml`.

Both shell suites are wired into `build.yml`, so they are not tests nobody runs.

**What I did NOT test.**

- **Nothing was published to the network**, and no real publish was run end to end. Every `fdev` invocation in the tests is the stub.
- The real `fdev` argument order and the real `Contract not found` text are matched against `freenet-core`'s `crates/fdev/src/commands.rs` and the existing pointer script, not against a live node. If `fdev`'s CLI or that error string changes, the rehearsal's stub changes with it and stays green — a real `fdev` regression would surface at publish time, not in CI.
- The `flock` path is exercised only in its "lock acquired" form; I did not rehearse two concurrent runs contending.
- The `RIVER_WC_ALLOW_FIRST_PUBLISH=1` path is covered only in the library test, not end to end.

## Residual risk after this change

- **Between the pre-flight read and the PUT there is still a window.** If something else published in that gap, we sign above a version that is no longer live. The read-back catches it (`superseded`) but only after the fact. In practice one key publishes this contract, so the window needs two concurrent operators — which the `flock` covers on one machine and nothing covers across machines.
- **An existing fork is not repaired by this change.** It prevents new ones and detects a live one (`FORKED` verdict, with the remedy: re-run, which signs above it and heals both branches). If the 2026-08-04 fork is somehow still live, it needs a publish, not this PR.
- **The override remains a footgun.** `RIVER_WC_ALLOW_UNVERIFIED=1` in a shell profile or a wrapper script would silently restore the old behaviour for the pre-flight half. Only the read-back would still guard.
- **`--sign-only` can still leave a signed artifact that is never published**, advancing the counter. That is now harmless (forward-only, and the floor recomputes) but it does widen counter gaps.
- **The test-contract path (`sign-webapp-test` / `publish-river-test`) still uses the bump-then-roll-back scheme.** Left deliberately — that contract is throwaway and a fork of it costs nothing — with a comment saying not to copy the pattern back.
- **This does not make the UI build reproducible.** A rebuild still produces a different archive; the fix ensures a different archive can never land at the same version, not that it is the same archive.

Refs the 2026-08-04 incident and commit `1032d373`.

https://claude.ai/code/session_0131oS6YAD9gMWigX8v4QR4J

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/river-publish.md
+++ b/.claude/rules/river-publish.md
@@ -81,17 +81,21 @@ cargo fmt
 ```bash
 git add legacy_delegates.toml ui/public/contracts/ cli/contracts/
 git commit -m "fix: <description> with delegate migration"
-cargo make build
-cargo make compress-webapp
-cargo make publish-river  # bumps published-contract/contract-version.txt
+cargo make publish-river  # builds, signs, publishes, and verifies the result
 ```
+
+`publish-river` owns the whole sequence in one process, under one lock. Do NOT
+run `cargo make compress-webapp` or `cargo make sign-webapp` beforehand: each is
+a separate `cargo make` process, so `publish-river` re-runs them as
+dependencies — rebuilding the UI (which is not reproducible, so the archive
+changes) and advancing the version counter a second time for a single publish.
+That is where the gaps in the counter's history came from.
 
 ### Step 6: Commit the version bump, verify, and push
 
 ```bash
-# `cargo make sign-webapp` (run transitively by publish-river) incremented
-# the counter. Commit it together with whatever other changes the publish
-# included.
+# The publish advanced published-contract/contract-version.txt. Commit it
+# together with whatever other changes the publish included.
 git add published-contract/contract-version.txt
 git commit -m "chore: bump web-container version after publish"
 curl -s http://127.0.0.1:7509/v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/ | head -5
@@ -116,7 +120,11 @@ and verifies it. Do not accept "the PUT returned OK": a publish at an
 already-used version is a no-op *success*, so the network will never tell you
 it was ignored.
 
-**Version policy:** the canonical source is `published-contract/contract-version.txt`, which `cargo make sign-webapp` increments and writes back each publish. NEVER base the version on wall-clock time (`date +%s / 60`) — the previous scheme bit us 2026-05-16 when the on-network version had drifted ahead of the timestamp-derived value. The counter file makes the version a strict monotonic local invariant; gaps are fine (the contract enforces monotonicity, not contiguity).
+**Version policy:** the counter at `published-contract/contract-version.txt` is local bookkeeping; the authority is the version **on the network**. `scripts/publish-web-container.sh` (which `cargo make publish-river` and `cargo make sign-webapp` both run) reads that version, verifies it under our own contract parameters, and signs at `max(counter, network) + 1`. NEVER base the version on wall-clock time (`date +%s / 60`) — that scheme bit us 2026-05-16 when the on-network version had drifted ahead of the timestamp-derived value. Gaps are fine (the contract enforces monotonicity, not contiguity).
+
+**Never reissue a version, and never roll the counter back.** 2026-08-04 (commit `1032d373`): a publish of version 30000377 reported `put timed out after 1 peer attempt(s)`, the counter was rolled back, and the retry rebuilt the UI — a *different* archive, because the build is not reproducible — signed it 30000377 again, and published it. The first PUT had landed. Two validly signed archives now existed at one version, and they do not converge: `update_state` rejects `version <= current` in **both** directions, and `summarize_state` emits only the u32 version, so anti-entropy between two split peers sees matching summaries and never heals. The second archive spreads because a peer new to the contract takes the initial-state bypass, which runs `validate_state` only. River self-heals at the next successful publish, so exposure is one release cycle — bounded, not harmless.
+
+The counter is therefore forward-only, and the publish script re-reads the state after publishing rather than trusting `fdev`'s exit code (a publish at an already-used version is a no-op *success*; a publish that reports a timeout may have landed). If a publish reports failure, **read the verdict block** — if it says PUBLISHED despite a non-zero fdev exit, do not retry.
 
 ## Two independent release surfaces
 
--- a/.claude/rules/river-publish.md
+++ b/.claude/rules/river-publish.md
@@ -76,11 +76,21 @@ cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync
 cargo fmt
 ```
 
-### Step 5: Commit, Build, and Publish
+### Step 5: Merge, then Build and Publish from main
+
+The web container is ONE contract key serving every River user, and there is no
+per-branch address to publish to instead — so the publish runs from a green
+`main`, never from the branch. `scripts/publish-web-container.sh` enforces this
+before it signs anything: clean tracked tree, on `main`, `HEAD == origin/main`,
+and CI green on that exact SHA (an empty list of runs is *not* green). It
+refuses otherwise. `RIVER_WC_ALLOW_UNPROVEN=1` skips the gate for one run; it
+is for an emergency you can justify, not for routine convenience.
 
 ```bash
 git add legacy_delegates.toml ui/public/contracts/ cli/contracts/
 git commit -m "fix: <description> with delegate migration"
+# open the PR, get it reviewed, wait for CI green, merge
+git checkout main && git pull
 cargo make publish-river  # builds, signs, publishes, and verifies the result
 ```
 
@@ -124,7 +134,9 @@ it was ignored.
 
 **Never reissue a version, and never roll the counter back.** 2026-08-04 (commit `1032d373`): a publish of version 30000377 reported `put timed out after 1 peer attempt(s)`, the counter was rolled back, and the retry rebuilt the UI — a *different* archive, because the build is not reproducible — signed it 30000377 again, and published it. The first PUT had landed. Two validly signed archives now existed at one version, and they do not converge: `update_state` rejects `version <= current` in **both** directions, and `summarize_state` emits only the u32 version, so anti-entropy between two split peers sees matching summaries and never heals. The second archive spreads because a peer new to the contract takes the initial-state bypass, which runs `validate_state` only. River self-heals at the next successful publish, so exposure is one release cycle — bounded, not harmless.
 
-The counter is therefore forward-only, and the publish script re-reads the state after publishing rather than trusting `fdev`'s exit code (a publish at an already-used version is a no-op *success*; a publish that reports a timeout may have landed). If a publish reports failure, **read the verdict block** — if it says PUBLISHED despite a non-zero fdev exit, do not retry.
+The counter is therefore forward-only, and the publish script re-reads the state after publishing rather than trusting `fdev`'s exit code. A zero exit says the node you published *through* accepted the PUT, not that the bytes users fetch are yours; and a publish that reports a timeout may have landed anyway, which is what 2026-08-04 actually was. If a publish reports failure, **read the verdict block** — if it says PUBLISHED despite a non-zero fdev exit, do not retry.
+
+Note the web container does **not** behave like the pointer contract here. An earlier version of this file said a web-container publish at an already-used version is a "no-op success"; that is wrong. `update_state` returns `InvalidUpdateWithInfo` for `version <= current` and freenet-core surfaces a failed PutResponse as an error to the publishing node — which is how the 2026-08-04 operator saw `New state version 30000377 must be higher than current version 30000377` at all. The silent-stale-update property belongs to the **pointer** contract (step 7 above), by design, and does not carry across.
 
 ## Two independent release surfaces
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -278,6 +278,30 @@ jobs:
       # tool and the verifying contract agree on the parameter encoding.
       run: cargo test -p web-container-contract -p web-container-tool
 
+    - name: Test web-container publish decisions
+      # The publish path cannot be rehearsed in CI — it needs a live network,
+      # and testing against the shared production contract is out of bounds. So
+      # the decisions it turns on (which version to sign at, whether an
+      # unreadable network may be published over, whether a publish actually
+      # landed) live in scripts/web-container-publish-lib.sh with no I/O in
+      # them, and this runs every branch with stubbed inputs. Without this step
+      # the file would be exactly the "tests nothing ever runs" shape the
+      # coverage guard above exists to prevent, just outside Cargo's reach.
+      run: ./scripts/tests/web-container-publish-lib-test.sh
+
+    - name: Rehearse the web-container publish against a stub node
+      env:
+        CARGO_TARGET_DIR: ${{ github.workspace }}/target
+      # Runs the REAL publish script and the REAL web-container-tool against a
+      # throwaway repo layout, a throwaway signing key and an `fdev` that is a
+      # shell script — including the 2026-08-04 shape (a publish that reports a
+      # timeout and lands anyway). This is the only coverage of the wiring:
+      # fdev argument order, the `Contract not found` string absence is keyed
+      # off, the `version=` line the script parses, and whether the verdict
+      # comes from the read-back rather than the exit code. It never touches
+      # the network or the production key.
+      run: ./scripts/tests/web-container-publish-rehearsal.sh
+
   ui-playwright-tests:
     # Ubicloud, NOT a GitHub-hosted larger runner -- see the `build` job above.
     runs-on: ubicloud-standard-8
--- a/.gitignore
+++ b/.gitignore
@@ -55,9 +55,10 @@ ui/tests/test-results/
 # an older checkout cannot be committed by accident.
 /published-contract/contract-version.txt.prev
 
-# Single-writer lock for the web-container publish. Two concurrent publishes
-# race the version counter.
-/published-contract/.publish.lock
+# (The single-writer lock for the web-container publish is deliberately NOT
+# here. It lives at $XDG_RUNTIME_DIR/river-web-container-<contract-id>.lock,
+# outside the checkout: a lock inside the repo gives two worktrees two
+# different lock files and serialises nothing.)
 
 # Syncthing in-flight transfer files (`.syncthing.<name>.tmp` are
 # created by the syncthing daemon while a file is being received).
--- a/.gitignore
+++ b/.gitignore
@@ -48,11 +48,16 @@ ui/tests/test-results/
 # Manual e2e test screenshots (see #244 review feedback)
 /dm-*.png
 
-# Web-container version-counter rollback snapshot. Created by
-# `cargo make sign-webapp` BEFORE bumping the counter; removed on
-# successful publish, or moved back over the counter on failure.
-# Should never appear in commits.
+# Web-container version-counter rollback snapshot. No longer written — the
+# counter is forward-only now, because rolling it back after a publish that had
+# actually landed is what forked the site on 2026-08-04 (see
+# scripts/publish-web-container.sh). Kept ignored so a stale one left over from
+# an older checkout cannot be committed by accident.
 /published-contract/contract-version.txt.prev
+
+# Single-writer lock for the web-container publish. Two concurrent publishes
+# race the version counter.
+/published-contract/.publish.lock
 
 # Syncthing in-flight transfer files (`.syncthing.<name>.tmp` are
 # created by the syncthing daemon while a file is being received).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -188,10 +188,15 @@ artifacts.
 
 Two things that follow from that, and matter if a publish goes wrong:
 
-- **`fdev`'s exit code decides nothing.** A publish at an already-used
-  version is a no-op *success*, and a publish that reports a timeout may
-  have landed anyway. Read the script's verdict block, not the exit status
-  of the publish inside it.
+- **`fdev`'s exit code decides nothing.** A zero exit says the node you
+  published *through* accepted the PUT, not that the bytes users fetch are
+  yours; and a publish that reports a timeout may have landed anyway, which
+  is what 2026-08-04 actually was. Read the script's verdict block, not the
+  exit status of the publish inside it. (The web container does **not**
+  silently accept a stale update: `update_state` returns
+  `InvalidUpdateWithInfo` for `version <= current`, which is how the
+  2026-08-04 operator saw the rejection message. That no-op-success property
+  belongs to the *pointer* contract, by design.)
 - **Never retry a publish by re-signing at the same version, and never roll
   the counter back.** The counter is forward-only on purpose; gaps are fine
   (the contract enforces monotonicity, not contiguity). Reissuing a version
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -179,9 +179,25 @@ cargo make publish-river                    # Publish UI release build to Freene
 The web container contract requires signed metadata with a version
 number strictly higher than the current published version. The version
 is tracked by a committed counter at
-`published-contract/contract-version.txt`; `cargo make sign-webapp`
-increments it. Commit the bumped counter alongside other publish
+`published-contract/contract-version.txt`, but the counter is bookkeeping,
+not the authority: `scripts/publish-web-container.sh` READS the version on
+the network before signing and signs at `max(counter, network) + 1`, then
+re-reads after publishing and compares the archive byte for byte before
+declaring success. Commit the bumped counter alongside other publish
 artifacts.
+
+Two things that follow from that, and matter if a publish goes wrong:
+
+- **`fdev`'s exit code decides nothing.** A publish at an already-used
+  version is a no-op *success*, and a publish that reports a timeout may
+  have landed anyway. Read the script's verdict block, not the exit status
+  of the publish inside it.
+- **Never retry a publish by re-signing at the same version, and never roll
+  the counter back.** The counter is forward-only on purpose; gaps are fine
+  (the contract enforces monotonicity, not contiguity). Reissuing a version
+  that may already be in use forked the site on 2026-08-04 — two validly
+  signed archives at version 30000377, which the update gate cannot converge
+  and anti-entropy cannot even see.
 
 **Contract ID:** `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`
 
--- a/DEVNOTES.md
+++ b/DEVNOTES.md
@@ -27,9 +27,14 @@ River is a decentralized chat application built on Freenet with the following ke
 ## Deployment Process
 
 1. **Build the UI:** `cargo make build-ui`
-2. **Compress the webapp:** `cargo make compress-webapp`
-3. **Sign the webapp:** `cargo make sign-webapp`
-4. **Publish to Freenet:** `cargo make publish-river`
+2. **Publish to Freenet:** `cargo make publish-river`
+
+`publish-river` builds, compresses, signs and publishes in one process, then
+reads the state back off the network to check it actually landed. Do not run
+`cargo make sign-webapp` first: it is a separate `cargo make` process, so
+`publish-river` would build and sign a second time, and the version counter
+would advance twice for one publish. The full workflow, including the
+delegate-migration coupling, is in `.claude/rules/river-publish.md`.
 
 ## Testing Challenges
 
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -200,50 +200,29 @@ tar -cJf ../../../../../webapp/webapp.tar.xz *
 '''
 
 [tasks.sign-webapp]
-description = "Sign the compressed webapp"
+# The version counter committed at published-contract/contract-version.txt is
+# still the local bookkeeping (the previous `seconds/60` timestamp scheme was
+# abandoned after the 2026-05-16 incident: the on-network version had drifted
+# to 30000208 while the timestamp gave 29649402, so publishes were rejected
+# with no way to catch up). What changed is that the counter is no longer
+# TRUSTED. The version signed here is
+#
+#     max(local counter, version currently on the network) + 1
+#
+# with the on-network version READ from the network and verified under our own
+# contract parameters before signing. It is also forward-only — nothing rolls
+# the counter back any more, because the rollback is what caused the 2026-08-04
+# fork (River commit 1032d373): a timed-out publish that had actually landed
+# was rolled back, and the retry re-signed a DIFFERENT archive at the same
+# version. Two states at one version never converge.
+#
+# All of it, including the flock the old comment here asked for and nothing
+# ever added, lives in scripts/publish-web-container.sh. Read that file's
+# header before changing any of this.
+description = "Sign the compressed webapp at a version proven to be above the network's"
 dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]
 script = '''
-# Bump the version counter committed at published-contract/contract-version.txt
-# and use that as the metadata version. The previous `seconds/60` timestamp
-# scheme was fragile — if any past publish ever ran with a clock skewed
-# ahead (or someone hand-bumped to a future round number), the on-network
-# version is stuck ahead and the timestamp-derived value can't catch up
-# without manual intervention. 2026-05-16 incident: the on-network version
-# was 30000208 while the timestamp gave 29649402, so the publish was
-# rejected.
-#
-# Counter lifecycle:
-# 1. `sign-webapp` here writes a `.prev-version` snapshot BEFORE bumping
-#    the counter file, so a downstream `publish-river` failure can roll
-#    back atomically (see [tasks.publish-river]'s `on-failure` arm).
-# 2. The counter file is bumped to `current+1` and the new metadata is
-#    signed with that value.
-# 3. After a successful `fdev publish`, the caller commits the bumped
-#    counter file (the snapshot is left for the publish task to clean
-#    up). Skeptical-review (#258) flagged that bumping before publish
-#    risks leaving the local file ahead of on-network if publish fails;
-#    the .prev-version snapshot + rollback closes that window.
-#
-# Locking: this script is NOT safe under two parallel `publish-river`
-# runs on the same workstation — they would race the read-modify-write.
-# Wrap in `flock` if you need that:
-#     flock published-contract/contract-version.txt cargo make publish-river
-version_file="published-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
-if [ ! -f "$version_file" ]; then
-    echo "ERROR: $version_file missing. Seed it with the current on-network version."
-    exit 1
-fi
-current=$(cat "$version_file" | tr -d '[:space:]')
-version=$((current + 1))
-cp "$version_file" "$prev_snapshot"
-echo "$version" > "$version_file"
-echo "Web-container version: $current -> $version (commit $version_file after a successful publish)"
-target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool sign \
-  --input target/webapp/webapp.tar.xz \
-  --output target/webapp/webapp.metadata \
-  --parameters target/webapp/webapp.parameters \
-  --version $version
+exec ./scripts/publish-web-container.sh --sign-only
 '''
 
 [tasks.build-ui-example]
@@ -328,9 +307,21 @@ cd ui/tests && npm install && npx playwright install --with-deps chromium firefo
 '''
 
 [tasks.update-published-contract]
-description = "Build, publish, and update the published contract files"
-dependencies = ["sign-webapp"]
+description = "Refresh published-contract/ (wasm, parameters, contract id)"
+# This needs the PARAMETERS (the 32-byte verifying key), not a signature, so it
+# uses `export-parameters` rather than depending on `sign-webapp`. It used to
+# depend on `sign-webapp`, which meant refreshing the committed contract files
+# silently burned a version counter value it never published — one of the
+# sources of the gaps visible in the counter's history. Since sign-webapp now
+# also reads the network before signing, that dependency would additionally
+# have made this task need a reachable node for no reason.
+dependencies = ["build-web-container", "build-web-container-tool"]
 script = '''
+set -e
+mkdir -p target/webapp
+target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool export-parameters \
+  --parameters target/webapp/webapp.parameters
+
 # Get the contract ID without publishing
 contract_id=$(fdev get-contract-id \
   --code target/wasm32-unknown-unknown/release/web_container_contract.wasm \
@@ -362,68 +353,25 @@ echo "Please commit the changes to git"
 # from main after the merge. Migration entries carry OUR users forward; the
 # pointer tells everyone ELSE where the artifact went. Same trigger, two
 # different sets of people who lose if it is skipped.
-description = "Publish River to Freenet using the committed contract version"
-dependencies = ["sign-webapp"]
+#
+# The sign-publish-verify sequence lives in scripts/publish-web-container.sh so
+# it can hold ONE lock across the whole thing and re-read the network after the
+# publish. `dependencies` deliberately no longer names `sign-webapp`: signing
+# happens inside that script, under the same lock as the publish, because the
+# gap between "sign at version N" and "publish version N" is where the
+# 2026-08-04 fork lived.
+description = "Publish River to Freenet, verifying the result against the network"
+dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]
 script = '''
-if [ ! -f "published-contract/contract-id.txt" ]; then
-    echo "No published contract found"
-    exit 1
-fi
-contract_id=$(cat published-contract/contract-id.txt)
-echo "Publishing using committed contract: $contract_id"
-version_file="published-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
-if fdev network publish \
-  --code published-contract/web_container_contract.wasm \
-  --parameters published-contract/webapp.parameters \
-  contract \
-  --webapp-archive target/webapp/webapp.tar.xz \
-  --webapp-metadata target/webapp/webapp.metadata; then
-    # Publish succeeded — the bumped counter file is the new truth.
-    # Discard the snapshot. Caller must commit $version_file.
-    rm -f "$prev_snapshot"
-    echo "Publish OK. Remember to commit $version_file."
-else
-    rc=$?
-    # Publish failed — roll the counter back so subsequent retries
-    # don't accumulate a drift. The on-network version is unchanged,
-    # and our local counter is now matched to it again.
-    if [ -f "$prev_snapshot" ]; then
-        mv "$prev_snapshot" "$version_file"
-        echo "Publish failed (exit $rc); rolled $version_file back to its pre-publish value." >&2
-    fi
-    exit $rc
-fi
+exec ./scripts/publish-web-container.sh
 '''
 
 [tasks.publish-river-debug]
 description = "Publish River to Freenet in debug mode"
 env = { BUILD_PROFILE = "dev" }
-dependencies = ["sign-webapp"]
+dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]
 script = '''
-if [ ! -f "published-contract/contract-id.txt" ]; then
-    echo "No published contract found"
-    exit 1
-fi
-contract_id=$(cat published-contract/contract-id.txt)
-echo "Publishing using committed contract in debug mode: $contract_id"
-version_file="published-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
-if fdev network publish \
-  --code published-contract/web_container_contract.wasm \
-  --parameters published-contract/webapp.parameters \
-  contract \
-  --webapp-archive target/webapp/webapp.tar.xz \
-  --webapp-metadata target/webapp/webapp.metadata; then
-    rm -f "$prev_snapshot"
-else
-    rc=$?
-    if [ -f "$prev_snapshot" ]; then
-        mv "$prev_snapshot" "$version_file"
-        echo "Publish failed (exit $rc); rolled $version_file back to its pre-publish value." >&2
-    fi
-    exit $rc
-fi
+exec ./scripts/publish-web-container.sh
 '''
 
 # Test publishing tasks - use separate keypair to avoid affecting production
@@ -550,6 +498,13 @@ fi
 # as the production publish (see [tasks.sign-webapp] for the 2026-05-16
 # incident). Test contract counter is NOT committed since the test
 # contract ID is regenerated freely.
+#
+# NOTE: this still uses the bump-then-roll-back scheme the PRODUCTION path
+# dropped, because rolling the counter back after a publish that had actually
+# landed is what forked the site on 2026-08-04. It is left here deliberately:
+# the test contract is throwaway, so a fork of it costs nothing, and nobody
+# reads it. Do NOT copy this pattern back to the production path — see
+# scripts/publish-web-container.sh.
 # .prev snapshot is captured here for the publish task to roll back on
 # failure — same pattern as the production sign/publish flow.
 version_file="test-contract/contract-version.txt"
@@ -659,30 +614,33 @@ echo "========================================="
 echo ""
 
 # Step 1: Build and publish River UI
-echo "▶ Step 1/5: Building River UI..."
-cargo make compress-webapp
-cargo make sign-webapp
-
-echo ""
-echo "▶ Step 2/5: Publishing River UI to Freenet..."
+#
+# One `cargo make publish-river`, not compress + sign + publish. Those two
+# extra lines used to be here and each ran in its OWN cargo-make process, so
+# `publish-river` re-ran them as dependencies: the UI was built twice and the
+# version counter was bumped TWICE for a single publish, while the rollback
+# snapshot was overwritten so a failure rolled back only one of the bumps.
+# That is where the gaps in the counter's history came from. `publish-river`
+# already depends on the build, and now does the signing itself.
+echo "▶ Step 1/4: Building and publishing River UI to Freenet..."
 cargo make publish-river
 
 # Step 2: Bump riverctl version
 echo ""
-echo "▶ Step 3/5: Bumping riverctl version..."
+echo "▶ Step 2/4: Bumping riverctl version..."
 cargo make bump-cli-version
 
 NEW_VERSION=$(grep '^version' cli/Cargo.toml | head -1 | sed 's/.*"\(.*\)".*/\1/')
 
 # Step 3: Commit the version bump
 echo ""
-echo "▶ Step 4/5: Committing version bump..."
+echo "▶ Step 3/4: Committing version bump..."
 git add cli/Cargo.toml
 git commit -m "chore: bump riverctl to v${NEW_VERSION} (WASM sync)"
 
 # Step 4: Publish to crates.io
 echo ""
-echo "▶ Step 5/5: Publishing riverctl v${NEW_VERSION} to crates.io..."
+echo "▶ Step 4/4: Publishing riverctl v${NEW_VERSION} to crates.io..."
 cd cli
 cargo publish
 cd ..
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -219,11 +219,33 @@ tar -cJf ../../../../../webapp/webapp.tar.xz *
 # All of it, including the flock the old comment here asked for and nothing
 # ever added, lives in scripts/publish-web-container.sh. Read that file's
 # header before changing any of this.
+#
+# Signing is gated the same way publishing is (clean tree, on main, at
+# origin/main, CI green on that SHA): it burns a version and hands back an
+# artifact `fdev network publish` will take by hand, so it is a release step,
+# not a build step. RIVER_WC_ALLOW_UNPROVEN=1 skips that gate for one run.
 description = "Sign the compressed webapp at a version proven to be above the network's"
-dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]
+# No `dependencies`. The script builds them itself via
+# [tasks.web-container-build-inputs], from INSIDE the publish lock — see that
+# task's comment.
 script = '''
-exec ./scripts/publish-web-container.sh --sign-only
+exec ./scripts/publish-web-container.sh --sign-only --build
 '''
+
+[tasks.web-container-build-inputs]
+# Deliberately a task with dependencies and NO script, invoked by
+# scripts/publish-web-container.sh once it holds the single-writer lock.
+#
+# These used to be `dependencies` of [tasks.publish-river] and
+# [tasks.sign-webapp]. cargo-make runs a task's dependencies to completion
+# BEFORE its script, so they ran before the script could take the lock: two
+# overlapping `cargo make publish-river` runs both executed `compress-webapp`
+# and both wrote the shared `target/webapp/webapp.tar.xz` with no lock held,
+# and the second could replace the archive while the first was signing,
+# publishing or byte-comparing it. The lock covered sign-to-verify but not the
+# artifact being verified. Do not move these back into a `dependencies` list.
+description = "Build the artifacts the web-container publish consumes"
+dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]
 
 [tasks.build-ui-example]
 description = "Build the Dioxus UI with example data"
@@ -360,18 +382,26 @@ echo "Please commit the changes to git"
 # happens inside that script, under the same lock as the publish, because the
 # gap between "sign at version N" and "publish version N" is where the
 # 2026-08-04 fork lived.
+#
+# Run it from a green `main`, after the PR has merged. This contract key is
+# shared by every River user with no per-branch address to publish to instead,
+# so the script refuses a dirty tree, a branch that is not main, a HEAD that is
+# not origin/main, or CI that is not green on that exact SHA — the same
+# pre-flight scripts/publish-pointer-records.sh makes.
 description = "Publish River to Freenet, verifying the result against the network"
-dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]
+# No `dependencies` — see [tasks.web-container-build-inputs].
 script = '''
-exec ./scripts/publish-web-container.sh
+exec ./scripts/publish-web-container.sh --build
 '''
 
 [tasks.publish-river-debug]
 description = "Publish River to Freenet in debug mode"
 env = { BUILD_PROFILE = "dev" }
-dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]
+# No `dependencies` — see [tasks.web-container-build-inputs]. The script passes
+# BUILD_PROFILE through with `cargo make --env`, because a cargo-make [env]
+# block overrides an inherited variable of the same name.
 script = '''
-exec ./scripts/publish-web-container.sh
+exec ./scripts/publish-web-container.sh --build
 '''
 
 # Test publishing tasks - use separate keypair to avoid affecting production
@@ -635,7 +665,13 @@ NEW_VERSION=$(grep '^version' cli/Cargo.toml | head -1 | sed 's/.*"\(.*\)".*/\1/
 # Step 3: Commit the version bump
 echo ""
 echo "▶ Step 3/4: Committing version bump..."
-git add cli/Cargo.toml
+# published-contract/contract-version.txt was advanced by step 1 and is
+# committed HERE. It used to be left dirty, which quietly undoes the reason the
+# counter is forward-only: a stray `git checkout`, `git stash` or discarded
+# working-tree change restores a version that has already been published, and
+# reissuing a used version is the 2026-08-04 fork. `git add` on it is a no-op
+# if step 1 refused before signing.
+git add cli/Cargo.toml published-contract/contract-version.txt
 git commit -m "chore: bump riverctl to v${NEW_VERSION} (WASM sync)"
 
 # Step 4: Publish to crates.io
--- a/contracts/web-container-contract/web-container-tool/src/main.rs
+++ b/contracts/web-container-contract/web-container-tool/src/main.rs
@@ -1,5 +1,5 @@
 use clap::{Parser, Subcommand};
-use ed25519_dalek::{Signer, SigningKey};
+use ed25519_dalek::{Signature, Signer, SigningKey, VerifyingKey};
 use river_core::crypto_values::CryptoValue;
 use river_core::web_container::WebContainerMetadata;
 use std::fs;
@@ -58,6 +58,38 @@ enum Commands {
         /// Key file to use (default: ~/.config/river/web-container-keys.toml)
         #[arg(long, short)]
         key_file: Option<String>,
+    },
+    /// Read a PACKED web-container state (the bytes `fdev execute get` returns)
+    /// and print its version, so the publish path can compare its intended
+    /// version against the one the network actually holds.
+    ///
+    /// This exists because the publish path had no way to answer "what version
+    /// is live right now?". Without it, `sign-webapp` signed at
+    /// `local_counter + 1` and `publish-river` trusted `fdev`'s exit code —
+    /// and on 2026-08-04 (River commit 1032d373) that combination signed two
+    /// DIFFERENT archives at the same version 30000377 and put both on the
+    /// network. See scripts/publish-web-container.sh for the whole story.
+    ///
+    /// Prints `version=<n>` on stdout, in the same `key=value` shape
+    /// `pointer-record verify` uses, so callers can `sed -n 's/^version=//p'`.
+    Inspect {
+        /// Packed state file (`[u64 metadata_len][metadata][u64 web_len][web]`)
+        #[arg(long, short)]
+        state: String,
+        /// Contract parameters (the raw 32-byte verifying key). When given,
+        /// the embedded signature MUST verify under it or this fails.
+        ///
+        /// Pass it. A version read out of unverified bytes is not evidence of
+        /// anything: it is whatever the responder chose to say.
+        #[arg(long)]
+        parameters: Option<String>,
+        /// Write the embedded webapp archive here, so the caller can compare
+        /// it byte-for-byte against the archive it published.
+        #[arg(long)]
+        archive_out: Option<String>,
+        /// Fail unless the state's version is exactly this.
+        #[arg(long)]
+        expect_version: Option<u32>,
     },
 }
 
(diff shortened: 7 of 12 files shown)


sanity (commented):
## Comprehensive PR Review: #634

### Summary

- **PR Title:** fix(publish): read the network before signing the web container
- **Type:** fix — release/publish path
- **Review tier:** **Full** — changes the publish path for the web container, an always-Full surface. Every publish hits one fixed shared contract key serving every River user.
- **HEAD SHA reviewed:** `c0846d1aa79b9672bcd44ca67a03408c3b956ed0`
- **CI at that SHA:** all 7 checks pass. Note `build` is **not** a required check — `repos/freenet/river/branches/main/protection` returns 404, so nothing mechanically blocks a merge on red. Pre-existing, not introduced here.
- **Reviewers run:** four independent Claude lenses (code-first, testing, skeptical, failure-mode-specific), each blind to the others, plus an external non-Claude pass (`codex review --base origin/main`, completed normally, no substitution). Findings deduplicated, contradictions reconciled, every cited `file:line` verified against the canonical head.

**The approach is right and the core mechanism works.** The byte comparison's provenance is sound — `$ARCHIVE` is the same file signed at step 3, handed to `fdev` at step 4, and compared at `:398` against what `inspect --archive-out` pulled from the state fetched back; nothing rebuilds or re-derives either side in between. `shellcheck -x` is clean on all four scripts. Both escape hatches fail closed (only the exact string `1` enables them; `allow_unverified` does not leak into `absent`). The decision logic in `web-container-publish-lib.sh` is genuinely pinned: **9 of 9 logic mutations were killed.**

The problems are in the wiring around that core, and they are consequential enough to block.

---

### Must Fix

**1. The script can inspect a different contract from the one it publishes.** *(external pass P1; independently found by code-first)*

`publish-web-container.sh:193` reads `CONTRACT_ID` from the committed `published-contract/contract-id.txt`, but `fdev network publish` targets the id derived from `$WASM_FILE` and `$PARAMS_FILE`. If `published-contract/` is internally inconsistent — stale wasm, stale params, or a stale id file — then the pre-flight GET, the version floor and the post-publish byte comparison all measure one contract while the publish writes another. The result is a silent publish to the wrong address, reported green.

This is worse than the fork the PR exists to prevent, because every guard it adds is pointed at the wrong target. Verified as **latent, not live**: recorded and derived ids match today. `fdev get-contract-id` is already used for exactly this at `Makefile.toml:335-337`, and `.claude/rules/river-publish.md:332` states the committed-parameters identity requirement explicitly.

**2. Single-shot network reads on an eventually-consistent network — one bug in two places.** *(four-way independent agreement: external pass P2, skeptical, code-first, and the coordinator's own trace; two lenses reproduced it experimentally)*

`publish-web-container.sh:413-418` — `if [ "$OUTCOME" != "unknown" ]; then break; fi`. The justification comment is true for `collision` and `superseded` and false for `not-landed`, which is exactly the answer you get when you ask before the state has propagated. Reproduced against the canonical head: a publish that landed at 30000386 reported `NOT PUBLISHED — the network is still at 30000384`, exit 5, with attempts 2 and 3 never used. `RIVER_WC_READBACK_ATTEMPTS` and the 15-second delay exist to absorb precisely this and are bypassed. The first read-back also fires with no delay at all, since `READBACK_DELAY` applies only *between* attempts.

The operator is then told to re-run. Re-running is safe under the forward-only counter, but it strands the state that did land and costs a full rebuild and republish — and it makes the script's headline verdict wrong in the most common transient.

The same shape appears at `:243` / `:209-241`: the pre-flight `read_network_version` is a single GET with no retry, so a transient failure classifies as `unknown` and refuses a legitimate publish.

The rehearsal cannot catch either: its stub applies state synchronously before `fdev` exits, so propagation lag is unrepresentable.

**3. The `flock` does not deliver whole-sequence locking — it is wrong in two independent ways.**

- **Wrong scope** *(failure-mode lens)*: `:135` sets `LOCK_FILE="published-contract/.publish.lock"`, resolved relative to `REPO_ROOT` after the `cd` at `:104-105`. River development is worktree-based by standing convention and nova routinely carries several River worktrees; two of them publishing take *different* lock files and race exactly as before. The header at `:144-155` claims "one operator on one machine"; the code delivers "one operator in one checkout".
- **Wrong position** *(external pass P2 — no Claude lens found this)*: `Makefile.toml:364` declares `dependencies = ["compress-webapp", "build-web-container-tool", "build-web-container"]`, which run to completion *before* the script is invoked and therefore before it takes the lock. Two overlapping `cargo make publish-river` runs both write the shared `target/webapp/webapp.tar.xz` outside any lock, so the second can replace the archive while the lock holder is between signing and comparing.

Traced consequence of the second: it does not fork — A signs archive X, B overwrites with Y, A publishes Y carrying metadata for X, `validate_state` rejects, read-back says not-landed — so it degrades to a confusing failure rather than a silent split. But it directly falsifies the "one lock across the whole sign-to-verify sequence" claim, and `.claude/rules/river-publish.md:87-91` states that invariant explicitly. **Fixing only the lock file path leaves the claim false.**

**4. Every publish-from-main and CI gate was dropped in the port.** *(failure-mode lens)*

The PR says the discipline was "ported, not invented" from `scripts/publish-pointer-records.sh`, whose pre-flight (`:105-125`) refuses on a dirty tracked tree, a branch that is not `main`, `HEAD != origin/main`, and absent-or-non-green CI on the exact HEAD SHA — including an explicit "an empty list of runs is not green" guard. Grepping both new shell files for `rev-parse` / `abbrev-ref` / `symbolic-ref` / `git status` / `gh run` returns **nothing**.

The web container is the higher-consequence of the two artifacts, `~/.claude/rules/publish-from-main.md` names this contract key explicitly, and river's `main` has no branch protection — so nothing else enforces it. This rewrite also newly exposes a standalone directly-runnable entry point where `cargo make` was previously the only route, which makes branch-publishing marginally easier rather than harder.

---

### Should Fix

- **`:312` writes the counter before signing (`:320`) and before the wrong-key parameters check that can `die` (`:331`)**, so a wrong-key abort burns a version and publishes nothing.
- **`:281` — `--sign-only` suppresses the unverified-network warning** and is the one path with no read-back, yet it burns the counter and emits a signed artifact directly consumable by a manual `fdev network publish`. The PR's justification for permitting the override is "the post-publish read-back still runs and still decides the exit code" — which is exactly the property this path lacks. The warning matters *most* here, not least.
- **No u32 ceiling on the counter.** Overflow leaves the forward-only counter poisoned at `4294967296`, a value that can never be signed; reproduced, `rc=2`.
- **`READBACK_ATTEMPTS=0` or a non-integer yields a false UNKNOWN** over a publish that landed; reproduced, `rc=6` while the network held the new version. Same shape as Must Fix 2 — a config value silently defeating the verification the script exists to perform.
- **`:162-163` — the flock re-exec uses `$0` after the script has `cd`'d**, so `cd scripts && ./publish-web-container.sh` dies with `flock: failed to execute …` at rc=69, outside the documented exit codes. Fail-closed, but opaque.
- **`:156` — `RIVER_WC_PUBLISH_LOCKED` preset in the inherited environment silently disables the single-writer lock**, with no warning and no documentation — not even the WARNING the missing-`flock` path prints.
- **`publish-all` never commits the version counter** (`Makefile.toml:638-639` commits only `cli/Cargo.toml`), leaving it dirty. That matters more now the forward-only counter is a defence: a stray `git checkout` of that file re-opens the reissue path this PR closes.
- **The rehearsal exits 0 when `python3` is missing** (`web-container-publish-rehearsal.sh:26`). The only end-to-end coverage of a release path is a CI step that reports success by skipping — a green signal no input can turn red.
- **The claim "a publish at an already-used version is a no-op SUCCESS" is over-general and appears four times**, plus `AGENTS.md:190-192`. Verified false: `put.rs:229-245` maps `PutResponse { new_value: Err(..) }` to `Err(OpError::..)`, so the rejection *is* surfaced to the originating node — which is what this PR's own incident narrative describes. It does not weaken the design, but it misleads future readers.

---

### Testing assessment

**Genuinely good where it is good, and honest about its gaps** — the PR's own "What I did NOT test" and "Residual risk" sections pre-disclose the stub tautology, the untested flock, and the untested first-publish path. Claimed counts are accurate and were independently reproduced twice: 7 Rust tests, 27 lib checks, 32 rehearsal checks across 10 scenarios, `shellcheck -x` clean.

**Decision logic: 9 of 9 mutations killed.** **Wiring: 5 of 7 mutations survive**, all green on both suites:

| Mutation | Line |
|---|---|
| `if command -v flock` → `if false` | `:157` |
| `"Contract not found"` → `"Contract NOT found"` | `:236` |
| key check → `cmp -s /dev/null /dev/null` | `:331` |
| read-back retry guard → `if true` | `:416` |
| counter rollback reintroduced in the `unknown` verdict arm | `:480` |

The last is the one to weight highest: it reintroduces precisely the behaviour this PR exists to remove, on the one path whose own text says the counter is "NOT rolled back, precisely because we cannot rule out that it landed". The `not-landed` equivalent *is* caught, by scenario 6.

**One vacuous assertion, and it is load-bearing.** `rehearsal.sh:272` checks `*"Contract not found"*` against `$RUN_OUT` to prove the script classified a NotFound as `absent` — but the `unknown` arm at `:248` echoes `tail -3 "$WORK/onnet.log"`, which carries the stub's own `Error: Contract not found: ...`. The assertion passes on stub output leakage regardless of classification, and its two sibling checks hold identically for both states. Scenario 5 cannot distinguish the two states it is named after. Assert instead on `the node answered: Contract not found` (`:247`), which only prints under `absent`.

Also: both suites are `set -uo pipefail` with no `set -e`, so a failed setup step inside a scenario silently mis-runs it rather than aborting.

---

### Review process disclosure

Two findings originally reported were **withdrawn**, and the reason matters for reading this review.

All four lenses were initially given one shared worktree, and one lens ran 15 mutate-run-revert cycles in it between roughly 08:24 and 08:33. Both suites read the *live* worktree files (`lib-test.sh:14` sources the library; `rehearsal.sh:61` copies the script into its sandbox), so other lenses briefly observed mutated code. Two of them independently reported the rehearsal suite as flaky, and one chased a pre-flight anomaly that was another lens's mutant.

On discovery, canonical copies were fetched from the GitHub API at `c0846d1a`, the mutating lens was moved to a dedicated worktree, and **all four lenses re-verified every citation against canon**, confirmed by `git hash-object` matching the HEAD blobs. Both flakiness findings were withdrawn; **no other finding changed or was dropped**, and all reproduced experiments were re-run against rigs rebuilt from canon.

Net: the suites are *not* flaky — across 50+ clean runs they proved correctly sensitive to mutations in the code they cover, which is incidental independent confirmation of the PR's mutation-testing claim. The setup error was the reviewer's, not the PR's.

One caveat carried through honestly: the external pass ran in a sandbox without network, so its own `cargo test` attempt failed on DNS and its findings are source-reading rather than test-derived. Its three cited line numbers were each verified independently.

---

### Out of scope, but should not go unaddressed

**`~/.claude/skills/river-publish/SKILL.md` is now stale and lives outside this repo, so this PR cannot fix it.** Lines 149-150 still describe `cargo make sign-webapp` as a transitive dependency of `publish-river`, which this PR makes false. More seriously, lines 157-177 give a manual fallback that hand-increments the counter and calls `fdev publish` directly — bypassing the entire new guard. An operator following that skill under pressure reintroduces the 2026-08-04 failure exactly. This needs fixing alongside the merge, not after it.

---

### Verdict

**State: Needs Changes — Re-review Required After Fix**

Four Must Fix findings. The strongest signal in this review is the **four-way independent agreement** on the read-back retry gap — the external model, two Claude lenses that each reproduced it experimentally, and a fourth independent trace, all arrived at it separately. The strongest single contribution from the external pass is the lock-position finding, which no Claude lens reached and which shows Must Fix 3 cannot be closed by relocating the lock file alone.

None of this says the approach is wrong. The design is sound, the decision logic is well pinned, and the incident analysis behind it is accurate. What is missing is that the guards around the core do not yet hold, and several of them cannot currently fail in tests.

Per `~/.claude/rules/multi-model-review.md`, blocking findings on a release surface require a fresh review on the updated HEAD; this review does not authorize merging a later SHA.

[AI-assisted - Claude]

sanity (commented):
## Re-review at `5d563436` — Full tier

This is a **re-review**, not a fresh one. The prior review ran against `c0846d1a` and found four Must Fix items; it does not authorize this SHA. The PR has grown from +776/−3 to **+2620/−130 across 12 files**, so roughly 1850 lines are new since that review — and per the "the fix round is where the next round's bugs live" heuristic, that new code got the same scrutiny as the original.

Lenses: **code-first**, **testing**, **skeptical / bash-adversarial**, **deploy-failure-mode** (four independent Claude reviewers, each blind to the others), plus an external non-Claude pass. Read-only lenses shared one checkout; every lens that mutated files to test a pin got its own worktree. Every `file:line` below was re-verified against the canonical head.

CI at `5d563436`: **all green** — `build`, `ui-playwright-tests`, `check-delegate-migration`, `check-pointer-freshness`, `check-room-contract-migration`, `license/cla`.

---

### Verdict

**Changes requested.** The rewrite is substantially better than what it replaces, and all four prior Must Fix findings are genuinely addressed in substance. But the fix round introduced one defect that is worse than any of the four it closed, and it is the same defect class the PR exists to eliminate: **the new provenance gate refuses the recovery the script itself tells the operator to perform, and the cheapest way past that refusal is the counter rollback that caused the 2026-08-04 fork.** Three of the four Claude lenses reached this independently, two demonstrated it end to end, and the external pass reached it as one of its two P1s.

Four Must Fix below. Nothing here argues against the design — the design is sound and the reasoning in the comments is unusually careful. These are all local fixes.

The external pass earned its place again: as on the prior head, it found a defect no Claude lens reached (**MF-4** — the provenance gate's "an untracked file cannot change what is published" assumption is false), and independently reached MF-1 as its other P1.

---

### Prior Must Fix findings — status

| # | Prior finding | Status |
|---|---|---|
| 1 | `flock` worktree-scoped, and positioned after the shared artifact build | **Resolved in substance, guard is weak.** Lock is now machine-scoped (`${XDG_RUNTIME_DIR:-/tmp}/river-web-container-<id>.lock`), held as fd 9, and the build moved inside it via the `--build` re-entry. Two residual gaps (SF-2, SF-4) and the pin that protects the structural half is bypassable — **MF-3**. |
| 2 | Single-shot network reads | **Resolved.** `wc_readback_is_final` terminates only on `landed`/`collision`/`superseded`; pre-flight retries `unknown`. Two asymmetries remain that contradict the header's own claim (SF-9, SF-10). |
| 3 | Missing publish-from-main / CI gates | **Resolved.** Transcription from `scripts/publish-pointer-records.sh:85-142` is faithful and slightly stronger (adds `|| die` on the fetch, adds a `jq` presence check). The `TOTAL > 0` "empty list is not green" guard is present at `:390-394` and pinned by scenario 28. But the gate is the direct cause of **MF-1**. |
| 4 | Contract-id drift | **Resolved.** `fdev get-contract-id` compared against `contract-id.txt` at `:248`, before both the counter write and the lock. Position is correct. |

---

## Must Fix

### MF-1. The dirty-tree gate blocks the recovery the script recommends, and the easiest way past it is the 2026-08-04 rollback

`scripts/publish-web-container.sh:616` writes the burned counter into `published-contract/contract-version.txt` — a **tracked** file (`git ls-files published-contract/` confirms). `scripts/publish-web-container.sh:351-354` refuses to run when any tracked file is modified, with no pathspec exclusion.

So after **every** outcome that reached step [3] — `not-landed`, `unknown`, `collision`, `superseded`, `--sign-only` — the tree is dirty and the next invocation dies with *"tracked files are modified. Publish only from a clean checkout of main."*

Meanwhile the verdict text at `:790` (`not-landed`) says *"Re-run the publish; it will sign above whatever the network reports"*, `:800` (`unknown`) says *"Re-run the publish once the node answers"*, and `:770` (`collision`) says *"re-run the publish"*. Every one of those advertises a re-run that is guaranteed to be refused. Demonstrated end to end by two lenses:

```
[0] provenance: clean tree, on main, at origin/main, CI green
     M published-contract/contract-version.txt

ERROR: tracked files are modified. Publish only from a clean checkout of main.
Set RIVER_WC_ALLOW_UNPROVEN=1 for this run if you accept that.
```

The operator now has exactly two doors, and the more natural one is the incident:

- **`git checkout -- published-contract/contract-version.txt`** — the reflex when a script says "clean checkout" and names the one dirty file. This is the counter rollback. If the node is still unreadable (the reason they are retrying), the pre-flight reports `unknown`, refuses, and offers `RIVER_WC_ALLOW_UNVERIFIED=1` in its own text at `:544-546`. With both set, against the real lib:
  ```
  wc_next_version 30000385 unknown -> 30000386   # REISSUES the burned version
  wc_next_version 30000386 unknown -> 30000387   # what would have happened without the revert
  ```
  `--build` has meanwhile rebuilt a different archive (the build is not reproducible). Two validly signed archives at one version. **Exact 2026-08-04 replay, reached by following the script's own instructions.**
- **`RIVER_WC_ALLOW_UNPROVEN=1`** — safe on the version axis, but one switch silently drops clean-tree, on-main, HEAD==origin/main *and* CI-green.

The PR already knows this hazard — `Makefile.toml:668-673` names it precisely (*"a stray `git checkout`, `git stash` or discarded working-tree change restores a version that has already been published, and reissuing a used version is the 2026-08-04 fork"*) — and then fixes it only for `[tasks.publish-all]` via the `git add` at `:674`, leaving the plain `publish-river` path, which is the documented normal path in `.claude/rules/river-publish.md:94`.

**Why the rehearsal cannot see it:** `run()` at `scripts/tests/web-container-publish-rehearsal.sh:397-407` calls `git_sync` before every scenario, committing the counter between runs. The harness supplies the missing operator step, so the trap is structurally invisible to the suite. Scenario 25 (`:813-822`) tests the refusal and treats it as correct without asking what the operator does next.

**Fixes, strongest first:**
1. Keep the burned-version high-water mark **outside the git tree** — beside the lock, at `${XDG_RUNTIME_DIR}/river-web-container-<id>.highwater` — and make the floor `max(counter, highwater, network) + 1`. No `git checkout`, `git stash`, branch switch or fresh clone can then reissue anything, which is the invariant the forward-only counter is actually trying to hold.
2. Make `[0]` tolerate the single case where the only modified tracked path is `contract-version.txt` **and** its working-tree value is strictly greater than `HEAD:`'s — that is provably a prior burn, not a provenance divergence.
3. At minimum: every non-`landed` verdict must say *"commit `published-contract/contract-version.txt` before re-running — do NOT `git checkout` it"*, and the `:351` refusal must special-case that filename with the same warning.

Two related text bugs on the same path, both worth fixing with it:
- `:758` (the `landed` arm) says `git add $VERSION_FILE`. After `git add`, `git status --porcelain --untracked-files=no` prints `M  ...` — still non-empty, so `:351` still refuses. It needs to say `git commit`.
- `:636` (`--sign-only`) says *"Commit $VERSION_FILE only after a successful publish."* That is the pre-PR doctrine surviving into the post-PR script: it implies "discard on failure", which is the rollback. An unpublished burn is harmless; a reissue is not. And `--sign-only` is the one path with no read-back at all.

---

### MF-2. `AGENTS.md:192` reinstates the exact false claim the rest of this PR retracts

```
- **`fdev`'s exit code decides nothing.** A publish at an already-used
  version is a no-op *success*, and a publish that reports a timeout may
  have landed anyway.
```

This sits in the web-container section of `AGENTS.md` (the surrounding lines at `:185-187` are about the byte-for-byte read-back). It is contradicted, in the same diff, by:
- `scripts/publish-web-container.sh:66-75` — *"That is WRONG for this contract and should not come back"*,
- `scripts/web-container-publish-lib.sh:119-124` — *"Do NOT restate the old ... here"*,
- `.claude/rules/river-publish.md:139` — an explicit retraction.

`update_state` returns `InvalidUpdateWithInfo` for `version <= current`, which is how the 2026-08-04 operator saw the rejection at all. `AGENTS.md` is the file every agent loads every session, so this is precisely how the retired belief gets resurrected — and an operator holding it reads a non-zero publish exit as "harmless, retry", which is the incident's reasoning. **The PR description repeats the same claim in Approach §3.** Delete both.

*(One lens also flagged `.claude/rules/river-publish.md:130` as self-contradictory. **Withdrawn on verification** — that paragraph is about the sibling POINTER contract, where the no-op-success property is real and deliberate, and `:139` says so explicitly. No change needed there.)*

---

### MF-3. Scenario 24 — the only guard on prior finding #1's structural half — does not fail on the most natural reintroduction of the bug

`scripts/tests/web-container-publish-rehearsal.sh:794-797`:
```sh
case "$blk" in
    *"--build"*) check "[tasks.$task] passes --build" yes yes ;;
    *)           check "[tasks.$task] passes --build" yes no  ;;
esac
```

Credit where due: this pin is **not** one of the un-failable source scrapes. Re-adding `dependencies = ["compress-webapp"]` and dropping `--build` turns it red (2 failures), and dropping `--build` alone turns it red (1 failure). Today the needle appears exactly once per scraped block.

But the `case` matches the **whole block including comments**, and the `dependencies` check greps only `^dependencies`. So the pin asserts *"no `dependencies` key, and the token `--build` appears somewhere"* — not the property it is named for, which is *"nothing builds the shared archive before the script takes the lock"*. I reproduced the bypass myself, in a dedicated worktree, on the real `Makefile.toml`:

```toml
description = "Publish River to Freenet, verifying the result against the network"
# No `dependencies` — the script used to be passed --build; we build here now.
script = '''
cargo make web-container-build-inputs
exec ./scripts/publish-web-container.sh
'''
```

The shared archive is built outside the lock again — the exact prior Must Fix — and the suite reports **`110 passed, 0 failed`, rc 0**.

Compounding it: the post-build tracked-file re-check at `scripts/publish-web-container.sh:438-442` — the guard on the seam between the provenance gate and the build-inside-lock, i.e. new code guarding two new features against each other — has **zero coverage**. Replacing its condition with `if false` also leaves the suite at 110/0, because the stub `cargo` (`rehearsal.sh:272-279`) only appends to a git-excluded log and can never dirty the tree.

So the structural half of prior finding #1 is currently unguarded in both directions.

**Fix:** scrape the `script = '''...'''` body only, excluding comments, and additionally assert that the body's non-comment lines contain no `cargo make` invocation other than the `exec ./scripts/publish-web-container.sh` line. Add a `cargo_dirties_tree` knob to the stub so `:438-442` can be turned red.

---

### MF-4. "An untracked file cannot change what is published" is false — untracked files do reach the archive

`scripts/publish-web-container.sh:348-350` states the assumption, and `:356-359` acts on it, printing a reassuring note and proceeding:

```sh
# TRACKED modifications only. An untracked file cannot change what is
# published; a modified tracked one means this checkout differs from the
# commit whose CI is verified below, which is the whole provenance claim.
...
UNTRACKED="$(git status --porcelain | grep '^??' || true)"
if [ -n "$UNTRACKED" ]; then
    say "note: untracked files present (they cannot change what is published):"
```

The assumption does not hold, because the build reaches files by **glob**, not by git. Verified chain, every link read at this head:

`[tasks.publish-river]` → `publish-web-container.sh --build` → `:433` → `[tasks.web-container-build-inputs]` (`Makefile.toml:235`, deps at `:248`) → `[tasks.compress-webapp]` (`:193`, dep `build-ui-with-cleanup`) → `[tasks.build-ui]` (`:147`, `dependencies = [..., "build-tailwind", ...]`) → `[tasks.build-tailwind]` (`:20`, `npm run build:css`) → `tailwindcss -i assets/tailwind.css`, whose line 6 is:

```css
@source "../src/**/*.rs";
```

Tailwind v4 scans that glob for class candidates. An **untracked** `.rs` file under `ui/src` — a scratch component, a half-finished module, a file another agent left behind — contributes class names to the generated `ui/assets/styles.css`, which `dx build` bundles and `compress-webapp` tars into `webapp.tar.xz`. The published archive then differs from both HEAD and the artifact CI tested, while the gate has just told the operator in writing that it cannot.

`dx build`'s asset bundling is a second vector of the same shape: it walks the asset directory (`WalkDir::new(&asset_dir)` in dioxus-cli 0.7.9 `build/request.rs:1456`), so an untracked file dropped in `ui/assets/` is bundled too.

This matters more than a normal provenance nit because it is exactly the claim the whole gate rests on: `[0]` exists to establish that "what is in this checkout is the commit whose CI is green", and the untracked carve-out is the one hole in it — asserted safe rather than checked. It also interacts with MF-1: the recovery path most likely to leave untracked files around is the one an operator is on after a failed publish.

**Fix:** either refuse untracked files under the build-input paths (`ui/src`, `ui/assets`, and anything else reached by glob), or narrow the carve-out to a known-inert allowlist and change the comment and the `say` text to state what is actually guaranteed. Asserting "they cannot change what is published" in a script whose job is provenance should require having checked.

*(Found by the external pass; verified independently against the head before inclusion.)*

---

## Should Fix

**SF-1. No `INT`/`TERM` trap over a multi-minute critical section.** `:409` is the only trap in the file and it is `EXIT`-only (`rm -rf "$WORK"`). Between `:616` (counter burned) and `:805` (verdict) sit a publish with a 300 s default timeout and a read-back loop of up to 3 × (180 s GET + 15 s). Ctrl-C — the exact operator reflex during a hang, and what 2026-08-04 was — leaves: counter burned, archive signed, PUT possibly in flight, tree dirty (MF-1), `pub.log` and `back.bin` deleted, and **no verdict block printed at all**. Add a trap on `INT TERM HUP` that prints the state: *"version $VERSION is BURNED; a publish may be in flight; commit $VERSION_FILE, do not revert it."*

**SF-2. fd 9 leaks into every child; a daemonizing grandchild holds the lock after the script exits.** `:306` `exec 9>>"$LOCK_FILE"`. Bash does not set `FD_CLOEXEC` on `exec`-opened fds, and `flock` releases only when *all* fds on the open file description close. Demonstrated:
```
$ bash -c 'exec 9>>/tmp/l; flock --nonblock 9 || exit 1; setsid bash -c "sleep 30" & disown'
$ flock --nonblock /tmp/l -c 'echo ok' || echo "LOCK STILL HELD BY ORPHAN"
LOCK STILL HELD BY ORPHAN
```
`:433` runs `cargo make web-container-build-inputs` while holding fd 9; its chain reaches `cargo build` → rustc → **sccache** (pre-configured on the build machines per `AGENTS.md`, auto-starts, persists), plus `dx build` and node. Any one that daemonizes wedges the lock permanently, and every later publish then dies at `:310` with *"another web-container publish is already running"* when there is none, with no PID and no diagnostic — pushing the operator to `rm` the lock or bypass it. Fix: `cargo make ... 9>&-` at `:433` (and on the `fdev` calls), and write the holder's PID into the lock file.

**SF-3. The u32 ceiling guard covers the counter but not the network floor, and the failure wedges the counter permanently.** `:222-231` tests `$COUNTER` only. `VERSION` is computed at `:562` as `max(counter, net) + 1` and written unguarded at `:616`. Verified directly against the real lib:
```
wc_next_version 30000385 known 4294967295  ->  4294967296
```
One lens reproduced the full wedge: the counter file is written to `4294967296`, *then* `sign --version` fails on clap's u32 parse with rc 2, and every later run dies at `:225` with *"Nothing has been written"* when something was — reproducing exactly the state `:216-221` claims the guard closes. Rehearsal scenarios 18/19 pin only the counter side of the boundary. Reachability is remote (~30M today, and `read_network_version` verifies under our own parameters so no third party can inflate it), but the fix is one line: re-apply the ceiling test to `$VERSION` after `:562`.

**SF-4. The counter is read outside the lock.** `:211` reads `$VERSION_FILE`; the lock is taken at `:302-321`. The comment at `:274-276` says the lock exists because *"Two concurrent runs would race the read-modify-write on the counter"* — but only the write half is inside. The network floor usually rescues this, so a fork needs a second fault, but re-reading the counter after `flock` succeeds costs one line.

**SF-5. The lock silently degrades to *no lock* when the lock file cannot be opened.** `:305` `if : >> "$LOCK_FILE" 2>/dev/null; then ... else` warns at `:315-317` and continues unlocked. Reachable via a stale inherited `XDG_RUNTIME_DIR` (routine under reattached `tmux`/`screen`, `sudo -i`, systemd units, cron), via one prior run under `sudo` leaving a root-owned lock file, or — with `XDG_RUNTIME_DIR` unset — via another user squatting a fixed name in world-writable `/tmp`. The comment at `:297-301` argues carefully for warn-don't-refuse in the missing-`flock`-*binary* case; the open-failure fallback is undocumented, more likely, and contradicts the file's own *"a knob that switches a check off is a hole in the check"* (`:188`). *(Verified safe by contrast: if `exec 9>>` itself fails, bash does exit.)*

**SF-6. `"$TOOL" sign` at `:620-625` is the one fallible step after the counter is burned with no error handling.** Every other risky call is wrapped (`:247`, `:598`, `:643`). A signing failure aborts under `set -e` with a raw Rust error, no verdict block, and the counter already advanced — the same terminal state as SF-1, reachable without any signal.

**SF-7. The bytes signed, published and compared are three independent reads of a live mutable path.** `ARCHIVE` (`target/webapp/webapp.tar.xz`) is read at `:621` (sign), `:649` (publish) and `:707` (`cmp`). The lock excludes another *publish*, but a bare `cargo make compress-webapp` in another shell writes that path with a non-atomic `tar -cJf` (`Makefile.toml:197-199`). A rewrite between `:649` and `:707` produces a **false " FORKED"** verdict — the loudest, most action-inducing message in the file. Snapshot the archive into `$WORK` once inside the lock and use the copy at all three sites.

**SF-8. `RIVER_WC_ALLOW_FIRST_PUBLISH=1` is an unconditional fork button, named in the refusal text, with no script-level test.** `web-container-publish-lib.sh:87-91` returns `proceed` for `absent` with no counter check, and `wc_next_version <counter> absent` has no floor. Given MF-1 makes "counter bumped but never committed" the *normal* state after a failed publish, the repo can legitimately be behind the network; one dead-ended GET then answers `Contract not found`, and the script hands the operator this flag at `:531-541`. The script already holds both facts that make a first publish impossible — `COUNTER` is non-zero and `contract-id.txt` is the production key. Make the flag refuse outright when `COUNTER > 0`. Also: `grep RIVER_WC_ALLOW_FIRST_PUBLISH scripts/tests/` returns nothing.

**SF-9. `absent` is never retried, which reaches for SF-8's button sooner than necessary.** `:499` `[ "$NET_STATUS" = "unknown" ] || break`. The script's entire case for distrusting `absent` (`lib:80-86`) is that a Freenet GET dead-ends transiently — which is exactly what retrying resolves. Retry `absent` on the same schedule as `unknown`.

**SF-10. "Every network read here is RETRIED, in both directions" (`:77-80`) is not true of the pre-flight.** `read_network_version` (`:470-475`) `die`s outright when `inspect` fails, so one response failing signature verification aborts the release with no retry, while a *timeout* on the same read gets `PREFLIGHT_ATTEMPTS` tries. The read-back path handles the same condition non-fatally at `:713-715`. The asymmetry may be deliberate; the header should not claim otherwise.

**SF-11. `contains "reports PUBLISHED" "PUBLISHED"` cannot distinguish PUBLISHED from NOT PUBLISHED.** `rehearsal.sh:432` (scenario 1) and `:589` (scenario 12). The banner at `publish-web-container.sh:782` is `" NOT PUBLISHED — ..."`, which contains the needle. Proven: a mutation making `not-landed` final produced the NOT PUBLISHED banner in scenario 12 and only the `exits 0` check failed. Scenario 12's whole purpose is that a stale first read must not be reported as NOT PUBLISHED, so its headline assertion is the one that cannot see it. Assert `"PUBLISHED — despite"`, or assert the output does *not* contain `"NOT PUBLISHED"`.

**SF-12. No scenario pins the documented exit code.** Everything goes through `not_zero` (`rehearsal.sh:348`), while the script defines a 0/3/4/5/6 vocabulary (`lib.sh:188-197`) and scenario 32's own comment cites "an exit code outside the documented set" as a bug it fixed. SF-3's real defect exits **rc 2** and would be reported as a pass. Replace `not_zero` with exact codes in scenarios 6/10/11/13.

**SF-13. `$WORK/back.bin` is not cleared between read-back attempts.** `:685` removes `back.tar.xz` but not `back.bin`, then guards on `[ -s "$WORK/back.bin" ]` at `:694`. The pre-flight deliberately does `rm -f "$out"` at `:461`, so staleness was on the author's mind; this is the miss. Bounded impact, gratuitous asymmetry.

**SF-14. The `EXIT` trap destroys the forensic evidence on exactly the outcomes that need it.** `:409` `rm -rf "$WORK"` runs unconditionally. On `collision`, `$WORK/back.tar.xz` is the only local copy of the foreign archive that forked the site, and it becomes unfetchable as soon as a higher version supersedes it. Preserve `$WORK` and print its path on any non-`landed` verdict.

**SF-15. The PR description is materially out of date and omits the largest new surface.** The provenance gate — which changes the preconditions of `cargo make publish-river` and `sign-webapp`, and adds `RIVER_WC_ALLOW_UNPROVEN` — appears nowhere in Approach, Testing or Residual Risk, though nine rehearsal scenarios cover it. Also stale: *"I did not rehearse two concurrent runs contending"* (scenario 22 at `:720-758` does exactly that), "10 scenarios, 32 checks" (32 scenarios, 110 checks), "27 checks" for the lib test (33), and the claim that the stub `fdev` models a stale publish as a no-op success (its own header at `:127-133` says the opposite). `--build`, the u32 guard, `require_count` and the `published-contract/` self-consistency check are also unmentioned.

---

## Consider

- **Two publishers on two machines is an unnamed residual fork path.** The lock is machine-local; `:277-301` reasons about worktrees and UNIX users and concludes *"one operator per machine"*, but the two-machine case appears nowhere in the script, `river-publish.md`, `AGENTS.md` or `DEVNOTES.md`. Two operators on green main both read `known N`, both sign `N+1` over independently-built archives, both publish inside the 300 s window; a peer new to the contract takes the initial-state bypass and accepts the loser. No override needed, every guard working as designed. It cannot be closed by a local lock — but it should be stated, with an out-of-band rule.
- `gh run list` at `:381` has an unpinned `--limit` (default 20) and sees Actions runs only — non-Actions commit statuses and external check-runs are invisible. Both inherited from `publish-pointer-records.sh`, both latent today (river's checks are all Actions). Pin `--limit 100`.
- `RIVER_WC_ALLOW_UNPROVEN` is all-or-nothing, and the common timing (merge, pull, publish while push-CI is still running) pushes operators toward it habitually — after which MF-1's refusal gets the same reflex. Consider per-check names.
- `:248` merges stderr into the derived contract id (`2>&1` then exact string compare), so any future `fdev` stderr line produces a false *"published-contract/ disagrees with itself"*. Latent, not live. The sibling parse at `:476` is line-anchored and immune.
- Bash arithmetic reads leading zeros as octal while `[ ]` reads base 10, and `require_count`'s digit check accepts them: `wc_next_version 030000377 absent` → `6291712`. The script never writes a leading zero, but `$(( 10#$var ))` closes it.
- `require_count` has no upper bound: `RIVER_WC_PREFLIGHT_ATTEMPTS=4000000000` passes and yields an unbounded loop with no overall deadline.
- `cargo make publish-river-debug` is broken — `BUILD_PROFILE = "dev"` but `cargo build --profile dev` emits to `debug/` while `:163` looks in `dev/`. Pre-existing; it now fails loudly at `:445` rather than after burning a version, which is an improvement, but `:433` forwards the wrong value deliberately.
- Stub `gh` emits `"conclusion":null` for pending CI (`rehearsal.sh:877-878`); real `gh` emits `""` (verified against the live repo). The verdict is identical either way, so the gate is right — but scenario 30 is not pinning the production shape.
- No total-check pin: `rehearsal.sh:917` prints `$PASS passed` and nothing asserts the expected count, so a scenario that silently stops running shrinks the count and the suite still exits 0. One line: `check "check count" 110 "$((PASS + FAIL))"`.
- The flock scenario skips silently and contributes zero checks (`rehearsal.sh:756-758`), while `missing_tool` (`:43-53`) exists precisely because a skip is "a green signal that no input can turn red". Add `flock` to the `missing_tool` list.
- Rehearsal litter: each run leaves a `$XDG_RUNTIME_DIR/river-web-container-<id>.lock` that the trap (`:85`, `$SANDBOX` only) never removes. Ephemeral on CI, unbounded on a dev box.
- `web-container-tool inspect --expect-version` (`main.rs:89-92`, `:436-443`) is implemented, documented and unit-tested but has no caller on this path — the script does the comparison in shell at `:720`. Wire it in as a second belt, or note it is deliberately caller-less.
- `AGENTS.md:209-211` points at a "manual fallback when `publish-river` fails" section in `.claude/rules/river-publish.md` that does not exist. Pre-existing, but more consequential now the old manual sequence has been removed.
- The Usage block (`:91-131`) documents none of the 0/3/4/5/6 exit codes that `wc_outcome_exit_code` defines and the rehearsal refers to as "the documented set".

---

## On splitting

**Recommendation: keep it as one change.** Rough breakdown:

| | lines |
|---|---|
| Publish path proper (`publish-web-container.sh` 805, lib 197, `Makefile.toml` ~212, `main.rs` 351) | ~1565 |
| Test harness + scenarios (`rehearsal.sh` 918, `lib-test.sh` 155) | ~1073 |
| Docs (`river-publish.md`, `AGENTS.md`, `DEVNOTES.md`) | ~67 |
| CI + misc (`build.yml`, `.gitignore`, `pointer-records.toml`) | ~45 |

The lock, the counter policy, the readback and the provenance gate are one interlocking mechanism — MF-1 exists precisely *because* the gate and the counter's on-disk location were designed against each other, and splitting them ships a half-guarded publish path to a key with no rollback and no second address. Deferring the provenance gate would also mean merging with a known blocking finding open, which is worse than the size. The tests are not separable either: they are the only thing standing between this script and the next fork.

The one thing that would genuinely reduce risk is not a split but a **rebase-and-squash of the fix round into a readable history** — the review burden here is that ~1850 lines arrived as fixes to fixes.

---

## What was verified and found sound

Recording these so they are not re-litigated:

- `shellcheck -S style` and `bash -n` are clean on both scripts. No unquoted expansion, no `[ ]`/`[[ ]]` bug, no space-in-path hazard. `${KEY_ARGS[@]+"${KEY_ARGS[@]}"}` is the correct `set -u`-safe idiom.
- **Bash-arithmetic code execution is not reachable** — every value entering `$(( ))` passes a digits-only `case` first.
- **The escape hatches are strict-equality `= "1"`** — `0`, `false`, `true`, `yes` do not enable any of them. Not trippable by accident.
- **The "empty list of runs is not green" guard is present and does guard** (`:390-394`), pinned by scenario 28; `jq`/`gh` absence is caught explicitly at `:378-379`.
- `set +e`/`set -e` pairs are balanced (16 occurrences); no `local x=$(cmd)` exit-status masking.
- **Both suites are wired into CI** — `.github/workflows/build.yml:290` and `:303`, in the `build` job, on `pull_request` and pushes to main — and exit codes propagate (each script ends `[ "$FAIL" -eq 0 ]` under `bash -e`).
- The harness itself reviewed as production code and found sound: `check`/`setup_failed`/`contains`/`not_zero` all increment `FAIL` correctly, no `|| true` masks a failing check, per-call knob prefixes do not leak between scenarios, and `set -u` aborting mid-file yields a red job rather than a silent pass.
- **The old `rehearsal.sh:272` stub-echo bug is gone**, and a systematic sweep — extracting every string the three stubs emit and matching it against every assertion needle — found no remaining instance of that shape. The only overlap is stub-`cargo` echoing the script's own argv, which is legitimate and load-bearing.
- All four of the incident's causal steps have coverage that a mutation turns red. 15 of 18 mutations were caught; the three survivors are MF-3 (×2) and SF-3.
- No infinite cargo-make recursion exists in the current task graph.
- `[tasks.publish-all]` has `set -e`, so a failed `publish-river` does abort before `cargo publish` reaches crates.io.
- `inspect_state` writes `--archive-out` only after parse+verify, so the `cmp` at `:707` can never compare against a file the tool declined to write.
- Twenty full rehearsal runs, zero spurious failures.

---

*Lenses: code-first, testing, skeptical/bash-adversarial, deploy-failure-mode (four blind Claude reviewers), plus an external non-Claude pass (`codex review --base origin/main`, from its own worktree, completed on the first attempt — no substitution needed). Codex returned two P1s: MF-4, which no Claude lens reached, and an independent arrival at MF-1. MF-3, MF-4 and SF-3 were re-reproduced by the synthesizer in a dedicated worktree before being reported.*

[AI-assisted - Claude]

sanity (commented):
## Round 3 re-review at `df1cb6f0` — Full tier

Third review of this PR. Round 1 (`c0846d1a`) found four Must Fix; the fix round (`5d563436`) closed all four and introduced four more; this round fixes those. CI at `df1cb6f0`: **all green** — `build`, `ui-playwright-tests`, `check-delegate-migration`, `check-pointer-freshness`, `check-room-contract-migration`, `license/cla`.

**The squash is faithful.** `git diff d313789e df1cb6f0` is empty, and `d313789e` is `5d563436` plus one `wip: round 3` commit — so the published head is byte-identical to the tree the author tested, and the round-3 delta is exactly `git diff 5d563436 df1cb6f0`: 3 files, +261/−28. Nothing was lost or duplicated in the rewrite. **No drive-by fixes**: nothing from the 15 open Should Fix items was folded in. The only changes beyond the four Must Fix are test-harness support the new scenarios require (`git_sync --allow-empty` plus a logged failure, the tracked `ui-source.txt` fixture, `/git-sync.log` in `.git/info/exclude`) and the scenario-25 rewrite that the counter exemption forced. All justified.

Lenses: **code-first**, **testing**, **skeptical/bash-adversarial**, **deploy-failure-mode** (four blind Claude reviewers), plus an external non-Claude pass. Read-only lenses shared one checkout; each mutating lens had its own worktree, as did the synthesizer. Every `file:line` re-verified against the canonical head.

---

### Verdict

**Changes requested.** All four prior Must Fix are addressed in substance, and three of them are properly closed. But **MF-1's fix does not reach the path any operator uses.** The exemption was added to `[0]` and not to the second dirty-tree check at `[0b]`, which runs 100 lines later and only when `--build` is passed — and `--build` is what all three make entry points pass. So the recovery that round 3 exists to permit is still refused, by a message that blames the build for a file the build never touched and names the counter as the evidence. The rehearsal cannot see it because scenario 33, the scenario written for exactly this, is the one provenance scenario that omits `--build`.

All four Claude lenses reached this independently, three reproduced it end to end, the external pass reached it as well, and I reproduced it myself in a dedicated worktree before including it.

Three new Must Fix. None argues against the design; all three are local.

**The external pass earned its place for the third round running.** `codex review --base origin/main` completed on the first attempt, no substitution needed. It independently reached MF-A, independently re-derived round 2's SF-3 (the u32 ceiling on the computed version), and found **MF-C** — a fork path in the pre-flight read that no Claude lens reached in any of the three rounds — plus SF-8.

---

### Prior Must Fix findings — status

| # | Prior finding | Status |
|---|---|---|
| **MF-1** | The dirty-tree gate blocked the re-run its own verdicts instruct | **Fixed in `[0]`, defeated at `[0b]`.** The exemption at `:367-390` is correct and strictly one-directional. It is not mirrored at `:494`, so on `--build` — every real entry point — the run still dies. → **new MF-A**. The two related text bugs ARE fixed: `:820` now says `git commit`, `:692-693` no longer says "commit only after a successful publish", and a new trailer at `:868-877` warns against `git checkout` on all four non-landed arms. |
| **MF-2** | `AGENTS.md` reinstated the retracted "no-op success" claim | **Resolved.** `AGENTS.md:189-198` now states the opposite and carries the correct pointer-contract carve-out. `grep -rn "no-op"` across the repo finds the claim surviving only where it is true: `check-pointer-freshness.sh:28,266,289`, `cli/src/pointer.rs:350`, `river-publish.md:130` — all pointer-contract, all carved out explicitly at `river-publish.md:139`. The PR description no longer repeats it. Verified against `contracts/web-container-contract/src/lib.rs:257-264`, which does return `InvalidUpdateWithInfo` for `version <= current`. |
| **MF-3** | Scenario 24's pin could not fail on the natural reintroduction | **Resolved for the named bypass.** I applied the exact prior bypass to `[tasks.publish-river]` in my own worktree — moving `cargo make web-container-build-inputs` into the script body ahead of the exec line — and the suite goes red, 2 failures (`body is a single command`, `body does not invoke cargo make`). Three other reintroductions still pass; see SF-2. The seam re-check at `:494-497`, which had zero coverage, is now covered: replacing its condition with `if false` gives 3 failures via scenario 24b. Verified myself. |
| **MF-4** | "An untracked file cannot change what is published" was false | **Resolved for the demonstrated vectors.** `:403-419` refuses untracked files under `ui/`, and the wording now names what was checked rather than asserting a universal. It does not over-refuse: scenario 36 pins that untracked files elsewhere stay a note, and mutating the refusal to cover all untracked files turns it red. One bypass remains — git-quoted filenames — see SF-1. |

---

## Must Fix

### MF-A. The counter exemption is not applied at `[0b]`, so the recommended re-run still fails on every real entry point — with a false message that names the counter

`scripts/publish-web-container.sh:494-497`:

```sh
if [ "$ALLOW_UNPROVEN" != "1" ] && [ -n "$(git status --porcelain --untracked-files=no)" ]; then
    git status --short --untracked-files=no | sed 's/^/    /'
    die "the build modified tracked files, so what is about to be published is no
longer the commit whose CI was verified in [0]. Nothing has been signed."
fi
```

This is the second dirty-tree check and it carries none of the exemption added at `:367-390`. It runs only under `--build`, and all three make entry points pass `--build`:

- `Makefile.toml:394` — `[tasks.publish-river]` → `exec ./scripts/publish-web-container.sh --build`
- `Makefile.toml:404` — `[tasks.publish-river-debug]` → same
- `Makefile.toml:232` — `[tasks.sign-webapp]` → `--sign-only --build`

Ordering is `[0]` (`:367`, exemption) → `[0b]` build + re-check (`:489-499`) → counter written (`:672`). The counter is written *after* `[0b]`, so this check can never observe a write from the current run. The only thing it can ever find in the counter slot is a burn left by a previous run — precisely the state round 3 exists to permit.

Reproduced in my own worktree, changing nothing but adding `--build` to scenario 33's second `run`, which is what `cargo make publish-river` does:

```
[0] provenance: clean tree, on main, at origin/main, CI green
  note: published-contract/contract-version.txt is ahead of HEAD (30000385 -> 30000386).
        That is a version an earlier run burned. COMMIT it. Never
        'git checkout' it — that reissues a version that may be live.
  HEAD == origin/main == af4fdaaa200cd9061809ec8ab32310a7e5eabc6f
  CI green on af4fdaaa200cd9061809ec8ab32310a7e5eabc6f (1 run(s), all successful)

[0b] building the inputs, under the same lock
     M published-contract/contract-version.txt

ERROR: the build modified tracked files, so what is about to be published is no
longer the commit whose CI was verified in [0]. Nothing has been signed.
```

`135 passed, 3 failed` — scenario 33's three assertions, all of them.

Three things make this worse than a fail-closed gap:

1. **The message is false.** The build modified nothing; the counter was dirty on entry. An operator reading "the build modified tracked files" above `M published-contract/contract-version.txt` concludes the build writes the counter, and goes looking for a build bug that does not exist.
2. **It contradicts the line printed twenty lines earlier.** `[0]` says "COMMIT it. Never 'git checkout' it". `[0b]` then refuses over that same file and repeats no such warning. One lens traced the three exits and confirmed `git checkout -- published-contract/contract-version.txt` **passes `[0b]`** — restoring the counter to `30000384` and making `30000386` reissuable. That is 2026-08-04, reached by following the script's own output. The comment at `:372-376` says a gate must not manufacture pressure toward the incident it exists to prevent; this is that gate doing it 100 lines later.
3. **It bites hardest in the `collision` arm.** The FORKED verdict's remedy is "re-run the publish" (`:812-813`) — the run this blocks — so the live fork persists while the operator fights the tooling, at the moment of maximum pressure.

**Why the suite is green:** every `--build` scenario (23, 24b, and the `--build` runs at `:780`, `:795`, `:882`) starts from a committed counter, and every `RUN_SYNC=0` provenance scenario (25, 26, 27, 33, 34) omits `--build`. There is no scenario at the intersection, which is the only invocation that exists in practice.

**Fix.** Factor `[0]`'s tracked-file adjudication into a function and call it from both sites. Better still, capture `git status --porcelain --untracked-files=no` immediately *before* the `cargo make` call and diff `[0b]` against that snapshot — the question `[0b]` is actually asking is "did the build change anything", not "is the tree clean", and the comment at `:491-493` already says so. That form is also strictly stronger: it would catch a build that modified the counter, which neither the current unexempted check nor a naive exemption can. And add `--build` to scenario 33, since no operator path omits it.

---

### MF-B. Obeying the new trailer's instruction produces a refusal that tells the operator to `pull` when they need to `push`

`scripts/publish-web-container.sh:868-877` (new this round) says, on every non-landed verdict:

```
  BEFORE re-running: commit published-contract/contract-version.txt (now 30000386).
```

`[0]` then requires `HEAD == origin/main` at `:430`. A local commit makes that false:

```
[0] provenance: clean tree, on main, at origin/main, CI green

ERROR: HEAD (0ce84fc4...) != origin/main (6b19fe29...). Pull first.
Set RIVER_WC_ALLOW_UNPROVEN=1 for this run if you accept that.
```

Reproduced by two lenses independently, doing exactly what the trailer says and nothing else.

"Pull first" is wrong: HEAD is *ahead*, so `git pull` changes nothing and the refusal repeats verbatim. The word `push` appears nowhere in the script. The real remedy is commit **and push**, then wait for a full CI cycle on the counter-bump SHA — `[0]` also demands CI green on that exact commit (`:437-459`), and `build.yml` has no path filter so a run does start, but not instantly.

The failure mode is the same shape as MF-A: given a message about diverging from origin, the reflex is `git reset --hard origin/main`, which discards the counter commit and restores the pre-publish value. That is the rollback, reached by following this PR's own instructions. It matters most in `not-landed`/`unknown`, where the burned version was never observed on the network so the floor cannot rescue it.

Composed with MF-A the operator is squeezed: don't commit → `[0b]` refuses claiming the build dirtied the tree; commit → `[0]` refuses telling you to pull. The two doors left are `git checkout`/`reset` on the counter, and `RIVER_WC_ALLOW_UNPROVEN=1` — which the round-3 comment at `:355-357` itself calls unacceptable because it drops every provenance check at once.

**Fix.** Make the trailer say what works. Once MF-A is fixed the honest text is "re-run now; the gate allows an uncommitted burned counter — commit and **push** it when convenient". And `:430` should branch on ahead/behind/diverged so it never says "pull" to someone who needs to push. The `landed` arm at `:820` has the same omission — `git commit` alone leaves HEAD ahead, so the *next* publish, days later by someone else, opens with "Pull first" and no memory of where the extra commit came from. Add the push to that line too.

---

### MF-C. The pre-flight accepts the first `known` read, so a stale-but-valid GET can make the script sign at a version that is already live

`scripts/publish-web-container.sh:552-563`:

```sh
preflight_attempt=1
while :; do
    read_network_version "$WORK/onnet.bin" "$WORK/onnet.log"
    [ "$NET_STATUS" = "unknown" ] || break
    ...
```

The loop breaks on the **first** `known`, so exactly one sample decides the network floor. `VERSION` is then `max(counter, that one sample) + 1` (`:617`).

The whole PR rests on the claim that a Freenet GET can return a correctly-signed state that is *behind* what other peers hold — that is why the read-back path retries through staleness and why scenario 12 ("the first read-back is stale, and the publish DID land") exists. The same mechanism applies to the same GET on the pre-flight side, where it decides what version gets signed, and there it is sampled once.

So if the local counter is behind the true network version **and** one GET answers with an older-but-validly-signed state, the script computes a version that is already live, signs a different archive at it, and publishes. Peers holding the live copy reject it (`version <= current`); peers new to the contract take the initial-state bypass and accept it. That is the 2026-08-04 fork, reached with no operator error and no override.

The counter is the thing normally preventing this, so the precondition is a counter that is behind: a fresh clone whose last counter bump was never pushed, a publish from a second machine, or an operator who gave up after a failed publish and never committed the burn. **MF-A makes that last case more likely, not less** — it is exactly the state an operator is left in when the recovery is refused.

Neither prior review caught this, and no Claude lens reached it this round. Round 2's SF-9/SF-10 flagged the pre-flight's handling of `absent` and of `inspect` failures; this is the third asymmetry on the same path and the only one that can fork the site.

**Fix.** Take the maximum over the configured `PREFLIGHT_ATTEMPTS` reads rather than the first — `read_network_version` already exists and the attempt budget is already configured, so this is a floor accumulated across the loop instead of a `break`. That does not make staleness impossible, but it removes the single-sample assumption the header's "every network read here is RETRIED, in both directions" (`:77-80`) already claims is not there.

---

## Should Fix

**SF-1. The untracked-under-`ui/` refusal is bypassed by any filename git quotes.** `:403-405` does `git status --porcelain --untracked-files=all | grep '^??' | sed 's/^?? //'` then `grep '^ui/'`. Git wraps any path containing a space, tab, backslash, quote or non-ASCII byte in double quotes, so the line starts with `"` and never matches. Verified myself in a scratch tree:

```
--- raw porcelain ---
?? "ui/src/component 2.rs"
?? ui/src/plain.rs
--- after sed, grep '^ui/' ---
ui/src/plain.rs
```

The quoted file then reaches the note branch and is printed under the words *"untracked files present, none under ui/"* — a statement contradicted by the line beneath it. `ui/src/component 2.rs` is what a GUI file-copy produces; no attacker needed. Tailwind's `@source "../src/**/*.rs"` (`ui/assets/tailwind.css:6`) globs it regardless of what git can spell, so the check fails **open** on exactly the vector it was written for. Note `-c core.quotePath=false` is not sufficient — one lens verified git still quotes the space case. Use `git ls-files --others --exclude-standard -z -- ui/`, which fixes the newline case and the not-at-git-top-level case in the same line.

**SF-2. Three reintroductions of "something builds before the lock" still pass scenario 24.** Each verified green by the testing lens against the real Makefile:

- **An indented `dependencies` key.** `rehearsal.sh:838` greps `'^dependencies'`, anchored at column 0; TOML permits leading whitespace. `tomllib` confirms `['web-container-build-inputs']` is parsed. Two spaces from the literal bug this pin exists to prevent — and this pin has already gone green over that bug once. → `138 passed, 0 failed`.
- **A single line with no literal `cargo make`.** `makers web-container-build-inputs && exec ./scripts/publish-web-container.sh --build` (`makers` is cargo-make's own binary). One line, contains the publish script, contains `--build`. → `138 passed, 0 failed`.
- **`[config] init_task`.** cargo-make runs it before every task; the pin reads only three `[tasks.*]` blocks. → `138 passed, 0 failed`.

Also uncovered: `[tasks.publish-all]` (`Makefile.toml:628`) is an operator entry point that calls `publish-river` and carries `dependencies = ["sync-cli-wasm"]`; scenario 24 never looks at it. Parse the file with `tomllib`/`taplo` rather than awk+grep, assert `[config]` carries no `init_task`/`end_task`, and match the body line for equality rather than containment.

**SF-3. `contains "reports PUBLISHED" "PUBLISHED"` cannot distinguish PUBLISHED from NOT PUBLISHED.** Carried over from round 2 and now proven at the strongest possible mutation: changing the landed banner at `:814` from `" PUBLISHED"` to `" NOT PUBLISHED"` — so a publish that landed reports that it did not — leaves **both suites fully green**, `138 passed, 0 failed`. The script's own comment at `:733` calls a NOT-PUBLISHED report on a landed publish "the report that invites the retry that forks the site". Assert the not-landed needle is absent alongside.

**SF-4. Every one of round 3's operator-guidance rewrites is untested.** The trailer at `:872-877`, the `--sign-only` reword at `:692-693` and the `git commit` line at `:820` are all new and carry zero assertions. Reverting **all three** to their pre-round-3 text leaves the suite at `138 passed, 0 failed`. Scenario 33's `contains "warns against git checkout" "Never"` (`rehearsal.sh:1026`) does not cover the trailer — instrumenting the run shows its only match is `[0]`'s note, and 33's re-run lands so the trailer never prints. A non-landed scenario asserting `"BEFORE re-running"` closes it.

**SF-5. `RIVER_WC_ALLOW_UNPROVEN` turns off more than its own text admits.** The Usage block at `:107-108` describes four checks; it also disables the post-build re-check at `:494`, a fifth, in a different phase, mentioned nowhere. Nine separate refusals end with the identical "Set RIVER_WC_ALLOW_UNPROVEN=1 for this run if you accept that" — "that" reads as the one check you just failed. An operator blocked only by a failed `git fetch` who sets it also loses clean-tree, CI-green, the counter-rollback detection at `:385` and `[0b]`. The counter-rollback refusal correctly does *not* offer it; keep that, and make the other nine name their real blast radius.

**SF-6. The untracked-`ui/` refusal fires even without `--build`.** `:415-419` refuses regardless of `DO_BUILD`, but with no build the archive is already on disk and an untracked source file provably cannot reach it. The only remedy offered is the five-check override. Gate it on `DO_BUILD -eq 1`.

**SF-7. `river-publish.md` still has no failure-recovery section, and `AGENTS.md:216-219` points at one.** For a PR whose entire subject is what an operator does after a publish that does not land, the one document they are sent to has nothing on the five verdicts, the counter left ahead, or the commit-and-push-then-wait-for-CI sequence the gate now requires. This was already flagged as a Consider in round 2; MF-A and MF-B make it the natural home for the answer.

**SF-8. "CI green" counts any workflow run on the SHA and never asks whether `Build` is one of them.** `:437-454` requires `TOTAL > 0` and `BAD == 0`, where `BAD` counts runs that are not completed-and-(success|skipped|neutral). Round 1 correctly closed "an empty list is not green"; this is the same defect one layer up. If `Build` has not been created yet — the seconds after a merge, which is exactly when an operator publishes — or if it stops being triggered at all (a `paths` filter, a workflow-file error), the remaining runs can all be green and the gate says green on code `Build` never touched. On the head I reviewed the four runs are `Build`, `Check Room-Contract Migration`, `Check Pointer Freshness`, `Check Delegate Migration`, all success, so the gate is correct today — but nothing makes it go red for the input that matters. Require a completed, successful run named `Build` by name, rather than inferring it from the absence of failures. *(Found by the external pass.)*

*(The 15 Should Fix items from round 2 remain open, as instructed. The skeptical lens re-confirmed all five standing hazards are unchanged and still live: fd 9 leaking into `cargo make` children, no INT/TERM trap over the critical section, the u32 ceiling covering `COUNTER` but not `VERSION` — `wc_next_version 1 known 4294967294` still returns `4294967295`, which has no successor — the counter read outside the lock, and the lock degrading silently to no-lock.)*

---

## Consider

- **The rejection of the out-of-tree high-water file was the right call for a slightly wrong reason.** Keeping the counter tracked is correct: it is the only home that survives a fresh clone or a second machine, which `/tmp` cannot. But the stated reason — that a `$XDG_RUNTIME_DIR` file "would silently vanish and keep reading as protection" — overstates it, because the round-2 proposal was `max(counter, highwater, network) + 1`, an *additional* floor. If it vanished the run would degrade to today's two floors, not to nothing. Worth correcting in the comment so a future editor does not reason from it.
- `:368` interpolates `$VERSION_FILE` raw into a grep BRE and matches a *suffix*, not the porcelain path field. `.` is a wildcard, so `contract-versionXtxt` is exempted too, and any tracked path ending `<space>published-contract/contract-version.txt` would be. Checked against the real repo: 302 tracked files, one match, none containing a space — hardening, not a live hole. Match the field explicitly.
- The exemption's stated safety rationale at `:359-362` ("re-floored against the version read off the network") holds only for `NET_STATUS=known`; `wc_next_version` returns `counter + 1` unfloored on the `absent` and `unknown` branches, both reachable via documented overrides. The *direction* is what makes it safe, which is the stronger argument and the one to write down.
- A `HEAD_COUNTER` over 10 digits falls into the `else` at `:385` and is told "Restore it to at least $HEAD_COUNTER" — a value the u32 guard at `:226` then refuses forever. Deserves its own message.
- Gitignored files under `ui/assets` are invisible to `--untracked-files=all`, and `dx build` walks that directory wholesale. Narrow in practice; the comment at `:396-401` is honestly scoped but does not name this blind spot.
- If `REPO_ROOT` is not the git top level (river vendored inside an outer repo), porcelain paths are prefixed and `grep '^ui/'` matches nothing. The counter grep fails closed in that layout; only the untracked check fails open. The `git ls-files -- ui/` fix covers it.
- Two numeric parsers on one value: `$COUNTER` passes a digits-only check permitting leading zeros at `:214`, is compared with `test -gt` (base 10) at `:379`, and incremented with `$(( ))` (octal) in the lib. `030000377` compares as 30000377 and increments as 6291712. Reachable only with a hand-edited counter plus an override.
- Still no assertion on the total check count (`rehearsal.sh:1071`), and still no scenario pinning the documented exit codes 3/4/5/6 — swapping `collision` and `not-landed`, two verdicts with opposite recovery actions, is green in both suites.
- Scenario 25's `not_zero` and `nothing was published` both still pass when the non-counter refusal is removed entirely, because the run then refuses via the counter branch instead. Only its `contains` is load-bearing. Worth knowing when adding the next provenance scenario.
- The PR body still omits the entire provenance gate — `[0]`, the counter exemption, the untracked refusal, `[0b]`, and `RIVER_WC_ALLOW_UNPROVEN`. It also says "27 checks" (33) and "36 scenarios" (37 since 24b). This is round 2's SF-15, still open.
- Ctrl-C between the counter write at `:670` and the verdict leaves the operator with a burned counter, one modified file, and nothing on screen that ever said the counter was burned. A single `say` at the point of the write would cover the interrupt path.
- The external pass argued `origin/main` should be re-fetched and re-compared after the multi-minute build, since it can advance mid-run. Recorded but **not adopted**: publishing a main that was green when checked, a few minutes stale, is a normal publish, not a provenance violation — the `publish-from-main` rule is about not shipping a feature branch. Re-fetching would also open a livelock on a busy day. Noted so it is not re-raised.
- The external pass independently re-derived round 2's SF-3 (the u32 ceiling guards `COUNTER` but not the computed `VERSION`, which is written to the tracked counter before the Rust signer refuses it). Unchanged this round, and the corroboration raises confidence that it is worth the one line.

---

*Lenses: code-first, testing, skeptical/bash-adversarial and deploy-failure-mode (four blind Claude reviewers), plus an external non-Claude pass (`codex review --base origin/main`, from its own worktree, completed on the first attempt — no substitution needed). MF-A was reached independently by all five. MF-C and SF-8 came from the external pass alone. MF-A, MF-3's bypass, 24b's coverage and SF-1 were re-reproduced by the synthesizer in a dedicated worktree before being reported.*

---

## What was verified and found sound

- The squash is byte-identical to the tested checkpoint; no fix lost or duplicated; no drive-by Should Fix work folded in.
- `bash -n` clean on both scripts and the rehearsal. `shellcheck -S style` clean on the lib and the rehearsal; only `SC1091` (info) on the publish script. The suite as shipped is `138 passed, 0 failed`.
- **The exemption's edge cases mostly fail closed, and were checked rather than assumed.** A deleted counter dies at the `[ -f "$f" ]` loop (`:184`) before the exemption is reached; an unmerged `UU` dies at the digit check (`:213`); a non-numeric `HEAD:` value refuses at `:377`; below-HEAD and equal-to-HEAD both refuse (scenario 34, and mutating the exemption to unconditional gives 3 failures); staged, unstaged and `MM` all exempt identically and all compare against the worktree value; leading zeros do not break `test -gt`, which parses base 10; and a rename *into* the counter path cannot occur, because the counter exists in HEAD so git records `M` plus a separate `D`, and the `D` line lands in `DIRTY_OTHER` and refuses.
- Mutation coverage of the new scenarios, run by the testing lens in its own worktree: 33 (exemption removed) → 5 failures; 34 (exemption unconditional) → 3; 35 (refusal disabled) → 4; 36 (refusal over-broad) → 3; 24b (`[0b]` disabled) → 3, which I re-ran myself. Every new scenario can fail.
- `git_sync`'s `--allow-empty` cannot mask a setup failure: the amend is `|| setup_failed`, and both postconditions still run afterwards. `git-sync.log` is correctly excluded so it does not perturb scenarios 35/36.
- The `web-container-tool` `inspect` path re-verified: both length fields bounds-checked, `verify` rebuilds `version || archive` exactly as `sign_webapp` and `validate_state` do, the archive is written only after parse+verify and not written when `--expect-version` fails, and the tests cover foreign key, swapped archive, truncation, an absurd `metadata_len`, and the same-version-different-archive case that is the fork.
- Documentation coherence outside the two Must Fix: `DEVNOTES.md`, the Usage block, `river-publish.md` and `AGENTS.md` agree with the code and with each other on the no-op-success question, the pointer carve-out is stated correctly in all four places, and `ui/public/contracts/*.wasm` — gitignored but force-added and tracked — does show as `M` and refuse, so a local `sync-wasm` cannot slip through.

[AI-assisted - Claude]

sanity:
Closing this in favour of #635, which fixes the same bug in a much smaller change.

The analysis here is sound and the incident write-up above is accurate. What changed is the judgement about how much machinery it warrants.

**The bug's blast radius is bounded.** Two states at one version never converge, but recovery is simply the next successful publish at a higher version, so exposure is one release cycle and no data is lost. This PR grew to +2620/-130 across 12 files and three review rounds producing 11 blocking findings — disproportionate to a bug whose fix is "publish again".

**The pre-flight read did not hold up.** The most recent round found that reading the network before signing, and taking a version floor from that read, can cause the very fork it is meant to prevent: it accepts the first `known` GET, and on an eventually-consistent network a stale-but-validly-signed read plus a counter that is behind produces a signature at a version the network has already accepted. That is the same failure arrived at more elaborately, and it is the piece the replacement deliberately does not have.

**#635 keeps the two parts that address the actual cause**, at the two links in the 2026-08-04 chain:

1. **The version counter is forward-only.** The rollback on a failed publish is what handed the retry an already-used version. It is removed from all four publish tasks. A burned version costs a gap in the sequence and nothing else — the contract enforces monotonicity, not contiguity.
2. **An advisory post-publish read-back.** It reports the on-network version and whether its archive is byte-for-byte the one just published. It reports; it does not gate: always exits 0, is called with `|| true`, and derives no version floor from what it reads. This is what breaks the operator side of the incident, which was *timeout → believed it failed → retried*.

Not carried over: the machine-global lock, the publish-from-main / CI provenance gate, the contract-id derivation, the pre-flight read and its version floor, and the escape-hatch env vars.

The `inspect` subcommand on `web-container-tool` and the packed-state parser from this branch were reused in #635, minus the `--expect-version` gate.

**The branch is deliberately left in place** — the review history and the incident analysis on it are worth keeping.

[AI-assisted - Claude]

PR #635, opened 2026-08-29
fix(publish): make the web-container version counter forward-only

sanity:
Replaces #634, which is being closed. Same bug, much smaller fix.

## Problem

On 2026-08-04 the River web container forked. The sequence:

1. `fdev network publish` reported `put timed out after 1 peer attempt(s)`.
2. The state had in fact landed. The exit code said otherwise.
3. `[tasks.publish-river]` treated that exit code as proof of failure and **rolled the version counter back**.
4. The operator retried. The rollback had handed the retry version 30000377, which the network had already accepted.
5. The rebuild is not byte-reproducible, so a **different** archive was signed at that same version, and both landed.

Two states at one version never converge: the contract rejects `version <= current` in both directions, and `summarize_state` emits only the version, so anti-entropy sees them as equal.

Recovery is simply the next successful publish at a higher version, so exposure is one release cycle and no data is lost. That bound is why this is a small fix rather than a large one.

## Approach

Two changes, at the two links in the chain above.

**1. The counter is forward-only (step 3).** No publish task snapshots it, and none rolls it back. A burned version costs a gap in the sequence and nothing else — the contract enforces monotonicity, not contiguity, so gaps have always been fine. That makes the rollback all cost and no benefit.

Five sites touch a counter, and all five are covered: the two signing tasks that bump it (`sign-webapp`, `sign-webapp-test`) and the three publish tasks that used to roll it back (`publish-river`, `publish-river-debug`, `publish-river-test`). `publish-river-debug` matters because it shares the production counter *and* the production contract; the test pair carried a copy-pasted duplicate of the same rollback.

**2. An advisory post-publish read-back (step 2).** `scripts/publish-readback.sh` fetches the state back and reports what is actually live: the on-network version, and whether its archive is byte-for-byte the one just published. It classifies as `LANDED` / `FORKED` / `SUPERSEDED` / `NOT SEEN YET` / unreadable.

This is the piece that addresses the operator's side of the incident, which was *timeout → believed it failed → retried*. A read-back that says "it landed despite the timeout, do not republish" stops that at the second step.

It **reports; it does not gate**, in both senses — exit code and wall clock:

- it always exits 0, whatever it finds and whatever goes wrong inside it;
- the network read is wrapped in `timeout(1)`, so a node that wedges cannot hang the publish. `|| true` rescues a non-zero exit, not a hang, so the exit-code discipline alone was not enough;
- `publish-river` calls it with `|| true`, so it cannot fail a publish that otherwise succeeded;
- it derives **no version floor** from what it reads. Reading the network *before* signing and signing at a floor taken from that read looks stronger and is not — Freenet is eventually consistent, so a stale-but-validly-signed read plus a counter that is behind produces a signature at a version the network has already accepted. That is the same fork reached more elaborately, and it is why #634's pre-flight read was abandoned;
- `NOT SEEN YET` explicitly advises re-reading rather than republishing, because one GET is a sample and not a verdict.

Reading the version back needs the packed state parsed and its signature checked, so `web-container-tool` gains an `inspect` subcommand: it parses `[metadata_len u64 BE][metadata CBOR][web_len u64 BE][web]`, verifies the signature against the contract parameters, prints `version=<n>`, and can write the embedded archive out for a byte comparison. `--parameters` matters: a version read out of unverified bytes is not evidence of anything, it is whatever the responder chose to say.

### What is deliberately not here

No machine-global publish lock, no CI provenance gate, no contract-id derivation, no pre-flight network read, no version floor, no new escape-hatch env vars. Those were #634's scope and are out of proportion to a bug whose recovery is a republish.

## Testing

**`contracts/.../web-container-tool/src/main.rs`** — three unit tests, run by the existing `cargo test -p web-container-tool` CI step:

- `sign_output_parses_back_and_verifies` — a state packed from `sign`'s own output parses back to the version signed, hands back the exact archive bytes, and verifies under the parameters `sign` wrote. This pins the read-back's parser against the other end of the same format.
- `verify_rejects_a_state_whose_archive_was_swapped` — same metadata, different archive: must not verify. A verifier that checked the version and ignored the archive would report the 2026-08-04 fork as a clean landing.
- `parse_refuses_truncated_states` — malformed or truncated answers are errors, never a version.

**`scripts/tests/publish-guards-test.sh`** — 11 checks, wired into `build.yml` (a guard no job runs is a guard that passes by default, per #614):

- two source scrapes asserting no publish task snapshots or rolls back the counter. A source scrape because the assertion is an *absence*, and the code lives in cargo-make script blocks that cannot run without a network, a signing key and a built UI. The realistic regression is someone reinstating one of the four near-identical copies;
- the read-back's five classifications, driven by stubbed `fdev` and `web-container-tool`, plus the invariant that every path exits 0.

**Mutation results** — each change was reverted, the test confirmed RED, then restored:

| Mutation | Test that caught it |
|---|---|
| Reinstate the `.prev` snapshot + rollback in `publish-river` | both counter scrapes FAIL |
| `cmp` always reports a match | `a version carrying different bytes is reported as a fork` FAIL |
| Read-back exits with the publish's code | `the read-back never propagates the publish's exit code` FAIL |
| Drop the "do NOT republish" guidance | `a publish that timed out but landed says do NOT republish` FAIL |
| Drop the not-yet-seen warning | `not-yet-seen warns against republishing on one early read` FAIL |
| Length fields read little-endian | `sign_output_parses_back_and_verifies` + swap test FAIL |
| `verify()` accepts anything | `verify_rejects_a_state_whose_archive_was_swapped` FAIL |

`shellcheck -x` clean on both scripts. `cargo fmt` and `cargo clippy -p web-container-tool --all-targets` clean.

Not exercised against a live node — the read-back's network call is `fdev network execute get`, stubbed in tests.

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/river-publish.md
+++ b/.claude/rules/river-publish.md
@@ -91,7 +91,9 @@ cargo make publish-river  # bumps published-contract/contract-version.txt
 ```bash
 # `cargo make sign-webapp` (run transitively by publish-river) incremented
 # the counter. Commit it together with whatever other changes the publish
-# included.
+# included — and commit it EVEN IF the publish reported a failure. See
+# "Version policy" below: the counter is forward-only, because a reported
+# failure is not evidence the state did not land.
 git add published-contract/contract-version.txt
 git commit -m "chore: bump web-container version after publish"
 curl -s http://127.0.0.1:7509/v1/contract/web/raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv/ | head -5
@@ -117,6 +119,10 @@ already-used version is a no-op *success*, so the network will never tell you
 it was ignored.
 
 **Version policy:** the canonical source is `published-contract/contract-version.txt`, which `cargo make sign-webapp` increments and writes back each publish. NEVER base the version on wall-clock time (`date +%s / 60`) — the previous scheme bit us 2026-05-16 when the on-network version had drifted ahead of the timestamp-derived value. The counter file makes the version a strict monotonic local invariant; gaps are fine (the contract enforces monotonicity, not contiguity).
+
+**The counter is forward-only. Never re-use a version, and never hand-edit it downwards.** A publish that reports a failure is not evidence the state did not land — on 2026-08-04 `fdev` reported `put timed out after 1 peer attempt(s)` while the state was on the network the whole time. The old flow rolled the counter back on that exit code, the operator retried, the rebuild was not byte-reproducible, and a *different* archive got signed at the *same* version. Two states at one version never converge (the contract rejects `version <= current` in both directions, and the state summary carries only the version, so anti-entropy cannot tell them apart), and the only recovery is the next publish at a higher version. Burning a version costs a gap in the sequence and nothing else.
+
+So: **commit the bumped counter whether or not the publish reported success.** `publish-river` then reads the state back and prints an advisory report — the live version, and whether its archive is byte-for-byte the one you just published. Read it before deciding anything. It is a report, not a gate: it cannot fail a publish, and `NOT SEEN YET` means "one GET did not find it yet", which on an eventually-consistent network is not the same as "it did not land". Re-read with `fdev network execute get <contract_id> -o /tmp/state.bin` rather than republishing on a single early read.
 
 ## Two independent release surfaces
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -278,6 +278,16 @@ jobs:
       # tool and the verifying contract agree on the parameter encoding.
       run: cargo test -p web-container-contract -p web-container-tool
 
+    - name: Test publish guards (forward-only counter + read-back)
+      # The publish path is shell inside cargo-make script blocks, so its
+      # tests are shell too. This asserts that no publish task rolls the
+      # web-container version counter back on a failed publish -- the step
+      # that handed a retry an already-used version and forked the site on
+      # 2026-08-04 -- and that the post-publish read-back classifies what the
+      # network answered without ever failing the publish. A guard no job runs
+      # is a guard that passes by default (freenet/river#614).
+      run: ./scripts/tests/publish-guards-test.sh
+
   ui-playwright-tests:
     # Ubicloud, NOT a GitHub-hosted larger runner -- see the `build` job above.
     runs-on: ubicloud-standard-8
--- a/.gitignore
+++ b/.gitignore
@@ -48,11 +48,12 @@ ui/tests/test-results/
 # Manual e2e test screenshots (see #244 review feedback)
 /dm-*.png
 
-# Web-container version-counter rollback snapshot. Created by
-# `cargo make sign-webapp` BEFORE bumping the counter; removed on
-# successful publish, or moved back over the counter on failure.
-# Should never appear in commits.
+# Web-container version-counter rollback snapshot. No longer produced —
+# the counter is forward-only (see [tasks.sign-webapp] in Makefile.toml),
+# so nothing writes this file. Kept ignored so a stale one left by an
+# older checkout cannot be committed by accident.
 /published-contract/contract-version.txt.prev
+/test-contract/contract-version.txt.prev
 
 # Syncthing in-flight transfer files (`.syncthing.<name>.tmp` are
 # created by the syncthing daemon while a file is being received).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -181,7 +181,11 @@ number strictly higher than the current published version. The version
 is tracked by a committed counter at
 `published-contract/contract-version.txt`; `cargo make sign-webapp`
 increments it. Commit the bumped counter alongside other publish
-artifacts.
+artifacts — **including when the publish reports a failure**. The counter
+is forward-only: a reported failure is not evidence the state did not
+land, and re-using a version is what forked the site on 2026-08-04.
+`publish-river` prints an advisory read-back of what is actually live;
+read it before retrying. See `.claude/rules/river-publish.md`.
 
 **Contract ID:** `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`
 
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -212,33 +212,37 @@ script = '''
 # was 30000208 while the timestamp gave 29649402, so the publish was
 # rejected.
 #
-# Counter lifecycle:
-# 1. `sign-webapp` here writes a `.prev-version` snapshot BEFORE bumping
-#    the counter file, so a downstream `publish-river` failure can roll
-#    back atomically (see [tasks.publish-river]'s `on-failure` arm).
-# 2. The counter file is bumped to `current+1` and the new metadata is
+# Counter lifecycle — FORWARD-ONLY. The counter is bumped here and is
+# NEVER rolled back, not even when the publish that follows fails:
+# 1. The counter file is bumped to `current+1` and the new metadata is
 #    signed with that value.
-# 3. After a successful `fdev publish`, the caller commits the bumped
-#    counter file (the snapshot is left for the publish task to clean
-#    up). Skeptical-review (#258) flagged that bumping before publish
-#    risks leaving the local file ahead of on-network if publish fails;
-#    the .prev-version snapshot + rollback closes that window.
+# 2. Whatever `publish-river` then reports, the bumped value stands. The
+#    caller commits the bumped counter file.
+#
+# Why forward-only (freenet/river#634, the 2026-08-04 fork): the counter
+# used to be snapshotted here and rolled back by `publish-river` when
+# `fdev network publish` exited non-zero. But that exit code is not
+# evidence the state did not land, so the rollback handed the operator's
+# retry a version the network had already accepted, and the rebuild was
+# not byte-reproducible. Full account in scripts/publish-readback.sh.
+#
+# A burned version costs a gap in the sequence and nothing else — the
+# contract enforces monotonicity, not contiguity — which makes rolling
+# back all cost and no benefit.
 #
 # Locking: this script is NOT safe under two parallel `publish-river`
 # runs on the same workstation — they would race the read-modify-write.
 # Wrap in `flock` if you need that:
 #     flock published-contract/contract-version.txt cargo make publish-river
 version_file="published-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
 if [ ! -f "$version_file" ]; then
     echo "ERROR: $version_file missing. Seed it with the current on-network version."
     exit 1
 fi
 current=$(cat "$version_file" | tr -d '[:space:]')
 version=$((current + 1))
-cp "$version_file" "$prev_snapshot"
 echo "$version" > "$version_file"
-echo "Web-container version: $current -> $version (commit $version_file after a successful publish)"
+echo "Web-container version: $current -> $version (commit $version_file — the bump stands whatever publish reports)"
 target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool sign \
   --input target/webapp/webapp.tar.xz \
   --output target/webapp/webapp.metadata \
@@ -372,27 +376,41 @@ fi
 contract_id=$(cat published-contract/contract-id.txt)
 echo "Publishing using committed contract: $contract_id"
 version_file="published-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
-if fdev network publish \
+version=$(tr -d '[:space:]' < "$version_file")
+
+fdev network publish \
   --code published-contract/web_container_contract.wasm \
   --parameters published-contract/webapp.parameters \
   contract \
   --webapp-archive target/webapp/webapp.tar.xz \
-  --webapp-metadata target/webapp/webapp.metadata; then
-    # Publish succeeded — the bumped counter file is the new truth.
-    # Discard the snapshot. Caller must commit $version_file.
-    rm -f "$prev_snapshot"
+  --webapp-metadata target/webapp/webapp.metadata \
+  && publish_rc=0 || publish_rc=$?
+
+# Advisory read-back. It REPORTS what the network actually holds; it
+# never gates, never refuses, and its own failure is not a publish
+# failure — hence the `|| true`. See scripts/publish-readback.sh.
+#
+# This is the piece that breaks the 2026-08-04 chain, which was
+# "timeout -> operator believed it failed -> operator retried". A
+# read-back that says "it landed despite the timeout, do not retry"
+# stops that at the second step.
+./scripts/publish-readback.sh \
+  "$contract_id" \
+  published-contract/webapp.parameters \
+  target/webapp/webapp.tar.xz \
+  "$version" \
+  "$publish_rc" \
+  "target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool" || true
+
+if [ "$publish_rc" -eq 0 ]; then
     echo "Publish OK. Remember to commit $version_file."
 else
-    rc=$?
-    # Publish failed — roll the counter back so subsequent retries
-    # don't accumulate a drift. The on-network version is unchanged,
-    # and our local counter is now matched to it again.
-    if [ -f "$prev_snapshot" ]; then
-        mv "$prev_snapshot" "$version_file"
-        echo "Publish failed (exit $rc); rolled $version_file back to its pre-publish value." >&2
-    fi
-    exit $rc
+    # The counter is NOT rolled back — see the lifecycle note in
+    # [tasks.sign-webapp]. Version $version is burned whether or not the
+    # publish landed, and burning one costs only a gap in the sequence.
+    echo "fdev network publish exited $publish_rc. $version_file stays at $version; commit it." >&2
+    echo "Read the read-back above before retrying: a timeout does not mean the state is not live." >&2
+    exit "$publish_rc"
 fi
 '''
 
@@ -407,22 +425,32 @@ if [ ! -f "published-contract/contract-id.txt" ]; then
 fi
 contract_id=$(cat published-contract/contract-id.txt)
 echo "Publishing using committed contract in debug mode: $contract_id"
+# Same production contract and same production counter as
+# [tasks.publish-river], so the same forward-only rule and the same
+# advisory read-back apply. See [tasks.sign-webapp] for why.
 version_file="published-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
-if fdev network publish \
+version=$(tr -d '[:space:]' < "$version_file")
+
+fdev network publish \
   --code published-contract/web_container_contract.wasm \
   --parameters published-contract/webapp.parameters \
   contract \
   --webapp-archive target/webapp/webapp.tar.xz \
-  --webapp-metadata target/webapp/webapp.metadata; then
-    rm -f "$prev_snapshot"
-else
-    rc=$?
-    if [ -f "$prev_snapshot" ]; then
-        mv "$prev_snapshot" "$version_file"
-        echo "Publish failed (exit $rc); rolled $version_file back to its pre-publish value." >&2
-    fi
-    exit $rc
+  --webapp-metadata target/webapp/webapp.metadata \
+  && publish_rc=0 || publish_rc=$?
+
+./scripts/publish-readback.sh \
+  "$contract_id" \
+  published-contract/webapp.parameters \
+  target/webapp/webapp.tar.xz \
+  "$version" \
+  "$publish_rc" \
+  "target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool" || true
+
+if [ "$publish_rc" -ne 0 ]; then
+    echo "fdev network publish exited $publish_rc. $version_file stays at $version; commit it." >&2
+    echo "Read the read-back above before retrying: a timeout does not mean the state is not live." >&2
+    exit "$publish_rc"
 fi
 '''
 
@@ -550,16 +578,17 @@ fi
 # as the production publish (see [tasks.sign-webapp] for the 2026-05-16
 # incident). Test contract counter is NOT committed since the test
 # contract ID is regenerated freely.
-# .prev snapshot is captured here for the publish task to roll back on
-# failure — same pattern as the production sign/publish flow.
+# Forward-only, same as the production flow: the bump is never rolled
+# back on a failed publish, because a failed publish is not evidence the
+# state did not land, and re-issuing a version the network already
+# accepted is what forked the site on 2026-08-04. See
+# [tasks.sign-webapp].
 version_file="test-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
 if [ ! -f "$version_file" ]; then
     echo "0" > "$version_file"
 fi
 current=$(cat "$version_file" | tr -d '[:space:]')
 version=$((current + 1))
-cp "$version_file" "$prev_snapshot"
 echo "$version" > "$version_file"
 echo "Test web-container version: $current -> $version"
 target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool sign \
@@ -600,21 +629,28 @@ contract_id=$(fdev get-contract-id \
 
 echo "Publishing TEST contract: $contract_id"
 version_file="test-contract/contract-version.txt"
-prev_snapshot="$version_file.prev"
-if fdev network publish \
+version=$(tr -d '[:space:]' < "$version_file")
+
+fdev network publish \
   --code target/wasm32-unknown-unknown/release/web_container_contract.wasm \
   --parameters target/webapp/webapp-test.parameters \
   contract \
   --webapp-archive target/webapp/webapp-test.tar.xz \
-  --webapp-metadata target/webapp/webapp-test.metadata; then
-    rm -f "$prev_snapshot"
-else
-    rc=$?
-    if [ -f "$prev_snapshot" ]; then
-        mv "$prev_snapshot" "$version_file"
-        echo "Test publish failed (exit $rc); rolled $version_file back." >&2
-    fi
-    exit $rc
+  --webapp-metadata target/webapp/webapp-test.metadata \
+  && publish_rc=0 || publish_rc=$?
+
+./scripts/publish-readback.sh \
+  "$contract_id" \
+  target/webapp/webapp-test.parameters \
+  target/webapp/webapp-test.tar.xz \
+  "$version" \
+  "$publish_rc" \
+  "target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool" || true
+
+if [ "$publish_rc" -ne 0 ]; then
+    # Forward-only: the counter is not rolled back. See [tasks.sign-webapp].
+    echo "Test publish exited $publish_rc. $version_file stays at $version." >&2
+    exit "$publish_rc"
 fi
 
 echo ""
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -400,7 +400,7 @@ fdev network publish \
   target/webapp/webapp.tar.xz \
   "$version" \
   "$publish_rc" \
-  "target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool" || true
+  "target/native/x86_64-unknown-linux-gnu/release/web-container-tool" || true
 
 if [ "$publish_rc" -eq 0 ]; then
     echo "Publish OK. Remember to commit $version_file."
@@ -439,13 +439,20 @@ fdev network publish \
   --webapp-metadata target/webapp/webapp.metadata \
   && publish_rc=0 || publish_rc=$?
 
+# `release`, NOT ${BUILD_PROFILE}. This task sets BUILD_PROFILE=dev, and
+# cargo's `dev` profile emits into `debug/`, never `dev/` — so
+# `${BUILD_PROFILE}` here names a path that cannot exist and the read-back
+# would silently short-circuit on "no web-container-tool", on the one task
+# that publishes to the PRODUCTION contract and counter. (The tool itself is
+# always the release build: a cargo-make task's `env` does not reach its
+# dependencies, so `sign-webapp` builds it under the default profile.)
 ./scripts/publish-readback.sh \
   "$contract_id" \
   published-contract/webapp.parameters \
   target/webapp/webapp.tar.xz \
   "$version" \
   "$publish_rc" \
-  "target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool" || true
+  "target/native/x86_64-unknown-linux-gnu/release/web-container-tool" || true
 
 if [ "$publish_rc" -ne 0 ]; then
     echo "fdev network publish exited $publish_rc. $version_file stays at $version; commit it." >&2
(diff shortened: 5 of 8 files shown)


sanity (commented):
## Multi-perspective review — PR #635

**Reviewed at `15336762d34a83b5261773c881dca7f293ffdaa5`.** CI green on that SHA (build, playwright, three migration/pointer checks, CLA).

**Lenses:** code-first, testing (mutation, own worktree), skeptical/bash-adversarial, plus `codex review --base origin/main` (external, non-Claude). Codex ran successfully and reported no actionable correctness issues; that is recorded as one lens, not as clearance. Every `file:line` below was re-verified against the canonical head via `gh api`, and the two headline defects were re-reproduced independently of the lens that first found them.

**Verdict: sound descope, approve after the Must Fix.** The forward-only counter is complete and correct — I enumerated every counter write independently and there is no rollback left anywhere. The read-back is genuinely advisory *by exit code*, which I confirmed under cargo-make's real `sh -e` runner. It is not advisory *by wall clock*, which is the one Must Fix.

---

### The property you asked to be verified: is the read-back provably advisory?

**By exit code: yes, proven.** Every exit routes through `done_` (`scripts/publish-readback.sh:73-82`), which `exit 0`s unconditionally. Verified empirically under cargo-make's actual script runner (`sh -e`, dash) that a read-back exiting 42, a non-executable read-back, and a missing read-back all leave the task at `Publish OK.` with rc 0. The script writes nothing outside its own `mktemp -d`, never touches the counter, git state, or the contract, and derives no version floor from the network read. No command injection: `contract_id` containing `$(id)` and backticks is printed literally.

**By wall clock: no.** See Must Fix 1.

---

### Must Fix

**1. The read-back can hang a publish indefinitely. `|| true` cannot rescue a hang.** `scripts/publish-readback.sh:112-113` delegates its entire time bound to `fdev --timeout`, and there is no `timeout(1)` wrapper there or at any of the three call sites (`Makefile.toml:397`, `:442`, `:642`).

In freenet-core, `--timeout` wraps **only** `client.recv()` (`crates/fdev/src/commands.rs:481`). Outside it: `start_api_client` at `:476` — whose `tokio_tungstenite::connect_async` (`crates/fdev/src/commands/v1.rs:50`) has no timeout at all — `execute_command` at `:479`, and `close_api_client` at `:515`. A node that is alive enough to complete the TCP handshake but never serves the WebSocket upgrade blocks forever. Reproduced with a stub `fdev` that ignores `--timeout`: `RIVER_READBACK_TIMEOUT=1` with a sleeping stub waited the full 8s, and with a long sleep an *external* `timeout 15` had to kill it (rc 124).

The precondition is narrow — the node must wedge between the publish and the read-back — but the consequence lands exactly where this PR is trying to help. The operator is left at a prompt that never returns, having never seen `Makefile.toml:406`'s "Publish OK. Remember to commit `$version_file`", and Ctrl-C leaves the counter bumped-but-uncommitted, which is precisely the state that invites the reflexive `git checkout -- published-contract/contract-version.txt` that forked the site.

Fix is one line and adds no gate — the timeout path already routes to `done_`:
```bash
timeout -k 5 "$((readback_timeout + 15))" fdev network execute get ... 
```
with a fallback when `timeout(1)` is absent.

---

### Should Fix

**2. `Makefile.toml:448` — the read-back is dead code in `publish-river-debug`.** The path interpolates `${BUILD_PROFILE}`, which `Makefile.toml:419` sets to `dev`; cargo's `dev` profile emits into `debug/`, never `dev/`. (Independently, a cargo-make task's `env` does not propagate to its dependencies, so `sign-webapp` builds the *release* tool regardless.) Either way `.../dev/web-container-tool` does not exist and the script always short-circuits at `:95-98`. That matters because this task publishes to the **production contract and the production counter** — its own comment at `:428-430` says so — so it is the one publish path where the advisory is silently absent. Hard-code `release`, or map `dev`→`debug`.

**3. `scripts/publish-readback.sh:172` — a missing local archive is reported as `FORKED`.** `cmp -s` returns 2 (not 1) when a file is unreadable, and `! cmp -s` treats that as "differs". Reproduced with `expected_archive` pointing at a nonexistent path: the script prints the full "Two different states now share one version, and they cannot converge" block and advises a fresh publish. A local tooling problem is thus rendered as the most alarming verdict the script has. Guard both operands exist and are non-empty first, and distinguish `cmp` rc 1 (differ → FORKED) from rc ≥ 2 (could not compare → unknown).

**4. `scripts/publish-readback.sh:87-93, 149, 158` — a failed comparison falls through to `LANDED`.** The integer guard validates the two fields *concatenated* (`case "$our_version$publish_rc"`), so an empty `our_version` with `publish_rc=0` concatenates to `"0"` and passes. Both comparisons then error, `if` treats the error as false, and control reaches the `LANDED` print at `:187`. Reproduced:
```
line 149: [: : integer expression expected
line 158: [: : integer expression expected
LANDED. The network is serving version 100 with the exact archive just published.
exit=0
```
Reachability through the shipped call sites is low — `sign-webapp` always writes a numeric value before `Makefile.toml:379` reads it — so this is about the *direction* of failure, not a live path: a check whose inputs are broken should report unknown, not report success. Validate the two fields separately, and resolve the comparison once into lt/eq/gt with "could not compare" as its own reported branch.

**5. `scripts/publish-readback.sh:149-156` — `SUPERSEDED` says "Nothing to do" for a case where your publish did not land.** `net_version > our_version` has a second cause the message does not consider: the local counter is behind the network, so the state signed at `our_version` was rejected (`contracts/web-container-contract/src/lib.rs:259-266`) and the operator's changes are *not* live. That is the 2026-05-16 drift incident the counter scheme was introduced to fix, still reachable from a stale counter, a second checkout, or a second machine — and because the fix is forward-only, the next run bumps by exactly one and can stay behind indefinitely, printing "Nothing to do" every time. Telling the operator this is not deriving a version floor, so it is compatible with the descope. Suggest adding: "If nobody else published, your counter is behind the network — set the counter above `$net_version` and publish again."

**6. `scripts/tests/publish-guards-test.sh:60-62` — the rollback guard matches the syntax that has already been deleted.** It flags `mv`/`cp` on lines mentioning `version_file`. Since the `.prev` file is gone, a reinstated rollback will not be written as a `mv`. Two mutations survived a full harness run:
```
echo "$version_before" > "$version_file"     → 11 checks, 0 failures
git checkout -- "$version_file"              → 11 checks, 0 failures
```
against the control `mv "$prev_snapshot" "$version_file"`, which is correctly killed. Two further mutations (`publish-river-debug`, `publish-river-test` using `$version_file.prev` rather than the literal filename) killed only 1 of the 2 assertions. Invert it into an allowlist: scan each publish/sign block for *any* write to the counter (`>`, `>>`, `mv`, `cp`, `tee`, `sed -i`, `git checkout`, `git restore`) and permit only the single known-good `echo "$version" > "$version_file"` in `sign-webapp`/`sign-webapp-test`. Then a novel rollback fails closed.

**7. The advisory property itself is untested at the call site.** Three mutations survived:
- removing `|| true` from **all three** call sites (`Makefile.toml:403, 448, 648`) → `11 checks, 0 failures`
- removing all three read-back invocations outright → `11 checks, 0 failures`
- six of the twelve exit-0 guards (`:49, 91, 97, 102, 105, 125`) made non-zero **simultaneously** → `11 checks, 0 failures`

The untested guards are the ones most likely to fire in practice (`fdev` not on PATH, tool not built at the expected path, empty state). Combined, these are the live risk: the read-back could become a publish gate with CI green — the one thing the design forbids. Add a source-scrape mirroring the counter one (assert each `fdev network publish` block is followed by a `./scripts/publish-readback.sh` call terminated by `|| true`, and that the count is 3), plus six one-line exit-0 cases.

**8. `main.rs:440` ↔ `publish-readback.sh:134` — the `version=<n>` stdout seam is untested on both sides.** Changing the format string to `Version: {}` survives *both* suites (`cargo test -p web-container-tool`: 4 passed; harness: 11 checks, 0 failures), while every publish would silently report "could not make sense of it" and the read-back would become a permanent no-op. The Rust tests call `parse`/`verify` directly and never exercise `inspect_state`; the shell harness stubs the tool and never asserts its argv. Neither side owns the contract. One Rust test capturing `inspect_state` stdout and asserting `^version=<n>$` closes it. Relatedly, dropping `--parameters` from `publish-readback.sh:132` — which the script's own comment at `:128-130` calls "not optional" — also survives, because the stub ignores argv.

---

### Consider

- **`main.rs:345-388`: the parser accepts trailing bytes.** `parse` never checks `off == bytes.len()`; a state with junk appended verifies and reports LANDED. Narrow, since the archive comparison is what carries the verdict, but "same version, different state" is the exact shape this script exists to detect.
- **`main.rs:411-413` uses `verify` where the contract uses `verify_strict`** (`contracts/web-container-contract/src/lib.rs:167`). The tool is strictly more permissive than the authority it claims to speak for. One-word change.
- **`LANDED` overstates what was measured.** The GET goes to the same node that just published, and freenet-core short-circuits a client GET to the local store when the node holds the contract with local interest (`crates/core/src/client_events.rs:186, 1013-1023`). So the common outcome is "my own node has version N", printed as "The network is serving version N" — and after a timeout, "It landed anyway… Do NOT republish." In the 2026-08-04 shape, propagation is exactly what is in doubt. Note the asymmetry: `NOT SEEN YET` correctly says "one GET is a sample, not a verdict"; `LANDED` carries no such caveat. Honest wording costs nothing.
- **`publish-readback.sh:158-170`: `NOT SEEN YET`'s rationale is now stale.** "Republishing on the strength of a single early read is how the same version gets signed twice" was true before this PR; with a forward-only counter a retry cannot re-use a version, so it costs one gap and nothing else. As written it may deter the correct action.
- **`publish-guards-test.sh:81-84`: `sandbox="$(mktemp -d)"` is unchecked** under `set -uo pipefail` (no `set -e`); on failure `bin` becomes `/bin` and the script `cat >"/bin/fdev"`. Harmless as non-root; CI containers often are not. `|| exit 1`.
- **`publish-guards-test.sh:192-193`: no check-count assertion** — the harness exits 0 on `failures == 0` regardless of how many assertions ran, so a deleted block passes silently. `[[ "$checks" -eq 11 ]]`.
- **`RIVER_READBACK_TIMEOUT=0` silently means 300s**, not "no wait" (`crates/fdev/src/commands.rs:64-65` ignores 0).
- **`Makefile.toml:699-704`: `publish-all` burns two versions per run** (`sign-webapp` directly, then again via `publish-river`'s dependency). Pre-existing and harmless under "gaps are fine", but the docs now tell operators to read and commit the counter, so a `+2` jump will be visible and confusing.

---

### Things checked and found clean

- **The rollback is gone from every path.** Independently enumerated: the only counter writes in the whole file are the bumps at `Makefile.toml:244` (production) and `:592` (test) plus the seed at `:588`. `grep 'prev_snapshot\|contract-version\.txt\.prev'` returns nothing repo-wide. `scripts/local-republish.sh` is a fifth publish path but uses the test contract and a `date`-derived version, touching neither counter — correctly unaffected. Five mutations re-introducing `.prev`-style rollbacks (one per site) were each killed.
- **No stub-output leakage in the harness** — the bug from #634. Both stub failure messages go to stderr and are redirected into `$log`; every `expect` string is text the script under test emits. Replacing `publish-readback.sh` with `exit 0` kills 8 of 9 classification assertions, so they are real.
- **The `inspect` parser is robust.** Built and run against 10 crafted states: empty, 4-byte, `metadata_len = 2^63`, `metadata_len = 2^64-1`, mid-payload truncation, zero-length metadata, empty archive, swapped archive. Every one is a clean `Err` with rc 1 — no panic, no attacker-sized allocation (it slices before `to_vec`), `checked_add` + `bytes.get(..)` throughout, no `unwrap`/`expect`/indexing outside `#[cfg(test)]`. `--archive-out` is correctly *not* written when verification fails. The signed message (`version.to_be_bytes() || archive`) is byte-identical across `sign_webapp:201-203`, `PackedWebApp::verify:399-413`, and the contract's `validate_state:117-118`.
- **The new CI step can go red.** `build.yml:281-289` is the last step of the `build` job: no `if:`, no `continue-on-error`, default `bash -e {0}`, script is `100755` in git. Deleting the harness gives 127; deleting `publish-readback.sh` gives 9 failures. Fails closed, does not skip-and-pass. (Two notes, neither introduced here: `main` has no branch protection, so no check is formally required; and a fast shell guard sitting behind a ~30-minute cargo chain is a guard you learn about late — moving it above the Rust steps costs nothing.)
- **Docs and code agree.** `.claude/rules/river-publish.md:94-96, 123-126` and `AGENTS.md:184-188` both say commit the counter even when the publish reports failure, and `Makefile.toml:411/451/652` print exactly that on the failure path. No contradiction of the kind #634 had. One pre-existing wrinkle: `river-publish.md:121` says "NEVER base the version on wall-clock time" while `scripts/local-republish.sh:110` does exactly that — safe (test contract, no counter), but the rule reads as absolute.
- Paths with spaces, unset/nonexistent `TMPDIR`, `fdev` absent, tool absent, garbage stdout, two `version=` lines: all handled, all exit 0.

---

### Scope

769 lines breaks down as roughly: **387 tests** (194 Rust in `main.rs:469-662`, 193 shell harness), **~120 prose** (the 39-line incident header in `publish-readback.sh`, the ~30-line counter-lifecycle comment at `Makefile.toml:206-236`, the CI step comment, the doc edits), and **~260 machinery** (~68 lines of Rust for the parser + `inspect` subcommand, ~36 net in `Makefile.toml`, ~158 non-comment lines of shell). The author's attribution is fair. Nothing needs to come out — but note that "50-100 lines" only ever described item 1: the forward-only counter is ~10 lines of deletion, and the other ~250 lines are the read-back, which is a second feature travelling with it. That is a reasonable pairing, not scope creep, and it is worth saying out loud so the next reader does not re-derive it.

Two small factual corrections for the PR description: there are **five** counter-touching sites, not four (`sign-webapp:237-245`, `sign-webapp-test:586-592`, `publish-river:371-415`, `publish-river-debug:421-455`, `publish-river-test:620-659`), and one of the four you list — `sign-webapp-test` — is a signing task, not a publish task. The code is right; the enumeration is what a reviewer checks against. Also `.gitignore:56` adds an ignore for `/test-contract/contract-version.txt.prev`, which nothing writes any more and which `.gitignore:37` already covers.

---

### Questions

1. Any reason not to wrap the `fdev` call in `timeout(1)`? It adds no gate — the timeout path already routes to `done_` and exits 0 — and it closes the only remaining way this script can affect a publish.
2. Given the `.prev` file no longer exists, what future regression is the `mv|cp` grep expected to catch? Would you take the allowlist form in Should Fix 6?
3. Is an offline check that the *committed* counter never decreases (`git show origin/main:published-contract/contract-version.txt` vs the PR's value) in scope, or deliberately deferred? It is purely local and is none of the five excluded mechanisms — and without it, "forward-only" is enforced by documentation alone against the exact reflex a failed publish invites: `git checkout` on the dirty counter file.

[AI-assisted - Claude]

sanity:
Review addressed in `0a806385`. Must Fix 1 plus Should Fix 2-5 and the `|| true` half of 7 are fixed; Should Fix 6 and 8 and the Consider items are deliberately not taken, listed below.

## Fixed

**Must Fix 1 — the read-back could hang a publish.** The `fdev` call is now wrapped in `timeout -k 5 $((readback_timeout + 10))`, falling back to a bare call where `timeout(1)` is absent, and rc 124/137 gets its own message. You are right that this matters more than its severity: `|| true` rescues a non-zero exit, not a hang, so "advisory" was only half true. The precondition is narrow but the consequence — a prompt that never returns with the counter bumped and uncommitted — is one Ctrl-C from the `git checkout` this PR exists to prevent.

**Should Fix 3 — `cmp` rc 2 rendered as FORKED.** Both operands are now checked readable and non-empty first, and rc 0 / rc 1 / rc >= 2 are three separate outcomes. A missing local archive reports as unknown and says so is a local problem.

**Should Fix 4 — fall-through to LANDED.** The two fields are validated separately rather than concatenated, and the version comparison is one explicit `gt` / `lt` / `ne` chain. LANDED is reached only by proving equality; the unreachable `ne` arm is kept so a future change that breaks the integer guarantee lands there instead of in LANDED.

**Should Fix 5 — SUPERSEDED's second cause.** It now names both: somebody else published (nothing to do), or your counter was behind, the state signed at your version was rejected, and your changes are not live. With the wording you suggested for the recovery.

**Should Fix 2 — `publish-river-debug`'s dead read-back.** All three call sites now name `release` literally instead of `${BUILD_PROFILE}`. I changed all three rather than only the broken one: the other two are safe only because they happen to run under the default profile, and a uniform rule is what makes it pinnable.

**Should Fix 7, the `|| true` half.** New scrape walks each `publish-readback.sh` invocation across its backslash continuations and requires the last line to end in `|| true`, and pins the invocation count at 3 against the `^fdev network publish` count. That also kills the "all three invocations deleted" mutation for free.

Plus a pin that the tool path is a literal profile, and one that SUPERSEDED names the rejected-publish case.

## Mutation results

Each applied, confirmed RED, restored:

| Mutation | Caught by |
|---|---|
| Drop the `timeout(1)` wrapper | `a node that never answers is abandoned, not waited on` |
| Any non-zero `cmp` treated as FORKED | `a missing local archive is reported as unknown, not as a fork` |
| Validate the integer fields concatenated | `a non-integer version is reported as unknown, never as LANDED` |
| Drop `|| true` from one invocation | `every read-back invocation is guarded by \|\| true` |
| Delete one invocation entirely | `every publish task calls the read-back` |
| `publish-river-debug` back to `${BUILD_PROFILE}` | `the read-back's tool path is a literal profile` |
| SUPERSEDED drops the rejected-publish cause | `superseded names the case where the publish was rejected` |

One of these initially passed for the wrong reason and is worth recording, since it is the same shape as the "stub-output leakage" you checked for: the broken-integer case first used a *mismatching* stub archive, so the mutant fell through to FORKED rather than LANDED and the `!= LANDED` assertion held. It now uses a matching archive, so the fall-through lands on LANDED and the assertion bites. The check that a test fails for the reason you think it does is not optional.

18 checks, 0 failures. `shellcheck -x` clean on both scripts, `cargo fmt` / `cargo clippy -p web-container-tool --all-targets` clean, `cargo test -p web-container-tool` 4 passed.

## Not taken, deliberately

- **Should Fix 6 (allowlist rollback guard).** You are right that the `mv`/`cp` grep matches syntax that no longer exists, and that `echo >` and `git checkout --` survive it. Left as-is this round to keep the change bounded.
- **Should Fix 8 (the `version=<n>` stdout seam).** Real gap — neither side owns the contract, and a format-string change would silently make the read-back a permanent no-op. Not taken this round.
- **All Consider items**, including the trailing-bytes check, `verify` vs `verify_strict`, the LANDED-overstates-propagation wording, the stale NOT-SEEN-YET rationale, the unchecked `mktemp -d` in the harness, and the check-count assertion. Noting the last two are one-liners and the `verify_strict` one is a single word.

## On your three questions

1. No reason at all — taken, as above.
2. Nothing, now that `.prev` is gone; it only kills the literal syntax it was written against. The allowlist form is the right shape. Not this round.
3. Deliberately deferred rather than judged out of scope. You are right that it is purely local and none of the five excluded mechanisms, and right that forward-only currently rests on documentation against exactly the reflex a failed publish invites.

Two corrections to the PR description taken: five counter-touching sites, not four, and `sign-webapp-test` is a signing task. The redundant `.gitignore` line for `/test-contract/contract-version.txt.prev` is left in place this round.

[AI-assisted - Claude]

sanity (commented):
## Focused re-check of the fix round — PR #635

**Not a full re-review.** This is the follow-up pass on the `+213/-23` delta from `15336762d34a83b5261773c881dca7f293ffdaa5` to **`0a806385ee908771dc55a5c314a0cf4dac6753cd`**, scoped to the four things the round set out to fix. Everything cleared in the earlier review was left alone.

**Lenses:** skeptical/bash-adversarial and testing/mutation, each blind and in its own worktree, plus my own mutation run in a third. `codex review --base origin/main` ran as the external non-Claude pass, first try, no substitution needed. Every `file:line` was verified against the canonical head via `git show 0a806385:<path>`.

All four passes independently landed on the same defect, and it was the *only* thing Codex reported:

> **[P3] Capture the GET status before negating it** — `scripts/publish-readback.sh:133-135`. When the read-back is killed by `timeout`, the `!` operator converts exit 124/137 to success before the `then` block runs, so `$?` here is always 0 and the timeout-specific branch is unreachable.

**Verdict: one small Must Fix, then merge.** Three of the four items are correct and well-pinned. The fourth — the hang — is genuinely closed by wall clock, but the operator message it was supposed to carry is dead code, and its test cannot see that.

---

### The thing you asked about: is the vacuous-test fix real?

**Yes, and the author's account is accurate.** Reproduced both forms. With the original mismatching stub archive, the concatenated-integer-guard mutant falls through to `FORKED` and the `!= LANDED` assertion holds for the wrong reason — `18 checks, 0 failures`, mutant survives. With the shipped matching archive, the same mutant reaches `LANDED.` after three `integer expression expected` errors and the assertion bites — `18 checks, 1 failures`. The fix is real and was necessary.

**But the same shape recurs in the new hang case**, which is Must Fix 1 below. That test asserts only an exit status and a wall-clock duration, never the message the code claims to print — so the branch under test can be deleted outright with the suite still green.

---

### Must Fix

**1. `scripts/publish-readback.sh:135` — `get_rc=$?` reads the negated status, so the 124/137 branch at `:136-138` never executes.**

`!` inverts the pipeline's status, so inside the `then` block `$?` is `0`, not the child's real rc:

```
$ if ! bash -c "exit 124"; then rc=$?; echo "rc=$rc"; fi
rc=0
```

`[ 0 -eq 124 ]` and `[ 0 -eq 137 ]` are therefore false unconditionally and the `else` arm always wins. Run against the shipped script with a hanging `fdev` stub and `RIVER_READBACK_TIMEOUT=1`:

```
=== post-publish read-back (advisory) ===
Could not read the network back (fdev execute get failed):

That says nothing about whether the publish landed. Re-read with
    fdev network execute get cid -o /tmp/river-state.bin
rc=0
```

The log body is **empty**, because `timeout` killed `fdev` before it wrote anything. So on the exact scenario this round was written for, the operator is told "fdev failed" and shown nothing, rather than "we gave up waiting; the publish is unaffected".

To be clear about what is and is not broken: **the hang itself is closed.** The wall clock is bounded (~11s here) and the script exits 0, and mutant *drop the `timeout(1)` wrapper* is killed by the new test. The verdict direction is also safe — "says nothing about whether the publish landed" is the honest answer either way, so this is not a misclassification. What does not work is the round's stated deliverable of a distinct timeout message, on a commit titled "bound the read-back's wall clock **and fix its classification**".

It is a Must Fix because it is three lines and because leaving it ships dead code the PR description says works. Verified fix:

```sh
get_rc=0
"${get_cmd[@]}" network execute get "$contract_id" \
        --timeout "$readback_timeout" -o "$state_file" >"$log" 2>&1 || get_rc=$?
if [ "$get_rc" -ne 0 ]; then
```

With that applied, same stub, same environment:

```
=== post-publish read-back (advisory) ===
Gave up reading the network back after 11s -- the node
did not answer. (The publish itself is unaffected by this.)
```
Harness stays at `18 checks, 0 failures`.

**The pin needs the second half.** `scripts/tests/publish-guards-test.sh:279-285` asserts `status -eq 0 && elapsed -lt 25` and nothing else. Deleting the whole 124/137 branch (`sed -i '142d;135,139d'`) leaves the suite at `18 checks, 0 failures`. Add `&& "$out" == *"did not answer"*` to that condition; it then kills the deletion.

`shellcheck` 0.9.0 is clean on this file, so static analysis does not catch the pattern either.

---

### Should Fix

**2. `scripts/tests/publish-guards-test.sh:127-128` — the `${BUILD_PROFILE}` pin is coupled to a line format its sibling pin deliberately tolerates, and the combination goes green over the exact defect this round fixed.**

The needle is `grep -n 'web-container-tool" || true'`, which requires the path and `|| true` on one physical line. The `|| true` scrape 40 lines above walks backslash continuations and accepts `|| true` on its own line — correctly, that is a legitimate format. So reformat the `publish-river-debug` invocation that way *and* revert its path to `${BUILD_PROFILE}`:

```
  "target/native/x86_64-unknown-linux-gnu/${BUILD_PROFILE}/web-container-tool" \
  || true
```

```
ok   every read-back invocation is guarded by || true
ok   every publish task calls the read-back (3 of 3)
ok   the read-back's tool path is a literal profile

18 checks, 0 failures
```

That is the dead read-back on the task that publishes to the production contract and counter, reinstated with all three pins green. Fix is to drop `" || true` from the needle: `grep -n 'web-container-tool"' "$makefile" | grep 'BUILD_PROFILE'` kills both this and the plain revert.

**3. `scripts/publish-readback.sh:127` — the new arithmetic breaks the script's own "always exits 0" promise.**

`hard_timeout=$((readback_timeout + 10))` is the first arithmetic use of `RIVER_READBACK_TIMEOUT`. Previously the value was only ever passed through as a string to `fdev --timeout`, where a bad one was caught by the failure branch and still exited 0. Under `set -u` (`:40`, no `set -e`):

```
--- RIVER_READBACK_TIMEOUT='60s' ---
publish-readback.sh: line 127: 60s: value too great for base (error token is "60s")
publish-readback.sh: line 130: hard_timeout: unbound variable
rc=1
--- RIVER_READBACK_TIMEOUT='abc' ---
publish-readback.sh: line 127: abc: unbound variable
rc=1
```

`|| true` at all three call sites means this cannot fail a publish, so impact is bounded. But the header at `:20-ish` claims it always exits 0, and what the operator sees is the advisory banner followed by two raw bash errors and no footer — `done_` never runs. That reads as "the publish tooling crashed", immediately after a publish, which is the state that invites the reflexive `git checkout` on the counter file. One line next to the existing guard at `:92`:

```sh
case "$readback_timeout" in *[!0-9]*|"") readback_timeout=60 ;; esac
```

(Related and pre-existing, noted last round: `RIVER_READBACK_TIMEOUT=0` silently means 300s, since `fdev` ignores 0.)

---

### Consider

- **`scripts/publish-readback.sh:128-131` — the `timeout(1)`-absent fallback restores the unbounded hang and says so nowhere.** Reproduced under a PATH without `timeout`: the script hung until an outer backstop killed it at 20s, having printed only the banner. Realistically unreachable, since the tool path hard-codes `x86_64-unknown-linux-gnu` so the publish host has coreutils. One `else echo` stating that the wait is bounded only by `fdev`'s own `--timeout` would close it.
- **`scripts/tests/publish-guards-test.sh:83-95` — the new awk scrape does not strip comments**, unlike the sibling scrape at `:60-62` which does and documents why. A comment containing `./scripts/publish-readback.sh` produces two false failures at once (an unguarded-invocation report and a count of 4). `Makefile.toml:227` and `:391` already reference the script and avoid this only by not writing `./`. False REDs, so safe — but the file solved this problem once already.
- **Still no check-count assertion** (`:308-309`). Deleting a check block yields `17 checks, 0 failures`, exit 0. This was a Consider last round and is on the deliberately-deferred list, which is fine — noting only that it matters slightly more now that four of the suite's checks are source scrapes whose whole value is failing when something is removed. `[[ "$checks" -eq 18 ]]` is one line.

---

### Mutation results

Run across three worktrees; the table is the union, with the survivals reproduced independently.

| Mutation | Result |
|---|---|
| Drop `\|\| true` from **one** site only (`Makefile.toml:403`, `:455`, `:655`) | **killed** each, independently — reports the invocation's start line |
| Drop `\|\| true` from all three | **killed**, all three listed |
| Delete one invocation / delete all three | **killed** by the count pin (2 of 3, 0 of 3) |
| Reformat a site to one line; move `\|\| true` to its own continuation line | no false failure; still catches a real removal afterwards |
| Plain `${BUILD_PROFILE}` revert | **killed** |
| Revert the integer guard to the concatenated `case` | **killed** (mutant reaches `LANDED.`) |
| Remove the `timeout(1)` wrapper | **killed** (exited 124 after 40s) |
| Disable the missing-archive guard | **killed** |
| Strip `YOUR CHANGES ARE NOT LIVE` from SUPERSEDED | **killed** |
| **Reformat + `${BUILD_PROFILE}` revert together** | **SURVIVED** — Should Fix 2 |
| **Delete the whole 124/137 branch** | **SURVIVED** — Must Fix 1 |
| **Delete an entire check block from the harness** | **SURVIVED** (17 checks, 0 failures) — Consider |

Needle counts are clean: `./scripts/publish-readback.sh` = 3, `^fdev network publish` = 3, `web-container-tool" || true` = 3. One per site, zero comment matches.

---

### Verified correct

- **The classification fixes are right, in both dangerous directions.** `LANDED` is now reachable only via proven version equality followed by `cmp_rc == 0`. A missing or empty local archive reports unknown and names it as a local problem, not `FORKED`. A missing tool `--archive-out` lands on `cmp` rc >= 2 (unknown), also not `FORKED` — `archive_file` is deliberately not pre-checked, which is the right call. `cmp -s` really does return 0/1/2 for same/differ/unreadable, and the `>= 2` arm was observed firing for real. Separate-field integer validation works in both directions of the concatenation trap: `('', 0)` and `(0, '')` both Skip. The `-ne` arm at `:208-213` is honestly labelled unreachable and is a reasonable defensive landing spot.
- **The Makefile change is complete and consistent.** All three sites (`:403`, `:455`, `:655`) name `release` literally. `publish-river-debug` spans `:417-464` and has no remaining `${BUILD_PROFILE}` on this path; the surviving hits at `:246`, `:470`, `:493`, `:601` are in tasks that run under the default profile. The comment's claim that a cargo-make task's `env` does not reach its dependencies was checked rather than assumed — a minimal `Makefile.toml` reproducing the dependency shape under cargo-make 0.37.24 showed the dependency seeing `BUILD_PROFILE=release` while the main task saw `dev`. So `sign-webapp:246`'s surviving interpolation is genuinely safe, not accidentally so.
- **Nothing crept in beyond the brief.** The delta touches exactly three files. `main.rs` is untouched, so the `verify`/`verify_strict` word, the trailing-bytes check and the `version=<n>` stdout seam are all unchanged; the rollback allowlist (Should Fix 6) and every Consider item are untouched. That matches what the author noted on the PR.
- 18/18 harness checks pass at this head; `shellcheck` clean on both scripts.

---

### CI

Green on `0a806385`: `build` (6m58s), `ui-playwright-tests` (13m40s), `check-delegate-migration`, `check-pointer-freshness`, `check-room-contract-migration`, CLA. The new `publish-guards-test.sh` step runs inside `build`, so the 18 checks passed in CI too. Note that a fix for Must Fix 1 changes the head SHA, so this needs re-confirming on whatever commit is actually merged.

[AI-assisted - Claude]

sanity:
Re-check addressed in `a76f554b`.

**Must Fix — `get_rc=$?` inside `if ! cmd`.** Correct, and the negation is exactly why: `$?` there reads the inverted status, so it was always 0 and the rc 124/137 branch never ran. A timed-out read reported the generic "fdev execute get failed" instead of saying the node never answered. Now `|| get_rc=$?` on the command itself, with a comment at the site naming the trap.

**And the reason it got through.** You are right that this is the more useful finding. The hang test asserted only exit status and elapsed time, so the whole 124/137 branch was deletable with the suite green — the assertion held because the code took the `else` arm, not because the behaviour was right. That is the same vacuous shape I caught in the integer test one round earlier and then reproduced immediately in the test written to cover the fix for it. The lesson is not "check the assertion" but "check *which path* makes it pass", and I did not apply it to my own new test. It now asserts on the timeout message.

**Should Fix — the defeatable profile pin.** The needle is now the tool-path argument line alone (`^[[:space:]]+"target/native/[^"]*web-container-tool"`), so the two scrapes no longer disagree about formatting. I also assert the match count is 3: a needle that stops matching passes vacuously, which is the same failure mode as the one above, and the bare needle I tried first silently picked up two unrelated lines.

**Should Fix — `RIVER_READBACK_TIMEOUT` validation.** Validated as a positive integer with a fallback to 60s and a warning. Confirmed the unvalidated form dies with `line 132: abc: unbound variable`, exit 1, no footer; the fixed form exits 0 with the warning.

**`timeout(1)`-absent fallback** left as-is with a one-line comment at the site saying why: the tool path hard-codes `x86_64-unknown-linux-gnu`, so a host without coreutils is not a case that occurs.

## Mutations

| Mutation | Result |
|---|---|
| Delete the rc 124/137 branch | RED — `a timed-out read says so, rather than reporting a generic failure` |
| Restore `if ! cmd; then get_rc=$?` | RED — same check |
| Reformat `\|\| true` onto its own line **and** revert the profile | RED — `the read-back's tool path is a literal profile` |
| Drop the timeout validation | RED — `a non-numeric read-back timeout falls back with a warning` |

**20 checks, 0 failures.** `shellcheck -x` clean on both scripts, `cargo make` parses, `cargo fmt` clean, `cargo test -p web-container-tool` 4 passed. Nothing added beyond the four items above.

[AI-assisted - Claude]

PR #661, opened 2026-08-31
feat(cli): surface member verifying keys in member list

sanity:
## Problem

River identifies members by an 8-character short id (`MemberId`'s `Display`).
That label is `truncated_base32(fast_hash(verifying_key))` — the first **40
bits** of a **64-bit non-cryptographic** polynomial hash. A *targeted*
collision on the 40-bit label costs roughly `2^40` keypair generations (hours
on a GPU), so the short id is not safe to use as a trust anchor.

This came up from a real user building a River bot that wants to accept
commands only from certain trusted members. `riverctl` gave them no
collision-proof identifier for *other* members: `member list` printed only the
short id, `ban`/`deputize` take the short id as input, and `debug` shows only
aggregate counts. `identity whoami` exposes the full key — but only for
*yourself*, which does not help an allow-list.

Every member's full ed25519 verifying key is already in room state; the CLI
just never surfaced it for anyone but you.

## Approach

Surface each member's full verifying key (base58) in `member list`, matching
the existing `identity whoami` `verifying_key` field so the two are
consistent:

- **Human output** gains an indented `key:` line per member, printed plain (no
  colour) so it copies cleanly into an allow-list.
- **JSON output** gains a `verifying_key` field: base58 when the key is known,
  or `null` when it is not (a member named only by a signed deputy grant, e.g.
  one pruned for inactivity). `null` rather than `""` so a consumer can tell
  "no key known" from an empty string it might compare against.

New `RoomDeputies::verifying_key(id)` accessor resolves the owner (kept
separately, since the owner is never in `members.members`) and current members,
and returns `None` for an id with no key in state — never a wrong key.

The full key is the collision-proof identity every signature check already
verifies against, so anything making a trust decision about a member should
compare **it**, not the 40-bit label. (Also fixes a misleading comment on
`MemberId` that claimed a collision would take ~3x10^59 years.)

No wire-format, contract, or delegate change — this is a read-only CLI output
addition, additive to the JSON shape.

## Testing

- `RoomDeputies::verifying_key` resolves owner + members and is `None` for an
  unknown id.
- `member_list_json` includes `verifying_key` (base58 present, and `null` when
  absent); the pre-existing pinned JSON shape test still passes (field is
  additive).
- Full `riverctl` lib suite green (`cargo test -p riverctl`), fmt + clippy
  clean (no new warnings).

[AI-assisted - Claude]

### DIFF
--- a/cli/src/commands/member.rs
+++ b/cli/src/commands/member.rs
@@ -127,7 +127,19 @@ pub async fn execute(command: MemberCommands, api: ApiClient, format: OutputForm
                     let party = deputies.party(id);
                     let granted_by: Vec<MemberId> =
                         deputized_by.get(&id).cloned().unwrap_or_default();
-                    (party, deputies.deputies_of(id).to_vec(), granted_by)
+                    // The full verifying key — the collision-proof identity to
+                    // compare when trusting a member (the `member_id` label is a
+                    // 40-bit truncation). `None` for a member with no key in
+                    // state (e.g. a pruned deputy named only by a signed grant).
+                    let verifying_key = deputies
+                        .verifying_key(id)
+                        .map(|vk| bs58::encode(vk.as_bytes()).into_string());
+                    (
+                        party,
+                        deputies.deputies_of(id).to_vec(),
+                        granted_by,
+                        verifying_key,
+                    )
                 })
                 .collect();
 
@@ -137,7 +149,7 @@ pub async fn execute(command: MemberCommands, api: ApiClient, format: OutputForm
                         println!("No members found in room.");
                     } else {
                         println!("\n{} member(s) found:\n", members.len());
-                        for (party, _own_deputies, granted_by) in &members {
+                        for (party, _own_deputies, granted_by, verifying_key) in &members {
                             // Escaped and quoted: an unescaped nickname can
                             // forge a row that reads as a real deputy grant,
                             // and `colored` drops the colour that would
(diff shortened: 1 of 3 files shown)


sanity:
## Multi-perspective review (2 independent Claude lenses)

Light tier + security-relevant (users will allow-list on this output), so two blind adversarial reviewers ran on the diff: **skeptical** (bug hunt) and **code-first** (read code before description). Per the external-model rule, Codex is opt-in for freenet and was not run; both lenses here are Claude.

**No High/Medium bugs.** Confirmed: `verifying_key` never returns a wrong key on verified state (owner branch is sound; member branch reads `member.member_vk` whose `id()` *is* that key's hash); the base58 encoding is byte-for-byte identical to `identity whoami`'s `verifying_key`, so the two are directly comparable; no privacy leak (public key already in state); JSON field is additive; the new tests are non-vacuous.

Findings addressed:

1. **Misleading `None`-branch comment** (code-first, low) — the reviewers disagreed on whether `None` is reachable in the `member list` path. Reconciled against `MemberInfoV1::verify` (`common/src/room_state/member_info.rs:355-357`): it rejects a non-owner `member_info` record whose member isn't current, so every id `member list` iterates resolves to `Some` on verified state. Fixed both the inline comment and the accessor rustdoc to describe this accurately; the `None` arm is defensive there. (commit e35fe671)

2. **Per-row `key:` line doubles human output** (code-first, design call) — kept **on by default deliberately**: the whole point is to make the collision-proof identity discoverable so people stop trusting the 40-bit short id; JSON (`-f json | jq -r '.[].verifying_key'`) is the tooling path. Documented as a deliberate choice in-code. (commit e35fe671)

3. **`member deputies` / `deputized-by` still show only short ids** (code-first, scope) — a genuine follow-up gap (those commands answer the same "who do I trust" question). Left out of this PR to keep scope tight; will fold into the UI/VK follow-up.

4. **`null` vs `""` for absent key** — intentional and documented: absence of a key is meaningful and must not compare-equal to an empty string a consumer might construct.

Skeptical's optional suggestion (a test for the lingering-`member_info` → `None` case) was skipped: that case requires unverified state, and the branch is already covered by the `stranger` → `None` case.

[AI-assisted - Claude]

PR #667, opened 2026-08-31
feat(cli): accept full verifying keys as member references; author key in message list

sanity:
## Problem

Follow-up to #661 (member-list verifying keys), from a bot author who needs to
identify members by a collision-proof key rather than the 8-char short id (a
40-bit truncation two members can share). Two gaps remained:

1. Member-targeting commands (`ban`, `deputize`, `revoke-deputy`, `deputies`,
   `deputized-by`) only accepted the short id as input — so even after you read
   a member's full key from `member list`, you couldn't *act* on them by it.
2. `message list` showed only the short id for a message's author, so a bot
   reading messages had no collision-proof way to check who sent a command.

## Approach

**Accept full verifying keys as member references.** New shared helper
`deputies::member_id_from_key_input` returns the derived `MemberId` when the
input parses as a full base58 32-byte ed25519 key, else `None`. All three
resolution sites use it as an exact-match fast path before the existing
prefix/first-8 short-id matching:
- `RoomDeputies::resolve_short_id` (drives `deputies` / `deputized-by`)
- `resolve_deputy_target` (drives `deputize` / `revoke-deputy`)
- `ban_member_with_safety`'s inline find

A full-key input names exactly one member, so it also satisfies
`--require-exact-member-id` by construction (the guard only rejects a non-exact
*short-id* match). Short-id behavior is unchanged. Help text and the ambiguity
error now mention the full-key option.

**Author key in `message list`.** `--format json` gains `author_verifying_key`
(base58, or `null` when the author is neither owner nor a current member — e.g.
a since-departed author of an older message, whose key is no longer in state).
JSON only: a bot allow-lists on this key, while the human chat view stays terse.

Bumps riverctl to 0.2.14. No wire-format, contract, or delegate change.

## Testing

- `member_id_from_key_input` parses full keys and rejects short ids / empty /
  non-base58 / wrong-length blobs.
- `resolve_short_id` and `resolve_deputy_target` resolve a member (and, for
  revoke, a pruned deputy) by full key, and return NotFound/None for a
  stranger's key — never a wrong member.
- Full `riverctl` lib suite green (351 tests), fmt + clippy clean (no new
  warnings).

[AI-assisted - Claude]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4782,7 +4782,7 @@ dependencies = [
 
 [[package]]
 name = "riverctl"
-version = "0.2.13"
+version = "0.2.14"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 All notable changes to riverctl will be documented in this file.
 
+## [0.2.14] - 2026-08-31
+
+### Added
+- Member-targeting commands (`member ban`, `member deputize`,
+  `member revoke-deputy`, `member deputies`, `member deputized-by`) now accept a
+  member's full base58 verifying key in place of the 8-character short ID. A
+  full key names exactly one member (the short ID is a 40-bit truncation that
+  two members can share), so it resolves unambiguously and satisfies
+  `--require-exact-member-id` by construction.
+- `message list --format json` now includes an `author_verifying_key` field
+  (base58, or `null` when the author is no longer in room state), so a bot can
+  identify a message's author by their collision-proof key.
+
 ## [0.2.13] - 2026-08-31
 
 ### Added
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "riverctl"
-version = "0.2.13"
+version = "0.2.14"
 edition = "2021"
 authors = ["Freenet Project"]
 description = "Command-line interface for River decentralized chat on Freenet"
--- a/cli/README.md
+++ b/cli/README.md
@@ -47,8 +47,9 @@ End-to-end-encrypted one-to-one messages between two members of the same room.
 They travel inside the room's contract state, so both people must already be
 members — there is no cross-room DM.
 
-The recipient is a member ID: either the short 8-character form that
-`riverctl member list` prints, or the full ID.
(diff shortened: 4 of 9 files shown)


sanity:
## Multi-perspective review (2 independent Claude lenses)

Security-sensitive (moderation input parsing), so two blind adversarial reviewers ran on the diff: **skeptical** (attack the resolution + `--require-exact` interaction) and **code-first** (read code before description). External model (Codex) is opt-in for freenet and not run; both lenses are Claude.

**No bugs found (no High/Medium).** Skeptical verified: no non-key input can resolve to a wrong member (key and short-id spaces are disjoint by length + ed25519 point validation), the `--require-exact` skip is safe (the full-key path is strictly more precise than the guard's own 40-bit check), no regression for existing short-id inputs (a 44-char key never matched before), ban still runs every authority/safety gate on the resolved id, and the tests are genuine.

Findings addressed (commit `82cf8278`):

1. **Missed resolution site: `dm.rs`** (code-first) — extended the full-key fast path to `resolve_recipient_vk`, so `dm send` / `dm list --with` accept a full key too. Added a test. It was the one member-reference input site left on prefix-only.
2. **`message list` duplicated `RoomDeputies::verifying_key`** (code-first) — now reuses the shared helper (whose owner/member/None behaviour is unit-tested), removing the drift risk.
3. **Ban path untested** (code-first) — ban now resolves via the shared, unit-tested `resolve_deputy_target` (no deputy fallback) instead of an inline find, so its key/short-id resolution is covered by that function's tests. Added a source-scrape pin that `message list` emits `author_verifying_key` via the shared helper.
4. **"collision-proof" wording** (skeptical L1) — tightened to state resolution is via the 64-bit `MemberId`, which is the room's own keying precision (two members can't occupy one `MemberId`), not literally the 32-byte key. Also fixed a doc-comment mis-attachment the helper insertion had caused.
5. **README** (code-first) — documented the full-key inputs, the `member list` `verifying_key` field (undocumented since #661), and `message list`'s `author_verifying_key`; removed the ambiguous "full ID" phrasing in the DM section.

Skeptical L2 (`.trim()` only on the key path) confirmed harmless and left as-is. 353 lib tests pass, fmt + clippy clean.

The delta since review is additive and follows the reviewed pattern (the dm fast path is the same helper + exact-match shape already validated for the member commands; the ban change swaps an inline matcher for the already-tested resolver).

[AI-assisted - Claude]

PR #668, opened 2026-08-31
feat(ui): show member's full verifying key in the member-info modal

sanity:
## Problem

The River member-info modal identified a member only by the 8-character short
Member ID — a 40-bit truncation of a non-cryptographic hash that two members can
share, so it's unsafe to trust for "is this really who I think it is?". The full
verifying key was already in state but never shown in the UI. This is the UI
counterpart to #661 / #667, which exposed the same key in `riverctl`.

## Approach

Add a read-only **Verifying key** field to the member-info modal, directly below
Member ID, showing the member's full base58 ed25519 key:
- the room owner key when the member is the owner (who isn't in the members
  list),
- the member's own `member_vk` otherwise.

It mirrors the existing read-only key field in `edit_room_modal` (same
`bs58::encode(vk.as_bytes())` + `readonly: true` input pattern), and is pinned in
the `volatile_value_binding_audit` as a display-only control (so it can't
silently drop out of the audit). `data-testid="member-info-verifying-key-input"`
for automation.

No contract or delegate WASM change — this is UI only. It does require a **Freenet
webapp republish** (`publish-river`) to reach users; that's a separate deploy
step, not part of this PR.

## Testing

- `cargo test -p river-ui --bins` green (932 tests), including the updated
  `volatile_value_binding_audit` (new binding pinned as display-only).
- `cargo check -p river-ui --target wasm32-unknown-unknown --features no-sync`
  compiles.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components.rs
+++ b/ui/src/components.rs
@@ -118,6 +118,7 @@ mod volatile_value_binding_audit {
         r#"components/members/invite_member_modal.rs <input> invitation_code"#,
         r#"components/members/invite_member_modal.rs <input> invitation_url"#,
         r#"components/members/member_info_modal.rs <input> "{member_id_str}""#,
+        r#"components/members/member_info_modal.rs <input> "{vk_str}""#,
         r#"components/room_list/edit_room_modal.rs <input> "{bs58::encode(room_data.owner_vk.as_bytes()).into_string()}""#,
         r#"components/room_list/edit_room_modal.rs <input> "{room_data.contract_key.id()}""#,
         r#"components/room_list/edit_room_modal.rs <input> "{secret_version}""#,
--- a/ui/src/components/members/member_info_modal.rs
+++ b/ui/src/components/members/member_info_modal.rs
@@ -425,6 +425,18 @@ pub fn MemberInfoModal() -> Element {
         // Get the member ID string to display
         let member_id_str = member_id.to_string();
 
+        // The member's full ed25519 verifying key (base58) — the collision-proof
+        // identity, matching what riverctl's `member list` / `identity whoami`
+        // print as `verifying_key`. The short "Member ID" above is only a 40-bit
+        // truncation two members can share; this key names exactly one member,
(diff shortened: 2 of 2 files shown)


sanity:
## Review (adversarial Claude lens)

Light tier (small, self-contained UI display field) but user-facing, so one blind adversarial reviewer attacked the VK resolution, Dioxus signal-safety, base58 parity, 320px layout, and the audit pin.

**Clean — no High/Medium findings.** Verified:
- **VK resolution is sound.** `is_owner` ⇒ `owner_key_signal` is `Some` and the owner's `MemberId` equals `member_id`, so showing the room key is correct; non-owner ⇒ the earlier "not found and not owner" early-return guarantees `member` is `Some`, and `member_vk` hashes to `member_id`. No panic path (no `unwrap`; both branches degrade to `None` → the field just doesn't render).
- **Exact base58 parity with riverctl** (`bs58::encode(vk.as_bytes())`, same owner-first precedence as `RoomDeputies::verifying_key`), so the UI and CLI show the identical key string for the same member.
- **Dioxus-safe:** only new reads are the infallible `CURRENT_ROOM` memo and a local ref; no fallible `ROOMS` read, no mutation, no `oninput`; literal `readonly: true` is exactly the volatile-value exemption.
- **No 320px overflow:** byte-identical classes to the existing Member ID field, and `edit_room_modal` already renders a ~44-char base58 key with the same pattern under the Playwright suite.
- **Audit pin matches** what the scanner extracts (ran `the_audit_still_sees_every_known_binding` → ok).

One low, non-blocking note: a 64-bit owner/member `MemberId` collision (~1e-12/member) would show the owner's key for that member — but this is pre-existing (the whole modal keys on the truncated id) and riverctl uses the same owner-first precedence, so the two stay consistent. No change needed.

No contract/delegate WASM change. Reaching users needs a separate `publish-river` webapp deploy.

[AI-assisted - Claude]

PR #670, opened 2026-09-06
fix(ui): gate outbound-DM save on delegate hydration to stop truncating saves

### LINKED ISSUE #530: fix(dm): a save before the delegate blob hydrates persists a truncated store, permanently losing archives and outbound-DM plaintexts
## Summary

`do_save_outbound_dms_to_delegate` serialises the **entire current** `HIDDEN_DM_THREADS` and `OUTBOUND_DMS` signals into a full-overwrite `StoreRequest`, with no guard that those signals have finished hydrating from the delegate. Any save that runs before hydration completes persists a truncated blob, permanently destroying every archive entry and every cached outbound-DM plaintext that had not yet loaded.

Found by two independent reviewers during the #526 review (PR #529). Distinct mechanism from #526, but it reproduces two of that issue's three reported symptoms — "many threads un-archive together" and "archives are gone after a reload" — so #526 being fixed does not fully close the user-visible complaint.

## Mechanism

`do_save_outbound_dms_to_delegate` (`ui/src/components/app/chat_delegate.rs`) snapshots both signals and writes them wholesale:

```rust
let mut entries: Vec<OutboundDmEntry> = cache.by_token.values().cloned().collect();
let mut hidden_threads = { HIDDEN_DM_THREADS.read().values().cloned().collect::<Vec<_>>() };
```

There is no "hydrated yet" latch. On cold start both signals are empty until the `OUTBOUND_DMS_STORAGE_KEY` `GetResponse` lands.

Reviewers traced all save call sites. The ones that run inside or after hydration are safe:

- `persist_sanitized_dm_store_after_tombstone_filter` runs inside the hydration `defer`, after the `with_mut`.
- `prune_outbound_dms_for_purges` early-returns on `to_remove.is_empty()`.
- The #414 identity-replace prune early-returns on `!has_entries && !has_hidden`.
- The legacy-migration save is ordered behind the hydration defers (both `defer` and `safe_spawn_local` route through `setTimeout(0)`, so they run FIFO).

The three **user-initiated** saves are reachable pre-hydration: `hide_dm_thread`, `unhide_dm_thread`, and `save_outbound_dm`.

## Failure scenario

1. User opens River. The rail renders with `HIDDEN_DM_THREADS` still empty, so every archived thread is visibly **un**-archived.
2. Seeing a thread they thought they had archived, the user clicks Archive (a very likely response to the state the bug itself creates).
3. `hide_dm_thread` queues a save. The blob written contains **only that one entry** — every other archive is gone from the delegate.
4. Hydration completes and repopulates memory correctly, but queues no save afterwards, so the blob stays truncated.
5. Next reload hydrates from the truncated blob. The archives are permanently lost.

The same shape applies to `save_outbound_dm`: sending a DM before hydration truncates `entries`, losing the cached plaintexts for every previously-sent DM — which surfaces as sent DMs rendering as the legacy `"sent — ciphertext only"` placeholder.

## Suggested fix

A hydration-complete latch gating `do_save_outbound_dms_to_delegate`: skip (or defer) the write until the `OUTBOUND_DMS_STORAGE_KEY` `GetResponse` has been processed, so a pre-hydration mutation cannot publish a truncated view. Because the store is a single blob shared by archives and outbound plaintexts, one latch covers both.

Worth checking whether the same shape exists for the per-room `room:<vk>` saves.

## Related

- #526 / PR #529 — archives destroyed by sync churn. Fixed separately; this path is untouched by that fix.
- The blob is `OutboundDmStore` under `OUTBOUND_DMS_STORAGE_KEY`; see `.claude/rules/direct-messages.md`.

[AI-assisted - Claude]

sanity:
## Problem

`do_save_outbound_dms_to_delegate` serialises the **entire current**
`HIDDEN_DM_THREADS` and `OUTBOUND_DMS` signals into a full-overwrite
`StoreRequest`, with no guard that those signals have finished hydrating
from the delegate. On cold start both signals are empty until the
`OUTBOUND_DMS_STORAGE_KEY` `GetResponse` from the current delegate lands.

Any of the three user-initiated saves (`hide_dm_thread`, `unhide_dm_thread`,
`save_outbound_dm`) reachable before that response arrives persists a
truncated blob, permanently destroying every archive entry and cached
outbound-DM plaintext that hadn't loaded yet.

## Approach

**Final design (after two review-driven redesigns — see the review comments
below for what changed and why):**

`save_outbound_dms_to_delegate` now unconditionally **awaits** a hydration
latch (`OUTBOUND_DMS_HYDRATED`) before proceeding to the real coalesced
save — there is no early-return path. The wait is bounded
(`await_flag_with_bound`, polling every 50ms up to 300 times / 15s) so a
lost `GetResponse` degrades to a bounded wait (logged) instead of hanging
forever. `Ok(())` from this function therefore always means the write was
genuinely attempted — this matters beyond user-triggered saves:
`LiveImportSeam::flush` (`freenet_api/delegate_migration.rs`) treats `Ok`
from this function as proof of durability before sealing a crate-migration
predecessor as done, so an early no-op `Ok` would have silently broken that
contract too.

The CURRENT (non-legacy) delegate's `OUTBOUND_DMS_STORAGE_KEY`
`GetResponse` handler (`handle_outbound_dms_get_response` in
`response_handler.rs`) sets the latch via `mark_outbound_dms_hydrated()` on
every return path (no blob, parsed OK, parse failure) — always scheduled
via `crate::util::defer`, never called synchronously, so the latch can only
flip to `true` AFTER `hydrate_hidden_dm_threads` / `hydrate_outbound_dms_cache`'s
own deferred signal merges have actually run (macrotask FIFO ordering
guarantees the ordering). A LEGACY delegate's response never sets the
latch — its data still needs merging into the current baseline first.

I also checked whether the same full-overwrite shape exists for the
per-room `room:<vk>` saves (as the issue suggested). It doesn't: those go
through a per-key read-merge-write CAS (`do_save_rooms_to_delegate`), so a
save before a room's own state is loaded simply never touches that room's
key rather than truncating a shared blob. No changes needed there.

## Testing

- `await_flag_with_bound_returns_immediately_when_already_set` /
  `_returns_once_flag_set_mid_wait` / `_gives_up_after_max_polls_when_never_set`:
  the generic bounded-poll primitive, with small/fast test parameters.
- `mark_outbound_dms_hydrated_sets_the_latch`: the latch itself.
- `outbound_dms_save_awaits_hydration_unconditionally_before_coalescing`:
  source-order pin that `save_outbound_dms_to_delegate` has no early-return
  or conditional bypass and awaits hydration before the real network call
  (the actual delegate I/O isn't unit-testable off-WASM — no `WEB_API` /
  Dioxus runtime in native tests).
- `outbound_dms_hydration_latch_is_marked_on_every_current_delegate_return_path`:
  pins that the latch is scheduled via `crate::util::defer` (never called
  bare/synchronously) on exactly the 3 CURRENT-delegate return paths, each
  gated by `!is_legacy_delegate`.

`cargo test -p river-ui --bins`: 946 passed, 0 failed.
`cargo fmt` / `cargo clippy -p river-ui --bins --tests`: clean (no new
warnings in the touched files).

Only `ui/` files changed — no delegate/contract WASM touched, no migration
needed.

## Review

Full-tier multi-perspective review ran across three rounds as the design
was hardened in response to findings — see the review comments on this PR
for the detailed history. External model (`codex review`) was unavailable
(usage limit, resets 2026-09-07) after one retry; substituted with four
parallel Claude reviewer lenses (code-first, testing, skeptical,
big-picture) per the project's review-fallback policy, each round.

Closes freenet/river#530

[AI-assisted - Claude]

### DIFF
--- a/.claude/rules/dioxus-signal-safety.md
+++ b/.claude/rules/dioxus-signal-safety.md
@@ -295,3 +295,44 @@ re-renders when the signal driving its open/close state changes), or
 reset per-open `use_signal` scratch state with a `use_effect` keyed on
 that open/close signal. freenet/river#291 (the invite-via-DM picker
 showing the previous invitee's name) was exactly this bug.
+
+## A readiness latch must flip from inside the same macrotask as the work it certifies
+
+A boolean "is X ready" flag that gates a later read of some state MUST be
+set from WITHIN the same deferred macrotask that actually performs the
+write making X ready — never set synchronously alongside code that only
+SCHEDULES that write via `defer()`/`safe_spawn_local()`.
+
+`defer()`'s `setTimeout(0)` guarantees FIFO ordering only among macrotasks
+registered against the SAME event-loop queue. It says nothing about a
+MICROTASK-resumed continuation (an already-suspended `.await` on a channel,
+a promise, another future) — that can run in between "a macrotask was
+registered" and "that macrotask actually executed," because microtasks
+drain before the next macrotask, not before the current synchronous call
+returns. Setting the flag synchronously, "because anything checking it must
+run later," conflates "runs later" with "runs after every already-scheduled
+macrotask" — those are not the same thing.
+
+The fix: schedule the flag-flip itself via `defer()`, positioned textually
+AFTER the `defer()` calls whose work it depends on, in the same synchronous
+function. Then the flag literally cannot become `true` until those merges
+have executed — the invariant is structural (a precondition of the flip
+existing at all), not an argument about scheduling order that has to be
+re-derived by every future reader.
+
+freenet/river#530 (PR #670, round 2 of review) hit exactly this: an
+outbound-DM hydration latch was set synchronously in
+`handle_outbound_dms_get_response`, in the same call as merely scheduling
+(not performing) the disk-baseline merge. `LiveImportSeam::flush` — already
+suspended on an unrelated `.await` — could resume via a microtask, observe
+the latch as `true`, and read the signals before the merge had landed,
+truncating the saved blob. See `direct-messages.md`'s "Hydration gate"
+section for the concrete fix and `chat_delegate.rs`'s
+`mark_outbound_dms_hydrated` doc comment for the full reasoning.
+
+**Test consequence:** a source-scrape pin that only checks a flag-setting
+call EXISTS and is gated correctly is not enough — also check it is
+scheduled via `defer()` (not called bare) AND that, textually, it comes
+AFTER the `defer()` calls for the work it depends on. A pin that misses the
+ordering check passes even after the ordering is silently broken by a later
+edit (e.g. hoisting the flag-flip above the merge calls "for readability").
--- a/.claude/rules/direct-messages.md
+++ b/.claude/rules/direct-messages.md
@@ -146,7 +146,17 @@ plaintext in the chat delegate.
   failure). `Ok(())` from `save_outbound_dms_to_delegate` always means the
   write was actually attempted — this is load-bearing for
   `LiveImportSeam::flush` in `freenet_api/delegate_migration.rs`, which seals
-  a predecessor as migrated on `Ok`.
+  a predecessor as migrated on `Ok`. **The latch flip itself must go through
+  `crate::util::defer`, never a bare synchronous call** — it is registered
+  AFTER `hydrate_hidden_dm_threads` / `hydrate_outbound_dms_cache` in the
+  same synchronous handler call, so `setTimeout(0)` FIFO ordering guarantees
+  it can't run (and the latch can't become `true`) until those two have
+  already executed their merges. A synchronous flip let a MICROTASK-resumed
+  caller (e.g. `LiveImportSeam::flush`, which awaits this function directly)
+  observe the latch as `true` before the merge had landed — reproducing the
+  truncation bug in a rarer, timing-dependent form (round-2/3 review on
+  PR #670; see `dioxus-signal-safety.md`'s readiness-latch rule for the
+  general pattern).
 - **Prune path**: `prune_outbound_dms_for_purges` (UI) and
   `prune_outbound_cache_for_room` (CLI) act ONLY on entries whose
   `(room, recipient, token)` appears in some recipient's
--- a/.claude/rules/direct-messages.md
+++ b/.claude/rules/direct-messages.md
@@ -135,6 +135,18 @@ plaintext in the chat delegate.
   delegate writes. `save_outbound_dms_to_delegate` and
   `save_rooms_to_delegate` both share the helper — grep `CoalesceState`
   to find every caller.
+  **Hydration gate (freenet/river#530):** before that, `save_outbound_dms_to_delegate`
+  awaits the `OUTBOUND_DMS_HYDRATED` latch (bounded by `await_flag_with_bound`,
+  15s worst case), so a save can never read `OUTBOUND_DMS`/`HIDDEN_DM_THREADS`
+  before they've merged whatever the CURRENT delegate already has on disk —
+  otherwise the save's full-blob overwrite would truncate it. The latch is set
+  by `mark_outbound_dms_hydrated()`, called only from the CURRENT (non-legacy)
+  delegate's `OUTBOUND_DMS_STORAGE_KEY` `GetResponse` handler in
+  `response_handler.rs`, on every return path (no blob / parsed / parse
+  failure). `Ok(())` from `save_outbound_dms_to_delegate` always means the
+  write was actually attempted — this is load-bearing for
+  `LiveImportSeam::flush` in `freenet_api/delegate_migration.rs`, which seals
+  a predecessor as migrated on `Ok`.
 - **Prune path**: `prune_outbound_dms_for_purges` (UI) and
   `prune_outbound_cache_for_room` (CLI) act ONLY on entries whose
   `(room, recipient, token)` appears in some recipient's
--- a/ui/src/components/app/chat_delegate.rs
+++ b/ui/src/components/app/chat_delegate.rs
@@ -3927,6 +3927,121 @@ mod tests {
         );
     }
 
+    // ===== A save before hydration must not truncate the delegate blob
+    // (freenet/river#530) =====
+    //
+    // `do_save_outbound_dms_to_delegate` snapshots and fully overwrites the
+    // `OutboundDmStore` blob. Before the CURRENT delegate's
+    // `OUTBOUND_DMS_STORAGE_KEY` `GetResponse` lands, `OUTBOUND_DMS` /
+    // `HIDDEN_DM_THREADS` haven't merged whatever is already on disk, so a
+    // save that races ahead of hydration would persist a truncated view and
+    // permanently destroy every entry the in-memory signals don't know about
+    // yet. `save_outbound_dms_to_delegate` now gates on
+    // `OUTBOUND_DMS_HYDRATED` and records a pending save instead;
+    // `mark_outbound_dms_hydrated` (called from the current-delegate
+    // `GetResponse` handler in `response_handler.rs`) drains it once the
+    // disk baseline is known.
+
+    /// Serializes the tests below that mutate the process-global
+    /// `OUTBOUND_DMS_HYDRATED` / `OUTBOUND_DMS_SAVE_PENDING_PRE_HYDRATION`
+    /// statics, mirroring `DEDUP_TEST_LOCK`'s rationale.
+    static OUTBOUND_DMS_HYDRATION_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+    /// A save requested before hydration must be a harmless no-op — it must
+    /// NOT reach `do_save_outbound_dms_to_delegate` (the real network write) —
+    /// and must record that a save is owed, or the mutation it was meant to
+    /// persist is silently lost forever.
+    #[test]
+    fn outbound_dms_save_defers_before_hydration_and_marks_pending() {
+        let _guard = OUTBOUND_DMS_HYDRATION_TEST_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
+        reset_outbound_dms_hydration_state_for_test();
+
+        let result = futures::executor::block_on(save_outbound_dms_to_delegate());
+        assert!(
+            result.is_ok(),
+            "a pre-hydration save must be a harmless no-op, not an error: {:?}",
+            result
+        );
+        assert!(
+            OUTBOUND_DMS_SAVE_PENDING_PRE_HYDRATION.load(Ordering::SeqCst),
+            "a save requested before hydration must be recorded as owed"
+        );
+
+        reset_outbound_dms_hydration_state_for_test();
+    }
+
+    /// Once hydration completes, a save that was deferred pre-hydration must
+    /// be drained — otherwise a user action taken before hydration (e.g.
+    /// archiving a thread) is remembered in memory but never reaches the
+    /// delegate, and the next reload reads the stale disk copy.
+    #[test]
+    fn mark_hydrated_drains_a_pending_pre_hydration_save() {
+        let _guard = OUTBOUND_DMS_HYDRATION_TEST_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
+        reset_outbound_dms_hydration_state_for_test();
+        OUTBOUND_DMS_SAVE_PENDING_PRE_HYDRATION.store(true, Ordering::SeqCst);
+
+        mark_outbound_dms_hydrated();
+
+        assert!(
+            OUTBOUND_DMS_HYDRATED.load(Ordering::SeqCst),
+            "the hydration latch must be set"
+        );
+        assert!(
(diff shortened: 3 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #670 (Round 1, on commit `5985e83d`)

### Summary
- **PR Title:** fix(ui): gate outbound-DM save on delegate hydration to stop truncating saves
- **Type:** fix
- **CI Status:** pending at review time
- **Linked Issues:** Closes freenet/river#530
- **Review tier:** Full (persistence/durability + async-ordering reasoning)
- **Reviewers run:** `freenet:code-first-reviewer`, `freenet:testing-reviewer`, `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer` (all Claude, background, parallel). External model: `codex review --base origin/main` attempted twice, both failed with a hit usage-limit error (resets 2026-09-07, outside a practical wait window for this session) — substituted with the four-lens Claude pass above per the multi-model-review fallback policy.

---

### Findings (on commit `5985e83d`)

| Finding | Severity | Source | Verified |
|---|---|---|---|
| `save_outbound_dms_to_delegate`'s early `Ok(())` (deferred pre-hydration) breaks `LiveImportSeam::flush`'s durability contract in `freenet_api/delegate_migration.rs` — `flush` seals a crate-migration predecessor as `Done { had_data: true }` on `Ok`, which would fire even when the write was skipped, reproducing the exact "sealed Done over data never saved" bug `merge_outbound_dms`'s own comment (lines ~816-837) was written to prevent | **Must Fix** | code-first-reviewer | Yes — read `flush()`/`merge_outbound_dms` directly |
| If the current delegate's `OUTBOUND_DMS_STORAGE_KEY` `GetResponse` never arrives, every subsequent save silently no-ops for the rest of the session with no user-visible signal | **Must Fix** | skeptical-reviewer (independently corroborated by code-first-reviewer) | Yes — traced `fire_load_outbound_dms_request`'s single-fire, no-timeout design |
| The `mark_outbound_dms_hydrated()` call-count pin test (`outbound_dms_hydration_latch_is_marked_on_every_current_delegate_return_path`) only counted call occurrences — it would still pass if the `!is_legacy_delegate` guard were inverted or removed | Should Fix | testing-reviewer | Yes — confirmed the test logic doesn't inspect the guard |
| `.claude/rules/direct-messages.md`'s "Save path" bullet didn't mention the new hydration latch | Should Fix | big-picture-reviewer | Yes |
| No test proving the actual merge-then-save interaction (vs. flag bookkeeping) | Consider | testing-reviewer | N/A — superseded by the redesign below |

**What was verified safe:** the macrotask-FIFO ordering claim in the original design (traced by code-first, skeptical, and testing-reviewer independently — all confirmed `defer()`/`safe_spawn_local()`'s `setTimeout(0)` FIFO ordering held); no double-drain/lost-wakeup race on the pending flag; no cross-tab issues; the per-room-saves claim ("no changes needed there" — verified by reading `do_save_rooms_to_delegate` directly, confirmed genuine per-key CAS, not a full-blob overwrite); no scope creep; no removed/weakened tests.

### Resolution (commit `dbe92661`)

Both Must Fix findings share a root cause: the "defer + replay + pending-flag" design let `save_outbound_dms_to_delegate` return `Ok(())` for a write that never happened, and had no bound on how long it could stay silent. Replaced with a bounded **await**: `save_outbound_dms_to_delegate` now polls the hydration latch (yielding via `crate::util::sleep` between checks, extracted as a generic testable `await_flag_with_bound` helper) for up to 15s, then proceeds **unconditionally** to the real coalesced save — no early-return shortcut exists anymore. This means:

- `Ok(())` now always means the write was attempted, so `LiveImportSeam::flush`'s durability contract holds.
- A lost `GetResponse` degrades to a bounded 15s wait (with a logged warning) instead of permanent silent no-ops.
- The macrotask-FIFO ordering argument the original design needed is no longer load-bearing (documented in the updated `handle_outbound_dms_get_response` doc comment) — a real `.await` yield naturally lets any pending hydration macrotasks complete first, so there's less subtle async reasoning for future readers to re-verify.

Also fixed: the pin test now verifies each `mark_outbound_dms_hydrated()` call is immediately preceded by the `!is_legacy_delegate` guard (not just counted), and `.claude/rules/direct-messages.md` documents the hydration gate and its `LiveImportSeam::flush` interaction.

`cargo test -p river-ui --bins`: 946 passed, 0 failed. `cargo fmt` / `cargo clippy -p river-ui --bins --tests`: clean, no new warnings.

A fresh review round is running against `dbe92661` to confirm the redesign and check for anything the rewrite introduced.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #670 (Round 2, on commit `dbe92661`)

### Summary
- **Review tier:** Full
- **Reviewers run:** `freenet:code-first-reviewer`, `freenet:testing-reviewer`, `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer` (Claude, background, parallel). External model still unavailable (see round-1 note); same four-lens substitution.

### Findings (on commit `dbe92661`)

| Finding | Severity | Source | Verified |
|---|---|---|---|
| Both round-1 Must-Fix findings genuinely resolved: no early-return path remains in `save_outbound_dms_to_delegate`, and `LiveImportSeam::flush` now gets a truthful `Ok`/`Err` in every case | — (confirmation) | code-first-reviewer, big-picture-reviewer | Yes |
| **The hydration latch's fast path (`await_flag_with_bound` returning immediately with zero yields when already `true`) could be observed as `true` by a caller resumed via a MICROTASK before the disk-baseline merge macrotasks (`hydrate_hidden_dm_threads_now`/`hydrate_outbound_dms_cache_now`, scheduled via `crate::util::defer`) had actually executed** — because `mark_outbound_dms_hydrated()` set the latch *synchronously*, in the same call as scheduling (not performing) those merges. The doc comment's reasoning ("yielded back to the event loop" implies "after every scheduled macrotask") conflated microtasks and macrotasks. Concretely reachable via `LiveImportSeam::flush` (`delegate_migration.rs:857`), which calls `save_outbound_dms_to_delegate` with no extra `defer`/`safe_spawn_local` hop of its own — reproducing #530 in a rarer, timing-dependent form | **Must Fix** | skeptical-reviewer | Yes — read `crate::util::sleep`/`defer` and the call graph directly |
| `outbound_dms_save_awaits_hydration_unconditionally_before_coalescing`'s `!body.contains("return Ok")` check doesn't catch a conditional tail-expression bypass (`if cond { Ok(()) } else { coalesce_save(...).await }`) with no `return` keyword at all | Should Fix | testing-reviewer | Yes |
| The `!is_legacy_delegate` guard-pin test is a text scan, not a parser — doesn't strip comments, so a guard reduced to a comment would still pass | Consider (low-likelihood, documented) | testing-reviewer | Yes |
| The 15s-timeout-then-proceed path is, if triggered, literally the original #530 bug in a much rarer form (accepted trade-off vs. waiting forever) | Consider (already a deliberate, now-documented trade-off) | skeptical-reviewer | Yes |
| PR body was stale (still described the round-1 defer/replay design after the round-2 rewrite) | Should Fix | code-first-reviewer | Yes |

**What was verified clean:** no scope creep, no removed/weakened tests, docs (`.claude/rules/direct-messages.md`) kept in sync with the round-2 redesign, no dead code left over from the superseded design (big-picture-reviewer confirmed zero hits for the old `OUTBOUND_DMS_SAVE_PENDING_PRE_HYDRATION`/replay symbols).

### Resolution (commit `1698437d`)

The Must-Fix finding and both Should-Fix findings share a root cause with round 1's: something was allowed to observe a "safe" state before the precondition for that state was actually true. Fixed by deferring `mark_outbound_dms_hydrated()` itself into its own macrotask, scheduled (via `crate::util::defer`) AFTER `hydrate_hidden_dm_threads`/`hydrate_outbound_dms_cache` in the same synchronous call. `setTimeout(0)` macrotasks run FIFO, so the latch-flip macrotask cannot execute — and the latch cannot become `true` — until those two have already run their merges. This makes the invariant structural rather than an argument about scheduling that has to be re-verified by every future reader: the merge is now a precondition of the flip existing at all, regardless of whether an observer is scheduled via a microtask or a macrotask.

Also fixed: the source-order pin test now additionally rejects any `if` in `save_outbound_dms_to_delegate`'s body (the correct implementation is an unconditional two-statement sequence, so this closes the tail-expression-bypass gap structurally rather than trying to string-match every possible bypass shape); the guard-pin test now also asserts the latch is never flipped via a bare synchronous call; the 15s-timeout trade-off is now explicitly documented as an accepted residual; and the PR body has been rewritten to describe the current design.

`cargo test -p river-ui --bins`: 946 passed, 0 failed. `cargo fmt` / `cargo clippy -p river-ui --bins --tests`: clean.

A fresh review round is running against `1698437d` to confirm this fix and check for anything the second redesign introduced.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #670 (Round 3, on commit `1698437d`) + light re-check (commit `0e831476`)

### Round 3 summary
- **Review tier:** Full
- **Reviewers run:** `freenet:code-first-reviewer`, `freenet:testing-reviewer`, `freenet:skeptical-reviewer`, `freenet:big-picture-reviewer`.
- **Verdict on the round-2 fix:** all four reviewers independently confirmed the microtask-race gap (round-2's Must Fix) is **genuinely closed** in `1698437d` — not relocated. `crate::util::defer`'s mechanics were re-verified directly (`ui/src/util.rs`): it always goes through a synchronous `set_timeout_with_callback` with no microtask hop, so FIFO ordering among the three `defer()` calls in `handle_outbound_dms_get_response`'s `Ok` arm genuinely holds today.

### Round 3 findings

| Finding | Severity | Source | Verified |
|---|---|---|---|
| The source-order pin `outbound_dms_save_awaits_hydration_unconditionally_before_coalescing`'s `!body.contains("if ")` check doesn't catch a `match`-arm bypass (`match cond { true => Ok(()), false => coalesce_save(...).await }`) — no `if`, no `return`, both anchor strings still present in order | Should Fix | testing-reviewer | Yes — constructed the exact bypass and traced it through every assertion |
| The response_handler.rs pin test never verified the `Ok`-arm's `defer(mark_outbound_dms_hydrated)` call is textually *ordered after* the two `hydrate_*` calls — the exact invariant the round-2 fix's correctness depends on. A future edit hoisting the guard block up would pass every existing assertion while reintroducing the round-2 race | Medium / Should Fix | skeptical-reviewer AND big-picture-reviewer (independently, same finding) | Yes |
| `.claude/rules/direct-messages.md` documented the hydration-await behavior but not the round-3 refinement (deferred-flip requirement) | Should Fix | code-first-reviewer | Yes |
| Suggested documenting the general "readiness latch must flip from inside the same macrotask as the work it certifies" pattern in `.claude/rules/dioxus-signal-safety.md`, since this PR is a clean case study of a failure mode not yet covered there | Consider | big-picture-reviewer | Judged worth doing |
| The 15s-timeout residual remains an accepted, documented trade-off (not a new finding — reconfirmed) | Consider | skeptical-reviewer | Yes |

**Also reconfirmed clean:** PR coherence across all 3 commits, PR description accuracy, no scope creep, no removed/weakened tests, `LiveImportSeam::flush`'s durability contract intact, all three return-path branches correctly deferred, no new issue for concurrent/duplicate `GetResponse`s.

### Resolution (commit `0e831476`)

Both Should-Fix test-rigor findings share the same shape: a blacklist-style textual check can always be defeated by a shape it didn't anticipate (this is the third round a blacklist approach was tried and defeated). Replaced with:

- **Whitelist, not blacklist**, for the source-order pin: normalize whitespace and require `save_outbound_dms_to_delegate`'s body be **byte-equal** to the one known-correct unconditional two-statement shape. Any control-flow construct at all — `if`, `match`, `?`, a second `return` — changes the normalized text and fails the comparison, regardless of form.
- **Explicit ordering assertion** scoped to the `Ok` arm: `hydrate_hidden_dm_threads(` and `hydrate_outbound_dms_cache(` must appear textually before `crate::util::defer(mark_outbound_dms_hydrated)`.
- Documented the general pattern in `.claude/rules/dioxus-signal-safety.md` (new section: "A readiness latch must flip from inside the same macrotask as the work it certifies") and extended `direct-messages.md`'s hydration-gate note.

This round's diff is **test-and-documentation only** — no production logic changed (the fix itself was already correct as of `1698437d`; round 3's findings were entirely about making the tests unable to be silently defeated by a future edit, which is exactly the failure mode this PR has been most at risk of across all three rounds). Per the multi-model-review policy's "lighter re-check" carve-out (no Must-Fix findings, purely additive, no rebase/amend), I read this diff myself rather than spawning a full round-4: confirmed the whitelist comparison and the ordering assertion are both correct against the current (correct) source, `cargo test -p river-ui --bins` is 946/946, and `cargo fmt`/`clippy` are clean.

**Full findings history across all 3 rounds + this light check:** 2 Must Fix (both resolved: the `LiveImportSeam::flush` durability lie in round 1, the microtask-race in round 2), 5 Should Fix / Medium (all resolved: guard-count robustness, PR-description staleness, tail-expression pin gap, ordering-pin gap, doc gaps), 2 accepted/documented residuals (the 15s-timeout trade-off, the comment-stripping limitation of a text-scan pin).

### Verdict
**State:** Ready to Merge, pending green CI on the current HEAD.
**HEAD SHA reviewed:** `0e831476` (test/doc round self-reviewed against the fully-reviewed `1698437d`).

[AI-assisted - Claude]

Issue #671, opened 2026-09-06
Room contract: whole-state merge is not commutative, and post_apply_cleanup is not idempotent (live Official-room state)


sanity:
> **Read the comments before acting on this body.** The root cause of defect 1 is
> traced in the second comment, and my first analysis of defect 2 was **wrong** and is
> corrected in a later comment: the DMs swept and the DMs lost are two *disjoint* sets
> of six, and the real defect is that a DM destined for deletion can win a global-cap
> slot and evict a legitimate message. Both defects turn out to share one root.

## Summary

The room contract's whole-state merge is not commutative on live Official-room state, and `ChatRoomStateV1::post_apply_cleanup` is not idempotent — the latter contradicting an invariant its own doc comment declares a MUST.

Found by `fdev verify-merge` against a capture bundle taken from the live network (freenet/freenet-core#5320, the merge-law conformance work), then reproduced in pure Rust against `river-core` at `main` (40ca5b05). It is not already fixed on main.

This is **not** freenet/river#413. That issue is about cap-eviction of bans; here the ban set is stable at 200 across every merge and every cleanup pass, and nothing is evicted.

## The counterexample

Contract `9rhuzMSn4v4AugF5FhrjuB1tGP936TaP6Dp7fXijNPUL`, owner `4uNUKFzZQCnzo4K2ecZ16cMsYEEfoaRS35z6exEsbvm4`, room name "Freenet" — the Official room. Corpus: 25 distinct states captured from live traffic by a shadow node running 0.2.131.

Taking two captured states A and B and merging them both ways, where `merge(X, Y)` is exactly what `update_state` does for `UpdateData::State` (`contracts/room-contract/src/lib.rs:139-146`):

| | members | messages | DMs | bytes |
|---|---|---|---|---|
| `merge(A, B)` | 122 | 200 | 300 | 418942 |
| `merge(B, A)` | 123 | 200 | 294 | 415940 |

The differing fields are always the same three: `members`, `member_info`, `direct_messages`. Never `bans`, never `recent_messages`, never `configuration`.

Across a 6-state sample of the corpus (15 unordered pairs), **5 pairs fail commutativity — and all 5 have that identical field signature.** That is what rules out the two paths one would suspect first: if the ban cap-eviction of #413 were firing, `bans` would differ in at least one pair, and it never does; if message retention were at fault, `recent_messages` would differ, and it never does. The defect is confined to member retention and DM retention, with `member_info` following `members`.

[...]

PR #672, opened 2026-09-06
fix: stop MemberInfo/message verify from permanently forking rooms

### LINKED ISSUE #423: Room fork oscillation: divergent states permanently reject each other's deltas (MemberInfo for non-existent member)
## Problem

Production forensics (freenet-core#4861) found River room contract instances in a **permanent two-fork oscillation** — divergent states that reject each other's deltas and can never reconcile:

- Contract `xqyyifd4LEGnZSMZaYGtahBBn5VfHgHLcMGWL83cYqt` (confirmed room via its contract-emitted error): every delta rejection is `MemberInfo exists for non-existent member: MemberId(CKEIWTUD)` — the **same member, 51 of 51 rejections** in the sampled hour. One fork's state contains MemberInfo for a member the other fork's state doesn't have; each side rejects the other's deltas forever.
- Contract `HQk7zGRQagL9AVUMM39C3W3nRVv81xk9BCAeY9MikirK` (same signature class): two stable state sizes (2,180 B seen 591×, 1,952 B seen 552× — ~50/50). Every full-state resync "applies successfully" (`changed=true`) but merely flips the node to the other fork, ~1 full oscillation/minute, indefinitely, with 45 distinct sender IPs participating. The gateway alone sent 9,733 full-state resync responses for this one contract in a day.

This violates the merge requirement that contract state form a commutative monoid: merging fork A and fork B must produce a state both sides accept, but here merge(A, B-delta) is rejected while replace-with-B "succeeds", so the network ping-pongs instead of converging.

freenet-core#4864 bounds the CPU/traffic cost of this class, but the rooms themselves remain forked.

## Ask

1. In `update_state`/merge: when the incoming state contains MemberInfo for a member absent from the local member set, the merge should either import the member (if the signature chain validates) or drop the orphaned MemberInfo deterministically on BOTH sides — not reject the whole delta. Rejection is what makes the fork permanent.
2. How did the fork arise? A member existed on one side but its removal/ban propagated to only one fork? (Possibly related to the deputy-ban work / #4857's propagation gap era, or the nondeterministic-summary period before the BTreeMap fix.)
3. Consider a room-contract invariant test: for arbitrary divergent states A, B built from valid op sequences, merge(A,B) == merge(B,A) and both sides accept it (property/fuzz test at the contract level).

Related: freenet/freenet-core#4861 (forensics), freenet/freenet-core#4864 (core containment), freenet/river#422 (merge cost).

[AI-assisted - Claude]

sanity:
## Problem

Production forensics (freenet-core#4861) found room contract instances stuck in a **permanent two-fork oscillation**: divergent states that reject each other's deltas/full-states forever. The confirmed error is `MemberInfo exists for non-existent member: MemberId(...)`.

`MemberInfoV1::verify` (and, same bug class, `MessagesV1::verify`) hard-rejects the *entire* state whenever it carries an entry (a `MemberInfo` record, or a message) whose author/member is absent from the local `members` set. But `apply_delta`'s unconditional retain-sweep and `post_apply_cleanup` step 4/4b already **tolerate and prune exactly this case** on every normal delta application — so the `verify()` path was strictly more conservative than the rest of the contract's own convergence logic, and that asymmetry is what makes a fork permanent: once fork A prunes a member that fork B still carries (with a `MemberInfo` record or message), *every* full-state PUT/resync between A and B is rejected outright by `verify()`, and the two forks can never reconcile.

## Approach

Mirror the existing precedent for the identical situation in `BansV1::verify_excluding_cap` ("banning member not in current members list"): skip re-verifying the signature for an entry whose member isn't currently known (we can't recover their key to check it anyway, and the signature was already checked when the entry was created), instead of rejecting the whole state. The entry carries no authority for a non-current member (`deputies_of`, ban enforcement, and nickname rendering are only ever consulted for ids present in `parent_state.members`), and it self-heals: `apply_delta` + `post_apply_cleanup` prune it on the very next applied delta.

Applied the identical fix to `MessagesV1::verify` ("Message author not found"), since it's the same asymmetry against the same `post_apply_cleanup` step 4b sweep, and is equally low-risk and well-precedented.

Left `DirectMessagesV1::verify`'s analogous "sender/recipient is not a current member" checks alone — DM tombstone/purge semantics make that a more delicate change, and it isn't what's implicated by the reported forensics; flagging it as a plausible follow-up rather than bundling it here.

This is a `common/` change compiled into both the room-contract and chat-delegate WASM, so it re-keys both:
- `legacy_delegates.toml` / `common/legacy_room_contracts.toml`: migration entries added (old hashes recorded) so existing rooms/secrets survive the re-key.
- `pointer-records.toml`: re-signed against the new WASM (third-party integrators resolve this instead of pinning a key).
- `ui/public/contracts/`, `cli/contracts/`: WASMs resynced via `cargo make sync-wasm`.

## Testing

- Updated the existing unit tests in `member_info.rs` / `message.rs` that previously asserted `verify()` *rejects* an orphaned entry — they now assert it's tolerated.
- Added `orphaned_member_info_and_message_do_not_reject_full_state` in `common/src/room_state.rs`: builds a full `ChatRoomStateV1` carrying a `MemberInfo` record and a message for a member absent from `members`, asserts the **top-level composed** `verify()` (what `validate_state` actually calls) still succeeds, then applies a no-op delta and asserts `post_apply_cleanup` prunes both orphans and the healed state still verifies — demonstrating the self-healing convergence path this fix unblocks.
- `cargo test -p river-core`: all 211+ tests pass (incl. `convergence_tests.rs`, `retention_proptest.rs`, `summary_determinism_test.rs`).
- `cargo make sync-wasm`, `check-migration`, `check-room-contract-migration`, `check-pointer-freshness`, `cargo test -p river-core --test migration_test --test room_contract_migration_test`: all pass.
- `cargo clippy -p river-core --all-targets`: no new warnings introduced by this change.
- Not verified in this PR: a full local UI build (`cargo check -p river-ui`) — this environment's `ui/assets/styles.css` (gitignored, tailwind-generated) isn't built and `tailwindcss`/npm deps aren't installed, so that check fails on an unrelated missing asset (`asset!("/assets/styles.css")`) rather than anything in this diff. `chat-delegate` and `room-contract`, which depend on the same `common` crate, both build and their WASMs were regenerated successfully.

## What I'd want a reviewer to check first

1. Whether skipping signature re-verification for an orphaned `MemberInfo`/message entry (rather than, say, still validating the signature shape/size while skipping only the "member exists" gate) is the right level of tolerance — I followed the `BansV1` precedent exactly since it's the same "can't recover the key" situation.
2. Whether `DirectMessagesV1::verify`'s analogous checks should be fixed in a follow-up now that this asymmetry is named, or left as-is given the tombstone/purge complexity.
3. The migration/pointer-signing bookkeeping (`legacy_delegates.toml`, `common/legacy_room_contracts.toml`, `pointer-records.toml`) — this re-keys both WASMs, so please double check the added registry entries look right before merge.

Closes freenet/river#423

[AI-assisted - Claude]


sanity:
Drive-by from the merge-law conformance work (#671) — one factual correction plus an argument I think strengthens this PR. Not a review, and I have no objection to the change.

**Correction:** the reasoning behind this PR included that `apply_delta` + `post_apply_cleanup` already prune orphans consistently within one pass. **`post_apply_cleanup` is measurably not idempotent.** Execution trace from live Official-room state: `merge(state_07, state_01)` returns 123 members, and running `post_apply_cleanup` on that result again drops it to 122 — the leftover member satisfies none of the five retention rules in the returned state. Its own doc comment (`common/src/room_state.rs:100`) declares idempotence a MUST, for exactly the reason that peers run it a variable number of times. Details and the trace are on #671.

**Where that conclusion still holds:** steps 3 and 4 prune `members` and `member_info` against the *same* `required_ids` (`room_state.rs:326-334`), so the two stay mutually consistent even on the non-idempotent pass. The residue is a member that should not be there together with its member_info — not an orphaned member_info. So I do not think #671's defect is the source of #423's orphans, and I would not claim it is.

**The argument I would put in the PR description instead**, because it does not depend on identifying the orphan's origin at all:

> A full-state PUT bypasses `post_apply_cleanup` entirely — it reaches state through `validate_state` → `verify` only. So peers legitimately hold states that cleanup would never have produced. A `verify` strict enough to reject them turns "this peer holds slightly unusual state" into "this state is permanently rejected", which is the fork.

Framed that way this is not tolerating a bug, it is `verify` accepting what the PUT path can legitimately deliver — and it stays correct after #671 is fixed. The non-idempotence above is useful here only as proof that at least one such path demonstrably exists.

Mirroring `BansV1::verify_excluding_cap` looks like the right precedent to me.

[AI-assisted - Claude]

sanity:
## Converted to draft — the tolerant branch admits unsigned `MemberInfo`, and that is reachable

CI is green and the convergence fix works, but an adversarial review found a privilege-escalation path introduced by the tolerance itself. Flagging before merge rather than after. Nothing here is exploitable today: this is unmerged, and `main` still has the old hard-reject.

### The change

Where a member is absent from `parent_state.members`, the old code errored; the new code accepts the record **with no signature check at all** (`common/src/room_state/member_info.rs:359-378`, and the same shape in `message.rs:139-154`). The stated justification is that such a record "carries no authority for a non-current member".

### Why that holds only while the id stays absent

Absence is not an invariant. An `AuthorizedMember` is public and replayable, so a pruned member can be re-added by anyone — and once re-added, the record is no longer an orphan.

`deputies_of` is what confers ban authority, and it is read in three places (`common/src/room_state/member.rs:378`, `:384`, `:393` — the last for every strict ancestor of the target). So the sequence is:

1. PUT a state where ancestor `A` is absent from `members`, carrying a forged `MemberInfo { member_id: A, version: u32::MAX, deputies: [attacker] }` with a garbage signature — now accepted.
2. `A` is later re-added, or the attacker replays `A`'s public member record.
3. Field order applies `members` before `member_info`, so the retain sweep (`member_info.rs:527`) now **keeps** the record rather than pruning it, and `canonical` selects it on rank `u32::MAX`.
4. `deputies_of(A)` returns `[attacker]`, granting ban authority over `A`'s entire subtree.

Pre-PR this was unreachable, because `verify` rejected the state outright.

### The cited precedent is real but the comment omits its second half

The code comment points at `BansV1::verify_excluding_cap` as doing "the identical" thing. It does skip the same signature check — but it compensates **at enforcement**: `ban_signature_matches_current_key` (`ban.rs:216`, enforced at `ban.rs:268`), whose own doc comment describes this exact re-add attack. `MemberInfoV1` and `MessagesV1` have no equivalent re-check, so the analogy imports the risk without the mitigation.

### Also worth fixing while here

`MemberInfoV1::verify` caps `deputies` per record (`member_info.rs:342`) but never the record **count** — previously bounded implicitly by `members`, now unbounded. `MessagesV1::verify` has no count cap at all (`max_recent_messages` is enforced only in `apply_delta`, `message.rs:346`), so unsigned orphan messages arrive unbounded on a single full-state PUT.

### Suggested direction

Keep the tolerance — the convergence fix is right and #423 is a real bug — but add the half the bans path has: an enforcement-time signature re-check before an orphaned record is allowed to confer anything, plus a record/message count cap on the `verify` path.

### Unrelated, and already fixed on this branch

The four red checks were migration bookkeeping, now green: the room-contract registry pin (31 → 32 + new blake3), the `riverctl` version bump for `check-wasm-sync`, and the two delegate pins. The migration substance was correct throughout — `legacy_delegates.toml` records the old delegate `code_hash`, so existing room keys stay findable, and `migration_test` and `check-migration` both pass. See sanity/torvan#39 for the pattern.

Draft rather than closed: the work is nearly there, and a dispatcher with landing authority would otherwise merge it on green.

[AI-assisted - Claude]

sanity:
One more instance of the same class, not covered here: **`DirectMessagesV1::verify` hard-rejects on a non-member DM endpoint.**

`common/src/room_state/direct_messages.rs:881-893`:

```rust
let sender_vk = resolve_member_vk(msg.message.sender, owner_id, parameters, &members_by_id)
    .ok_or_else(|| format!("DM sender {:?} is not a current member", msg.message.sender))?;

if resolve_member_vk(msg.message.recipient, owner_id, parameters, &members_by_id).is_none() {
    return Err(format!("DM recipient {:?} is not a current member", msg.message.recipient));
}
```

One unresolvable endpoint rejects the **entire state**, exactly as `MemberInfoV1::verify` and `MessagesV1::verify` did before this PR. This diff touches `member_info.rs` and `message.rs` but not `direct_messages.rs`.

Note bans are correctly *not* enforced here — the module comment says so deliberately, so `verify` stays stable across ban churn. It is the **membership** check that has the hard-reject shape.

Why it is reachable rather than theoretical: a full-state PUT bypasses `post_apply_cleanup` entirely, which is the same gap that motivates this PR. Step 6's own comment says as much — *"Without this, a fresh ban (or member-prune) would leave the DMs in state but break `verify` because the sender/recipient can no longer be resolved."* That sentence is describing this hazard from the other side.

It is also directly on the path of the #671 fix, which will re-key the contract and therefore migrate every room's state through a PUT.

Your call whether to fold it into this PR (it is the same three-line shape and would share the re-key you are already paying for) or take it separately. Flagging rather than assuming — I have not touched your branch.

[AI-assisted - Claude]

sanity:
## 🔴 Merge-blocking: this PR would publish riverctl 0.2.15 against a stale, immutable river-core

Found while reviewing #673, which has the **identical** defect. Flagging here because whichever merges first ships it.

`cli/Cargo.toml:53` on `d8ca9182`:

```toml
river-core = { version = "=0.1.20", path = "../common", features = [… "migration" …] }
```

and workspace `Cargo.toml:67` is still `0.1.20`, while `cli` bumps 0.2.14 → 0.2.15 and the PR adds a **V32** entry to `common/legacy_room_contracts.toml`.

**river-core 0.1.20 is already on crates.io (2026-08-23) and immutable.** A reviewer downloaded and inspected the published crate:

```
river-core-0.1.20/legacy_room_contracts.toml → 31 entries, last = V31 dd63bcc9…
grep -c e765339b…  →  0
```

`cargo publish` rewrites the path dependency to the registry dependency. So riverctl 0.2.15 ships bundling the **new** room contract against a registry that **omits the generation live on the network right now**. A `cargo install riverctl` build derives the new key, probes back to V31 — dead since 2026-07-30 — and skips `e765339b` entirely. **It could not find any existing room, the Official room included**, until a UI client migrated that room forward.

It is silent by construction. `.github/workflows/release-riverctl.yml:130` is *"Publish river-core then riverctl (**skip if already on crates.io**)"* — `publish_if_needed river-core` sees 0.1.20 present, prints "already on crates.io — skipping", exits 0, and publishes riverctl anyway. In-workspace builds resolve via `path`, so nothing in CI ever sees it, and `check-cli-wasm` only gates the *cli* version.

### The fix

Bump **all three** together, as the V31 precedent commit `394c27dd` (#572) did — workspace `0.1.18→0.1.19`, cli `0.2.9→0.2.10`, and the `=` pin. Here that is workspace `0.1.20 → 0.1.21` and the pin `=0.1.20 → =0.1.21`.

`0.1.21` rather than `0.2.0`: crates.io reverse-dependencies for river-core is `total: 1` (riverctl, in-workspace), and `394c27dd` already shipped a genuinely breaking public type change as a patch bump.

**It has to be in the same commit as the artifacts.** A river-core version bump re-keys the delegate on its own, so doing it afterwards invalidates `pointer-records.toml`, both legacy registries' recorded hashes, and the two WASM pins — the whole migration payload.

### Why merge-blocking rather than publish-blocking

crates.io is the one irreversible step here. A wrong riverctl 0.2.15 can be **yanked but not replaced**, and yanking does nothing for anyone who already installed. Everything else in this change is recoverable via the append-only registry.

### Also relevant to both PRs

- **You and #673 both bump riverctl to 0.2.15.** Whichever merges second needs 0.2.16, and a fresh river-core bump with it.
- Second to merge must also: **keep** the first's V32 entry and **append** a new one for the first PR's published hash (append-only — both generations may hold rooms), regenerate both WASMs via the **co-build** (`-p room-contract` alone yields a different key, because `chat-delegate` unifies river-core's `ecies` feature), re-sign both pointer records at **version 3**, and re-run review and CI on the rebased head.
- Corroboration in your favour: your rebuilt delegate hash is **byte-identical** to #673's, and your delegate pointer record is byte-identical too. Two unrelated branches independently signing the same delegate generation is good evidence the delegate move is toolchain drift rather than either fix.
- Heads-up unrelated to this PR: the **delegate pointer record `6qF2H5JR…` resolves to "Contract not found"** on the live network — its committed v1 apparently never landed, so integrators fall back to a baked-in key. `publish-pointer-records.sh` reads back after writing; do not skip that step this time.

Nothing here is a criticism of the change itself — the migration payload is otherwise correct and was verified against the live network's signed pointer record.

[AI-assisted - Claude]

sanity:
## 🔴 New blocker: riverctl 0.2.15 is already published — `origin/main` moved

`origin/main` advanced from `40ca5b05` to `73a6ef3f` (#680, `fix(cli): emit author_verifying_key on every message stream JSON event`). That PR bumped **riverctl to 0.2.15**, and it is now **on crates.io and immutable**:

```
riverctl published: 0.2.11, 0.2.12, 0.2.13, 0.2.14, 0.2.15
origin/main cli/Cargo.toml: version = "0.2.15"
```

This PR also sets riverctl `0.2.15`, so the merge result carries a version that is already taken.

**Consequence — the same defect as the river-core one, from a different direction.** `publish_if_needed riverctl` queries the index, sees 0.2.15, prints "already on crates.io — skipping", exits 0. **riverctl is never republished.** The CLI on crates.io stays the artifact #680 built, carrying the pre-fix room contract and a registry without the current generation. A `cargo install riverctl` build then cannot find the Official room. Green pipeline, wrong artifact.

**Fix: bump riverctl to `0.2.16`.** river-core `0.1.21` is still unpublished and remains correct.

Then rebase onto `73a6ef3f` and **rebuild-and-compare** rather than assuming the bump is hash-neutral. It should not reach the contract build — the WASM comes from `room-contract` + `chat-delegate` via `river-core`, not riverctl — but `Cargo.lock` changes, and this review has already established that a single **comment line** re-keys the room contract. If either hash moves, re-sign the pointer records.

**Both #672 and #673 are affected**, and both currently claim 0.2.15. Whichever merges second needs a further bump again, plus the registry **append** (not replace), regenerated artifacts via the **co-build**, and pointer records re-signed at the next version.

### How this was found, since it is the interesting part

Not by checking this branch. A reviewer noticed `origin/main` had moved, then built the **actual auto-merge result** rather than the branch tip — parents `73a6ef3f` + `8c833270`, merge `008d3021`. Git auto-merged `Cargo.lock` and `cli/Cargo.toml`, both of which carry release payload state. The merge itself was clean and reproduced both artifacts byte-identically with the `=0.1.21` pin intact — but the version collision is only visible once you look at the merge, not the branch.

**Verifying a branch tip does not verify what lands on main.** Worth adding to the release checklist, alongside the fact that nothing in CI compares a committed artifact to a build of its source (#678) and that two `check-pointer-freshness` clauses never execute (#677).

[AI-assisted - Claude]

sanity:
### Heads-up: #673 has merged, and this PR's release metadata is now stale in four ways

#673 merged as `8c0ca8c3` and re-keyed both WASMs. This PR was prepared against the pre-#673 world, so a rebase alone is not enough — four recorded values are now wrong, and one of them loses user data silently.

**1. The migration predecessor is the dangerous one.** This PR's `legacy_room_contracts.toml` entry records `e765339b…` as the predecessor. That was correct yesterday; it is not now. After #673 publishes, the live room-contract generation is `a3e63c8c0426255376f9f6f00948512e185bd0961875e43cd9e3e479ccaa0ce1`. If this PR re-keys from that and the registry never names it, every room touched between the two publishes is stranded on an unprobed key.

The registry is **append-only**. On rebase, keep #673's entry and add a new one below it — do not resolve the conflict by keeping this branch's version of the block, which is what a conflict resolution naturally does here and what would drop `a3e63c8c` on the floor. Same for `legacy_delegates.toml` (`c2e60638…` is the new delegate predecessor).

**2. The registry labels collide.** #673 used `V32` (room contract) and `V31` (delegate). This PR also uses `V32`/`V31`, so the next ones are `V33`/`V32`.

**3. Both crates.io versions are taken.** This PR claims riverctl `0.2.15`, which was published before #673 was written; #673 then took `river-core 0.1.21` and `riverctl 0.2.16`. So this needs `river-core 0.1.22` and `riverctl 0.2.17`. Worth re-checking against crates.io immediately before merge rather than trusting this comment — that is the failure this PR's sibling hit twice (#685).

**4. Pointer records need re-signing at version 3.** #673 moved them 1 → 2. `cargo make sign-pointer-records` after the rebase; `check-pointer-freshness` will fail the PR otherwise.

Because the room-contract WASM changes again, the review is stale under the per-code-content rule once the rebase lands — the conflict resolution in the registries is itself an unreviewed judgment call, and it is precisely the part where getting it wrong is silent.

[AI-assisted - Claude]

PR #680, opened 2026-09-06
fix(cli): emit author_verifying_key on every `message stream` JSON event

### LINKED ISSUE #679: riverctl: `message stream` JSON omits `author_verifying_key` that `message list` provides
`riverctl -f JSON message list <ROOM>` includes `author_verifying_key` — the full base58 Ed25519 verifying key of each message's author — but `riverctl -f JSON message stream <ROOM>` does not. The stream's JSON events carry only `author`, the 8-character `MemberId` short form, plus `nickname`.

Both of those are unsafe as a bot's identity key:

- `MemberId`'s printed form is 40 bits of a 64-bit hash, so two distinct members can print the same short id (the CLI's own `resolve_short_id` documents exactly this).
- `nickname` is member-controlled and can be changed to impersonate anyone.

So a bot driven off `message stream` cannot allow-list or attribute on a collision-proof identity, while the same bot reading `message list` can. That forces a stream consumer to fall back to a `message list` call per event just to recover the author key.

Reported by Ivvor in the Official River room.

### Scope

The gap is in the shared stream output path (`cli/src/api.rs`), so it affects **both** stream modes (`--subscribe` and the default polling path) and **all** JSON event types:

- `output_message` — `type: "message"` and `type: "edit"`
- `output_deletion` — `type: "delete"`
- `output_reaction_change` — `type: "reaction"`

### Origin

`author_verifying_key` was added to `message list` in #667; the stream path was not updated at the same time.

### Fix

Emit `author_verifying_key` on every stream JSON event, resolved through the same `RoomDeputies::verifying_key` helper `message list` uses (so the two cannot drift), and `null` for an author who is no longer the owner or a current member — matching `message list`'s existing semantics.

Note on `type: "reaction"`: `author` there is the author of the message reacted to, not the reactor, so `author_verifying_key` follows the same rule. Giving reactors their full keys is a separate gap that `message list` shares (its `reactors` map is short ids too) and is not part of this fix.

[AI-assisted - Claude]

sanity:
Fixes #679, reported by Ivvor in the Official River room.

## The bug

`riverctl -f JSON message list <ROOM>` includes `author_verifying_key` — the author's full base58 Ed25519 key — but `riverctl -f JSON message stream <ROOM>` does not. The stream's events carried only `author` (the 8-character `MemberId`, a 40-bit truncation of a 64-bit hash that two members can share) and `nickname` (member-controlled, so freely impersonable).

So a bot driven off the live stream could not allow-list or attribute on a collision-proof identity, while the same bot reading the backfill could — forcing a `message list` call per event just to recover the key. `author_verifying_key` was added to `message list` in #667 and the stream path was not updated alongside it.

## The fix

Emit `author_verifying_key` on **all four** stream JSON event types — `message`, `edit`, `delete`, `reaction` — which also covers **both** stream modes, since polling and `--subscribe` funnel through the same output path.

- Resolution goes through the same `RoomDeputies::verifying_key` helper `message list` uses, so backfill and live feed cannot drift.
- `null` when the author is neither the owner nor a current member (banned, or pruned for inactivity), matching `message list`'s documented semantics — never a wrong key.
- The owner resolves correctly, which is the case a members-only lookup would get wrong: the owner is structurally absent from `members.members`.
- On a `reaction` event the field names the author of the message **reacted to**, not the reactor, following the existing `author` field. Reactors are in `reactors`.

Not in scope: giving *reactors* their full keys. `message list` has the same gap (its `reactors` map is short ids too), so it is a shared limitation rather than a list-vs-stream discrepancy — noted in #679.

## Refactor, and why

Each event's JSON payload is split out of its `OutputFormat::Json` arm into a function returning the `serde_json::Value` (`message_event_json`, `reaction_event_json`, `deletion_event_json`); the arms now only serialize and print. The output functions write to stdout, so without this the only available test is scraping the source that builds the JSON. This lets the tests assert on the object actually emitted. Human output paths are untouched, and the existing raw-vs-escaped nickname pin still holds (its wording was updated to name the new locations).

## Verification

- **Three new tests**, all confirmed to FAIL with the field removed (not just watched to pass):
  - every event type carries the author's real key, and the fixture asserts up front that the short id differs from the full key so the test cannot pass trivially;
  - the stream's value equals what `message list` resolves, for a member *and* for the owner;
  - a departed author yields `null`. This one uses `.get(...) == Some(&Value::Null)` rather than indexing, because indexing yields `Null` for an *absent* key too — under the indexed form the original bug would have satisfied it.
- Full lib suite green (356 tests).
- **Live against the Official room**: all 165 streamed events carried a non-null key, and for the 5 messages also returned by `message list` the `author` / `author_verifying_key` pairs were byte-identical.

Version bumped to 0.2.15 with a CHANGELOG entry; README's stream event-type table now documents the field (and gains the `reaction` row, which was missing).

[AI-assisted - Claude]

https://claude.ai/code/session_01GNpAZJgzRRaGr3puE7HBPi

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4782,7 +4782,7 @@ dependencies = [
 
 [[package]]
 name = "riverctl"
-version = "0.2.14"
+version = "0.2.15"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to riverctl will be documented in this file.
 
+## [0.2.15] - 2026-09-06
+
+### Fixed
+- `message stream --format json` now includes `author_verifying_key` on every
+  event (`message`, `edit`, `delete`, `reaction`), matching `message list`.
+  Previously only the backfill carried the author's collision-proof key, so a
+  bot driven off the live stream had to fall back to a `message list` call per
+  event to recover it, or trust the 8-character `author` short id (a 40-bit
+  truncation two members can share) or the member-controlled `nickname`. Both
+  stream modes are covered (polling and `--subscribe`). The value is resolved
+  through the same helper `message list` uses, and is `null` for an author who
+  is no longer the owner or a current member. On a `reaction` event it names the
+  author of the message reacted to, not the reactor. (freenet/river#679)
+
 ## [0.2.14] - 2026-08-31
 
 ### Added
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -7,14 +7,17 @@ All notable changes to riverctl will be documented in this file.
 ### Fixed
 - `message stream --format json` now includes `author_verifying_key` on every
   event (`message`, `edit`, `delete`, `reaction`), matching `message list`.
-  Previously only the backfill carried the author's collision-proof key, so a
+  Previously only the `message list` backfill carried the author's full key, so a
   bot driven off the live stream had to fall back to a `message list` call per
   event to recover it, or trust the 8-character `author` short id (a 40-bit
-  truncation two members can share) or the member-controlled `nickname`. Both
-  stream modes are covered (polling and `--subscribe`). The value is resolved
-  through the same helper `message list` uses, and is `null` for an author who
-  is no longer the owner or a current member. On a `reaction` event it names the
-  author of the message reacted to, not the reactor. (freenet/river#679)
+  truncation of a 64-bit non-cryptographic hash, which two members can share) or
+  the member-controlled `nickname`. Both stream modes are covered (polling and
+  `--subscribe`). Resolution and base58 encoding now come from a single shared
+  helper, so the backfill and the live feed cannot drift. The value is `null`
+  when the author is not in the room state riverctl holds — departed, pruned, or
+  a members delta not yet merged — and never a wrong key. On a `reaction` event
+  it names the author of the message reacted to, not the reactor; `reactors`
+  remains short ids only. (freenet/river#679)
 
 ## [0.2.14] - 2026-08-31
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "riverctl"
-version = "0.2.14"
+version = "0.2.15"
 edition = "2021"
 authors = ["Freenet Project"]
 description = "Command-line interface for River decentralized chat on Freenet"
--- a/cli/README.md
+++ b/cli/README.md
@@ -115,22 +115,28 @@ riverctl identity whoami <room-owner-vk>         # Your member ID in one room.
 riverctl identity whoami                         # Every room you're in.
 ```
 
-The `member_id` it reports is exactly the top-level `author` value your own
-messages carry in `message list` / `message stream --format json`, which is what
-a bridge needs to filter out its own echo:
+`whoami` reports both identifiers, and they pair with two different fields — a
+bridge filtering out its own echo should use the full key:
 
 ```bash
-me=$(riverctl identity whoami <room-owner-vk> --format json | jq -r .member_id)
+me=$(riverctl identity whoami <room-owner-vk> --format json | jq -r .verifying_key)
 riverctl message stream <room-owner-vk> --format json --no-version-check |
-  jq -c --arg me "$me" 'select(.author != $me)'
+  jq -c --arg me "$me" 'select(.author_verifying_key != $me)'
 ```
 
 `message list` and `message stream --format json` both carry
-`author_verifying_key` — the author's full base58 key (or `null` when the author
-is no longer in the room). The `author` short id can be filtered on cheaply as
-above, but a bot deciding whether to *trust* a command should compare this full
-key, since the short id is a 40-bit truncation. Get the trusted keys from
-`member list`'s `verifying_key`.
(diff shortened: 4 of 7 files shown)


sanity:
## Multi-perspective review — Full tier

Five independent lenses, each blind to the others: **code-first**, **skeptical**, **testing**, **big-picture**, and a fifth **downstream-consumer** lens.

**External model:** `codex review --base origin/main` was attempted twice and failed both times with `You've hit your usage limit … try again at Sep 7th, 2026 8:14 AM`. Per the project's review rule the external pass was substituted with the extra Claude lens above (the consumer-contract one), and that substitution is recorded here rather than silently dropped. Gemini remains unavailable (deprecated free tier, since 2026-06-24).

No High-severity defect was found. Every finding is fixed in `401dc62e` or answered below.

### The one that mattered: the tests did not cover the site the bug lived at

Two lenses, independently, ran the same mutation — take the payload the builder produced and strip the field at the emission site:

```rust
let mut v = message_event_json(room_state, msg, room_owner_key, is_edit, secrets);
v.as_object_mut().unwrap().remove("author_verifying_key");
```

`message stream --format json` then emits no `author_verifying_key` at all, and **all 356 tests passed.** The refactor had made the payload testable and, in the same motion, moved the test seam one level *below* where #679 actually lived. As one lens put it, the stream ended up less guarded than the `message list` surface it was brought to parity with.

Fixed by moving the seam up to the line each output function prints — `message_event_line` / `reaction_event_line` / `deletion_event_line` — with the tests parsing that line back into JSON. Six mutations, each confirmed to turn the suite red:

| # | Mutation | Tests failed |
|---|---|---|
| E | field stripped at the call site (the one that defeated the old suite) | 4 |
| A | field absent from all three builders | 4 |
| C | helper resolves everyone to the owner's key | 3 |
| H | the shared base58 encoding drifts | 2 |
| G | reaction event names the reactor instead of the message author | 3 |
| J | the two signature-compatible builders swapped (compiles cleanly) | 4 |

G and J were both raised as hazards the compiler cannot catch; they are now caught by tests.

### "Cannot drift" was only half true

Both surfaces resolved through `RoomDeputies::verifying_key`, but each did its **own** base58 encoding — and the parity test compared the helper against its own body, so it was an identity by construction and passed even with the field removed from every event. Resolution *and* encoding are now one shared helper that `message list` calls too, pinned at both ends, and the parity test asserts against the fixture's own signing keys instead.

### Per-event cost

A lens benchmarked `RoomDeputies::new`, which was being built once per emitted event:

| members | `RoomDeputies::new` + lookup | the lookup alone |
|---|---|---|
| 200 (default `max_members`) | 137 µs | 12 µs |
| 5000 | 45.0 ms | 234 µs |

and showed the doc comment defended it with the wrong comparison — per-*poll*, when the backfill and the first poll emit in a burst. Resolution is now `deputies::verifying_key_of`, a scan with no index build, and it is the single definition `RoomDeputies::verifying_key` delegates to.

### Also fixed

- **Duplicated work on the JSON path.** The output functions computed `reply`, `nickname`, `reactions` and `edited` before the match; the JSON path discarded and recomputed them — re-decrypting a private-room reply and re-unsealing a nickname — while a comment two lines away claimed exactly this had been avoided. They now live in the human arm, the only arm that reads them.
- **The docs oversold the guarantee.** `MemberId` is a 64-bit *non-cryptographic* hash, so "collision-proof" was wrong. What actually makes the field trustworthy is that `MessagesV1::verify` checks the signature against an entry from the same member map, so the key reported is the key the message was verified under. Now written down rather than inferred. Also documented: `None` does not mean "not banned" — enforcement runs in `post_apply_cleanup`, not `verify`.
- **The README taught the collision-prone recipe.** The self-echo filter used the 40-bit id although `whoami` already emits a comparable `verifying_key`. It now uses the full key and names the crossed-pairing trap: `member_id` compared against `author_verifying_key` is never equal, which silently disables the filter and can put a bridge in a feedback loop.
- **The new reaction caution steered readers onto a weak field.** It sent them to `reactors`, which is short ids only. Now said explicitly.
- Corrected `reply_to`'s documented field list (it has four keys, not two), and the forward-compat note now asks consumers to tolerate unknown **fields**, not only unknown types.
- Field-set assertions per event type, and null coverage extended to the `reaction` and `delete` events.

### Deliberately not changed

Full keys for `reactors`. Both lenses that raised it are right that it is a live trust surface — `river-moderator` compares reactor short ids against a ban protect-list (`runtime.rs:208-221` → `ban_member_safely`) — but it is a different field, pre-existing on **both** surfaces rather than a list-vs-stream gap, and it has three plausible shapes with different compatibility and payload-size costs. Filed as #682 with the evidence and the options, rather than grown into this PR. Say the word if you would rather it landed here.

Also filed: #683 (`member deputies` / `member deputized-by` JSON has no `verifying_key`, though a deputy holds ban authority) and #684 (two remaining list-vs-stream asymmetries).

### Verified, not assumed

- Human/terminal output is unchanged — checked by reading each arm and by running `message stream` against the Official room.
- Both stream modes route through the three emitters; there is no fourth emission site.
- Backward compatibility: the two real consumers on this machine (`river-mirror`, `river-moderator`) parse permissively; no `deny_unknown_fields` struct touches riverctl wire JSON.
- Live re-check after the shared-helper change: 165 streamed events all carried a non-null key, and for the 6 messages also returned by `message list` the `author` / `author_verifying_key` pairs were identical.
- 358 tests green; no new clippy warnings (the 10 that remain are all pre-existing and outside this diff).

[AI-assisted - Claude]

PR #688, opened 2026-09-10
fix(riverctl): route every wait for a node response through one shared waiter

sanity:
## Problem

riverctl and the freenet node share one WebSocket connection, and the node
multiplexes every response for every subscription onto it. A response that is
not the answer to the request just sent can therefore arrive first, most
commonly an `UpdateNotification` for a room this process is already subscribed
to.

Ten call sites read that first message as their answer and failed if it was
not.

This broke the freenet v0.2.135 release announcement. The node had restarted
onto the new binary moments earlier, a notification landed inside the send
window, and `announce-to-river.sh` exited 1 with
`Unexpected response type: ContractResponse(UpdateNotification ...)`. Every CI
job reported success, because the release workflow only sees the webhook's HTTP
202 and the post fails afterwards on nova.

## Approach

This is the fourth time this class has been found. The pointer GET got a
step-over loop, the SUBSCRIBE acknowledgement got `SubscribeAck`
(freenet-core#4970), the streaming monitor got its own, and each fix was written
at one call site and left every other site doing the old thing. So this is not
another per-site loop.

**One waiter.** `await_response` reads until it gets the answer or its deadline
expires, stepping over anything else. Its classifier reports three outcomes, not
two:

| Outcome | Meaning |
|---|---|
| `Some(Ok(v))` | this is our answer |
| `Some(Err(e))` | the node answered, and refused. Definitive, stop |
| `None` | not our answer. Step over it and keep waiting |

The third is load-bearing rather than decorative: without a definitive-refusal
outcome, an answered `NotFound` would be stepped over and surface as a timeout
instead of as what the node actually said.

**Two waiters on top of it.** `await_update_response` is used by the seven
paths that send a `ContractRequest::Update`; `await_get_response` by
`get_state_from_contract` and `accept_invitation_struct`'s GET.
`republish_room` uses `await_response` directly.

**An `UpdateResponse` for a different key is stepped over, not accepted.** The
base accepted any key and logged the *wrong* one; the first version of this PR
accepted it with a `warn!`. Both were wrong, and the path is reachable:
`send_message_with_key` calls `get_room`, which calls `heal_member_info`, which
calls `send_delta`, whose error is swallowed with `warn!` and leaves its own
`UpdateResponse` queued.

This does **not** close the same-key case, and review established that case is
reachable rather than theoretical: `heal_member_info` resolves the *same*
contract key as its caller, so the chain above is itself a same-key case. The
lock stops two sends overlapping but is released when a wait *fails*, and a
failed wait leaves its answer outstanding. Filed as **#691** and stated at the
classifier rather than waved away as "this should not happen".

**No cap on responses stepped over.** The deadline is the whole bound. The
reason is the cost of *giving up*, not memory: `MAX_UNRELATED_RESPONSES_DURING_POINTER_GET`
also drops what it steps over and is capped anyway, because bailing early is
nearly free there (it answers `Unreachable` and falls back to the bundled hash).
These callers have no fallback, so a cap would only add a second way to fail a
send that the deadline already bounds, firing exactly when the room is busiest.
What *is* unbounded without a cap is log volume, which is why the step-over line
names the response kind instead of formatting a delta's bytes.

## Testing the wait, not just the predicate

Review round 2 found the gap that mattered: reverting the step-over inside
`await_response` left **all 368 tests green**. The classifier tests bound the
predicate and the pin bound the routing; nothing bound the loop joining them,
and the loop is where the defect lived. Consolidating ten sites into one waiter
had made that worse, not better, since one line could then reinstate the bug
everywhere at once.

`await_response` now takes `&mut S: ResponseSource`, a two-line trait
implemented for the `MutexGuard` every caller already holds, so **no call site
changed**. `mod await_response_tests` drives the loop against a scripted
connection under `start_paused` virtual time: 9 tests, 0.00s, no socket, no
node, no real clock. The first is the v0.2.135 regression end to end.

## The pin

`mod response_multiplexing_pin` binds the CLASS rather than any one site:

1. each path in `UPDATE_SEND_PATHS` waits through `await_update_response`,
   searched in that function's own body, so lifting the wait into a sibling does
   not satisfy it (seven of the eight production sends; the join delta is #690's);
2. the number of `ContractRequest::Update` sends is accounted for;
3. the number of `.recv()` calls is accounted for;
4. the number of times the connection lock is TAKEN is accounted for.

Checks 2 to 4 count strings, and the pin's doc now says so rather than claiming
proof. Check 3 originally counted the literal `web_api.recv()` and a reviewer
walked a new raw read past it by naming the guard `api`: **a count keyed on a
caller-chosen name is a spelling ban wearing a count's clothes.** Check 4 is the
backstop, and the one count not keyed on a name, since `self.web_api` is private
to the file and every use of the connection must pass through it.

Six mutation meta-tests check the pin bites, including a raw read bound to
`conn` rather than `web_api`, and a **send-only** call site that checks 1 to 3
all miss by construction.

## Deliberately not in this change

Six call sites are left alone: five tracked in **#690**, one in **#689**. Four
of them are *tolerant*, folding an unrelated response into a success or a "not
found" rather than an error, so fixing them changes what counts as success and
needs its own review (`ensure_room_migrated` is the worst: a non-`GetResponse`
is read as "the contract does not exist" and starts a full migration PUT). The
other two are fatal like the ten converted here, but held back for their own
reasons: `create_room`'s two success arms persist different things, and
`migrate_room_to_new_contract` writes an unrelated notification's key into local
storage, which is data corruption rather than a tolerance bug (#689).

**#691** is the residual this PR narrows but does not close: a timed-out wait
leaves its answer outstanding, and the next same-key send can consume it.

## Testing

- 9 loop tests through the `ResponseSource` seam, including the v0.2.135
  regression end to end.
- Classifier unit tests for `classify_update_response` and `classify_put_response`
  alongside the `SubscribeAck` ones they mirror.
- 6 mutation meta-tests on the pin.
- **Red-checked, not merely written.** Every test above was verified to fail
  against a mutation of the real source: reverting the step-over (7 of 9 fail),
  deleting the deadline guard, recomputing the deadline per message, deleting
  the `NotFound` arm, dropping the stepped-over count, deleting
  `await_get_response`'s key guard (it returns another room's state), deleting
  `PutAck::Merged`, dropping the PUT key guards, and removing pin check 4.
- 385 tests pass (358 on the base). `cargo fmt --check` clean. Clippy adds no
  new `api.rs` warning and removes one pre-existing one.

## Review history

- **`0fd8da98`** fixed one call site and was rejected: 1 of 7 update sites, a
  silent-false-success regression from any-key acceptance, four tests that
  stayed green with the fix reverted, and a step-over counter that would have
  failed a room emitting more than 16 notifications in the window. Force-pushed
  away, so this is a different change rather than an increment on it.
- **`9028ccf1`** answered those, and round 2 (codex plus three blind Claude
  lenses) found the untested loop, the missing `UpdateResponse` answer for a PUT
  onto an existing contract, two defeats of the pin, two untested classifiers,
  and five source comments claiming more than the code delivers.
- **`3e8ecaf1` / `2e2013eb`** answered all of those.
- Round 3 (codex plus a fresh skeptical lens) came back clean from codex, and
  the lens found four more, **three of them in code added to answer round 2**:
  the PUT path's wiring to its classifier was bound by nothing (rewiring it to
  accept any response left 383 of 383 green), check 4 was a spelling ban that an
  `Arc` clone walked past, `PutAck::Merged` was described as the common answer
  when it is the local executor's only, and my claim that the cancel-safety
  exposure was unchanged from base was falsified.
- **`6615c356` / `087bf3b9`** answer those.

Both consolidated reviews are posted as comments on this PR.

**Three defeats of the pin in three rounds**, each by a spelling its needles did
not anticipate, and **three instances of one shape**: the pin binds routing, the
unit tests bind the predicate, and the join between them goes unbound. That is
what #692 is for, and why the structural fix is not in this PR.

Closes nothing on its own; #689, #690 and #691 track the remainder.

https://claude.ai/code/session_015k3AmivdXMHWaZb8qGuQQz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4782,7 +4782,7 @@ dependencies = [
 
 [[package]]
 name = "riverctl"
-version = "0.2.16"
+version = "0.2.17"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -101,3 +101,6 @@ ureq = { version = "2", default-features = false, features = ["tls"] }
 freenet-test-network = "0.1.24"
 tempfile = "3"
 assert_cmd = "2"
+# `test-util` gives `#[tokio::test(start_paused = true)]`, so the response-wait
+# tests can exercise 60-second deadlines in virtual time instead of real time.
+tokio = { version = "1.42", features = ["full", "test-util"] }
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -104,7 +104,11 @@ assert_cmd = "2"
 # `test-util` gives `#[tokio::test(start_paused = true)]`, so the response-wait
 # tests exercise 60-second deadlines in virtual time rather than sleeping.
 #
-# It does NOT reach the shipped binary. The workspace sets `resolver = "2"`,
+# It does not reach the SHIPPED BINARY, which is a narrower claim than "is never
+# compiled into the binary": `cargo test --all-targets` does build the bin target
+# with the feature on, because test-time features unify across targets. That is
+# harmless, since `test-util` only changes behaviour when the clock is explicitly
+# paused. The workspace sets `resolver = "2"`,
 # which stops dev-dependency features being unified into the normal build;
 # verified rather than assumed with
 # `cargo tree -p riverctl -e features,no-dev | grep test-util`, which finds
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "riverctl"
-version = "0.2.16"
+version = "0.2.17"
 edition = "2021"
 authors = ["Freenet Project"]
 description = "Command-line interface for River decentralized chat on Freenet"
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -102,5 +102,11 @@ freenet-test-network = "0.1.24"
 tempfile = "3"
 assert_cmd = "2"
 # `test-util` gives `#[tokio::test(start_paused = true)]`, so the response-wait
-# tests can exercise 60-second deadlines in virtual time instead of real time.
+# tests exercise 60-second deadlines in virtual time rather than sleeping.
+#
+# It does NOT reach the shipped binary. The workspace sets `resolver = "2"`,
+# which stops dev-dependency features being unified into the normal build;
+# verified rather than assumed with
+# `cargo tree -p riverctl -e features,no-dev | grep test-util`, which finds
+# nothing, against one hit for the same command with dev edges included.
 tokio = { version = "1.42", features = ["full", "test-util"] }
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -930,10 +930,26 @@ async fn await_response<T, S: ResponseSource>(
         //     of budget and cancelled part-way through reassembling a streamed
         //     `GetResponse`. The old shape could not produce that.
         //
-        // Neither is a desync: the loss is bounded to the one response. Both
-        // are properties of wrapping a non-cancel-safe `recv` in a `timeout`,
-        // which every call site here did before and after; freenet/river#692
-        // is where that stops being the shape of this code at all.
+        // Neither is a REGRESSION, and that conclusion matters more than the
+        // shape, because without it this note reads like a bug someone should
+        // go and fix. The precondition for a short final budget is at least one
+        // step-over, and a step-over means a non-answer arrived first, which on
+        // every one of these call sites is precisely what made the old code
+        // return `Err` at t=0. The two conditions cannot both be exercised:
+        // there is no input on a converted path where the old code succeeds and
+        // this one fails. A GET with no interleaving is identical either way,
+        // since the first iteration gets the whole window.
+        //
+        // What changes is DIAGNOSIS, not outcome: the failure moves from "fails
+        // at once, naming the response" to "fails at the deadline", and nothing
+        // in the error says a reassembly was cut. The count and last-kind in
+        // the timeout message above are the partial answer to that.
+        //
+        // A minimum-budget floor was considered and rejected: it would make the
+        // common case worse to improve a rare one, refusing an already-available
+        // response when the remaining budget is small, and it would only move
+        // the threshold rather than remove the hazard. The loss is bounded to
+        // one response and is not a desync.
         //
         // What IS avoidable is constructing one more `recv` that is guaranteed
         // to be cancelled, so avoid that:
@@ -12050,13 +12066,25 @@ mod response_multiplexing_pin {
     /// `CONNECTION_ACQUISITIONS` counts regardless of receiver. The Arc-clone
     /// shape moves both.
     ///
+    /// Measured against evasion shapes rather than argued. Planting each into
+    /// production source and running the live pin:
+    ///
+    /// | shape | caught by |
+    /// |---|---|
+    /// | cloned `Arc`, locked under another name | field, lock |
+    /// | `try_lock()`, no await at all | field ONLY |
+    /// | `tokio::sync::Mutex::lock(&self.web_api).await` | field, recv |
+    /// | helper accessor returning the `Arc`, two users | field, lock, recv |
+    ///
+    /// The `try_lock` row is why both counts run rather than whichever looks
+    /// stronger: it moves no lock count and no recv count, and the field count
+    /// is the only thing that sees it.
+    ///
     /// Residual, since claiming completeness here is what went wrong twice
-    /// already: these are still string counts. `try_lock()` and
-    /// `blocking_lock()` would evade the second (both are absent today, and
-    /// either would still have to name the field). The counts are also NET, so
-    /// a one-for-one swap passes. The check that would actually bind this is
-    /// not a count at all: make the raw connection unreachable by wrapping it
-    /// in a type whose only methods are the waiters. That is freenet/river#692.
+    /// already: these are still string counts, and they are NET, so a
+    /// one-for-one swap passes. The check that would actually bind this is not
+    /// a count at all: make the raw connection unreachable by wrapping it in a
+    /// type whose only methods are the waiters. That is freenet/river#692.
     const CONNECTION_ACQUISITIONS: usize = 20;
 
     /// See [`CONNECTION_ACQUISITIONS`]. One higher than the lock count, because
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -64,6 +64,14 @@ const POINTER_GET_TIMEOUT: Duration = Duration::from_secs(10);
 /// we asked for (freenet-core#4970 is the same race on the SUBSCRIBE
 /// acknowledgement). Bounded so a chatty subscription cannot starve the wait.
 const MAX_UNRELATED_RESPONSES_DURING_POINTER_GET: usize = 16;
+/// How long to wait for the node to answer an update we sent.
+const UPDATE_RESPONSE_TIMEOUT: Duration = Duration::from_secs(60);
+/// How long to wait for the node to answer a PUT we sent.
+const PUT_RESPONSE_TIMEOUT: Duration = Duration::from_secs(60);
+/// How long to wait for the room-state GET behind `accept invitation`.
+const ACCEPT_INVITATION_GET_TIMEOUT: Duration = Duration::from_secs(60);
+/// How long to wait for a room-state GET addressed by contract id.
+const GET_STATE_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// The BLAKE3 code hash of the room-contract WASM this binary bundles.
 ///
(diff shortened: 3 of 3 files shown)


sanity:
## Review round 2, and what it changed

Four independent reviewers on `9028ccf1`: `codex` (external, non-Claude) plus
three blind Claude lenses (skeptical, testing, code-first). Every finding is
either fixed in `3e8ecaf1` / `2e2013eb` or answered below. Nothing was
dismissed.

The short version: **the code mostly survived, the claims about it did not, and
the tests did not.**

### The finding that mattered

The testing lens replaced `await_response`'s step-over arm with the old
`return Err(anyhow!("Unexpected response type"))` and **all 368 tests stayed
green**. That is the same objection that rejected the first version of this PR
("the tests do not bind the fix"), restated one level up, and consolidating ten
call sites into one waiter had made it *worse*: the bug went from ten homes to
one, and that one was as unguarded as the ten had been, so a single line could
reinstate it everywhere at once.

The loop was untestable because it took `&mut WebApi`, and `WebApi` needs a real
socket. It now takes `&mut S: ResponseSource`, a two-line trait implemented for
the `MutexGuard` every caller already holds, so **no call site changed** and the
lines the pin scans for are byte-identical. `mod await_response_tests` drives it
against a scripted connection under `start_paused` virtual time: 9 tests, 0.00s,
no socket, no node, no real clock. The first is the v0.2.135 regression end to
end, which nothing reproduced until now.

Every one was red-checked against a mutation of the real source:

| Mutation | Before | Now |
|---|---|---|
| revert the step-over | 0 failed | 7 of 9 fail |
| delete the deadline guard | 0 failed | 1 fails |
| recompute the deadline per message | 0 failed | 1 fails |
| delete the `NotFound` arm | 0 failed | 1 fails |
| drop the stepped-over count | n/a | 1 fails |

The deadline guard turned out to be load-bearing rather than defensive:
`tokio::time::timeout` polls the inner future *before* checking its own
deadline, so with zero remaining an immediately-available response is returned
rather than timed out, and without the guard the loop keeps consuming messages
after it has given up.

### A real defect, from codex

`republish_room`'s classifier omitted `UpdateResponse`. Confirmed in freenet
0.1.188 `contract/executor/runtime.rs::perform_contract_put`: when the node
already holds the contract it merges rather than stores, and **both** exits of
that branch answer `UpdateResponse`, not `PutResponse`. Republishing is by
definition a PUT onto a contract that already exists, so that is the common
answer there. The old code called it "Unexpected response type"; this one would
have waited out the full 60 seconds. Fixed, and the classifier is now
`classify_put_response` / `PutAck` so it can be tested at all.

### The pin was defeated, twice

Check 3 searched for the literal `web_api.recv()`. The skeptical lens added a
raw read in production code and bound the guard to `api`: count stayed at 11,
pin reported clean. **A count keyed on a caller-chosen name is a spelling ban
wearing a count's clothes**. That is the exact trap the membership pin
documents for its own scans, twenty lines above where I wrote one. It counts
`.recv()` now, and its meta-test plants a read named `conn`.

Check 2 (`ContractRequest::Update {`) is defeatable the same way via a `use`
import or a shared builder, and that one cannot be fixed by picking a better
string. So there is now a **check 4** counting `self.web_api.lock().await`,
the one count not keyed on a caller-chosen name, since the field is private to
the file and every use of the connection must pass through it. Its meta-test
plants a *send-only* call site, which checks 1 to 3 all miss by construction.

Also fixed: `mutate()` used `str::replace` and fired three times where one was
intended; `assert_caught` accepted a broken *scan* as a fired *check*.

### Two classifiers had no coverage at all

`republish_room`'s (a closure, so unreachable, meaning "any response is success" would
have passed) and `await_get_response`'s key guard. The second is the worse one:
without it a GET returns **another room's state**, which is worse than the
failure this PR set out to fix. Both now have tests, both red-checked.

### Claims corrected

Five source comments asserted more than the code delivers. Listing them because
the pattern is the interesting part, not any one of them:

1. *"riverctl never has two updates in flight"* is false on the timeout path, and
   the code-first lens went further than the finding: `heal_member_info`
   resolves the **same** contract key as its caller and swallows its error, so
   the PR's own worked example is a same-key case the key guard does not cover.
   Pre-existing; now stated plainly and filed as #691.
2. *"every path that sends an `Update` waits through `await_update_response`"*:
   seven of the eight, and the constant twenty lines below said so.
3. The no-cap justification appealed to drop-versus-collect, which does not
   distinguish it from the pointer GET's cap. The real reason is the cost of
   giving up: that one has a free fallback, these have none.
4. The inventory said all six deferred sites are deferred for being *tolerant*.
   Two are not: `create_room` is fatal (held back because its two success arms
   persist different things), and `migrate_room_to_new_contract` writes an
   unrelated notification's key into local storage, which is data corruption.
5. `await_response`'s doc read as though it closed the multiplexing class. It
   closes half: a recv **error** carries no key, so an error belonging to
   somebody else's request still fails ours.

### Not fixed here, deliberately

- **#691**: a timed-out wait leaves its answer outstanding; the next same-key
  send can consume it. Needs the connection drained on failure, or a request id
  on the wire. Documented at `classify_update_response`.
- **#690 / #689**: the six call sites that still read the first message.
- The counts are **net**: converting one raw read while adding another leaves
  the total unchanged. Stated in the pin's doc rather than papered over.

### Verification

383 tests pass (368 at `9028ccf1`, 358 on base). `cargo fmt --check` clean.
Clippy on `api.rs`: 4 warnings, all pre-existing, one fewer than base.

Reviewers: `codex review --base origin/main`, `freenet:skeptical-reviewer`,
`freenet:testing-reviewer`, `freenet:code-first-reviewer`. The testing lens
supplied a working proof-of-concept patch for the seam, which this adopts with
one change: its own stated limitation (a per-iteration deadline reset still
passed) is closed by giving the scripted connection per-response arrival delays.

[AI-assisted - Claude]

sanity:
## Review round 3

`codex review --base origin/main` plus a fresh `freenet:skeptical-reviewer` on
`82c8e273`, then a follow-up pass on `78ba95a3`.

**Codex: clean.** *"The shared response waiters correctly filter multiplexed
responses by contract identity, preserve bounded deadlines, and are applied
consistently to the targeted request paths. No actionable regressions."*

**The skeptical lens found four Should-Fix, and three of the four were in code
added to answer round 2.** That is the signal this PR should stop growing, and
it is why the remaining structural work is #692 rather than more commits here.

### 1. The PUT path's wiring was bound by nothing

Replacing `republish_room`'s classifier call with `|_| Some(Ok(PutAck::Stored))`,
which accepts **any** response as a successful republish, left 383 of 383 green.
`classify_put_response` has unit tests; nothing said the call site still called
it.

This is the **third** instance of one shape in this change: the pin binds the
routing, the unit tests bind the predicate, and the join between them is
unbound. First the loop, then `await_get_response`'s key guard, now this. Each
time it reappeared in the code written to fix the previous one.

Fixed by check 1b (`CLASSIFIER_WIRING`), covering the PUT and both GETs, with
the reviewer's mutation as its meta-test.

### 2. Check 4 was a spelling ban, and its comment said it was not

It claimed to be "the one count not keyed on a caller-chosen name", justified by
the field being private. Private is not unaliasable: it is an
`Arc<Mutex<WebApi>>`, so

```rust
let api = self.web_api.clone();
let mut conn = api.lock().await;
```

reaches the connection under any name, and a site that sends *and* reads raw
through it was invisible to all four checks. `NodePointerIo` is that pattern
already, counted only because its field happens to be called `web_api`.

**Three defeats in three rounds, each by a spelling the needle did not
anticipate.** So check 4 now counts both chokepoints every path must pass:
naming the field, and taking the lock. Verified by planting each shape into
production source and running the live pin:

| shape | caught by |
|---|---|
| cloned `Arc`, locked under another name | field, lock |
| `try_lock()`, no await at all | **field only** |
| `Mutex::lock(&self.web_api).await` | field, recv |
| helper accessor returning the `Arc`, two users | field, lock, recv |

The `try_lock` row is why both run rather than whichever looks stronger.

More importantly the comment stopped claiming completeness. The check that would
actually bind this is not a count: **#692** proposes making the raw connection
unreachable behind a type whose only methods are the waiters.

### 3. `PutAck::Merged` is not "the common answer"

Measured in freenet 0.2.133: `contract/executor/runtime/contract_ops.rs` returns
`UpdateResponse` from both exits of its already-exists branch, carrying the
comment *"Re-PUT into an existing contract returns an UpdateResponse"*. But
every **network** exit answers `PutResponse` (`operations/put/op_ctx_task.rs`).
So it is the local executor's answer only. Accepting it is still necessary
(`freenet local` is River's routine dev mode), but the comment was wrong for a
networked node, and cited a stale 0.1.188 path.

### 4. The cancellation claim was falsified

My comment said the non-cancel-safe-`recv` exposure was *unchanged* from base.
It is not, in two ways: base cancelled **nothing** on an unrelated response (its
`recv` completed and it errored at once), and base always gave its single `recv`
the whole window where this gives the last one a remainder, which can cut a
streamed `GetResponse` off with milliseconds left.

The reviewer then settled whether that matters, and the answer is worth
recording: **it is not a regression.** The precondition for a short final budget
is at least one step-over, and a step-over means a non-answer arrived first,
which is exactly what made base fail at t=0 on every one of these sites. The two
conditions cannot both be exercised, so there is no input on a converted path
where base succeeds and this fails. What changes is *diagnosis*, not outcome.

A minimum-budget floor was considered and **rejected**: it would refuse an
already-available response when the remaining budget is small, making the common
case worse to improve a rare one, and would only move the threshold rather than
remove the hazard. Deliberately not filed as an issue, since filing one for a
non-regression invites someone to land exactly that floor. The reasoning is in
the comment instead.

### Also fixed

- `await_response`'s 37-line doc comment had ended up documenting
  `trait ResponseSource` (round 2 inserted the trait between the doc and its
  function). Nothing catches this: no blank line, so clippy stays quiet, and
  rustdoc renders it onto the wrong item.
- `?Sized` was dead syntax on all three signatures.
- `unreachable!` was a production panic on a state only inspection rules out.
- `nothing_is_read_after_the_deadline_passes` died at `unwrap_err` before its own
  message could print. That mattered, because under the mutation it guards, the
  wait *succeeds*.
- Four pure classifier tests sat in the module documented as testing the wait
  rather than the classifiers, one of them `async` with nothing to await.
- "checkable against a plain grep" was false (grep gives 22; the pin strips test
  literals first).
- `create_room` has three success arms, not two, and the bare-`Ok` one silently
  drops invitation secrets.
- The tokio note now claims only "does not reach the shipped binary", since
  `--all-targets` does build the bin with `test-util` on.

### Verification

385 tests pass (358 on base). `cargo fmt --check` clean. Clippy on `api.rs`: 4
warnings, all pre-existing, one fewer than base. Every new test red-checked
against a mutation of the real source.

[AI-assisted - Claude]

sanity:
## Correction: the v0.2.135 announcement was NOT lost

This PR's description, its commit messages, and two source comments say the bug
"broke the freenet v0.2.135 release announcement". **That is wrong, and I am
correcting it here rather than leaving it in the record.**

The announcement posted. It is in the room, once, at 12:49:43 CDT, and a member
replied to it 29 minutes later. I found this by reading the room back before
re-running the announcement, which I had been about to do.

The unfiltered journal gives the exact sequence:

```
[17:49:41Z] posting to River room 4uNUKFzZQCnzo4K2ecZ16cMsYEEfoaRS35z6exEsbvm4 (msg 88 bytes)
17:49:54  Error: Unexpected response type: ContractResponse(UpdateNotification { ... })
[17:49:54Z] FAILED with rc=1
```

`FAILED with rc=1` is `send_message_checked`'s log, so the SEND is what reported
failure, and the script exited there without reaching `verify_converged`.

### What the bug actually is

A **false negative**: the update was transmitted, applied by the node, and
converged into room state, and riverctl reported it as a failure.

That is the inverse of freenet-core#4318, which is the failure this tooling was
hardened against: riverctl printing "Message sent successfully" while the
contract silently dropped the delta, losing three announcements. Every safeguard
in `announce-to-river.sh` points at that direction. Nothing points at this one.

### Does the fix still matter? Yes, and arguably more

The severity moves rather than disappearing. A send that reports failure on
success is worse than a cosmetic bug, because **the operator retries**, and a
retry posts a duplicate release announcement to a public room. That retry was
one command away here.

Everything else in this PR stands: the race is real, the fatal `_` arm was real
at ten call sites, and the interleaving that triggered it is ordinary traffic
rather than a corner case.

### What I got wrong, and how

I inherited "the announcement never posted" from a handoff summary and carried
it through three review rounds without once checking the room. It was an
inference from a script's exit code, and the exit code was exactly the thing
under investigation. Four reviewers examined the code and none of them could
have caught this, because it was never a claim about the code.

The lesson is the one in this repo's own testing rules: **a true artifact can
stand in for a fact it does not establish.** `announce-to-river.sh exited 1` was
true. "The announcement is missing" did not follow from it, and the room was
readable the whole time.

### Follow-ups

- **freenet-core#5640** filed: on a send failure the script should run its
  existing convergence check before reporting failure, so a false negative
  cannot invite a duplicate retry.
- The two source comments in `cli/src/api.rs` that repeat the wrong claim will
  be corrected in a small follow-up PR.

[AI-assisted - Claude]

Issue #695, opened 2026-09-12
riverctl: a room-contract re-key hard-stops every bot's writes, because nothing publishes a write-compatibility floor


sanity:
## The question

@sanity, 2026-09-12: riverctl should warn when it needs upgrading — but if it is using freenet-migrate, it ought to be able to handle new contract generations *without* necessarily upgrading.

The first half is done. The second half is not, and the reason is worth stating precisely.

## What freenet-migrate's pointer already solves, and what it does not

Since #627, riverctl resolves River's pointer record and re-anchors its room-contract code hash at runtime (`cli/src/pointer.rs`). That solves **addressing**: riverctl can derive the correct key for a generation whose WASM it has never seen, which is exactly why `KeyIntent::Read` is permitted against an unknown generation (`pointer.rs:258`).

It does not solve **compatibility**. Nothing in the pointer record says whether a delta encoded by an older `river-core` is still accepted by the new contract code. So `authorize()` refuses every write against an unknown generation (`pointer.rs:275-280`):

> Refusing to write with a room-contract generation this riverctl does not know. Reads still work; run `cargo install riverctl` (or update your package) to write.

## Why the refusal is defensible, and why it is still too blunt

Defensible, because the failure it prevents is silent. On the UPDATE path the contract runs `apply_delta`, never `verify` (`contracts/room-contract/src/lib.rs::update_state`), and `MessagesV1::apply_delta`'s author retain **drops** content rather than erroring — see the analysis already written up at `cli/src/api.rs:1275-1283`. A delta a newer contract does not like is discarded on the network with a successful-looking `UpdateResponse`. Refusing beats writing blind.

Too blunt, because the evidence says these re-keys have never actually broken the delta format. `CURRENT_STATE_VERSION` is still `1` (`common/src/room_state/version.rs:13`) after **32 room-contract re-keys** (`common/legacy_room_contracts.toml` holds 32 superseded generations). The re-keys are rustc bumps, dependency bumps, and contract logic fixes (#673 being the most recent). So today every routine release hard-stops every riverctl bot's writes until its operator upgrades, to guard against a breakage that has not once occurred — and the operator has no way to tell a benign re-key from a real one, because nobody publishes that distinction.

## What is missing: a published write-compatibility floor

The knowledge of whether a re-key breaks older writers exists exactly once, in the head of whoever cuts the release. It is never written down anywhere a client can read. Two directions, ideally both:

### 1. Publish the floor next to the code hash

[...]

PR #696, opened 2026-09-12
fix(cli): follow a mid-run room-contract re-key instead of going silently deaf

### LINKED ISSUE #694: riverctl: long-running `message stream` never re-resolves the room-contract pointer, so bots go silently deaf across a re-key
## Symptom

Reported by @Ivvor on Matrix (2026-09-12):

> Might be useful if River could warn users when the underlying contract ID for a room changes. It screws up riverctl-based bots and they usually need to be restarted to work properly.

The restart is the tell. A bot that is restarted starts working again, which means nothing is wrong with the bot or the room — riverctl simply never notices the re-key within the life of one process.

## Why it happens

riverctl resolves River's room-contract pointer record **at most once per process**, and caches the result for the process lifetime:

- `ApiClient::room_anchor()` wraps `resolve_room_anchor()` in a `tokio::sync::OnceCell` (`cli/src/api.rs:2384-2388`, field declared `api.rs:2334`).
- `main.rs` builds exactly one `ApiClient` per invocation (`cli/src/main.rs:173-179`) and runs one subcommand against it.
- No timer or interval re-invokes resolution. Every other call site (`api.rs:2471, 2490, 2663, 3037, 3289, 3826, 4169`) reads the cached value.

That is correct for a one-shot command and wrong for the two commands that are explicitly documented as the bot integration surface (`cli/README.md:127,294`, `cli/CHANGELOG.md:11`):

**`message stream --subscribe`** (`cli/src/api.rs:6203-6504`) derives the contract key once at `api.rs:6244-6247`, SUBSCRIBEs to that instance id (`api.rs:6317-6388`), then loops forever (`api.rs:6399`). After a re-key the subscription stays bound to the retired instance id. New messages are written under the new key, which this process never subscribes to, so the node has nothing to send. The loop's receive timeout arm (`api.rs:6499-6501`) treats that as "no new messages" — **indistinguishable from a quiet room, forever**.

**`message stream` (polling mode)** (`cli/src/api.rs:5114-…`) is no better. Each poll calls `get_room` (`api.rs:5222`) → `ensure_room_migrated` (`api.rs:2816`) → the same cached anchor (`api.rs:3826`), so every poll re-derives the same stale key. GETs against the retired contract keep **succeeding** — it still exists, it is just frozen — so polling never errors either.

The auto-migration path is inert here too: it keys off `room_code_hash`, itself a `OnceLock` set once from the same frozen anchor (`cli/src/storage.rs:319-323, 369-371`; set at `api.rs:2455`).

## Why this is worse than it looks

The failure is silent in the direction that matters. The bot's process is healthy, its socket is open, its GETs return 200-equivalents, and its logs show nothing. An operator has no signal short of noticing that a busy room went quiet. Re-keys are routine — any `common/` change, stdlib bump, or contract fix moves the WASM — so this recurs on roughly every release.

## Possible directions

1. **Re-resolve the pointer periodically in long-running commands**, and on a change: log it loudly, re-SUBSCRIBE to the new instance id, and continue. This is the fix that matches what the user is asking for — the bot keeps working without a restart.
2. **At minimum, detect and fail loudly.** Re-resolve on an interval; if the anchor moved, print a clear message naming both code hashes and exit non-zero, so a supervised bot restarts itself. Strictly worse than (1) but much better than silence, and far smaller.
3. **A liveness signal on the subscription.** Independent of the pointer: a stream that has received nothing for N minutes could re-check that the key it is subscribed to is still the live one. Catches the same failure from the other side.

Note that (1) only keeps a bot fully working if it can still **write** after the re-key, which today it cannot — see the companion issue on the missing write-compatibility signal. Reads recover under (1) on their own.

## Related

- #580 — the same failure in the UI (a stranded browser on a cached webapp), still open.
- The pointer machinery itself is live and correct: I confirmed a GET of `Ai4VLoC2jGdhpcB2UU8VPo3efUoxjm1Ju9VKXqRC63Az` returns version 2 / code hash `a3e63c8c…`, matching `pointer-records.toml` and the WASM riverctl 0.2.16/0.2.17 bundles. This issue is about *when* riverctl consults it, not whether it works.

[AI-assisted - Claude]

sanity:
Closes #694.

## The bug

riverctl resolves River's room-contract pointer **once per process** and never again. That is right for a one-shot command and wrong for the two commands documented as the bot integration surface.

After a re-key:

- **`message stream --subscribe`** stays attached to the retired contract instance id. The node has nothing left to send for a generation nobody writes to, and the loop's receive-timeout arm treats that as "no new messages" — indefinitely.
- **`message stream` (polling)** re-derives the same stale key on every poll, and GETs against the retired contract keep **succeeding**, because that contract still exists; it is just frozen.

Neither surfaces an error. The process is healthy, the socket is open, the logs are clean, and the room simply appears to go quiet. Restarting fixes it, which is exactly the symptom @Ivvor reported on Matrix.

## The fix

Both loops re-resolve the pointer every five minutes. When the generation actually moves:

- the subscription re-subscribes to the new instance id,
- a notice naming both generations goes to stderr,
- and the stream explicitly catches up on what was written during the gap — the node only notifies about changes *after* a subscription, so skipping that would leave a silent hole of precisely the kind being fixed.

Reads follow the move even when the new generation is one this riverctl is too old to write to. That is the case the pointer record exists to make survivable, and refusing to follow it would be the original bug wearing a safety justification. (Writes against an unknown generation are still refused — that is #695, handled separately.)

## Supporting changes

| Change | Why it is load-bearing |
|---|---|
| `Storage::room_code_hash`: `OnceLock` → `RwLock<Option<..>>` | The second resolution was silently dropped, pinning key regeneration to startup. |
| `ApiClient::room_anchor`: `OnceCell<RoomAnchor>` → `RwLock<Option<RoomAnchor>>` returning a clone | The cache has to be replaceable. One-shot commands still resolve exactly once. |
| SUBSCRIBE handshake and fetch-and-emit extracted into helpers | So the second subscribe site cannot reimplement the ack-race handling (freenet-core#4970) slightly differently, and the three monitor paths cannot drift on what counts as a reportable change. |

The advisory is compared against the previous one before printing, so a five-minute refresh cannot turn a real warning into background noise.

## On the modified guard

`monitor_edit_detection_wiring_pinned` counted two emitter call sites, one per loop. Routing both loops through one shared step reduces that to one, which would read as a regression while the property is in fact better enforced. It is **re-pointed, not relaxed**: it now pins both halves — every monitor path reaches the shared step, and the shared step still runs all three emitters — which additionally covers the third caller this PR adds.

## Verification

- `cargo test -p riverctl`: 389 + 19 pass, 0 fail.
- Both new guards were verified by reintroducing the bug and watching them fail:
  - restoring `OnceLock` semantics to `set_room_code_hash` fails `a_second_resolved_generation_replaces_the_first`;
  - deleting one emitter from the shared step fails `monitor_edit_detection_wiring_pinned`.
- No new clippy warnings (16 before, 16 after; all pre-existing).
- Separately confirmed the live pointer record is current: a GET of `Ai4VLoC2jGdhpcB2UU8VPo3efUoxjm1Ju9VKXqRC63Az` returns version 2 / code hash `a3e63c8c…`, matching `pointer-records.toml` and the bundled WASM.

## What is NOT covered

The five-minute window is a bound, not an elimination: a bot can still talk to a retired generation for up to five minutes after a re-key. Reads recover on their own; writes do not, because riverctl still refuses to write against a generation it does not know — see #695.

[AI-assisted - Claude]

https://claude.ai/code/session_01LSj17em5VJvYN4ZYCRJVdF

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -18,6 +18,24 @@ All notable changes to riverctl will be documented in this file.
   on anything written during the gap. Reads follow the move even when the new
   generation is one this riverctl is too old to write to. (freenet/river#694)
 
+  Details worth knowing if you run a bot:
+  - The catch-up fetch runs after **every** refresh, not only after a re-key,
+    because the pointer GET shares the node connection with the subscription and
+    steps over (discarding) frames while it waits for its own answer. One of
+    those can be an update notification for the room being streamed.
+  - A re-SUBSCRIBE the node will not yet accept is **not** fatal. The node most
+    likely to refuse is one that does not hold the newly-published generation
+    yet, so riverctl retries every 30s and keeps polling meanwhile rather than
+    exiting.
+  - Only a signature-verified pointer record can move a stream. A refresh that
+    merely timed out can name a different hash (the fallback is whatever was
+    last persisted, or the bundled one), and acting on that would announce a
+    re-key that never happened and re-subscribe to a retired generation.
+  - A signed **withdrawal** of the pointer record ends the stream, rather than
+    being swallowed as a transient failure like every other resolution error.
+  - The interval carries ±20% jitter, so a fleet of bots started together does
+    not hit the network in one synchronised burst after a re-key.
+
 ## [0.2.15] - 2026-09-06
 
 ### Fixed
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -27,10 +27,16 @@ All notable changes to riverctl will be documented in this file.
     likely to refuse is one that does not hold the newly-published generation
     yet, so riverctl retries every 30s and keeps polling meanwhile rather than
     exiting.
-  - Only a signature-verified pointer record can move a stream. A refresh that
-    merely timed out can name a different hash (the fallback is whatever was
-    last persisted, or the bundled one), and acting on that would announce a
-    re-key that never happened and re-subscribe to a retired generation.
+  - A stream moves only when the move is both **evidenced and actionable**. A
+    refresh that merely timed out can name a different hash, and acting on that
+    would announce a re-key that never happened and re-subscribe to a retired
+    generation. And a *verified* re-key to a generation this riverctl does not
+    know is reported but **not** followed: the new key holds nothing this binary
+    can reach (the backward probe will not search from a generation it does not
+    know, and writing is refused), so following it would trade a subscription
+    that is still delivering for a key it cannot act on — leaving the stream
+    silent on both. In that case riverctl stays where the data is and tells you
+    to run `cargo install riverctl --force`.
   - A signed **withdrawal** of the pointer record ends the stream, rather than
     being swallowed as a transient failure like every other resolution error.
   - The interval carries ±20% jitter, so a fleet of bots started together does
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 All notable changes to riverctl will be documented in this file.
 
+## [Unreleased]
+
+### Fixed
+- `message stream` (both `--subscribe` and polling modes) now re-checks River's
+  room-contract pointer every five minutes instead of once at startup, and
+  follows a re-key while running. Previously a long-running bot resolved the
+  pointer once, stayed bound to the retired contract for the life of the
+  process, and went silently deaf: the node has nothing left to send for a
+  generation nobody writes to, and a poll against the retired contract still
+  succeeds because that contract still exists. Neither mode surfaced any error,
+  so the only symptom was a busy room appearing to go quiet, and the only fix
+  was restarting the bot. On a re-key the subscription now re-subscribes to the
+  new generation, prints a stderr notice naming both generations, and catches up
+  on anything written during the gap. Reads follow the move even when the new
+  generation is one this riverctl is too old to write to. (freenet/river#694)
+
 ## [0.2.15] - 2026-09-06
 
 ### Fixed
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -35,6 +35,11 @@ All notable changes to riverctl will be documented in this file.
     being swallowed as a transient failure like every other resolution error.
   - The interval carries ±20% jitter, so a fleet of bots started together does
     not hit the network in one synchronised burst after a re-key.
+  - The stream's **periodic** fetch no longer migrates the room or self-heals
+    `member_info`. Those are writes, they still happen on the stream's first
+    fetch and in every one-shot command, and issuing them on a timer as a side
+    effect of reading was both surprising and — because of a known hazard when a
+    migration's GET races a live notification — unsafe.
 
 ## [0.2.15] - 2026-09-06
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -35,11 +35,17 @@ All notable changes to riverctl will be documented in this file.
     being swallowed as a transient failure like every other resolution error.
   - The interval carries ±20% jitter, so a fleet of bots started together does
     not hit the network in one synchronised burst after a re-key.
-  - The stream's **periodic** fetch no longer migrates the room or self-heals
-    `member_info`. Those are writes, they still happen on the stream's first
-    fetch and in every one-shot command, and issuing them on a timer as a side
-    effect of reading was both surprising and — because of a known hazard when a
-    migration's GET races a live notification — unsafe.
+  - The stream's **periodic** fetch no longer writes: it does not migrate the
+    room, does not self-heal `member_info`, and does not republish state it
+    recovered from an older generation. All three still happen on the stream's
+    first fetch and in every one-shot command. Issuing them on a timer as a side
+    effect of reading was surprising, and — because of a known hazard when a
+    migration's GET races a live notification — unsafe. The periodic fetch still
+    READS across older generations, which is what makes a catch-up straight
+    after a re-key return anything at all.
+  - `message stream` without `--subscribe` now does one full fetch at startup
+    regardless of `--initial-messages`, so migration and the `member_info`
+    self-heal still run in the default polling mode.
 
 ## [0.2.15] - 2026-09-06
 
--- a/cli/src/api.rs
+++ b/cli/src/api.rs
@@ -775,6 +775,23 @@ impl Retry {
         self.state = None;
         self.failures = 0;
     }
+
+    fn is_armed(&self) -> bool {
+        self.state.is_some()
+    }
+
+    /// Keep the work armed, but with the cadence starting from `now` and the
+    /// failure penalty cleared — "succeeded, and needed again one interval from
+    /// now" rather than "succeeded, stop".
+    ///
+    /// Distinct from [`Self::arm`], which makes the next attempt IMMEDIATE.
+    /// Using `arm` here would poll on the loop's 500ms tick; using nothing would
+    /// let the work idle. This is the steady-state case: a job that has to keep
+    /// running on its own schedule for as long as something else is unfinished.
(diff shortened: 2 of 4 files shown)


sanity (commented):
## Comprehensive PR Review: #696

### Summary
- **PR Title:** fix(cli): follow a mid-run room-contract re-key instead of going silently deaf
- **Type:** fix
- **Linked Issues:** #694
- **Review tier:** **Full** — concurrency/async (a `OnceCell`/`OnceLock` → `RwLock` cache replaced mid-run, a periodic refresh racing a live push subscription) is a high-risk surface.
- **Reviewers run:** `code-first`, `skeptical`, `testing`, `big-picture` (all four, blind, in parallel) + **external model: `codex review --base origin/main`** (ran successfully; no fallback needed).
- **HEAD reviewed:** `9c1cb3a4`. **All findings below are now fixed in `65839168`**, which requires a fresh review pass before merge.

---

### Code-First Analysis
**Alignment:** matches. Read blind, the code reads as "make the pointer anchor replaceable, re-resolve it periodically in both stream loops, re-subscribe on a move" — which is what the description claims and what #694 asks for. The PR's own "What is NOT covered" section was accurate as far as it went.

**Gap:** it did not mention that the refresh mechanism can itself eat a live update notification. That is a distinct failure mode from plain staleness (finding 2).

---

### Testing Assessment

| Unit | ⚠️→✅ | Five new tests, none vacuous; two gaps found and now closed |
| Integration | ❌ | No mock transport exists in this codebase; the loop sequence remains untested end-to-end |
| Simulation / E2E | N/A | |

**Regression test:** present and valid. `a_second_resolved_generation_replaces_the_first` was confirmed to fail against the old `OnceLock` semantics.

**Gaps found (both now closed):** no test distinguished "compares `code_hash` only" from "compares the whole struct" — a mutation to `p != anchor` passed everything; and nothing pinned catch-up continuity across the re-key boundary, the behaviour an operator notices first (duplicate spam, or missed messages).

---

### Skeptical Findings
**Risk Level:** high (as submitted)

| Concern | Severity | Location | Status |
|---|---|---|---|
| A refused/timed-out re-SUBSCRIBE propagated out and **killed the bot** — one line after printing "no restart needed". The node most likely to refuse is one that does not hold the just-published generation yet, so the first bot to notice a re-key was the one that exited. | **High** | `api.rs` re-subscribe call | Fixed: retried on a 30s cadence, never fatal, and the catch-up is now sequenced *before* it since it needs no subscription |
| The 5-minute pointer GET **discards** `UpdateNotification`s for the room being streamed (`Ok(Ok(_)) => continue`). Harmless before this PR, because resolution only ran before anything was subscribed. | Medium | `api.rs:121-128` | Fixed: catch-up now runs after *every* refresh, so a full re-fetch supersedes anything dropped |
| A **failed** resolution could masquerade as a re-key. `moved` compared hashes only, but a failed refresh still yields an anchor on `last_known()` — the floor's hash, or the *bundled* one if the floor never persisted (best-effort). Result: a false "River re-keyed" notice, a re-subscribe to the **retired** generation, and a flap every 5 minutes, rewriting `rooms.json` each cycle. | Medium | `AnchorRefresh::new` | Fixed: a move now requires `AnchorSource::Pointer` |
| The guard **did not enforce what its docstring claimed** — proven empirically. | Medium | `storage.rs` | Fixed; see below |
| Refresh not interruptible (Ctrl+C latency 500ms → potentially ~100s) and no jitter on the interval (fleets burst in lockstep). | Medium | both loops | Fixed: `select!` on shutdown, ±20% jitter |
| Control flow gated on `move_notice()` — a *presentation* string. A later "don't repeat this line" tweak would silently disable re-subscription. | Low | subscribe loop | Fixed: gated on `moved()` |
| Catch-up failure logged at `debug!` and never retried | Low | subscribe loop | Fixed: `warn!`, and retried until one succeeds |
| Poisoning-recovery comment misstated how poisoning works (per-lock, not global) | Low | `storage.rs` | Fixed |

**Deadlock: none found.** The reviewer traced all 20 `web_api.lock()` sites and confirmed every one that re-enters anchor-touching code drops the guard first, and that the only `tokio::spawn`s are the two Ctrl+C handlers, neither holding an `ApiClient`. The `room_anchor.write() → web_api.lock()` ordering is real but unreachable without a second task. Storage's fs2 no-nesting rule is not violated.

---

### External Model (Codex)
Three P1s, converging independently with the Claude lenses on the dropped-notification hole, and adding two:
- catch-up must be **retried**, not one-shot — a transient failure right after a re-key loses the gap permanently, since the anchor has already advanced and no later refresh would look again;
- a **signed withdrawal** must stop the stream rather than being swallowed as a transient failure. It is now a distinct `PointerWithdrawn` error type, recognised by type rather than by matching its text.

---

### The modified guard — the thing most worth reading

`monitor_edit_detection_wiring_pinned` was re-pointed because routing both loops through one shared step reduces the per-loop emitter count from two to one. Big-picture read it as a legitimate strengthening. **The skeptical lens disproved that empirically**: the `contains` half searched the whole file, so deleting `Self::emit_deletions(...)` from the shared step and adding one call back in the polling loop alone left the test **passing** while deletions silently stopped surfacing on the other two paths — the exact #323 regression the guard exists to catch.

That is an ornamental guard: satisfiable without achieving its purpose. It now **slices the body of `emit_stream_changes`** and searches within it, and separately pins the load-bearing emitter ordering. Re-verified against the reviewer's exact mutation, which it now catches.

Two lessons recorded rather than glossed: a whole-file source grep is not containment, and a guard verified only by watching it pass is unverified.

---

### Documentation
- Code docs: complete; every non-obvious decision carries its reasoning.
- No stale rules/skills/docs found (`.claude/rules/*`, `FREENET.md`, `cli/README.md` all checked — none documents the once-per-invocation cadence).
- CHANGELOG updated with the operator-facing consequences.

---

### Recommendations

#### Must Fix (Blocking) — all now addressed in `65839168`
1. Non-fatal, retried re-subscribe, ordered after the catch-up.
2. Catch up after every refresh, retried until it succeeds.
3. A move requires signature-verified provenance.
4. A signed withdrawal ends the stream.
5. Slice the guard to the shared step's body.

#### Consider (not done, deliberately)
- **The retired subscription is never torn down.** `freenet-stdlib` 0.8.5 has no unsubscribe variant in `ContractRequest` (checked), so a stream following several re-keys leaves one idle subscription per retired generation. Documented in the code as an accepted leak rather than an oversight; worth an upstream request.
- **No end-to-end test of the loop sequence.** No mock transport exists in this codebase, and the two async tests that exist avoid the network via injected closures. The pieces are unit-tested and the wiring is pinned by source-grep; closing this properly means extracting the sequence behind an injectable closure, which is a larger refactor than this fix warrants.

---

### Verdict
**HEAD SHA reviewed:** `9c1cb3a4`

Findings 1-5 above are blocking and have been fixed in `65839168`. Per the per-code-content rule, that SHA has not been reviewed, so a fresh full pass runs before merge.

[AI-assisted - Claude]

sanity:
Closing without merging, in favour of a much smaller fix.

This PR tried to keep a long-running `message stream` alive across a room-contract re-key: re-check the pointer, re-subscribe to the new key, catch up on the gap, retry when the node refuses. Seven review rounds each found real bugs, and from round 4 on most of them were introduced by the previous round's fixes. The diff reached ~2,300 added lines for what is conceptually "check again every few minutes".

The root cause was the goal of "no restart needed". A fresh riverctl process already does everything a re-key requires, correctly, through code that has worked for months: it looks up the current generation, migrates the room, heals membership, and subscribes. This PR built a second, mid-run copy of all of that, and the bugs lived in the interactions inside it.

The replacement detects the move and exits with a distinct status so a supervisor restarts the bot, which then does the right thing through the existing startup path. That was option 2 in #694 all along.

The review threads here remain a useful record of the failure modes a mid-run approach has to handle, if anyone revisits it.

[AI-assisted - Claude]

Issue #689, opened 2026-09-10
migrate_room_to_new_contract persists a contract key taken from an unrelated UpdateNotification


sanity:
## `migrate_room_to_new_contract` can write the wrong contract key into local storage

`cli/src/api.rs:3932-3958`.

While awaiting the migration's response, the match accepts an `UpdateNotification` as "migrated successfully via merge" and then calls:

```rust
storage.update_contract_key(room_owner_key, &key)
```

**with the key taken from the notification, never compared against `new_contract_key`.**

The node multiplexes every response onto one connection, so an `UpdateNotification` for a *different* contract can arrive while this code is waiting. When it does, the wrong contract key is persisted against this room.

This is not a misreport that a retry corrects. It **corrupts local state**, and every later operation for that room resolves through the stored key.

## Why it is being filed now

Found during review of #688, which fixes a related race in `send_message_with_key`: that path took the first message off the connection as its answer, so an overtaking `UpdateNotification` made the send fail. That failure broke the freenet v0.2.135 release announcement.

This one is worse in kind. The send-path bug fails loudly. This one succeeds, silently, with wrong data written to disk.

It is filed separately because it is **not** the same fix. #688 and its follow-ups make call sites keep waiting instead of treating an unrelated response as the answer. Here the response is already tolerated by design (a merge genuinely can be reported as a notification); the defect is that the key inside it is trusted without comparison.

## Suggested fix

Compare before persisting: accept the notification as the migration's answer only when its key matches `new_contract_key`, and keep waiting otherwise. If a merge legitimately reports a different key, that case needs to be stated explicitly rather than reached by accident.

A regression test is straightforward once `classify_update_response`-style extraction exists (see #688): hand the classifier a notification for an unrelated contract and assert the migration does not treat it as its answer.

## Related, from the same review

Two further sites tolerate an unexpected response in a way that changes what counts as success, both pre-existing and neither corrupting state:

- `ensure_room_migrated` (`api.rs:3590`): the `_` arm at 3683 reads any non-`GetResponse` as "contract does not exist" and starts a full migration PUT.
- `accept_invitation_struct` (`api.rs:3455`): the join-delta update warns and proceeds as accepted (3463-3465), reporting "invitation accepted and membership published" with no acknowledgement.

Worth tracking together with this one; they are the same family as #688's finding 2 (a false success where there used to be a loud failure).

[AI-assisted - Claude]

PR #698, opened 2026-09-12
fix(cli): exit so a supervisor restarts `message stream` when River re-keys

### LINKED ISSUE #694: riverctl: long-running `message stream` never re-resolves the room-contract pointer, so bots go silently deaf across a re-key

sanity:
## Problem

`riverctl message stream` looks up River's room-contract address once, at startup. When River re-keys the room contract, every room's address changes and a running stream keeps listening to the old one. Nothing errors: the room just appears to go quiet until someone restarts the bot. @Ivvor reported exactly this on Matrix. (#694)

## Approach

Every few minutes the stream re-checks River's pointer record. If it names a different, **signature-verified** generation, the stream exits with **status 75**. A restarted riverctl then uses the current generation through the same startup path every command already uses.

- **Any verified move exits**, including to a generation newer than this riverctl. A stream only reads, and reads are permitted against any generation, so a restarted stale riverctl can stream the new one once the room has moved there. It also prints riverctl's existing "older than River's room contract" upgrade advisory.
- **An unverified hash change is ignored.** A re-check that can't reach the pointer falls back to a hash that can differ with no re-key at all, and exiting on it would turn a timeout into a restart.
- **The anti-rollback floor is also kept in memory**, so a failed save (a read-only config directory) can't let an older signed record look like a move. At equal versions it uses the same tiebreak as River's resolver.

Status 75 is `EX_TEMPFAIL`, distinct from the status 1 every other failure uses. The README tells bot operators to **restart on any failure with a delay**, shows systemd with `RestartSec`, and states that a bot with no restart wrapper now **stops** after a re-key instead of silently going deaf.

### Why exit instead of following the move

#696 tried to follow the re-key inside the running process. It reached ~2,300 lines and seven review rounds, and nearly every defect was in the interactions inside the mid-run state it needed. A fresh process already does all of it correctly. This is option 2 from #694.

### A pre-existing bug this makes matter

A polling `message stream` (the default, without `--subscribe`) re-emitted the room's recent history on start: it recorded only the messages it displayed, so with `--initial-messages 0` the first poll reported everything as new. That was a nuisance while restarts were rare. Now that a stream restarts on every re-key, it would flood a bot each time, so polling now records the whole window at startup and shows only the last N, as `--subscribe` always has.

### Smaller points

- In `--subscribe` mode, the re-check's pointer GET shares the connection and discards what it steps over, so the loop queues a full-state re-fetch after each re-check.
- A polling stream's sleep is capped at the next re-check, so a long `--poll-interval` can't defeat it.

## Testing

- `pointer::recheck`: reconfirmed generation, unverified hash change, verified move to a known generation, verified move past this binary.
- `pointer::highest_floor`: disk behind, disk ahead, withdrawal at equal versions, lower-hash tiebreak at equal versions.
- `error::report`: a re-key prints its own message with status 75; every other error prints `Error: {err:?}` with status 1, exactly as before.
- Stream behaviour: a seeded poll of an unchanged room emits nothing, for `-i 0`, `1` and `3`; a move ends the stream with status 75; a failed re-check is an ordinary failure; jitter stays within ±20% and actually varies.
- Source pins: both loops re-check inside their timer gate and reset it; the subscription loop re-fetches after re-checking; the polling loop seeds before emitting and caps its sleep; `resolve_pointer` remembers each floor before saving and consults that memory.

Every guard was verified by a **compiling** mutation that reintroduces its bug (13 in total, all caught). `cargo test -p riverctl` passes, with no new clippy warnings.

Reviewed at Full tier over two rounds (four Claude lenses plus Codex each round). Findings and dispositions, including those declined and why, are in the review comments.

Bumps riverctl to 0.2.18 (0.2.17 is the latest on crates.io). No `common/` or WASM changes.

Closes #694

[AI-assisted - Claude]

https://claude.ai/code/session_01LSj17em5VJvYN4ZYCRJVdF

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4782,7 +4782,7 @@ dependencies = [
 
 [[package]]
 name = "riverctl"
-version = "0.2.17"
+version = "0.2.18"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -18,6 +18,12 @@ All notable changes to riverctl will be documented in this file.
   with a short delay** (systemd `Restart=on-failure` plus `RestartSec=`), and
   note that a restart with `--initial-messages N` re-emits the last N messages.
   See "Running `message stream` as a long-lived bot" in the README.
+- `message stream` without `--subscribe` no longer re-emits the room's recent
+  history when it starts. It recorded only the messages it displayed, so with the
+  default `--initial-messages 0` the first poll reported every recent message as
+  new, and with `-i N` it reported all but N of them. It now records everything
+  already in the room and shows only the last N, as `--subscribe` always has.
+  This matters more now that streams restart on every re-key.
 
 ## [0.2.15] - 2026-09-06
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -24,6 +24,9 @@ All notable changes to riverctl will be documented in this file.
   new, and with `-i N` it reported all but N of them. It now records everything
   already in the room and shows only the last N, as `--subscribe` always has.
   This matters more now that streams restart on every re-key.
+- A polling `message stream` whose `--poll-interval` is longer than the re-check
+  interval (a few minutes) now also polls at each re-check, so a long interval
+  cannot delay noticing a re-key.
 
 ## [0.2.15] - 2026-09-06
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 All notable changes to riverctl will be documented in this file.
 
+## [0.2.18] - 2026-09-12
+
+### Fixed
+- `message stream` no longer goes silently deaf when River re-keys the room
+  contract while it is running. It used to look up the room's address once at
+  startup and never again, so after a re-key it kept listening to a contract
+  nobody writes to: no error, no output, just a room that appeared to go quiet
+  until the bot was restarted. It now re-checks every few minutes and, when
+  River has re-keyed, **exits with status 75** so a supervisor restarts it — a
+  restarted riverctl follows the new address through the normal startup path.
+  If River has re-keyed to a version newer than this riverctl, it keeps running
+  and prints an upgrade warning instead, since restarting would not help.
+  (freenet/river#694)
+
+  **If you run a bot, run it under something that restarts it**, and treat
+  status 75 as "restart me" rather than as a failure. See "Running
+  `message stream` as a long-lived bot" in the README.
+
 ## [0.2.15] - 2026-09-06
 
 ### Fixed
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -10,15 +10,14 @@ All notable changes to riverctl will be documented in this file.
   startup and never again, so after a re-key it kept listening to a contract
   nobody writes to: no error, no output, just a room that appeared to go quiet
   until the bot was restarted. It now re-checks every few minutes and, when
-  River has re-keyed, **exits with status 75** so a supervisor restarts it — a
-  restarted riverctl follows the new address through the normal startup path.
-  If River has re-keyed to a version newer than this riverctl, it keeps running
-  and prints an upgrade warning instead, since restarting would not help.
+  River's pointer names a different generation, **exits with status 75** so a
+  supervisor restarts it. A restarted riverctl uses the current generation.
   (freenet/river#694)
 
-  **If you run a bot, run it under something that restarts it**, and treat
-  status 75 as "restart me" rather than as a failure. See "Running
-  `message stream` as a long-lived bot" in the README.
+  **If you run a bot, run it under something that restarts it on any failure,
+  with a short delay** (systemd `Restart=on-failure` plus `RestartSec=`), and
+  note that a restart with `--initial-messages N` re-emits the last N messages.
+  See "Running `message stream` as a long-lived bot" in the README.
 
 ## [0.2.15] - 2026-09-06
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "riverctl"
-version = "0.2.17"
+version = "0.2.18"
 edition = "2021"
 authors = ["Freenet Project"]
 description = "Command-line interface for River decentralized chat on Freenet"
--- a/cli/README.md
+++ b/cli/README.md
@@ -365,16 +365,17 @@ failures should be left stopped.
 
 **Restarting can re-deliver messages.** A restarted stream starts with no memory
 of what it already emitted. With `--initial-messages N`, it re-emits the last N
-messages as new, which covers anything sent while it was restarting but repeats
-what you had already seen. If the bot acts on messages, deduplicate on
+messages as new, which covers up to N messages sent while it was restarting but
+repeats what you had already seen. If the bot acts on messages, deduplicate on
 `message_id`. Without `--initial-messages`, nothing is repeated, but messages sent
 during the restart are not shown.
 
 If River has re-keyed to a version **newer than your riverctl**, the restarted
 process still streams the room once it has been moved to the new address, and
 prints a warning telling you to upgrade (`cargo install riverctl`). Until the room
-has been moved — which an up-to-date River client does when someone opens it — it
-keeps failing and being restarted.
+has been moved — which an up-to-date River client does when someone opens it —
+`--subscribe` keeps failing and being restarted, while polling mode stays running
+and keeps retrying.
 
 ## Configuration
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -125,7 +125,9 @@ riverctl message stream <room-owner-vk> --format json --no-version-check |
 ```
 
 For a bot that runs unattended, wrap this in something that restarts it — see
-"Running `message stream` as a long-lived bot" below.
+"Running `message stream` as a long-lived bot" below. Note that a pipeline's exit
+status is the LAST command's, so a wrapper around this one sees `jq`'s status, not
+riverctl's; use `set -o pipefail` in bash, or restart on any exit.
 
 `message list` and `message stream --format json` both carry
 `author_verifying_key` — the author's full base58 key, comparable to `whoami`'s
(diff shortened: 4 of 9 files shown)


sanity (commented):
## Comprehensive PR Review: #698

### Summary
- **Type:** fix
- **Linked issue:** #694
- **Review tier:** Full (touches the long-running async stream loops and process exit semantics)
- **Reviewers run:** code-first, skeptical, testing, big-picture (blind, in parallel) + external `codex review --base origin/main` (ran successfully)
- **HEAD reviewed:** `869fd209`. Findings below are addressed in `31eddd07`, which is being re-reviewed.

### Alignment
No reviewer found a discrepancy between the code and the PR body, CHANGELOG or doc comments. Both stream modes are covered, and `main`'s change preserves output and exit status for every other command.

### Findings

| # | Finding | Source | Disposition |
|---|---|---|---|
| 1 | Keeping a stale riverctl running on a newer generation is wrong for a stream. Reads are permitted against any generation, so a restart can stream it once the room has moved, while the process that stays put listens to the retired key. | skeptical (M1) | **Fixed.** Exits on any verified re-key; the branch and its warn-once state are deleted. |
| 2 | With a read-only config dir, the anti-rollback floor re-read from disk can be behind what the process verified, so an older signed record could look like a move. | Codex (P1) | **Fixed.** `pointer::highest_floor` combines the in-memory and on-disk floors; a withdrawal wins at equal versions. |
| 3 | Removing the `>= next_recheck` timer gate passed every test. | testing | **Fixed.** Source pin requires the call inside the gate. |
| 4 | `main`'s stderr/exit wiring was untested; `eprintln!` could panic on a closed stderr. | testing, skeptical (L1) | **Fixed.** Extracted and tested `error::report`; written with errors ignored, like std's `Termination`. |
| 5 | README said to restart only on status 75, which would permanently stop a bot on a transient post-re-key failure; systemd example lacked `RestartSec`. | skeptical (M2, M3), code-first | **Fixed.** Restart on any failure with a delay; `RestartSec` documented. |
| 6 | `--initial-messages N` re-emits N messages on every restart. | skeptical (L2), code-first | **Fixed.** Documented, with dedup-on-`message_id` advice. |
| 7 | Exit message said "re-keyed while this command was running", untrue when the stream started on a fallback anchor. | skeptical (L3) | **Fixed.** Reworded. |
| 8 | Earlier README bot example had no restart wrapper. | big-picture | **Fixed.** Cross-reference added. |
| 9 | Pin sliced `subscribe_and_stream` to the impl block's close, not its own. | testing, code-first | **Fixed.** |
| 10 | Re-check can delay Ctrl+C / `--timeout` by up to 10s. | Codex (P2) | **Declined.** Both loops already make bare room fetches with a 30s timeout on every poll or notification, so the worst case is unchanged. |
| 11 | A late error reply to the pointer GET can end `--subscribe` with status 1. | skeptical (M2) | **Declined.** The stdlib cannot attribute errors to requests (see its own TODO), so this can't be told apart from a broken connection. Exiting costs a restart; swallowing risks the silent deafness #694 is about. Mitigated by the restart-on-any-failure guidance. |
| 12 | Floor rewritten every re-check even when unchanged. | skeptical (L4) | **Declined.** One small atomic write every few minutes. |
| 13 | Narrow restart loop if floor save fails *and* the pointer is unreachable at startup; Ctrl+C during a re-key exits 75 not 0; polling now ends on a withdrawn pointer. | skeptical (L5-L7) | Accepted as documented behaviour. |

### Verification
Every new guard was checked by a compiling mutation that reintroduces its bug: timer gate removed, unverified hash treated as a move, withdrawal tiebreak dropped, remembered floor ignored, re-key reported as a crash, ordinary errors reported differently. All six fail. `cargo test -p riverctl`: 399 lib + integration tests pass.

Not covered by tests: `resolve_pointer`'s call to `remember_floor` (an `ApiClient` can't be constructed without a live connection), and the synthetic re-fetch's delivery through the notification handler (pinned by source position only).

### Verdict
**Needs Changes — Re-review Required After Fix** (findings 1 and 2 changed behaviour). Re-review running on `31eddd07`.

[AI-assisted - Claude]

sanity (commented):
## Review rounds 2 and 3: #698

Round 1 was posted earlier, reviewing `869fd209`. This comment records rounds 2 and 3, which were run but not posted until now. That was an oversight, flagged by the big-picture lens.

**Tier:** Full. **Each round:** code-first, skeptical, testing, big-picture (blind, in parallel), plus `codex review --base origin/main`. Codex ran successfully both times.

---

### Round 2, HEAD `31eddd07`

| Finding | Source | Disposition |
|---|---|---|
| Polling `message stream` re-emitted the room's recent history on start. It recorded only the messages it displayed, so the first poll reported the rest as new (all of them with `-i 0`). Pre-existing, but now routine because streams restart on every re-key. The README also claimed the opposite. | code-first (High), skeptical (H1), Codex (P1) | **Fixed** in `2e926122`. `seed_stream` records the whole window and shows only the last N, matching `--subscribe`. |
| `resolve_pointer`'s `remember_floor` call was unpinned. | testing | **Fixed**, source pin. |
| Deleting the `next_recheck` reset still passed. | skeptical (L1) | **Fixed**, source pin. |
| Equal-version floors didn't use the resolver's lower-hash tiebreak. | Codex (P2), skeptical (L3) | **Fixed** in `920ac751`. I had declined this in round 1; two reviewers converging on it plus a three-line fix changed the call. |
| A `--poll-interval` longer than the re-check interval delayed noticing a re-key. | Codex (P2) | **Fixed** in `920ac751`: the sleep is capped at the next re-check. |
| README: "subscribes", `--force`, "briefly fail", the jq pipeline hiding exit status, unsupervised bots now stopping. | code-first, skeptical, big-picture | **Fixed.** |
| A late pointer reply can be misattributed to the next room GET, causing a spurious recovery and a redundant republish. | skeptical (M1) | **Declined.** Existing first-response-wins class tracked in #689/#690. The consequence here is a merge-safe no-op, and a real fix needs request correlation the stdlib lacks. |
| A late error reply can still end `--subscribe` with status 1. | skeptical (M2) | **Declined.** It can't be told apart from a broken connection. Exiting costs one restart, and subscribe mode replays nothing. |
| Status-1 restarts continue on a withdrawn pointer or persistently unreadable floor. | skeptical (L2) | **Accepted.** These are permanent conditions a fresh process refuses at startup too. |

### Round 3, HEAD `920ac751`

| Finding | Source | Disposition |
|---|---|---|
| Seeding reintroduced the #324 spurious-`delete` bug. Every message became eligible for a delete event, so deleting a message the stream never showed printed a `delete` for it. | skeptical (M1), Codex (P2) | **Fixed** in `a389324f`. `seed_stream` suppresses deletions for everything outside the shown window, exactly as the subscription path does. |
| `save_pointer_floor` inserted blindly, so a concurrent riverctl's newer floor could be overwritten with an older one. | Codex (P2) | **Fixed** in `a389324f`. The save decides under its lock via `highest_floor`, so the stored floor only moves forward. |
| Nothing verified that `seeded = true` follows each seed. | testing | **Fixed**, source pin. |
| Startup fetch failure in polling mode became silent (it used to exit 1 with `-i N`). | skeptical (M2) | **Fixed**: one stderr note; README corrected for polling vs subscribe. |
| New test inserted between a helper and its doc comment. | skeptical (L1) | **Fixed.** |
| `--initial-messages N` covers *up to* N messages from a restart, not "anything". | code-first | **Fixed** (README). |
| The PR description claimed two posted rounds that weren't posted. | big-picture | **Fixed** by this comment. |

Code-first reported no discrepancies between the code and its claims in round 3. Big-picture found the code sound and scope justified, and flagged only process gates: this record, and CI.

### Verification
Every guard added across rounds 2 and 3 was checked with a mutation that compiles and reintroduces its bug, and all were caught. The list: seeding only displayed messages, `remember_floor` dropped, remembered floor ignored, unseeded poll emitting, polling gate never re-armed, higher hash kept at equal versions, poll sleep uncapped, unshown deletions not suppressed, fallback seed not marking `seeded`, and a blind floor save. `cargo test -p riverctl`: 402 lib tests plus the integration suites pass, with no new clippy warnings.

### Next
The round 3 fixes each address a specific finding and are mutation-verified, but they add new logic (deletion suppression, the monotonic save). So a final skeptical + Codex pass is running on `a389324f` before merge, scoped to that delta.

[AI-assisted - Claude]

